PR #14, opened 2022-04-03
Remove duplicated counters and filter group id in kernel

marceloamaral:
**What this PR does / why we need it:**
Currently, in the eBPF code, when a new `cgroup_time` (i.e., `new_cgroup`) is created, it is initialized with counter values, and shortly after the initialization, the structure instance is also updated with the same counter, doubling the values.
image
While this only happens at the initialization, we shouldn't duplicate the counters. Also, for the sake of performance and memory usage into the kernel, we should only create a new structure instance when it does not exist, not every time that the existing structure instance is updated.

Additionally, the cgroup id 0 is filtered in the user space, however, it is possible to filter it in the kernel space, which also reduce the in-kernel computation and memory usage.

Finally, we just need to read the process's command (`bpf_get_current_comm`) during the  structure instance initialization.


rootfs:
Thank you @marceloamaral 

I should have documented this, hopefully this is not too late :D 

To update the ebpf module, please modify it [here](https://github.com/sustainable-computing-io/kepler/blob/main/bpf_assets/perf_event/perf_event.c), then generate the bindata using [this script](https://github.com/sustainable-computing-io/kepler/blob/main/scripts/bindata.sh). You'll then have two modified file: the .c and .go files.

Once you update the PR, I'll happily merge it.

marceloamaral:
Thanks @rootfs! I updated this PR according to your comments. Could you please review it again?

rootfs:
thank you @marceloamaral !

PR #15, opened 2022-04-03
Update rapl sysfs event detection and measurement

marceloamaral:
**Why we need this PR:**
Before this PR, the logs were showing that the DRAM energy consumption was null.

image
After this PR, we can see accurate cpu and dram energy consumption:
image

**What this PR does:**
Investigating and analyzing both the [`rapl-read.c`](https://github.com/deater/uarch-configure/blob/master/rapl-read/rapl-read.c) code and the `sysfs` rapl sub-directory, the dram energy consumption file exist and the core energy consumption file that is missing in the measured system. Each `sysfs` rapl sub-directory has a file called `name` with the rapl event name.
image

In the old Kepler logic, the rapl `sysfs` event path was defined regarding a sub-folder id:
image
However, as we saw before, the ID varies. For example, in the old Kepler logic, `intel-rapl:%d:0/` is related to core power consumption, but looking at the sub-folder on the measured system, `intel-rapl:%d:0/` is related to dram power consumption.

Additionally, as the system is missing core power consumption, we calculate it from ` core = package - dram`.

Energy measurement is collected considering the CPU topology, ie., considering the packet id as we are planning to use this information in another PR.

### DIFF
--- a/pkg/power/sysfs.go
+++ b/pkg/power/sysfs.go
@@ -19,71 +19,117 @@ package power
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"strconv"
 	"strings"
 )
 
 const (
-	corePath    = "/sys/class/powercap/intel-rapl/intel-rapl:%d/intel-rapl:%d:0/energy_uj"
-	dramPath    = "/sys/class/powercap/intel-rapl/intel-rapl:%d/intel-rapl:%d:1/energy_uj"
-	uncorePath  = "/sys/class/powercap/intel-rapl/intel-rapl:%d/intel-rapl:%d:2/energy_uj"
-	packagePath = "/sys/class/powercap/intel-rapl/intel-rapl:%d/energy_uj"
+	// sysfs path templates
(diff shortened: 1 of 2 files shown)


pkg/power/sysfs.go
```diff
@@ -19,71 +19,117 @@ package power
...
+			packageEnergy := readEventEnergy(packageEvent)
+			coreEnergy := readEventEnergy(coreEvent)
+			for id, e := range packageEnergy {
+				energy += e - coreEnergy[id]
+			}
+			return energy, nil
+		}
+	}
+
+	return energy, fmt.Errorf("cloud not read RAPL energy for %s", event)
```
rootfs, line 84: nit: `could`

rootfs (approved):

rootfs:
Thank you @marceloamaral This looks great!

pkg/power/sysfs.go
marceloamaral: oww, good catch! Since this PR got merged, I will submit another one to fix the typo.....

marceloamaral: Never mind, I saw that you just did it! Thanks!

PR #16, opened 2022-04-06
Fix counters overflow and cgroup id to pod id

marceloamaral:
**Why we need this PR:**

- Problem 1:
Before this PR, the logs were showing that the `cpuTime` and the `cycles` of a given pod was with quintillion values (20 digits).
And after reaching this value, the counter does not update anymore because it reached the `uint64 max value`: 18,446,744,073,709,551,615 (20 digits).
image

- Problem 2:
There is a bug when calculating the delta of eBPF table counters. The delta should be the current accumulated eBPF table counter `ct.Time` minus the previous `ct.Time`. However, the calculation was `ct.Time - podEnergy[podName].CPUTime` and the `podEnergy[podName].CPUTime` is a delta, not the absolute value. So it wasn't calculating the difference between the current value and the last seen value.
image

- Problem 3:
The matching of cgroup id to pod id was not working in the measured system (`baremetal01`). The function [CgroupIdToName()](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/pod_lister/resolve_container.go#L87) was not finding the cgroup path with the cgroup id reported by the eBPF code. Because of that, all pods reported by kepler were identified as system processes.
Not even a cpu intensive pod (i.e., running sysbench) was identified by kepler.
image

**What this PR does:**
1. fix the counter overflow by resetting the eBPF table counters every time the value is read. With this, the reader code no longer needs the logic to calculate the delta values. So, also simplifying the code.
2. use pid instead of cgroup id to match pid -> container id -> pod id. 
3. fix Prometheus metrics reporting the current values as gauge metrics, only cumulative metrics (i.e., `_total`) should be used as Prometheus counter
4. refactor the name of some variables, for example, change the pod variable `LastEnergyInDram` to `AggEnergyInDram`, since it reports the aggregated used energy, and changed variable `EnergyInDram` to `CurrEnergyInDram`

I tried to break the changes in several commits to be easier to review.


rootfs:
@marceloamaral thanks for the fixes!

A couple of high level comments/discussions. Current pod lister queries kubelet endpoint (not the kube-apierserver) to obtain the Pod spec, this is inline with your scalability concern. The Pod spec contains the entire Pod information. Currently only the name and namespace are used but there are plans to use annotations and labels for high level aggregations in the future. These info are not available from the `crio` endpoint. In addition, accessing kubelet endpoint can be accessed controlled, while accessing crio doesn't seem to have that level of control.

marceloamaral:
@rootfs because of the the security breach of directly accessing the container runtime from the pod that you mentioned, I reverted the runtime access to ListPods back to how it was before.
Could you please review it again?

PR #22, opened 2022-04-11
[WIP] using cgroup id to resolve pod name

rootfs:


marceloamaral:
Thanks @rootfs for finding this bug...

Actually the high CPU usage is related to not updating the cache properly... But I agree that exposing the `/proc` folder can become a security issue.

However, the previous update not only updated the usage of the cgroup id but also refactored the function and variable name.
So, I'm proposing another update on a new PR[#23](https://github.com/sustainable-computing-io/kepler/pull/23), reverting pod matching back to using the cgroup id. But, the new PR retains all the refactoring part, plus additional fine-tuning performance improvements.

PR #25, opened 2022-04-11
[WIP] load cpu_frequency tracepoint

rootfs:
thanks to @marceloamaral suggestion

### DIFF
--- a/bpf_assets/perf_event/perf_event.c
+++ b/bpf_assets/perf_event/perf_event.c
@@ -166,7 +166,9 @@ int sched_switch(switch_args *ctx)
         new_process.cache_misses = cache_miss_delta;
         bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
         new_process.start_time = time;
-        new_process.last_freq = last_freq;
+        if (last_freq > init_freq) {
+            new_process.last_freq = last_freq;
+        }
         new_process.last_avg_freq_update_time = time;
         new_process.avg_freq = last_freq;
         processes.update(&pid, &new_process);
@@ -180,7 +182,9 @@ int sched_switch(switch_args *ctx)
         process_time->cache_misses += cache_miss_delta;
 
         // calculate runtime cpu frequency average
-        process_time->last_freq = last_freq;        
+        if (last_freq > init_freq) {
+            process_time->last_freq = last_freq;
+        }
         u64 last_freq_total_weight = (process_time->last_avg_freq_update_time - process_time->start_time)* process_time->avg_freq;
         u64 freq_time_delta = time - process_time->last_avg_freq_update_time;
         u64 last_freq_weight = process_time->last_freq * freq_time_delta;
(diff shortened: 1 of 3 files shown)


marceloamaral:
After some investigation, I think the ftrace event of 'trace_cpu_frequency' is not working as we expect. 

For example, I am printing only the cpu frequency of the first 9 cpus:

```
freq.CPUID 0 freq.AvgFreq 0
freq.CPUID 1 freq.AvgFreq 0
freq.CPUID 2 freq.AvgFreq 0
freq.CPUID 3 freq.AvgFreq 0
freq.CPUID 4 freq.AvgFreq 0
freq.CPUID 5 freq.AvgFreq 0
freq.CPUID 6 freq.AvgFreq 0
freq.CPUID 7 freq.AvgFreq 0
freq.CPUID 8 freq.AvgFreq 0
```

As suggested by @rootfs, for triggering ftrace event 'trace_cpu_frequency', [the kernel says](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/samples/bpf/cpustat_user.c#n154) to forces to set 'scaling_max_freq' to "Y"MHz and then recovery back to
 the maximum frequency value "X"GHz. 
```
for i in $(seq 0 9) ; do 
   echo 800000 > /sys/devices/system/cpu/cpu${i}/cpufreq/scaling_max_freq; 
   echo 3000000 > /sys/devices/system/cpu/cpu${i}/cpufreq/scaling_max_freq ; 
done
```

Then, after that, we start to see values from the eBPF code, but only the scaling_max_freq value. However, what we want to see is the current frequency, not the maximum.
```
freq.CPUID 0 freq.AvgFreq 3000000
freq.CPUID 1 freq.AvgFreq 3000000
freq.CPUID 2 freq.AvgFreq 3000000
freq.CPUID 3 freq.AvgFreq 3000000
freq.CPUID 4 freq.AvgFreq 3000000
freq.CPUID 5 freq.AvgFreq 3000000
freq.CPUID 6 freq.AvgFreq 3000000
freq.CPUID 7 freq.AvgFreq 3000000
freq.CPUID 8 freq.AvgFreq 3000000
```

At the same time, I also measured the current frequency from sysfs:
```
for j in $(seq 0 1); do 
   for i in $(seq 0 8); do 
      echo $i - $(cat /sys/devices/system/cpu/cpufreq/policy$i/scaling_cur_freq)
   done
   echo "--"
done
```
```
0 - 826151
1 - 1666225
2 - 2390428
3 - 1218803
4 - 800324
5 - 1176160
6 - 1149860
7 - 823022
8 - 814367
--
0 - 1847530
1 - 2348989
2 - 928625
3 - 997615
4 - 1338714
5 - 961773
6 - 1265506
7 - 800180
8 - 1211796
```

One might think that instead of changing the maximum frequency to force trigger the event, we could update the current frequency. However, the kernel does not allow us to change the current frequency, like `echo "1356788" > /sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq`. Which makes sense, as the [`cpuinfo_cur_freq`] is the current CPU frequency taken from the hardware.

Also, note that the frequency at which the CPU actually runs changes very often within a very small range. The previous code called it twice with no sleep in between. What doesn't change very often is `scaling_max_freq` and `scaling_min_freq`.

PR #38, opened 2022-04-16
collect cpu time per cpu in ebpf, and cpu frequency from acpi sysfs

marceloamaral:
**Why we need this PR:**
The ftrace event of 'trace_cpu_frequency' is not working as expected. Even forcing triggering the event, the ebpf code was not collecting the current CPU frequency for the cores. But only the maximum frequency as the trigger was just updating that value.

Actually, ftrace is only tracking when a kernel function is called to change the CPU frequency (max, min or curr).
But, during the development experiments, these changes were not happening because the OS is using [CPU governor]'s "performance" policy (https://wiki.alpinelinux.org/wiki/CPU_frequency_scaling). To see OS CPU frequency changes, one need to use a different policy such as "ondemand".

However, even if the operating system is not forcing the CPU frequency to change, it does not mean that the hardware is not changing it. In fact, we found that the CPU frequency is always varying within a very small range.

**What this PR does:**
This PR implements another logic to collect the average CPU frequency per pod.
- First the ebpf code starts tracking cpu time by cpu. Before, it was only the aggregation. Colleting per cpu not only allows for a lower level of granularity, but also defines which CPU the process used.
- Second, we dynamically collect the acpi CPU frequency in the background, averaging every 1 second.
- Third, we calculate the weighted average of the CPU frequency for each pod,  were the weights are the cpu time.

After this PR, you will be able to see the average used cpu frequency per pod:
image

This PR is related to PR#[25](https://github.com/sustainable-computing-io/kepler/pull/25).


bpf_assets/perf_event/perf_event.c
```diff
@@ -114,44 +130,34 @@ int sched_switch(switch_args *ctx)
...
     if (((s64)val > 0) || ((s64)val < -256))
     {
-        prev = prev_cpu_cycles.lookup(&cpu);
+        prev = prev_cpu_cycles.lookup(&cpu_id);
         if (prev)
         {
             cpu_cycles_delta = val - *prev;
         }
-        prev_cpu_cycles.update(&cpu, &val);
+        prev_cpu_cycles.update(&cpu_id, &val);
```
rootfs, line 138: just a thought, would cpu_cycles/delta a good proxy of cpu frequency?

rootfs: potentially, the two perf counters can also be used to estimate frequency
https://man7.org/linux/man-pages/man2/perf_event_open.2.html
```
PERF_COUNT_HW_CPU_CYCLES
                          Total cycles.  Be wary of what happens during
                          CPU frequency scaling.
PERF_COUNT_HW_REF_CPU_CYCLES (since Linux 3.3)
                          Total cycles; not affected by CPU frequency
                          scaling.
```

marceloamaral: Okay, in theory we can calculate the CPU frequency as the ratio of the number of cycles per second.

I will implement it and validate it with the current CPU frequency information from sysfs.

marceloamaral:
As suggested by @rootfs, I tried to use `PERF_COUNT_HW_CPU_CYCLES` to determine the CPU frequency. However, while the CPU frequency is the number of cycles per second, the `PERF_COUNT_HW_CPU_CYCLES` is application specific metric.

For example, suppose a CPU with a certain frequency can process 100 cycles per second, but if the application only needs 10 cycles, the application’s cycle usage will wrongly determine the CPU frequency….. So, if a process is not using 100% of the CPU, we cannot determine the current CPU frequency only accounting the `PERF_COUNT_HW_CPU_CYCLES`.


To verify that, I implemented a test using two different methods to measure the CPU frequency as show in the table as follows.

The second column is the cpu frequency from the sysfs (the real frequency).
The third column is the total number of CPU cycles used per second.
The fourth column is the second/first ratio, which means the error.

```
PU_ID:  CPU frequency from sysfs -          total number of cycles/second  -  ratio (second / first)
0     : 1255300                 -      0                              -  0
1     : 1157684                 -      894982                         -  0.77307975
2     : 1199974                 -      788081                         -  0.6567484
3     : 1373484                 -      847687                         -  0.6171805
4     : 1325083                 -      880425                         -  0.6644308
5     : 974409                  -      891067                         -  0.9144698
6     : 926092                  -      860142                         -  0.92878777
7     : 1741005                 -      838235                         -  0.48146653
8     : 1445783                 -      845703                         -  0.5849452
9     : 2143484                 -      749640                         -  0.3497301
10     : 1435002                -      265362                         -  0.18492141
11     : 911185                 -      262604                         -  0.2882016
12     : 882749                 -      257340                         -  0.29152203
13     : 1041426                -      894863                         -  0.8592673
14     : 1120397                -      259279                         -  0.23141746
15     : 991192                 -      258806                         -  0.26110664
16     : 1038305                -      260831                         -  0.25120908
17     : 1265184                -      861106                         -  0.6806177
18     : 854519                 -      862724                         -  1.0096024
19     : 1033330                -      870989                         -  0.842896
20     : 2002683                -      879473                         -  0.4391474
21     : 1508103                -      833973                         -  0.55299485
22     : 1445540                -      885438                         -  0.6125315
23     : 1207195                -      889627                         -  0.7369378
24     : 927276                 -      877977                         -  0.9468348
25     : 1014826                -      274150                         -  0.2701456
26     : 1026141                -      259792                         -  0.25317407
27     : 1713776                -      879575                         -  0.51323855
28     : 1189053                -      320778                         -  0.2697763
29     : 821138                 -      890963                         -  1.0850348
30     : 957966                 -      855121                         -  0.89264315
31     : 1736413                -      258451                         -  0.14884245
```

We can see that the error can be more than 70%. This is because calculating CPU frequency by counting cycles used is not an accurate proxy.

marceloamaral:
This PR doesn't change the CPU utilization, Kepler remains using between 0.3 and 0.5%:
image

PR #39, opened 2022-04-20
update power model including the other energy consumption besides CPU and memory for each pod

marceloamaral:
**Why we need this PR:**
image
Currently, the power model to calculate pod's energy consumption is only accounting for CPU and DRAM energy from RAPL, missing all other energy consuming parts in the host, i.e., fans, power supply, PCI, drivers, devices, mboard, etc. .

Also, to validate the power model, one should sum the power consumption of all pods plus the OS and be equal the power consumption of the node.

**What this PR does:**
This PR introduces a new Power Model for calculating the pod's energy consumption. The new Power Model adds a proportionate share of the energy consumed by the other parts to each pod. The "other" energy consumption will be distributed equally across all pods + the OS. Where the OS represents all processes running outside of the pods in the OS (not just the kernel process).

After this PR, it will be possible to see the pod's core, dram, and other energy consumption:
image

After the tests, this PR did not introduce any additional cpu usage.


rootfs:
@williamcaban yet another time to update the dashboard :D

williamcaban:
ack.

PR #50, opened 2022-05-04
DRAM energy consumption estimate

rootfs:
The DRAM energy consumption consists of background (or static/leakage), estimated by resident memory in this PR, and activation (or dynamic) power that is estimated by L3 cache miss PMC .


marceloamaral (commented):
@rootfs What is the main motivation for this change? Is it because we don't have cache miss metric on the VMs?

Is MemBackground meant to be idle energy consumption?
I see value in separating the idle and dynamic part, and using the memory footprint to assosiate the idle DRAM energy consumption to containers. However, I am not comfortable with the Coefficient part. It is currently hardcoded as 50% of the DRAM power consumption is idle and 50% is dynamic.

That said, if MemBackground is the idle power consumption, we can calculate idle as the minimum DRAM power consumption during the execution, and then calculate the delta.

rootfs:
Yes, the coefficients are hardcoded but in the path to a model server. We'll use the data points to train the model and get the best estimate from there.

PR #51, opened 2022-05-06
work with model server to update estimate coefficient

rootfs:


rootfs:
tested with the [simple model server](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/server/server.py)

PR #56, opened 2022-05-10
add model server sidecar container

rootfs:


rootfs:
Add as a sidecar container for dev purpose

PR #78, opened 2022-06-08
Updated dashboard with Carbon footprint 

williamcaban:
Including an example of dashboard reporting Carbon Emissions based on Coal, Petroleum and Natural Gas for the selected Namespace based on its 24hrs energy consumption.


rootfs:
@williamcaban cool, do you mind following up with a PR to add the screenshot of the dashboard?

PR #79, opened 2022-06-08
screenshot of dashboard reporting carbon

williamcaban:


rootfs:
would be great if you have a dynamic gif :D

PR #85, opened 2022-06-17
fix format of freq/cpu time, handle zero total cpu time

sunya-ch:
This PR fixes format issue when CPU time = 0.
- Cause: due to typo fmt.Sprint instead of fmt.Sprintf, exported value of v.CurrCPUTime and v.AvgCPUFreq contains %f.
   Change: use the same standard strconv.FormatFloat instead
- Cause: totalCPUTime = 0, avgFreq is NaN
   Change: compute avgFreq equally for all core when totalCPUTime = 0


rootfs:
thank you @sunya-ch for catching up these! Would you please resolve the conflicts?

PR #87, opened 2022-06-20
Fix value export in wrong order and continue handling due to energy read failure

sunya-ch:
This PR contains the following changes:

**required changes:**
- fix export order
  some export variables and labels are in a wrong order (specifically, total_energy and curr_energy).
- unlock and add consideration of delta period when calling continue after failed energy read
  I found that the program is locked forever if it calls continue in the for loop. 
  for example, the program is locked after printing this.
  ```bash
  2022/06/20 06:33:11 failed to get latest core or dram energy. Core energy 44590920 should be more than 18446744073688380290; Dram energy 34647631 should be more than 100237883
  ```

**optional changes:**
- change the order of bytes_read and bytes_write to be same as the other (i.e., total, curr) 
- define image version tag in Makefile


marceloamaral:
Seems like a great starting plan to deal with the problem of reading a value less than the last measured....
As we can imagine, this probably happens after the RAPL counters overflow

So we should actually handle this exception by resetting the delta calculation rather than keeping propagating the problem across different iterations.

So my suggestion is that when there is an error reading the RAPL, we should set the lastEnergy* counters to 0.
The subsequent iteration will not have a valid lastEnergy and will not calculate the delta, but will update the lastEnergy to calculate the delta in the next subsequent iteration.

For example:
- at t, energy = 10
     - calculate delta
     - lastEnergy = energy
- at t + 3s, cannot read (e.g., RAPL overflow)
     - no delta
     - lastEnergy = 0
- at t + 6s, skip energy calculation since lastEnergy = 0, energy = 16
     - no delta
     - lastEnergy = energy
- at t + 9s, energy = 20
     - delta = energy - lastEnergy
     - lastEnergy = energy

rootfs:
@sunya-ch @marceloamaral good ideas! let's sync up on these.

sunya-ch:
Regarding to the PR, I simplify the case when RAPL counter reaches overflow (e < lastEnergy).
There are three valid methods. This PR implements the method 1.
method 2 is invalid in the case that RAPL counter handles the overflow but continue counting from overflow point.
Also, I don't think method 3 is a good idea because it can affect the other application that refers to the same RAPL counter file.
power_counter_reset_issue

I also found there is someone implements the method 4 by using **Max** value which we can also do the same thing.
https://github.com/cs-21-pt-9-01/rapl.rs/commit/5ab636f0b8834cda89251881a58ea9e4c8123668

In the case that we cannot read RAPL counter file,
case 1: file is just temporarily unavailable but RAPL counter is still counting and can later update to the file (e > lastEnergy over x unavailable sampling time)
- This PR takes an average over sampling time = (e-lastEnergy)/x

case 2: RAPL is broken and counter is forcefully reset to 0
- This PR expects reset value < lastEnergy and back to above case (e < lastEnergy)

PR #88, opened 2022-07-05
Refactor pod_lister and cgroup proposal

sunya-ch:
Hi, Team.

There are two code-refactoring proposals in this PR.

1. pod_lister
**issue:** Some short-lived pods cannot be detected at the time of ticker (as kubelet does not report deleted pods).
**proposal:** 
    1. Detect create/delete pod events from API server with kubernetes API
    2. Keep the pod/container info until x seconds after the pod is deleted. (now x is set to 30s but we might change it to ticker time).

2. cgroup metrics from file systems
**issue:** It is difficult to extend function to get more cgroup metrics from file systems and also from the different file system structure. 
**proposal**: 
   1. Separate cgroup as a new module
   2. Define SliceHandler to handle different file hierarchy
   3. Define StatReader to read different kinds of metric files
   4. Define Converter map to map heterogenous metrics read from different file system to the same standard value such as CPU time from nanoseconds to microseconds.   

**notices for both proposals:**
   1.  Multiple changes to dependencies (vendors), Need to run go mod vendor to update
   2. Both take container ID as an input. So, It is still rely on `pod_lister.GetContainerIDFromcGroupID` function from previous implementation to convert cGroup ID to container ID. However, for older kernel version, we may need to handle Process ID in stead of cGroup ID.
   3. The modified code comes with unit test and is confirmed functionality on the experimental environment. Nevertheless, it causes relatively major changes, I believe discussion is needed.
   
Best,


sunya-ch:
separate to two PRs

PR #95, opened 2022-07-18
plugin regression models

rootfs:
regression coefficients are from a linear regression model running on a cascade lake setup


rootfs:
there are quite gofmt changes

PR #98, opened 2022-07-22
add cgroup to collector with refactored code and unit test

sunya-ch:
This PR is to export cgroup metrics of pods to the prometheus format. 
In addition to that, I refactor the collector struct for properly add desc of each metric to prometheus desc channel. 

Also, I implement the unit test for the collector too.
To run the test independently, go into collector folder and run `../../test-bin/ginkgo`. 
> note: this must be run after ginkgo is installed by `make ginkgo-set`

### DIFF
--- a/pkg/cgroup/slice_handler.go
+++ b/pkg/cgroup/slice_handler.go
@@ -32,6 +32,11 @@ import 	(
 	"strings"
 )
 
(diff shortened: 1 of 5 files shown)


rootfs:
thank you @sunya-ch

PR #101, opened 2022-07-28
fix issues in collector.go (including missing podEnergy stats) 

sunya-ch:
This PR is to fix the bug from the previous PR https://github.com/sustainable-computing-io/kepler/pull/98

**Bug:** When func (c *Collector) Describe(ch chan<- *prometheus.Desc) is called, there is no data in the map.
Consequently, no description available when Collect is called.

**Fix:** Generate descriptions by function and calls them at both calls.

### DIFF
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


rootfs:
thank you @sunya-ch

PR #102, opened 2022-08-03
Enable Kepler to use PID instead of cGroupID when needed

marceloamaral:
#### Why we need this PR:
Kernel versions prior to 4.18 do not support collecting cgroup id in eBPF code.
To support these kernels and other scenarios, we need to extract the container ID from the cGroupID or PID.

#### What this PR does:
This PR replaces the use of cGroupID for PID when the kernel is older than 4.18. 
It is also possible to enable the use of PID via command line parameter.

I also extended the resolve container code to extract the container ID from not only OpenShift systems (i.e. focused on crio) to also work on Kubernetes systems with other runtimes (e.g. docker and containerd).

I don't have access to push the new image to `quay.io/sustainable_computing_io/kepler:latest`
Someone else needs to do this.


rootfs:
@marceloamaral I'll run some local test and get back to you

manifests/openshift/kepler/01-kepler-install.yaml
```diff
@@ -72,8 +72,7 @@ spec:
         - /usr/bin/kepler
         - -address
         - 0.0.0.0:9102
-        - -enable-gpu
-        - "false"
+        - -enable-gpu=false
```
rootfs, line 75: keep it `true`

pkg/config/config.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	CGROUP_ID_MIN_KERNEL_VERSION = 4.18
+)
+
+var (
+	EnabledEBPFCgroupID = false
+)
+
+// EnableEBPFCgroupID enables the eBPF code to collect cgroup id if the system has kernel version > 4.18
```
rootfs, line 36: I don't think kernel version is reliable to determine if cgroup id is supported. For instance, RHEL 8 kernel supports both cgroup v1 and v2 yet the kernel version is 4.18

A better way is to following the instruction [here](https://docs.okd.io/latest/post_installation_configuration/machine-configuration-tasks.html#nodes-nodes-cgroups-2_post-install-machine-configuration-tasks):

```
you can verify that cgroups v2 is enabled by checking that the /sys/fs/cgroup/cgroup.controllers file is present on the node. This file is created by cgroups v2.
```

rootfs:
@marceloamaral I've ran some tests. Here are the comments, rest look good.

bpf_assets/perf_event/perf_event.c
```diff
@@ -76,7 +76,7 @@ BPF_ARRAY(prev_cache_miss, u64, NUM_CPUS);
 static void safe_array_add(u32 idx, u16 *array, u16 value)
 {
 #pragma clang loop unroll(full)
-    for (int array_index = 0; array_index < CPU_VECTOR_SIZE; array_index++)
+    for (int array_index = 0; array_index < CPU_VECTOR_SIZE-1; array_index++)
```
rootfs, line 79: after this change, please run `scripts/bindata.sh` to update the assets

bpf_assets/perf_event/perf_event.c
```diff
@@ -88,8 +88,12 @@ static void safe_array_add(u32 idx, u16 *array, u16 value)
 
 int sched_switch(switch_args *ctx)
 {
-    u64 pid = bpf_get_current_pid_tgid() & 0xffffffff;
+    u64 pid = bpf_get_current_pid_tgid();
```
rootfs, line 91: the pid is in the [low 32 bits](https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#4-bpf_get_current_pid_tgid)

marceloamaral: The `bpf_get_current_pid_tgid() & 0xffffffff` was not giving to me the right pid of the process. 
Then I saw other eBPF examples, like this one [here](https://github.com/iovisor/bcc/blob/master/tools/cpudist.py) that are not using a mask to get the pid. 
After many tests, using it without the mask, the command was giving the right pid of the process.

pkg/config/config.go
marceloamaral: Initially the PR target to solve the kernel problem does not support the eBPF command ` bpf_get_current_cgroup_id()`. Only kernel >= 4.18 supports this command.

But testing if the system has cgroups v1 or v2 is also important, I will include that.

marceloamaral:
@rootfs, I updated the PR, could you please do another review?

rootfs:
@marceloamaral this looks good to me, do you mind double checking the vendor files again? They look quite volatile after the go.mod update.

marceloamaral:
@rootfs updated, could you please review again?

rootfs:
Thank you! @marceloamaral

PR #105, opened 2022-08-10
export package energy and CPU time per CPU core of each pod

sunya-ch:
This PR is composed of two commits to
1.  defines PackageEnergy struct to keep RAPL energy per package and add GetPackageEnergy() function in the powerInterface for getting RAPL energy per package.  
2. add two new metrics to collector module. These values can be further used in model training.
    - node_package_energy_joule: Current package energy by RAPL in joules
    - pod_cpu_cpu_time_us:           Current CPU time per CPU of each Pod

### DIFF
--- a/pkg/collector/reader.go
+++ b/pkg/collector/reader.go
@@ -88,6 +88,7 @@ type PodEnergy struct {
 	AvgCPUFreq    float64
 	CurrProcesses int
 	CgroupFSStats map[string]*UInt64Stat
+	CurrCPUTimePerCPU map[uint32]uint64
 }
 
 type UInt64Stat struct {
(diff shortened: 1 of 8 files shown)


pkg/collector/reader.go
```diff
@@ -251,7 +254,11 @@ func (c *Collector) reader() {
 					v.CurrCPUInstr = 0
 					v.CurrBytesRead = 0
 					v.CurrBytesWrite = 0
+<<<<<<< HEAD
```
rootfs, line 257: looks a merge conflict

sunya-ch: @rootfs sorry for that. just amend the update to solve the conflict.

PR #108, opened 2022-08-16
fix kernel version parsing position

sunya-ch:
This PR fixes the missing position when getting kernel release version to second decimal number (from x.x to x.xx). 
Additionally, to avoid unnecessary fatal failure when it cannot detect the kernel version from the system info, this PR returns -1 value instead.

_note: the current minimum kernel version check is down to second decimal number (4.18)._


rootfs:
lgtm

PR #110, opened 2022-08-22
dev: create secret manually

ruomengh:
From Kubernetes 1.24, secret will not be create automatically with service account. Thus create_k8s_token.sh would fail due to no secret is created.
Create secret manually to fix the issue.

Signed-off-by: Hao, Ruomeng <ruomeng.hao@intel.com>


rootfs:
Thank you @ruomengh

PR #112, opened 2022-08-23
refactor reader for readability and flexible sets of features

sunya-ch:
This PR refactor the source reader.go with the following items.
- separate pod_energy.go and node_energy.go to handle PodEnergy and CurrNodeEnergy struct separately from the reader.
- use UInt64Stat (Curr,Aggr) struct instead of two static variables to handle these two values systematically. (including overflow checking)
- use map of metric(feature) name for not only cgroup but also hardware counters and kubelet metrics.
- use model/estimator.go to compute pod power

There are two kinds of powers computed.
- **Ratio power:** This is a package power computed by ratio of usage metric defined in config module to the RAPL power.
- **Dynamic power:** This is a dynamic part of package power predicted by customized or trained model if available using only usage metric .

remarks:
- Instead of compute core power from pkg power, I think we should report as it is. If RAPL energy for core component is not available, we should not report it.
- This should work even no estimator sidecar container. (no dynamic power reporter)


rootfs:
thank you @sunya-ch 

can you post the the prometheus query result sample?

pkg/collector/collector.go
```diff
@@ -286,131 +230,60 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) {
...
+				v.PodName, v.Namespace,
+			)
+			ch <- edesc
+		}
+		// dyn curr pod energy
+		edesc := prometheus.MustNewConstMetric(
+			podDesc.CurrEnergy[DYN_ENERGY_LABEL],
 			prometheus.GaugeValue,
-			float64(v.CurrEnergyInOther),
+			float64(v.DynEnergy.Curr)/1000.0, //J
```
rootfs, line 265: If switching to J from mJ, please also update the dashboard unit

sunya-ch:
> thank you @sunya-ch
> 
> can you post the node and pod prometheus query result sample?

Theses are pod's result. (total and curr in RAPL pkg)
Screenshot 2022-08-23 at 21 50 15

Screenshot 2022-08-23 at 22 17 07

And, I just add node energy report with amended commit. (at the beginning, it has only node stat)
Screenshot 2022-08-23 at 22 12 53

pkg/collector/collector.go
sunya-ch: I'm sorry but could you clarify where do you mean the dashboard unit?

rootfs: the dashboard uses mW https://github.com/sustainable-computing-io/kepler/blob/main/grafana-dashboards/Kepler-Exporter-Dashboard-openshift.json#L133

sunya-ch: I see. Do you think should we export energy in J or mJ? 
If you agree to change to J, I will update the grafana. But, if you think it is better to keep the decimal, I will modify the collector sourcecode.

rootfs: let's keep mJ then, thanks

sunya-ch: Thanks. I see that originally you report node power as J. Should I keep it like that? 
So, the units will be
- node power -> J
- package power -> mJ
- pod power -> mJ 

do I understand it correctly?

rootfs: that sounds right

rootfs:
thank you @sunya-ch

pkg/collector/reader.go
```diff
@@ -52,459 +52,331 @@ type CgroupTime struct {
...
-				coreDelta := float64(energyCore-lastEnergyCore) / samples
-				dramDelta := float64(energyDram-lastEnergyDram) / samples
-				if coreDelta == 0 && dramDelta == 0 {
-					log.Printf("power reading not changed, retry\n")
-					samples += 1
-					lock.Unlock()
-					continue
-				}
-
-				if coreDelta > maxEnergyDelta || dramDelta > maxEnergyDelta {
```
rootfs, line 204: is the sanity checking still applicable?

PR #118, opened 2022-08-30
Generalize Grafana dashboard

Feelas:
Kepler-Exporter.json dashboard has been generalized taking OpenShift Carbon USA legacy dashboard as an example. This PR partially addresses #119 issue.

Dashboard is now possible to import by hand and opens up correctly, not throwing any errors. This change doesn't include the URL changes necessary in manifests/openshift/04-grafana-dashboard.yaml.

### DIFF
--- a/grafana-dashboards/Kepler-Exporter.json
+++ b/grafana-dashboards/Kepler-Exporter.json
@@ -1,82 +1,9 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "Prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    },
-    {
-      "name": "VAR_COAL",
-      "type": "constant",
-      "label": "coal",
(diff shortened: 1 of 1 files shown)


rootfs:
thank you @Feelas, this looks good to me, can you post a screen shot of your dashboard?

rootfs:
@Feelas I loaded the dashboard, it looks something missing. The datasource in my environment is prometheus, but it is not showing in the dashboard. The A/B/C series labels need a update too.

image

Feelas:
This is how it looks for me - notice that the Namespace and Pod fields are loaded properly.
Zrzut ekranu (101)
I have removed all of the hardcoded datasource references, so interesting what the difference is, since I don't have any import problems with this definition.

I tried to load up Your modified dashboard after replacing all Prometheus UIDs by hand and then export it using native Grafana exporting feature (Share -> Export -> Export for sharing externally) but afterwards it asks for the datasource when importing manually & doesn't work when importing thru GrafanaDashboard CR, thus no-go there.

rootfs:
cool! Let's merge it

Issue #126, opened 2022-09-02
03-grafana-datasource-define.sh uses deprecated 'oc serviceaccounts get-token'


Feelas:
**Describe the bug**
As can be seen in [OpenShift 4.11 release notes](https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html), the dashboard creation command found in `manifests/openshift/dashboard/03-grafana-datasource-define.sh` uses deprecated `oc serviceaccounts get-token grafana-serviceaccount -n monitoring` command. As a result, the command does not return a valid token. This causes the Grafana "Prometheus" datasource to not be created.

Example output form the command (using oc cli 4.11):
```
[root@workstation dashboard]$ oc serviceaccounts get-token grafana-serviceaccount -n monitoring
Command "get-token" is deprecated, and will be removed in the future version. Use oc create token instead.
error: could not find a service account token for service account "grafana-serviceaccount"
```

**To Reproduce**
1. Use an OpenShift 4.11 cluster
2. Run manifests/openshift/dashboard/03-grafana-datasource-define.sh script and note the output of the script.

**Expected behavior**
Token is created successfully and 03-grafana-datasource-define.sh works properly, creating a Prometheus datasource.

**Screenshots**

**Desktop (please complete the following information):**
- OpenShift 4.11 cluster
- oc tool 4.11.0-rc.1 (any version of 4.11 oc cli can be used)

**Additional context**
I used a straight replacement command `oc create token grafana-serviceaccount --duration 31536000s -n monitoring` to create a 365d token for the dashboard. Not sure whether this is the right duration, since I didn't ever use the old `get-token` command.

rootfs:
OpenShift 4.11 is based on kubernetes 1.24, this token issue is similar to #110 
cc @ruomengh

sallyom:
will need to update the manifests to create the secret similar to the dev/role.yaml also https://github.com/sustainable-computing-io/kepler/blob/main/dev/role.yaml#L23-#L34

Feelas:
Closing due to #128 being merged.

PR #128, opened 2022-09-02
Update manifests and dashboard deploy script

sallyom:
Signed-off-by: Sally O'Malley <somalley@redhat.com>

For https://github.com/sustainable-computing-io/kepler/issues/126


rootfs:
@williamcaban can you take a look? thanks

sallyom:
(the bash script to define grafana dashboard I'm still working on)

marceloamaral:
@sallyom in this case, could you please update the title with the work in progress status? [WIP]

sallyom:
> @sallyom in this case, could you please update the title with the work in progress status? [WIP]

will do - all seems to be working now, ready for review, thanks!

manifests/openshift/kepler/01-kepler-install.yaml
```diff
@@ -60,7 +60,8 @@ spec:
         app.kubernetes.io/name: kepler-exporter
     spec:
       nodeSelector:
-        sustainable-computing.io/kepler: ""
+        #remove and add node annotation if running in non-development
+        #sustainable-computing.io/kepler: ""
```
williamcaban, line 64: If we we are commenting out this nodeSelector it will result in an invalid array so we will need to comment out `spec.nodeSelector` or put something like `kubernetes.io/os: linux`. Probably adding a default nodeSelector for it to be `kubernetes.io/os: linux` will also help with mixed cluster (linux and non-linux nodes) as currently Kepler is known to work only on Linux based OS.

williamcaban (commented):
Consider the comment for `01-kepler-install.yaml`, any of the option should work. The rest LGTM.

manifests/openshift/kepler/01-kepler-install.yaml
sallyom: I added `kubernetes.io/os: linux` there, thanks!

williamcaban (approved):

rootfs:
thank you @sallyom

PR #134, opened 2022-09-05
update ginkgo-set target for go version 1.18, delete vendor folder

sunya-ch:
This PR 

- updates Makefile to fix ginkgo-set target by setting both GOPATH and GOBIN and use `go install` with version tag to be compatible with version 1.18 which is used in the build Dockerfile.
[more details for `go get`/`go install` changes](https://tip.golang.org/doc/go1.18)

- removes the vendor folder to avoid vendor conflict from auto-generated modules.txt

- separates `test` and `test-verbose` target


rootfs:
@sunya-ch I prefer to keep the vendor directory, in case the vendor pkgs are missing and it saves time on CI. 

Can you vendor in the ginkgo pkg?

sunya-ch:
> @sunya-ch I prefer to keep the vendor directory, in case the vendor pkgs are missing and it saves time on CI.
> 
> Can you vendor in the ginkgo pkg?

In this [amend](https://github.com/sustainable-computing-io/kepler/compare/0f5ec496ac74467f4fad52e65b88178e13e3a2e6..df1773f0e26c15eb7ed1b1aa43f81506d443c21f), I rerun the `go mod vendor` and have vendor folder back. 
The previous generated vendor causes some permission issues but now it is fine I guess.

rootfs (approved):

PR #137, opened 2022-09-06
copy files for dev local build

suigh:
After local build for keper and run it, there will an error reports that `normalized_cpu_arch.csv` deson't exist:

Here is the message
```
[root@xxx kepler]# ./_output/bin/_/kepler
2022/09/06 00:58:56 InitSliceHandler: &{map[] /sys/fs/cgroup/cpu /sys/fs/cgroup/memory /sys/fs/cgroup/blkio}
no cpu info: open /var/lib/kepler/data/normalized_cpu_arch.csv: no such file or directory
power not supported
```

To resolve this issue, this pull request will:
1. Add a script named `dev/prepare_dev_env.sh` to copy the needed files for local build to the target folder
2. Update readme to remind developer to run this script

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -58,10 +58,11 @@ If successful, the binary is at `_output/bin/_/kepler`
 
 ### Test
 
-Create the k8s role and token, this is only needed once.
+Create the k8s role and token, copy data files, this is only needed once.
 ```bash
 cd dev/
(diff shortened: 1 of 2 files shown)


rootfs (approved):

rootfs:
thank you @suigh

PR #145, opened 2022-09-07
Model Server Energy Prediction

KaiyiLiu1234:
Retrieves model weights from the Kepler Model Server and uses the weights for energy prediction.

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -162,12 +162,12 @@ func SetModelServerEndpoint(ep string) {
 func getRequest(endpoint string) (*http.Response, error) {
 	req, err := http.NewRequest("GET", endpoint, http.NoBody)
 	if err != nil {
-		return nil, errors.New("could not create request for Model Server Endpoint:" + endpoint)
+		return nil, errors.New("could not create request for Model Server Endpoint " + endpoint)
 	}
 
 	res, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return nil, errors.New("request for Model Server Endpoint failed:" + endpoint)
+		return nil, errors.New("request for Model Server Endpoint failed " + endpoint)
 	}
 	return res, nil
 }
@@ -205,14 +205,13 @@ func retrieveCoeffForCategoricalVariable(categoricalPrediction string, allCatego
 			return architecture.Weight, nil
 		}
 	}
-	return -1, errors.New("architecture feature does not exist.")
+	return -1, errors.New("architecture feature does not exist")
 }
 
-/*
- Using Direct Access instead of Dynamic lookup to retrieve Numerical Weights. Direct access is more efficient and easier
- to implement, but it makes the code less flexible if more fields need to be added to DramModelServerCoeff or
- CoreModelServerCoeff.
-*/
+// Using Direct Access instead of Dynamic lookup to retrieve Numerical Weights. Direct access is more efficient and easier
+// to implement, but it makes the code less flexible if more fields need to be added to DramModelServerCoeff or
+// CoreModelServerCoeff.
+
 func predictLinearDramEnergyConsumption(prediction *EnergyPrediction, dramModelServerCoeff *DramModelServerCoeff) (float64, error) {
 	var energyPrediction float64 = 0
 	dramCPUArchitectureWeights := dramModelServerCoeff.AllWeights.CategoricalVariables.CPUArchitecture
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -38,6 +38,15 @@ type Coeff struct {
 	InterceptDram float64 `csv:"intercept_dram"`
 }
 
+type EnergyPrediction struct {
+	Architecture   string
+	CPUTime        float64
+	CPUCycle       float64
+	CPUInstr       float64
+	ResidentMemory float64
+	CacheMisses    float64
+}
+
 type CategoricalFeature struct {
 	Name   string  `json:"name"`
 	Weight float64 `json:"weight"`
@@ -60,7 +69,7 @@ type CoreModelServerCoeff struct {
 			CPUTime  NormalizedNumericalFeature `json:"cpu_time"`
 			CPUInstr NormalizedNumericalFeature `json:"cpu_instr"`
 		} `json:"Numerical_Variables"`
-	} `json:"All_Weights"`
+	} `json:"core_All_Weights"`
 }
 
 type DramModelServerCoeff struct {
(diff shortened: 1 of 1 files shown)


pkg/model/model.go
```diff
@@ -38,6 +40,103 @@ type Coeff struct {
 	InterceptDram float64 `csv:"intercept_dram"`
 }
 
+type CoreModelServerCoeff struct {
+	AllWeights struct {
+		BiasWeight           float64 `json:"Bias_Weight"`
+		CategoricalVariables struct {
+			CPUArchitecture struct {
+				AlderLake struct {
```
marceloamaral, line 48: Instead of having the Architecture'a namea encoded in structures, it might be better to have a map where the key is the architecture name and the value is the weight...

Architecture's names can be constants to be used to fill the map.

marceloamaral (changes requested):

pkg/model/model.go
KaiyiLiu1234: Thank you for the reply! I was thinking that the structs might be better in this case mainly because all of these fields will all be populated and used. The struct is to store a fully populated json request body. I do plan to pull out some of the nested structs and make them into their own types for readability and ease of use. What are your thoughts?

marceloamaral: Right, but the name of the architectures must be in the variables and not in the name of the fields.
Having the value of variables as field names of the structure is not a good practice, the structure must be generic.

So, we need to have something like:
```golang
type CoreModelServerCoeff struct {
	AllWeights struct {
		BiasWeight           float64 `json:"Bias_Weight"`
		CategoricalVariables struct {
			CPUArchitecture []struct {
				Name   string  `json:"name"`
				Weight float64 `json:"weight"`
			}
		} `json:"cpu_architecture"`
	} `json:"Categorical_Variables"`
}
```

And then the json file will be something like:
```json
{
	"Categorical_Variables": {
		"Bias_Weight": 0,
		"cpu_architecture": [
		{
			"name": "Broadwell",
			"weight": 9
		},
		{
			"name": "Cascade Lake",
			"weight": 127
		}]
}
```

Please, update that accordingly

KaiyiLiu1234: I have updated the structs and incoming json file as requested!

rootfs:
thank you @KaiyiLiu1234

rootfs (approved):

rootfs:
looking forward to integrating this to the collector
cc @sunya-ch

rootfs (approved):

pkg/model/model.go
```diff
@@ -106,6 +161,115 @@ func SetModelServerEndpoint(ep string) {
...
+	energyPrediction += numericalWeights.CPUTime.Weight * normalizedCPUTimePredict
+	normalizedCPUCyclePredict := (prediction.CPUCycle - numericalWeights.CPUCycle.Mean) / math.Sqrt(numericalWeights.CPUCycle.Variance)
+	energyPrediction += numericalWeights.CPUCycle.Weight * normalizedCPUCyclePredict
+	normalizedCPUInstrPredict := (prediction.CPUInstr - numericalWeights.CPUInstr.Mean) / math.Sqrt(numericalWeights.CPUInstr.Variance)
+	energyPrediction += numericalWeights.CPUInstr.Weight * normalizedCPUInstrPredict
+
+	return energyPrediction, nil
+}
+
 /*
```
marceloamaral, line 273: Same here as I commented before

marceloamaral (commented):
Looks good to me now!
@sunya-ch could you please review this PR?

pkg/model/model.go
```diff
@@ -40,53 +41,36 @@ type Coeff struct {
...
-				Weight   float64 `json:"weight"`
-			} `json:"curr_cpu_time"`
-			CurrentCPUInstructions struct {
-				Mean     float64 `json:"mean"`
-				Variance int64   `json:"variance"`
-				Weight   float64 `json:"weight"`
-			} `json:"current_cpu_instructions"`
+			CPUCycle NormalizedNumericalFeature `json:"curr_cpu_cycles"`
+			CPUTime  NormalizedNumericalFeature `json:"curr_cpu_time"`
+			CPUInstr NormalizedNumericalFeature `json:"current_cpu_instructions"`
```
sunya-ch, line 73: just a small comment, we should change json key to the same format and same name as up-to-date exported metrics (cpu_intruction --> cpu_instr, cache_misses  to cache_miss - this is for attacher package). Also, we may remove the `curr` or `current` prefix.

pkg/model/model.go
```diff
@@ -95,48 +79,20 @@ type DramModelServerCoeff struct {
...
-				Mean     float64 `json:"mean"`
-				Variance float64 `json:"variance"`
-				Weight   float64 `json:"weight"`
-			} `json:"curr_cache_misses"`
-			CurrResidentMemory struct {
-				Mean     float64 `json:"mean"`
-				Variance float64 `json:"variance"`
-				Weight   float64 `json:"weight"`
-			} `json:"curr_resident_memory"`
+			CacheMisses    NormalizedNumericalFeature `json:"curr_cache_misses"`
```
sunya-ch, line 85: same comment as line 71.

cache_misses --> cache_miss from attacher
resident_memory --> container_memory_working_set_bytes from pod_lister

sunya-ch (approved):
I think we also need a function to link between collector module to this file.
something like 

```golang
func GetNodePower(metrics []string, values []float64) (corePower, dramPower []float64)
```

rootfs:
@KaiyiLiu1234 can you rebase?

KaiyiLiu1234:
Ok will do. I will also make a small change to how the model server exports the weights.

pkg/model/model.go
```diff
@@ -38,8 +41,60 @@ type Coeff struct {
...
+
+type NormalizedNumericalFeature struct {
+	Mean     float64 `json:"mean"`
+	Variance float64 `json:"variance"`
+	Weight   float64 `json:"weight"`
+}
+
+type CoreModelServerCoeff struct {
+	AllWeights struct {
+		BiasWeight           float64 `json:"Bias_Weight"`
```
rootfs, line 66: there are mixed upper and lower cases here, are they designed to be that way?

KaiyiLiu1234: I believe the naming convention is camel case but start upper case if public. Acronyms are also always capitalized. https://stackoverflow.com/questions/52358247/what-are-golang-struct-field-naming-conventions

KaiyiLiu1234:
Working to resolve.

rootfs (approved):

PR #146, opened 2022-09-08
Add ut for byte order

jichenjc:


pkg/pod_lister/byte_order_test.go
```diff
@@ -0,0 +1,21 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Byte order test", func() {
+	It("Byte order test", func() {
+		order := determineHostByteOrder()
+		switch runtime.GOARCH {
+			case "amd64":
+				Expect(order).To(Equal(binary.LittleEndian))
+			case "s390x":
```
rootfs, line 17: just out of curiosity, any difference between ppc64le and s390x?

jichenjc: I am s390x guy :) 

so I didn't count others (e.g ARM, ppc etc)
will update this

sunya-ch:
May add the other cases refer to this.
https://github.com/xelaj/go-dry/blob/21c77821487c0b6177fd22a7ad91db062ee4f91f/endian.go#L18

pkg/pod_lister/byte_order_test.go
rootfs: np, looking forward to a s390x port!

jichenjc:
>>>Just wonder why we don't use this as a function to determine byte order instead of the test case.

has no clue .. as we have a function to decide the order ,a UT is reasonable?
or we can use library function to decide so we can avoid both function and UT?

pkg/pod_lister/byte_order_test.go
sunya-ch: > > > > Just wonder why we don't use this as a function to determine byte order instead of the test case.
> 
> has no clue .. as we have a function to decide the order ,a UT is reasonable? or we can use library function to decide so we can avoid both function and UT?

Actually, in my understanding, using unsafe pointer is more valid than GORACH as GOARCH can be overrode by environment variable. (on the other hand, the default value (amd64) might be applied if it is unset)

Anyway, I think it is a good UT for checking whether it is consistent on endianness.

sunya-ch (approved):

PR #138, opened 2022-09-06
Fix issue 107 in cgroup v1 environment

ruomengh:
1. User space PID should take upper 32 bit of bpf_get_current_pid_tgid(). Please refer to https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#4-bpf_get_current_pid_tgid
2. Update path regex and containerID regex for cgroup v1 env. Note that this has only been tested with cgroup v1 + containerd

Signed-off-by: Hao, Ruomeng <ruomeng.hao@intel.com>

PR #148, opened 2022-09-08
Revert back the logic to extract container id from path

marceloamaral:
#### Why we need this PR:
Previous PR #138 updated the logic to extract the container ID from a cgroup path.

Unfortunately, I wrongly revised the PR #138 as it had a problem. With the changes, Kepler stopped collecting the container IDs in my environment....

My system has cgroup v1 and the OS is ubuntu, and after creating a kepler deployment, the cgroup path is as follows:
```
11:blkio:/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod481c0ae9_7d40_46dd_b6ca_ba27cb64f87e.slice/docker-28a5e57257f81fcd6d592647dde27e06b53944d58af4fa546ad77a12ce8b41c2.scope
```

Note that the chunk that has `.slice` is actually the pod id and not the container id. The container id is in the chunk that has `.scope`. The previous PR introduced the logic to extract the id from the `.slice` chunk.
Also, in the previous PR, the regex is not extracting an ID but is matching the entire string for that given cgroup path in my environment. You can test the regex using this site [here](https://regex101.com/).

#### What this PR does:
Revert back the logic to extract container id from the cgroup path

@rootfs I suggest reverting the changes now and work on another PR with @ruomengh to make it work in his environment.
Actually @ruomengh, could you share the cgroup path of a kubernetes container running in your environment?

### DIFF
--- a/pkg/pod_lister/resolve_container.go
+++ b/pkg/pod_lister/resolve_container.go
@@ -55,13 +55,12 @@ var (
 	containerIDToContainerInfo = map[string]*ContainerInfo{}
 	cGroupIDToPath             = map[uint64]string{}
 
-
 	//regex to extract container ID from path, default value is for cgroup v2
-	regexFindContainerIDPath = regexp.MustCompile(`.*-(.*?)\.scope`)
+	regexFindContainerIDPath          = regexp.MustCompile(`.*-(.*?)\.scope`)
 	regexReplaceContainerIDPathPrefix = regexp.MustCompile(`.*-`)
 
-	regexReplaceContainerIDPathSufix  = regexp.MustCompile(`\..*`)
-	regexReplaceContainerIDPrefix     = regexp.MustCompile(`.*//`)
+	regexReplaceContainerIDPathSufix = regexp.MustCompile(`\..*`)
+	regexReplaceContainerIDPrefix    = regexp.MustCompile(`.*//`)
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


rootfs:
@marceloamaral thanks for reporting this. In this case, we probably have different container paths. @ruomengh do you happen to have the container path? We'll find a common regex to identify the paths. 

For verification, we need to unit test the regex as well. cc @jichenjc @SamYuan1990

rootfs (approved):

ruomengh:
@marceloamaral @rootfs  I'm running Kepler on RHEL8.6 with cgroup v1. The path is something like following
`13:memory:/system.slice/containerd.service/kubepods-besteffort-pod0043435f_1854_4327_b76b_730f681a781d.slice:cri-containerd:01fd96f7ad292b02a8317cde4ecb8c7ef3cc06ffdd113f13410e0837eb2b2a20`
Seems the path is different on RHEL and Ubuntu. Unfortunately I don't have Ubuntu + cgroup v1 environment. I'll improve the regex to fit all the path pattern, at least the ones we already noticed.

rootfs:
it is indeed a platform issue, thank you @ruomengh

marceloamaral:
@ruomengh could you please check if the PR #152 fix the issue in your platform?

ruomengh:
> @ruomengh could you please check if the PR #152 fix the issue in your platform?

Thanks for prompt fix! It works well as long as PID is correct. So I'll re-submit the changes on PID as I did in PR #138

PR #157, opened 2022-09-09
add integration test after unit test and just test for deployment

SamYuan1990:
add integration test after unit test and just test for deployment

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -56,3 +56,25 @@ jobs:
           GOBIN: /home/runner/go/bin
     - name: Run 
       run: make test-verbose
+
+  integration_test:
+    runs-on: ubuntu-latest
+    needs: unit_test_without_bcc
+    strategy:
+      matrix:
+        kube_provider: [kind]
+    steps:
+      - uses: actions/checkout@main
+      - uses: actions/setup-go@main
+        with:
+          go-version: 1.18
+      - name: install kubectl
+        run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+      - name: build image locally
+        run: export CTR_CMD=docker && make
+      - name: infra up
+        run: ./hack/integration_cluster.sh ${{matrix.kube_provider}}
+      - name: deploy test
+        run: kubectl create -f manifests/kubernetes/deployment.yaml 
+      - name: checking
+        run: until kubectl get svc --all-namespaces ; do date; sleep 1; echo ""; done  
--- a/Developer.md
+++ b/Developer.md
@@ -1,0 +1,21 @@
+# CI test
+our CI test has two phases for now
+- unit test
+- integration test
+
+## unit test
+We run go test base on specific build tag for different conditions.
+
+## integration test
+It should base on the miminal scope of unit test succeeded.
+the logic is
+- download tools as kubectl.
+- build kepler image with specific pr code.
+----------------------------------------------------------------
+base on different k8s cluster, for example kind(k8s in docker)
+- start KIND cluster locally with a local image registry.
+- upload kepler image build by specific pr code to local image registry for kind cluster.
+--------------------------------------------------------------------------------
+back to common checking process
+- deploy a kepler at local kind cluster with image stored at local image registry.
+- check via kubectl command for ...
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ SRC_ROOT :=$(shell pwd)
 OUTPUT_DIR :=_output
 CROSS_BUILD_BINDIR :=$(OUTPUT_DIR)/bin
 FROM_SOURCE :=false
-CTR_CMD :=$(or $(shell which podman 2>/dev/null), $(shell which docker 2>/dev/null))
 ARCH :=$(shell uname -m |sed -e "s/x86_64/amd64/" |sed -e "s/aarch64/arm64/")
 
 ifdef IMAGE_REPO
@@ -21,6 +20,12 @@ ifdef IMAGE_TAG
 	IMAGE_TAG := $(IMAGE_TAG)
 else
 	IMAGE_TAG := latest
+endif
+
+ifdef CTR_CMD
+	CTR_CMD := $(CTR_CMD)
+else 
+	CTR_CMD :=$(or $(shell which podman 2>/dev/null), $(shell which docker 2>/dev/null))
 endif
 
 # restrict included verify-* targets to only process project files
--- a/hack/integration_cluster.sh
+++ b/hack/integration_cluster.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+# SPDX-License-Identifier: Apache-2.0
+set -o errexit
+#set -x
+CLUSTER_NAME=${ENV_KIND_CLUSTER_NAME:-kind}
+
+LOCAL_REGISTRY_NAME=${ENV_LOCAL_REGISTRY_NAME:-kind-registry}
+LOCAL_REGISTRY_PORT=${ENV_LOCAL_REGISTRY_PORT:-5000}
+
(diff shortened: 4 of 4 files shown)


.github/workflows/unit_test.yml
```diff
@@ -56,3 +56,28 @@ jobs:
...
+      - name: install kubectl
+        run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+      - name: check version
+        run: kubectl version || true
+      - name: infra up
+        run: ./scripts/kind.sh
+      - name: build image
+        run: make
+      - name: upload image to local registry
+        run: kind load docker-image quay.io/sustainable_computing_io/kepler:latest
```
rootfs, line 81: @marceloamaral has a recent change to build local images

.github/workflows/unit_test.yml
```diff
@@ -56,3 +56,28 @@ jobs:
           GOBIN: /home/runner/go/bin
     - name: Run 
       run: make test-verbose
+
+  integration_test:
+    runs-on: ubuntu-latest
+    needs: unit_test_without_bcc
```
rootfs, line 62: what is this for?

scripts/kind.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# create a cluster with the local registry enabled in containerd
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+    endpoint = ["http://${reg_name}:${reg_port}"]
+
+EOF
+
+    kubectl apply -f kube/ingress-nginx.yaml
```
rootfs, line 55: where is the yaml?

rootfs:
thank you @SamYuan1990 this is really cool!

.github/workflows/unit_test.yml
SamYuan1990: I suppose for this PR we can ignore the change for local image build?
the logic:
use local docker to build an image with tag `quay.io/sustainable_computing_io/kepler:latest` and upload to local registry here.

.github/workflows/unit_test.yml
SamYuan1990: with this needs, integration_test will start just after one for unit test as unit_test_without_bcc passed.

scripts/kind.sh
SamYuan1990: fixed

.github/workflows/unit_test.yml
SamYuan1990: as all image files at github agent locally, hence I support no need to worry too much?

.github/workflows/unit_test.yml
```diff
@@ -56,3 +56,30 @@ jobs:
...
+      - name: install kubectl
+        run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+      - name: check version
+        run: kubectl version || true
+      - name: build image
+        run: export CTR_CMD=docker && make
+      - name: infra up
+        run: ./scripts/kind.sh
+      - name: upload image to local registry
+        run: kind load docker-image quay.io/sustainable_computing_io/kepler:latest
```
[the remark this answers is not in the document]
SamYuan1990: > will this command load the image into docker registry or kubernetes/openshift registry?
this command will load image into local docker registry for KIND(k8s in docker cluster)

.github/workflows/unit_test.yml
```diff
@@ -56,3 +56,30 @@ jobs:
...
+      - name: deploy test
+        run: kubectl create -f manifests/kubernetes/deployment.yaml
```
marceloamaral, line 83: The `make cluster-sync` command deletes a running kepler instance, creates a new kepler image, pushes it to the repository and deploys kepler to the cluster with the newer image.

It will be great if you use this, which will reuse all the scripts already created for this...
Also, using the scripts we will make sure that everyone will be using the same flow...

.github/workflows/unit_test.yml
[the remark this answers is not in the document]
SamYuan1990: > do you need to load it locally? it will be downloaded when deploying kepler...
yes, we need to load it into local cluster.
as
for PR level, even if we don't have any integration test case for now, but
the logic for this PR at integration part is:
- download tools as KIND(k8s in docker) and kubectl.
- build kepler image with specific pr code.
- start KIND cluster locally with a local image registry.
- upload kepler image build by specific pr code to local image registry for kind cluster.
- deploy a kepler at local kind cluster with image stored at local image registry.
- check via kubectl command for ...

.github/workflows/unit_test.yml
SamYuan1990: again, we can't reuse current script for now.
with reason
`make cluster-sync` is used for release with upstream and for this script just used for KIND as local k8s cluster for testing.
https://github.com/sustainable-computing-io/kepler/blob/main/Makefile#L105
we can't upload pr level image as a temp image to over right any publish registry without test it.

marceloamaral: That is why we need to carefully define the env variables:
`IMAGE_REPO="quay.io/sustainable_computing_io/kepler"` (already the default)
`IMAGE_TAG="$(git describe --tags)"`

But, you will need to create a new script to push the image, which will read the `CLUSTER_PROVIDER` and if it is `"kind"`, it will push the image to the kind cluster

Note that the `build-manifest` will already generate the yam `_output/manifests/kubernetes/generated/deployment.yaml` with the custom image...

.github/workflows/unit_test.yml
SamYuan1990: resolve as out of date

hack/ingress-nginx.yaml
```diff
@@ -0,0 +1,687 @@
+
```
marceloamaral, line 1: As @rootfs has asked before, why do you need the ingress?

And in the case it is necessary, this yaml should be in the `manifests/kind` folder, not in the `hack`...

hack/kind.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# create a cluster with the local registry enabled in containerd
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+    endpoint = ["http://${reg_name}:${reg_port}"]
+
+EOF
+
+    kubectl apply -f hack/ingress-nginx.yaml
```
marceloamaral: Is the ingress necessary to expose the container registry?

hack/ingress-nginx.yaml
SamYuan1990: removed

marceloamaral:
@rootfs this is good to go for me

We can improve it later in another PR....

SamYuan1990:
I am try to remove un used nginx setting for kind cluster and then set pr ready for review.

SamYuan1990:
recording some of our discussion in a developer.md for others to understand our CI process. just in case...

SamYuan1990:
> @rootfs this is good to go for me
> 
> We can improve it later in another PR....

thanks @marceloamaral , so far I put all kind specific script in shall command. 
pre discussed with @rootfs offline, we may need to add more k8s cluster in CI for integration test.
hence, please take a look at https://github.com/sustainable-computing-io/kepler/pull/157/files#diff-012d76a293bfa67c5d65f938b15705520eef288139f3b513cd16266e6f780a4cR65
I put all the common steps and specific steps away.
in another pr, for example we add [microshift](https://github.com/rootfs/microshift) we can extend here with a new value as microshift. and put the script in shell as implementation.

if we can summarized common part for docker upload, then we can optimize at that pr.

SamYuan1990:
@rootfs , @marceloamaral  I am going to enjoy local holiday and weekend, may check msg randomly or next Tuesday night. :-)

rootfs (approved):

rootfs:
this is very helpful, thank you!

PR #158, opened 2022-09-10
Add UT for convertToStandard

jichenjc:


jichenjc:
need https://github.com/sustainable-computing-io/kepler/pull/159 to solve the FIXME

pkg/cgroup/stat_reader_test.go
```diff
@@ -0,0 +1,149 @@
...
+	It("Test converter cgroupfs_user_cpu_usage_us with cpuacct.usage_user", func() {
+		imap := make(map[string]interface{})
+		imap["cpuacct.usage_user"] = uint64(100000)
+		out := convertToStandard(imap)
+		Expect(len(out)).To(Equal(1))
+		v, ok := out["cgroupfs_user_cpu_usage_us"]
+		Expect(ok).To(Equal(true))
+		Expect(v).To(Equal(uint64(100)))
+	})
+	It("Test converter cgroupfs_system_cpu_usage_us with usage_usec", func() {
```
sunya-ch, line 108: Do you mean `Test converter cgroupfs_user_cpu_usage_us`?

pkg/cgroup/stat_reader_test.go
```diff
@@ -0,0 +1,149 @@
...
+		imap := make(map[string]interface{})
+		imap["usage_usec"] = 100
```
sunya-ch, line 110: Sorry. It was my type here. 
Your PR https://github.com/sustainable-computing-io/kepler/pull/159 fixed the issue.
Please also update the test case to use `user_usec`

pkg/cgroup/stat_reader_test.go
```diff
@@ -0,0 +1,149 @@
...
+		v, ok := out["cgroupfs_cpu_usage_us"]
+		Expect(ok).To(Equal(true))
+		Expect(v).To(Equal(uint64(100)))
+	})
+	It("Test converter cgroupfs_cpu_usage_us with usage_usec", func() {
+		imap := make(map[string]interface{})
+		imap["usage_usec"] = 100
+		out := convertToStandard(imap)
+		// FIXME
+		Expect(len(out)).To(Equal(2))
```
sunya-ch, line 76: I think we can just remove FIXME and Let Expect(len(out)).To(Equal(1)).
If the test case is failed, we should just let failed and fix it.

sunya-ch (changes requested):

pkg/cgroup/stat_reader_test.go
jichenjc: done, updated

sunya-ch: @jichenjc 
Thank you for the updates. 
I think you may misunderstand my review on the line 107 (previously 108).  

I mean changing from `cgroupfs_system_cpu_usage_us ` to `cgroupfs_user_cpu_usage_us ` not `usage_us` to `user_us`. 

The other changes look good to me.

pkg/cgroup/stat_reader_test.go
```diff
@@ -0,0 +1,147 @@
...
+	It("Test converter cgroupfs_user_cpu_usage_us with cpuacct.usage_user", func() {
+		imap := make(map[string]interface{})
+		imap["cpuacct.usage_user"] = uint64(100000)
+		out := convertToStandard(imap)
+		Expect(len(out)).To(Equal(1))
+		v, ok := out["cgroupfs_user_cpu_usage_us"]
+		Expect(ok).To(Equal(true))
+		Expect(v).To(Equal(uint64(100)))
+	})
+	It("Test converter cgroupfs_system_cpu_usage_us with user_usec", func() {
```
sunya-ch, line 107: could you also change `cgroupfs_system_cpu_usage_us` to `cgroupfs_user_cpu_usage_us `?

The other updates look good to me.

pkg/cgroup/stat_reader_test.go
jichenjc: yes, I forgot that part ..

sunya-ch (approved):

PR #164, opened 2022-09-12
Adding support to easily create local kubernetes cluster for testing purposes

marceloamaral:
### What this PR does / why we need it:

After making changes to a kepler code, you should test it locally and the github should also trigger integration tests.

This PR introduce the code to provision and start containerized Kubernetes clusters in order to test Kepler on different Kubernetes configurations.

Currently, there is only support for creating a k8s cluster using the `kind` tool, but we will add support for other tools later such as `microshift`.

The developer will be able to simply run:
```bash
export CLUSTER_PROVIDER=`kind`
make cluster-up
```

### Special notes for your reviewer:
After this PR, we will be able to update the github workflow that runs the integration test.
This PR is related to the PR #157.

### DIFF
--- a/cluster-up/cluster-up.sh
+++ b/cluster-up/cluster-up.sh
@@ -17,27 +17,10 @@
 # Copyright 2022 The Kepler Contributors
 #
 
-set -e
+if [ -z "$CLUSTER_PROVIDER" ]; then
+    echo 'You must define a cluster provider (e.g. CLUSTER_PROVIDER="kind")'
+    exit 1
+fi
 
-source cluster-up/common.sh
-
-CLUSTER_PROVIDER=${CLUSTER_PROVIDER:-kubernetes}
-MANIFESTS_OUT_DIR=${MANIFESTS_OUT_DIR:-"_output/manifests/${CLUSTER_PROVIDER}/generated"}
-
-function main() {
-    echo "Cleaning up ..."
-
-    if [ ! -d "${MANIFESTS_OUT_DIR}" ]; then
-        echo "Directory ${MANIFESTS_OUT_DIR} DOES NOT exists. Run make generate first."
-        exit
-    fi
-
-    # Ignore errors because some clusters might not have prometheus operator
-    kubectl delete --ignore-not-found=true -f ${MANIFESTS_OUT_DIR} || true
-
-    sleep 2
-
-    echo "Done $0"
-}
-
-main "$@"
+source cluster-up/cluster/$CLUSTER_PROVIDER/common.sh
+up
--- a/cluster-up/common.sh
+++ b/cluster-up/common.sh
@@ -19,25 +19,7 @@
 
 set -e
 
-source cluster-up/common.sh
-
-CLUSTER_PROVIDER=${CLUSTER_PROVIDER:-kubernetes}
-MANIFESTS_OUT_DIR=${MANIFESTS_OUT_DIR:-"_output/manifests/${CLUSTER_PROVIDER}/generated"}
-
-function main() {
-    echo "Cleaning up ..."
-
-    if [ ! -d "${MANIFESTS_OUT_DIR}" ]; then
-        echo "Directory ${MANIFESTS_OUT_DIR} DOES NOT exists. Run make generate first."
-        exit
-    fi
-
-    # Ignore errors because some clusters might not have prometheus operator
-    kubectl delete --ignore-not-found=true -f ${MANIFESTS_OUT_DIR} || true
-
-    sleep 2
-
-    echo "Done $0"
-}
-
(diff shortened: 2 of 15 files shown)


rootfs (approved):

cluster-up/common.sh
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2022 The Kepler Contributors
+#
+
+set -e
+
+# the cluster kind is a kubernetes cluster
+if [ ${CLUSTER_PROVIDER} = "kind" ]; then
```
SamYuan1990, line 23: I don't want `CLUSTER_PROVIDER` check been appears in many shell scripts. as duplicate validation?

dev/README.md
```diff
@@ -2,7 +2,16 @@
 
 A quick start guide to get Kepler up and running inside your container-based development cluster.
 
-## I just want it built and run it on my cluster
+## To create a new emphemeral local kubernetes cluster
+You can bring the cluster up with:
```
SamYuan1990, line 6: just fw to the How to use cluster-up

hack/cluster-down.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2022 The Kepler Contributors
+#
+./cluster-up/cluster-down.sh
```
SamYuan1990, line 19: why we keep this file?

hack/cluster-up.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2022 The Kepler Contributors
+#
+./cluster-up/cluster-up.sh
```
SamYuan1990, line 19: why we keep this single line file?

cluster-up/cluster-up.sh
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2022 The Kepler Contributors
+#
+
+if [ -z "$CLUSTER_PROVIDER" ]; then
```
SamYuan1990, line 20: if we remove duplicate validation.

cluster-up/cluster-up.sh
```diff
@@ -0,0 +1,26 @@
...
+    echo 'You must define a cluster provider (e.g. CLUSTER_PROVIDER="kind")'
+    exit 1
+fi
+
+source cluster-up/cluster/$CLUSTER_PROVIDER/common.sh
+up
```
SamYuan1990, line 26: then this file will only invokes common.sh up?

hack/cluster-down.sh
marceloamaral: This is because to make the Makefile only execute the scripts in `hack/` folder....
Then, the those scripts can execute scripts in other folders....

It is not 100% necessary, but a good way to structure the project....

hack/cluster-up.sh
marceloamaral: same as before

cluster-up/common.sh
marceloamaral: I understand the concern, we could set this validation directly in the other scripts.
I just wanted to make it a generic function, for example, we will need later to also say that `microshit` is an `openshift` cluster.
But I agree that we can do it in a different way...

hack/cluster-down.sh
SamYuan1990: ok

dev/README.md
marceloamaral: Right, we redirect

cluster-up/cluster-up.sh
marceloamaral: Yes, this script loads the functions of a cluster provider, e.g., "kind" or "microshift", and then start the cluster
With that, we have the scripts well structure in the project

Issue #161, opened 2022-09-11
Add lint and memory leak detection as a basic test case in CI 


SamYuan1990:
**Is your feature request related to a problem? Please describe.**
Before unit test running, we'd better have a lint test and memory leak detection as a part of CI.

**Describe the solution you'd like**
`go fmt, vet, goimports ....`
```
go build -gcflags="-m -l" ./... | grep "escapes to heap" || true
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rootfs:
in addition to memory leak, please also consider [vulnerability detection](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck)

PR #167, opened 2022-09-12
Update go test to run all go vet checks

marceloamaral:
Add all go vet checks as a basic test case in CI.

As part of building a test binary, go test runs go vet on the package and its test source files to identify significant problems. To run all go vet checks, we can use the -vet=all flag. More info [here](https://manpages.debian.org/testing/golang-go/go-test.1.en.html).

This is PR is part of the effort for testing the code quality, as describe in the issue #161


rootfs (approved):

SamYuan1990:
LGTM

Issue #179, opened 2022-09-14
getKernelVersion doesn't work at all


jichenjc:
**Describe the bug**

https://github.com/sustainable-computing-io/kepler/blob/main/pkg/config/config.go#L65

paste those to https://go.dev/play/ and run it 
```
package main

import (
	"encoding/json"
	"fmt"

	"github.com/zcalusic/sysinfo"
)

func main() {

	var si sysinfo.SysInfo

	si.GetSysInfo()

	data, err := json.MarshalIndent(&si, "", "  ")
	if err == nil {
		var result map[string]map[string]string
		if err = json.Unmarshal(data, &result); err != nil {
			fmt.Println("----")
			fmt.Println(err)
			fmt.Println("----")
		}
	}
	fmt.Println("done")
}

```

```
----
json: cannot unmarshal number into Go value of type string
----
done

```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

jichenjc:
I will submit a PR to fix this soon

jichenjc:
https://github.com/zcalusic/sysinfo/blob/99e836ba64f229922382eecd8a346e6b3d1b560a/kernel_linux.go#L22

we can refer to this function and just use it instead of using the whole library

SamYuan1990:
but `fmt.Println(si.Kernel.Version)` is empty at go playground.

jichenjc:
I don't think we should use this library (at least for now) at all
as you can see , they just read a file and we should get from that file directly
https://github.com/zcalusic/sysinfo/blob/99e836ba64f229922382eecd8a346e6b3d1b560a/kernel_linux.go#L22

SamYuan1990:
> I don't think we should use this library (at least for now) at all as you can see , they just read a file and we should get from that file directly https://github.com/zcalusic/sysinfo/blob/99e836ba64f229922382eecd8a346e6b3d1b560a/kernel_linux.go#L22

I am not sure if this one looks better 
https://github.com/cilium/ebpf/blob/b81c37125bd70a4f086b073e821a5ea79b31e394/features/version.go#L12
...// ignore for code trace stack
https://github.com/cilium/ebpf/blob/3418f341ac0460cf4164146e0c1ec6c40156b4fc/internal/vdso.go#L21

marceloamaral:
We can read the os-release files instead of use the library.
[...]

PR #180, opened 2022-09-14
Add UT for config:isCGroupV2

jichenjc:
will wait for merge this then follow up with https://github.com/sustainable-computing-io/kepler/issues/179

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -32,6 +32,16 @@ const (
 	cGroupV2Path = "/sys/fs/cgroup/cgroup.controllers"
 )
 
+type Client interface {
+	getSysInfo() ([]byte, error)
+	getCgroupV2File() string
+}
+
+type config struct {
+}
+
+var c config
+
 var (
 	EnabledEBPFCgroupID = false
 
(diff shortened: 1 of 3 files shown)


pkg/config/config_test.go
```diff
@@ -0,0 +1,66 @@
...
+		file, err := createTempFile("")
+		Expect(err).NotTo(HaveOccurred())
+		defer os.Remove(file)
+		tempfile = file
+
+		Expect(true).To(Equal(isCGroupV2(mockc)))
+
+		tempfile = "/tmp/do_not_exist"
+		Expect(false).To(Equal(isCGroupV2(mockc)))
+	})
```
SamYuan1990, line 65: @jichenjc , do you think we need add a test to cover `getKernelVersion` as #179 reports for `getKernelVersion`?

jichenjc: yes, I plan to add UT in this PR and modify code+UT for that issue with follow up PR

pkg/config/config_test.go
```diff
@@ -0,0 +1,66 @@
...
+var _ = Describe("Test Configuration", func() {
+	It("Test cgroup version", func() {
+		file, err := createTempFile("")
+		Expect(err).NotTo(HaveOccurred())
+		defer os.Remove(file)
+		tempfile = file
+
+		Expect(true).To(Equal(isCGroupV2(mockc)))
+
+		tempfile = "/tmp/do_not_exist"
```
rootfs, line 63: a bit paranoid but what if this file does exist in the test environment?

pkg/config/config.go
```diff
@@ -32,6 +32,16 @@ const (
 	cGroupV2Path = "/sys/fs/cgroup/cgroup.controllers"
 )
 
+type Client interface {
+	getSysInfo() ([]byte, error)
+	getCgroupV2file() string
```
rootfs, line 37: nit: `getCgroupV2File`

rootfs:
@jichenjc can you sign your commits? https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

pkg/config/config_test.go
jichenjc: updated with a more `unlikely name` :)

PR #187, opened 2022-09-16
detect kernel version through unix way

### LINKED ISSUE #179: getKernelVersion doesn't work at all

jichenjc:
Fixes #179


pkg/config/config.go
```diff
@@ -17,12 +17,12 @@ limitations under the License.
 package config
 
 import (
-	"encoding/json"
+	"bytes"
 	"fmt"
 	"os"
 	"strconv"
 
-	"github.com/zcalusic/sysinfo"
```
rootfs, line 25: can you remove this pkg from go.mod and vendor too?

rootfs (approved):

PR #201, opened 2022-09-18
grafana: Set the default datasource to "prometheus"

kenplusplus:
Without this patch, the default datasource is "Grafana" itself. So it requires to fix the datasource for each panel in kepler dashboard. This patch forge to set the data source to "prometheus".

It has been tested on a kubernetes cluster (v1.25.1).

Signed-off-by: Lu Ken <ken.lu@intel.com>

### DIFF
--- a/grafana-dashboards/Kepler-Exporter.json
+++ b/grafana-dashboards/Kepler-Exporter.json
@@ -28,7 +28,7 @@
   "panels": [
     {
       "collapsed": false,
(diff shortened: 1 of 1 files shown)


rootfs:
cc @Feelas

rootfs (approved):

rootfs:
the datasource could vary among platforms. If you can follow up with a README instruction, that'll be great.

PR #202, opened 2022-09-18
multi-arch: create kepler_base multi-arch image

huoqifeng:
- Cross-platform `kepler_base` image comes from different Dockerfile for different `arch` for maintain purpose. 
- It's a nightmare to maintain all arches in a single `Docekerfile`, so we build arch specific images first and then create a manifest then.
- arm64 base image is just a base, some necessary RPM might be needed in future
- Trigger it when there is a tag generated.

Signed-off-by: huoqifeng <huoqif@cn.ibm.com>

### DIFF
--- a/.github/workflows/image-base.yml
+++ b/.github/workflows/image-base.yml
@@ -3,9 +3,7 @@ name: base
 on:
   push:
     branches: 
-      - 'main'
-    tags:
-      - 'v*'
+      - 'base'
 
 jobs:
   docker:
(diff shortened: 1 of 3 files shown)


huoqifeng:
@SamYuan1990 @rootfs should we run https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/image-base.yml in a separate branch to create the kepler_base multi-arch images separately from `main`?

.github/workflows/image-base.yml
```diff
@@ -24,15 +27,29 @@ jobs:
...
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: ./build/Dockerfile.base.arm64
+          platforms: linux/arm64
+          push: true
+          tags: quay.io/sustainable_computing_io/kepler_base:latest-arm64
+      - name: Create and upload multi-arch image
+        run: |
+            docker pull --platform=linux/s390x quay.io/sustainable_computing_io/kepler_base:latest-s390x
```
rootfs, line 51: Can this be embedded in a script and called out by `Makefile`? That'll help a local dev build.

rootfs:
@huoqifeng thanks for taking this action! Having a separate branch is a good idea. It could well be the case other platform development can work in their own branch without having to break the `main` branch. The image tag should pick up the branch name in that case. Would you like to make that happen?

.github/workflows/image-base.yml
huoqifeng: @rootfs done

huoqifeng:
@rootfs maybe, we can merge this PR to `main` and create a `base` branch to generate the multi-arch base images?

Makefile
```diff
@@ -110,6 +110,14 @@ push-image:
 	$(CTR_CMD) push $(IMAGE_REPO):$(IMAGE_TAG)
 .PHONY: push-image
 
+multi-arch-image-base:	
+	docker pull --platform=linux/s390x quay.io/sustainable_computing_io/kepler_base:latest-s390x; \
```
rootfs, line 114: can you replace `docker` with `${CTR_CMD}`? We usually run `podman` rather than docker during development.

huoqifeng: Done

rootfs (approved):

PR #213, opened 2022-09-20
Update perf err (or it will print nil always)

jichenjc:


rootfs:
can linter capture this type of error?

jichenjc:
check https://github.com/sustainable-computing-io/kepler/pull/213/files#diff-9460a34504f0c66b69c7bd5890c338e717e9abcc29d33dad961bd31d0edaf9bbL71

err is defined ,the reason is perfErr actually has more userful info and err is always nil from context

rootfs:
@jichenjc can you see if [go static checker](https://staticcheck.io/) can detect mistakes like this?

jichenjc:
not sure. .. @rootfs 
need some analysis later on , thanks for the link

PR #223, opened 2022-09-21
follow up getKernel version (e.g version 5.4.x.x.x)

jichenjc:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 
 	"golang.org/x/sys/unix"
@@ -51,6 +52,8 @@ var (
 	DRAMUsageMetric       = "curr_cache_miss"
 	UncoreUsageMetric     = ""                // no metric (evenly divided)
 	GeneralUsageMetric    = "curr_cpu_cycles" // for uncategorized energy; pkg - core - uncore
+
+	versionRegex = regexp.MustCompile(`^(\d+)\.(\d+).`)
 )
 
 // EnableEBPFCgroupID enables the eBPF code to collect cgroup id if the system has kernel version > 4.18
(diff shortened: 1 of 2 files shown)


jichenjc:
after change, it print 

config getKernelVersion:  5.4

pkg/config/config.go
```diff
@@ -82,11 +82,19 @@ func getKernelVersion(c Client) float32 {
 	}
 	// per https://github.com/google/cadvisor/blob/master/machine/info.go#L164
 	kv := utsname.Release[:bytes.IndexByte(utsname.Release[:], 0)]
+	// for something like 4.18.10-24-generic
 	val, err := strconv.ParseFloat(string(kv[:4]), 32)
 	if err == nil {
 		return float32(val)
+	} else {
+		// for something like 5.4.0-48-generic
```
marceloamaral, line 90: `5.4.0-48-generic` and `4.18.10-24-generic` looks very similar to me
What is the problem here?

jichenjc: check https://github.com/sustainable-computing-io/kepler/issues/179#issuecomment-1253069991

rootfs: what about kernel `5.0, 6.0`?
https://en.wikipedia.org/wiki/Linux_kernel_version_history

how about splitting the string by `.`,  taking the first two substrings, and recreating a new kernel string?

pkg/config/config.go
```diff
@@ -82,11 +82,19 @@ func getKernelVersion(c Client) float32 {
 	}
 	// per https://github.com/google/cadvisor/blob/master/machine/info.go#L164
 	kv := utsname.Release[:bytes.IndexByte(utsname.Release[:], 0)]
+	// for something like 4.18.10-24-generic
 	val, err := strconv.ParseFloat(string(kv[:4]), 32)
 	if err == nil {
 		return float32(val)
+	} else {
```
rootfs, line 89: how about splitting the string by `.`, and taking the first two substrings, and recreating the new kernel string?

jichenjc:
@rootfs done

pkg/config/config_test.go
```diff
@@ -70,11 +70,27 @@ var _ = Describe("Test Configuration", func() {
...
 		Expect(true).To(Equal(valid))
 
 		tempUnixName = "3.10"
-		valid = getKernelVersion(mockc) > cGroupIDMinKernelVersion
+		v = getKernelVersion(mockc)
+		valid = v > cGroupIDMinKernelVersion
+		Expect(-1).NotTo(Equal(v))
+		Expect(false).To(Equal(valid))
+
+		tempUnixName = "3.1"
```
rootfs, line 90: a bit paranoid but can you add a case `6.0-rc6`?

rootfs:
thank you @jichenjc, one small suggestion, ready to go

pkg/config/config.go
marceloamaral: +1

marceloamaral (approved):
Good to go now
/lgtm

PR #225, opened 2022-09-21
generalize power model

sunya-ch:
The main changes are in `pkg/model`. 

The logic follows the discussion https://github.com/sustainable-computing-io/kepler/discussions/266.
In this package, there will be three main functions to be called by collector package.
- _GetNodeTotalPower_: to get node total power if energy sensor is not available
```go
// pkg/collector/node_energy.go
func (v *NodeEnergy) SetValues(sensorEnergy map[string]float64, pkgEnergy map[int]source.RAPLEnergy, totalGPUDelta uint64, podMetricValues [][]float64) {
...
	if len(sensorEnergy) == 0 && model.NodeTotalPowerModelValid {
		// no energy sensor
		valid, estimatedTotalPower := model.GetNodeTotalPower(nodeUsageValues, systemValues)
		if valid {
			v.SensorEnergy.AddCurrStat(estimatorID, estimatedTotalPower)
		}
	}
```
- _GetNodeComponentPowers_: to get node component power (RAPL socket) if RAPL is not available
```go
// pkg/collector/node_energy.go
func (v *NodeEnergy) SetValues(sensorEnergy map[string]float64, pkgEnergy map[int]source.RAPLEnergy, totalGPUDelta uint64, podMetricValues [][]float64) {
...
	if len(pkgEnergy) == 0 && model.NodeComponentPowerModelValid {
		// no RAPL
		valid, estimatedComponentPower := model.GetNodeComponentPowers(nodeUsageValues, systemValues)
		if valid {
			v.EnergyInCore.AddCurrStat(estimatorPkgKey, estimatedComponentPower.Core)
			v.EnergyInDRAM.AddCurrStat(estimatorPkgKey, estimatedComponentPower.DRAM)
			v.EnergyInUncore.AddCurrStat(estimatorPkgKey, estimatedComponentPower.Uncore)
			v.EnergyInPkg.AddCurrStat(estimatorPkgKey, estimatedComponentPower.Pkg)
		}
	}
```
- _GetPodPowers_: to get pod component + other power
```go
// pkg/collector/reader.go
c.readEnergy()
nodeEnergy.SetValues(sensorEnergy, pkgEnergy, totalGPUPower, podMetricValues)
podComponentPowers, podOtherPowers := model.GetPodPowers(podMetricValues, systemValues, nodeEnergy.Curr(), totalGPUPower, nodeEnergy.GetNodeComponentPower())
```

Specifically, to estimate the power by trained models, there are two kinds of estimator. 
1. By trained weight, applying linear regression locally in Kepler main container
2. By archived model, applying arbitrary model by Kepler estimator sidecar

For simplicity of reviews, I separate this PR to 5 commits.
1. [implement local estimator](https://github.com/sustainable-computing-io/kepler/pull/225/commits/d0b3bcbcbc0c40e4662c4e72c1d576bb7b52e399): power models for estimating or calculating power locally (in Kepler container) including ratio power model and linear regression.
2. [implement sidecar estimator](https://github.com/sustainable-computing-io/kepler/pull/225/commits/05861b634b915b906a1ac9dc5375a66887a16c74): power models for making a request to kepler estimator to apply arbitrary models aiming at higher accuracy. 
3. [implement power model call](https://github.com/sustainable-computing-io/kepler/pull/225/commits/ee9c21292566a3d20ce892099c2bb8cf323ad17c): main function called by collector package
4. [integrate model to collector package](https://github.com/sustainable-computing-io/kepler/pull/225/commits/f5cd97963b9cc9911d7a2cbeec3377a18603e9a1): updates on collector package to call functions in model package
5. [update config related to model package](https://github.com/sustainable-computing-io/kepler/pull/225/commits/ce3ae214ca161bdb45a01a42f781ad35f99ed31c): updates on other config related to the model package

############################################
notes:

**previous:** 
```bash
pkg
├── model
│   ├── estimate.go
│   ├── estimate_test.go
│   ├── model.go
│   └── suite_test.go
└── power
  └── rapl
      └── source
          └── estimate.go
```
   
**this PR:** 
```bash
pkg
|   |-- model
|   |   |-- estimator
|   |   |   |-- local
|   |   |   |   |-- lr.go
|   |   |   |   |-- lr_test.go
|   |   |   |   |-- ratio.go
|   |   |   |   `-- ratio_test.go
|   |   |   `-- sidecar
|   |   |       |-- estimate.go
|   |   |       `-- estimate_test.go
|   |   |-- model.go
|   |   |-- model_test.go
|   |   |-- node_component_power.go
|   |   |-- node_total_power.go
|   |   |-- pod_power.go
|   |   |-- suite_test.go
|   |   `-- types
|   |       `-- types.go
```

**credit:**
I would like to give a credit to @marceloamaral for helping brushing up the above package design.


marceloamaral:
Could you please run `make golint` and fix the lint issues?

rootfs:
@KaiyiLiu1234

pkg/model/model.go
```diff
@@ -1,260 +1,86 @@
...
+
+const (
+	AbsPower ModelOutputType = iota + 1
+	AbsModelWeight
+	AbsComponentPower
+	AbsComponentModelWeight
+	DynPower
+	DynModelWeight
+	DynComponentPower
+	DynComponentModelWeight
```
marceloamaral, line 24: I think we need to description about the `AbsPower` and the `DynPower`
Both a comment here and a small documentation

pkg/model/model.go
```diff
@@ -1,260 +1,86 @@
...
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+model package provides NodePackedPower, PodPowerRatio, and PodDynamicPower
+- NodePackedPower: node-level power in core, uncore, dram, and package
+- PodPowerRatio: power that each pod or system should account for when considering usage shares (ratio)
+- PodDynamicPower: power that each pod should account for when considering only its usage values
```
marceloamaral, line 5: This is when the hardware counter is not available, right?
I think we can make this comment clearer

marceloamaral (changes requested):
This PR will change the VM's power model to a different method, right?
We need to provide a simple description of the power model in the Documentation to keep tracking this.

For the sake of tracking this discussion. Could you please add here in this PR a brief description of the old and the new VM Power Model?

rootfs:
@sunya-ch is it still WIP?

sunya-ch:
> @sunya-ch is it still WIP?

Yes, please let me do the following checklist before changing this to ready-to-merge PR.

- [ ] documenting power model choices
- [ ] documenting training pipeline availability and road map on kepler-model-server
- [ ] implement integration test case for kepler-estimator to kepler-model-server  
- [ ] implement integration test case for kepler to kepler-estimator and to kepler-model-server

These are planned to be done by this week.

pkg/model/estimate.go
```diff
@@ -3,165 +3,81 @@ package model
...
+		MetricNames:    metricNames,
+		MetricValues:   metricValues,
+		SelectFilter:   config.ModelSelectFilter,
+		OutputType:     outputTypeStr,
+		SystemFeatures: systemFeatures,
+		SystemValues:   systemValues,
 	}
 	powerRequestJSON, err := json.Marshal(powerRequest)
 	if err != nil {
 		log.Printf("marshal error: %v (%v)", err, powerRequest)
```
marceloamaral, line 48: We have changed all the logs to use klog with verbosity support #238

The general rule of thumb is:
- verbosity V(1): must always be printed
- verbosity V(3): extra info
- verbosity V(4): only debug info
- verbosity V(5): only extra debug info

marceloamaral (changes requested):

jichenjc:
FYI

>>>verbosity V(1): must always be printed
klog.log (w/o V) will print always
klog.V(1) will print things when -v=1+

sunya-ch:
I rebase and update the PR description/document. Now this PR is ready for reviews.

rootfs:
lgtm
@marceloamaral any change needed?

marceloamaral:
@rootfs I prefer to merge the PR #287 first...

pkg/collector/metrics.go
```diff
@@ -43,10 +45,12 @@ var (
 	availableCgroupMetrics  []string = cgroup.GetAvailableCgroupMetrics()
 	availableKubeletMetrics []string = podlister.GetAvailableKubeletMetrics()
 	// TO-DO: merge to cgroup stat and remove hard-code metric list
-	iostatMetrics []string = []string{ByteReadLabel, ByteWriteLabel}
-	uintFeatures  []string = getUIntFeatures()
-	features      []string = append(FloatFeatures, uintFeatures...)
-	metricNames   []string = getEstimatorMetrics()
+	iostatMetrics  []string = []string{ByteReadLabel, ByteWriteLabel}
+	uintFeatures   []string = getUIntFeatures()
```
marceloamaral, line 49: The concept of "[feature](https://en.wikipedia.org/wiki/Feature_(machine_learning))" is very much related to machine learning....
Where `feature` is an individual measurable property or characteristic of a phenomenon...

So, in the end, what is the difference between Metrics and Features?

I think we should only use metric, having both makes the code a little confusing...

I understand that the the Linear Model and the other machine learning packages can have the concept of feature, where both numerical metrics and string metadata are used...
But the metric package should only have numerical metrics and metadata...

marceloamaral (changes requested):

pkg/collector/metrics.go
marceloamaral: But of course, we can update it later in another PR

sunya-ch: Exactly, when I named, I think the metrics are for reader and collector but the features are more for modelling. 
We can move the feature parts to model package later.

e2e/healthcheck_test.go
```diff
@@ -4,23 +4,28 @@ import (
...
+
 var _ = Describe("healthz check should pass", func() {
 	Context("start server with health check", func() {
 		It("should work properly", func() {
 			address, ok := os.LookupEnv("kepler_address")
 			if !ok {
 				address = "localhost:8888"
 				cmd := exec.Command(keplerBin)
-				keplerSession, err = gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
+				keplerSession, err = gexec.Start(cmd, nil, nil)
```
marceloamaral, line 26: Why did you remove the GinkgoWriter?

cmd/exporter.go
```diff
@@ -51,6 +55,10 @@ func healthProbe(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func finalizer() {
+	fmt.Println(kFlag)
```
marceloamaral, line 59: We should not use `fmt` package to print anymore....
The PR #238 moved all prints and logs to `klog`, so for consistency we should use klog

cmd/exporter.go
```diff
@@ -73,13 +82,14 @@ func main() {
...
 	if modelServerEndpoint != nil {
-		model.SetModelServerEndpoint(*modelServerEndpoint)
+		config.SetModelServerEndpoint(*modelServerEndpoint)
 	}
 
 	newCollector, err := collector.New()
 	defer newCollector.Destroy()
 	defer rapl.StopPower()
 	if err != nil {
+		finalizer()
```
marceloamaral, line 92: Isn't the log enough? The log will already print a string....

In the case we need the function, the `finalizer()` looks like something that will run when the program exit, so we could defer as the first line of the main

e2e/healthcheck_test.go
```diff
@@ -4,23 +4,28 @@ import (
...
 				Expect(err).NotTo(HaveOccurred())
-				time.Sleep(5 * time.Second) // wait for server start up
+				Eventually(keplerSession.Out, 30, 5).Should(gbytes.Say(kFlag))
```
marceloamaral, line 28: I think we should write a comment here explaining why we need this
```golang
// verifying if kepler is still alive
```

Do we need "Eventually"? 
Or we can just sleep 30s and check 
```golang
Expect(KeplerSession.Out).ShouldNot(gbytes.Say(finalizerKeyWord))
```

cmd/exporter.go
marceloamaral: This also should be a more meaningful word, for example:
```golang
klog.InfoS("exiting...")
exitCode := 10
klog.FlushAndExit(klog.ExitFlushTimeout, exitCode)
```
As in the example of the library [here](https://pkg.go.dev/k8s.io/klog/v2#FlushAndExit).

e2e/healthcheck_test.go
```diff
@@ -4,23 +4,28 @@ import (
...
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 )
 
+const (
+	// kFlag is a keyword to check whether the healthcheck is ready (or process is died)
+	kFlag = "kflag"
```
marceloamaral, line 16:
```golang
finalizerKeyWord = "exiting"
```

e2e/healthcheck_test.go
sunya-ch: This is to speed up the other case. 
Otherwise all cases have to to wait for 30s without necessity.
If we use this periodic check, the other case can finish by just 5s.

e2e/healthcheck_test.go
sunya-ch: Actually, we could have the other wording better than finalizer because it includes the case that start listening too.

cmd/exporter.go
sunya-ch: Agree. Just does not come out with other function name.

e2e/healthcheck_test.go
marceloamaral: Start listening?

We should only print it when it fails with a fatal error, so that kepler will not be running

e2e/healthcheck_test.go
sunya-ch: Sorry I just didn't merge with the latest PR.

e2e/healthcheck_test.go
marceloamaral: So, there are two logics here

1. verify if the server started
2. verify if the server stopped running

In this case, we should have different strings

e2e/healthcheck_test.go
marceloamaral: This is true and that's why the health checking exists!

So we don't need the string to say kepler failed, but to say it has started... 
And then the test will try the health check to see if the kepler is still alive...

What we want then is a log that prints after the server is started

Anyway, as we discussed before let's correct the logic in another PR....

marceloamaral (approved):
This PR introduces a major design change... which is fine merging it now, but for future design changes we should have a design proposal before...


As we discussed before, to speed up the process let's merge this PR to enable other PRs get merged as well as...

We will fix the the issues in other PRs...

pkg/collector/node_energy.go
```diff
@@ -80,16 +135,49 @@ func (v *NodeEnergy) ResetCurr() {
...
+	}
+	return
+}
+
+func (v *NodeEnergy) SetValues(sensorEnergy map[string]float64, pkgEnergy map[int]source.RAPLEnergy, totalGPUDelta uint64, podMetricValues [][]float64) {
+	nodeUsageValues, nodeUsageMap := v.sumUsage(podMetricValues)
+	v.Usage = nodeUsageMap
 	for sensorID, energy := range sensorEnergy {
-		v.SensorEnergy.AddStat(sensorID, uint64(energy))
+		v.SensorEnergy.AddAggrStat(sensorID, uint64(energy))
```
marceloamaral, line 156: @sunya-ch 
The measured data added as AggrStat and the estimated as the CurrStat
So, we only estimate the current but measure the aggregation?
Is that consistent?

sunya-ch: Value from sensor is aggregated value. Value from estimation is current value (estimated from current value).

marceloamaral: Ok, thanks!
Maybe we should consider collect the CurrStat in both cases for consistency...

PR #233, opened 2022-09-23
acpi power sensor: check all the sensor file before using them

rootfs:


rootfs:
on two separate setup, the power meters are different
```console
# ls /sys/class/hwmon/hwmon2/device/power*_average
/sys/class/hwmon/hwmon2/device/power1_average
```

the other

rootfs:
this looks a regression from https://github.com/sustainable-computing-io/kepler/commit/75f60857a022410fb03f6488c6a137527cae785f#diff-01d4e5e0e8a39b2a514a24ebfd7524d6d692272ef2c75a615bd7435f1b097931L177

PR #234, opened 2022-09-23
acpi power sensordon't continue if the power meter doesn't exist

rootfs:
fix the regression in https://github.com/sustainable-computing-io/kepler/commit/75f60857a022410fb03f6488c6a137527cae785f#diff-01d4e5e0e8a39b2a514a24ebfd7524d6d692272ef2c75a615bd7435f1b097931L177


rootfs:
the loop break causes the [Fatalf](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/acpi/acpi.go#L90)

marceloamaral (approved):
Sorry, it was my mistake fixing the lint issues!
Thanks for fixing this

PR #235, opened 2022-09-23
Add support to deploy prometheus operator via make cluster-up

marceloamaral:
#### What this PR does / why we need it:
The current integration test tries to deploy the prometheus operator, but it doesn't work because the cluster doesn't have the docker images.
We don't see errors in the test because the test is not verifying if all containers were created successfully....

To deploy the prometheus operator on the kind cluster, we need to load the images into the cluster.

This PR adds that logic and wait all containers to be ready.
Also, I applied logic to deploy the minimum version of the prometheus operator to speed up the test execution.

After this PR anyone can create a kind cluster with prometheus using:
```
CLUSTER_PROVIDER="kind" PROMETHEUS_ENABLE="true" cluster-up
```

I don't think pushed a new image to quay should be part of an integration test.
We need to fix this in another PR.

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -16,25 +16,23 @@ jobs:
       - uses: actions/setup-go@main
         with:
           go-version: 1.18
+          
       - name: install kubectl
         run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+
       - name: start local k8s cluster
         run: make cluster-up
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
-      - name: Checkout kube-operator repo
-        uses: actions/checkout@v3
-        with:
-          repository: prometheus-operator/kube-prometheus
-          path: kube-prometheus
-      - name: deploy kube operator
-        run:  ls && kubectl create -f manifests/setup && until kubectl get servicemonitors --all-namespaces ; do date; sleep 1; echo ""; done && kubectl create -f manifests/
-        working-directory: kube-prometheus
+            PROMETHEUS_ENABLE: "true"
+            GRAFANA_ENABLE: "false"
+
       - name: simple test - deploy kepler
         run: make cluster-sync
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
             CTR_CMD: docker
+
       - name: Login to Quay
         if: github.event_name == 'push'
         uses: docker/login-action@v1
--- a/cluster-up/cluster/kind/common.sh
+++ b/cluster-up/cluster/kind/common.sh
@@ -22,6 +22,8 @@ set -ex pipefail
 _registry_port="5001"
 _registry_name="kind-registry"
 
+CTR_CMD=${CTR_CMD-docker}
+
 CONFIG_PATH="cluster-up/cluster"
 KIND_VERSION=${KIND_VERSION:-0.15.0}
 KIND_MANIFESTS_DIR="$CONFIG_PATH/${CLUSTER_PROVIDER}/manifests"
@@ -32,6 +34,11 @@ KIND_DEFAULT_NETWORK="kind"
 
 IMAGE_REPO=${IMAGE_REPO:-localhost:5001/kepler}
 IMAGE_TAG=${IMAGE_TAG:-devel}
+
+PROMETHEUS_OPERATOR_VERSION=${PROMETHEUS_OPERATOR_VERSION:-v0.11.0}
+PROMETHEUS_ENABLE=${PROMETHEUS_ENABLE:-false}
+PROMETHEUS_REPLICAS=${PROMETHEUS_REPLICAS:-1}
+GRAFANA_ENABLE=${GRAFANA_ENABLE:-false}
 
 CONFIG_OUT_DIR=${CONFIG_OUT_DIR:-"_output/manifests/${CLUSTER_PROVIDER}/generated"}
 rm -rf ${CONFIG_OUT_DIR}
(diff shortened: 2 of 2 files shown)


.github/workflows/integration_test.yml
```diff
@@ -16,37 +16,35 @@ jobs:
...
-        working-directory: kube-prometheus
+            PROMETHEUS_ENABLE: "true"
+            GRAFANA_ENABLE: "false"
+
       - name: simple test - deploy kepler
         run: make cluster-sync
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
             CTR_CMD: docker
-      - name: Login to Quay
```
rootfs, line 38: do you have another GH action to push the image?

cluster-up/cluster/kind/common.sh
```diff
@@ -55,10 +62,48 @@ aarch64* | arm64*)
...
+} 
+
+function _deploy_prometheus_operator {
+    git clone -b ${PROMETHEUS_OPERATOR_VERSION} --depth 1 https://github.com/prometheus-operator/kube-prometheus.git
+    sed -i -e "s/replicas: 2/replicas: ${PROMETHEUS_REPLICAS}/g" kube-prometheus/manifests/prometheus-prometheus.yaml
+    _load_prometheus_operator_images_to_local_registry
+    kubectl create -f kube-prometheus/manifests/setup
+    # wait k8s create resources
+    sleep 5
+    for file in $(ls kube-prometheus/manifests/prometheusOperator-*); do
```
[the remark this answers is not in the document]
marceloamaral: This is better! Will update!

.github/workflows/integration_test.yml
marceloamaral: I will uncomment this, but should it be the image-base.yml?
Push a new image shouldn’t be an action of integration test....

rootfs: sounds another PR :D

rootfs:
@marceloamaral @SamYuan1990 please follow up with the quay push GH action, thank you!

.github/workflows/integration_test.yml
```diff
@@ -16,25 +16,23 @@ jobs:
...
           go-version: 1.18
+          
       - name: install kubectl
         run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+
       - name: start local k8s cluster
         run: make cluster-up
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
-      - name: Checkout kube-operator repo
```
SamYuan1990, line 25: wait a min @marceloamaral , @rootfs 
1st of all, I don't mind we switch to prometheus operator instead of an all in one deployment.
but, some thing I wanna to make it clear.

> To deploy the prometheus operator on the kind cluster, we need to load the images into the cluster.

- [ ] Kind as a k8s cluster will download the image from docker registry no matter local or public right?
- [ ] if some pod waiting for download, via this pr, we should know ... and if kind not download the docker image, we are expecting the CI fails right?

with the my personal confusing points as above, I did a test as https://github.com/sustainable-computing-io/kepler/actions/runs/3118522667/jobs/5057872188
and it seems that if we remove the `_load_prometheus_operator_images_to_local_registry` function invoke, the kind cluster is still able to download the images and start ... correctly?

SamYuan1990: Hence, in this pr, do we have covered any check points to ensure we can see some/any error in test by verifying if all pods were created successfully?
> We don't see errors in the test because the test is not verifying if all containers were created successfully....

SamYuan1990:
> @marceloamaral @SamYuan1990 please follow up with the quay push GH action, thank you!

think twice before act
1. should we reuse image-base.yml by `use` keyword to avoid copy paste ourself?
2. meanwhile enhance image-base.yml to image.yaml and cover all image build process? 
3. if so, should we remove limitation in image-base.yaml from base branch?

.github/workflows/integration_test.yml
marceloamaral: >Kind as a k8s cluster will download the image from docker registry no matter local or public right?

No. You must push the images to the kind cluster....
I tried many different ways to enable kind download the image from public repository without success...:/

marceloamaral: > if some pod waiting for download, via this pr, we should know ... and if kind not download the docker image, we are expecting the CI fails right?

Yes! But the old integration test was not verifying if the container were successfully created, so we did not see any error...

marceloamaral: > with the my personal confusing points as above, I did a test as https://github.com/sustainable-computing-io/kepler/actions/runs/3118522667/jobs/5057872188
and it seems that if we remove the _load_prometheus_operator_images_to_local_registry function invoke, the kind cluster is still able to download the images and start ... correctly?

I made a comment in this PR :)

marceloamaral:
> should we reuse image-base.yml by use keyword to avoid copy paste ourself?
> meanwhile enhance image-base.yml to image.yaml and cover all image build process?
> if so, should we remove limitation in image-base.yaml from base branch?

@SamYuan1990 
Right, we should have another job to push the latest image, but only after a PR is merged, not during integration testing.

SamYuan1990:
> > should we reuse image-base.yml by use keyword to avoid copy paste ourself?
> > meanwhile enhance image-base.yml to image.yaml and cover all image build process?
> > if so, should we remove limitation in image-base.yaml from base branch?
> 
> @SamYuan1990 Right, we should have another job to push the latest image, but only after a PR is merged, not during integration testing.

one question, on this.
for push job triggered by pr merge back to default branch. why can we skip the test(as integration) and push image directly?

marceloamaral:
- We need the integration to validate the PR

- A PR will only be merged if all tests succeed 

- We will only publish a new image if the PR gets merged

SamYuan1990:
> * We need the integration to validate the PR
> * A PR will only be merged if all tests succeed
> * We will only publish a new image if the PR gets merged

in that case after merged integration can be remove.
why?
so far we push the image after integration test passed.(to ensure end user can follow same steps to deploy kepler and use)
the integration test can be covered by pr level?

marceloamaral:
> after merged integration

I see that as a double checking. But don't see the need for our use case right now...
There is a similar discussion [here](https://github.com/keptn/keptn/issues/2748).

> the integration test can be covered by pr level?

Yes, we will test the Kepler integration with the other components
If it fail, we should not merge the PR....

SamYuan1990:
> > after merged integration
> 
> I see that as a double checking. But don't see the need for our use case right now... There is a similar discussion [here](https://github.com/keptn/keptn/issues/2748).
> 
> > the integration test can be covered by pr level?
> 
> Yes, we will test the Kepler integration with the other components If it fail, we should not merge the PR....

I am worry about case below
pr A is bug free.
pr B is bug free.
A+B merge together will case a bug.
if we removed integration test before push.
Will bug as we merged A then B been skipped? or say will the merge for A trigger PR B rerun or not?

SamYuan1990:
as the [link](https://github.com/keptn/keptn/issues/2748) I do agree that even if we have an integration test at after merge phase. and it breaks CI, which is right. 
however, we can find bug hidden behind A+B at 1st time by CI break. (is CI break so un acceptable?)

Issue #212, opened 2022-09-20
log/println usage


jichenjc:
**Is your feature request related to a problem? Please describe.**

we use fmt.Println , log.Println etc in code

we should follow best practice to include log only 
and use different level such as Log.V(3) / V(4) etc to show different logs with different level

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

marceloamaral:
Right, this was in my TODO list...

It will be great to have different verbosity levels....

rootfs:
closed by #238

PR #238, opened 2022-09-24
Update log lib to klog to enable log verbosity support

marceloamaral:
There are a lot of prints currently in the code generating a lot of log data.
To fix this, we must use a different log verbosity.

Also, as indicated in #212, we should standardize the use of fmt and log prints to just one way.

This PR is moving all prints from fmt or log to klog.

Klog was developed by the Kubernetes community and has some advantages over glog as described [here](https://pkg.go.dev/k8s.io/klog).

Signed by: Marcelo Amaral <marcelo.amaral1@ibm.com>

### DIFF
--- a/cmd/exporter.go
+++ b/cmd/exporter.go
@@ -19,7 +19,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"log"
 	"net/http"
 
 	"github.com/sustainable-computing-io/kepler/pkg/collector"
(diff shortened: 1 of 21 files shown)


cmd/exporter.go
```diff
@@ -31,6 +30,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/version"
+
+	"k8s.io/klog/v2"
```
rootfs, line 34: does klog pkg exist in go.mod and vendor pkg?

marceloamaral: @rootfs updated!

rootfs (approved):

PR #239, opened 2022-09-24
[don't merge]CI:just have a try, if kind will upload images or not

SamYuan1990:

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -24,8 +24,15 @@ jobs:
         run: make cluster-up
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
-            PROMETHEUS_ENABLE: "true"
-            GRAFANA_ENABLE: "false"
+            
+      - name: Checkout kube-operator repo
+        uses: actions/checkout@v3
+        with:
+          repository: prometheus-operator/kube-prometheus
+          path: kube-prometheus
+      - name: deploy kube operator
+        run:  ls && kubectl create -f manifests/setup && until kubectl get servicemonitors --all-namespaces ; do date; sleep 1; echo ""; done && kubectl create -f manifests/
+        working-directory: kube-prometheus
 
       - name: simple test - deploy kepler
         run: make cluster-sync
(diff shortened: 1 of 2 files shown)


.github/workflows/integration_test.yml
```diff
@@ -24,8 +24,15 @@ jobs:
...
-            PROMETHEUS_ENABLE: "true"
-            GRAFANA_ENABLE: "false"
+            
+      - name: Checkout kube-operator repo
+        uses: actions/checkout@v3
+        with:
+          repository: prometheus-operator/kube-prometheus
+          path: kube-prometheus
+      - name: deploy kube operator
+        run:  ls && kubectl create -f manifests/setup && until kubectl get servicemonitors --all-namespaces ; do date; sleep 1; echo ""; done && kubectl create -f manifests/
```
marceloamaral, line 34: You need to check if the containers were created!

For example, you can add after `kubectl create -f manifests/`:
`; kubectl wait --for=condition=Ready pod --all -n monitoring --timeout 12m`

marceloamaral (changes requested):

.github/workflows/integration_test.yml
SamYuan1990: you mean, if not, kepler pod will success but namespace is monitoring may not ready?

marceloamaral: prometheus operator containers will not be ready

if we don't care if the prometheus operator containers are running we don't need to deploy it....

SamYuan1990: will if we have integration test case. them will test rest api for kepler. and should we need test output from promethes?

marceloamaral: Yes, that is why we need to make sure that the prometheus operator containers are running
As I am suggesting to do :)

marceloamaral: Note that, prometheus operator containers are created in the monitoring namespace, then we need to verify if those container a Ready

SamYuan1990: are we force kepler bind with prometheus?

> Yes, that is why we need to make sure that the prometheus operator containers are running As I am suggesting to do :)

marceloamaral: Did not understand your question...

Prometheus operator must successfully run

PR #243, opened 2022-09-26
update loglevel of some lines to make it always print

jichenjc:


marceloamaral (changes requested):
Can you please also remove the `-v=1` from the manifest?

rootfs (approved):

jichenjc:
Sorry for confusion, I forgot to update the manifests file in my test env.. thanks for @marceloamaral for point that out
so this PR is not needed

Issue #237, opened 2022-09-23
Need health checks


rootfs:
**Is your feature request related to a problem? Please describe.**
Kepler currently only exports metrics, it doesn't have other health checks to report internal status

**Describe the solution you'd like**
A health check endpoint to report internal component health status, including ebpf collector, model server/estimator, model accuracy

SamYuan1990:
I suppose this one can be closed?

PR #257, opened 2022-09-28
add healthy for exporter

jichenjc:
related https://github.com/sustainable-computing-io/kepler/issues/237

### DIFF
--- a/cmd/exporter.go
+++ b/cmd/exporter.go
@@ -42,6 +42,14 @@ var (
 	enabledEBPFCgroupID = flag.Bool("enable-cgroup-id", true, "whether enable eBPF to collect cgroup id (must have kernel version >= 4.18 and cGroup v2)")
 )
 
+func healthProbe(w http.ResponseWriter, req *http.Request) {
+	w.WriteHeader(http.StatusOK)
+	_, err := w.Write([]byte(`ok`))
+	if err != nil {
+		klog.Fatalf("%s", fmt.Sprintf("failed to write response: %v", err))
+	}
+}
+
 func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
@@ -80,6 +88,7 @@ func main() {
 	}
 
 	http.Handle(*metricsPath, promhttp.Handler())
+	http.HandleFunc("/healthz", healthProbe)
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		_, err = w.Write([]byte(`<html>
 			<head><title>Energy Stats Exporter</title></head>
--- a/manifests/kubernetes/deployment.yaml
+++ b/manifests/kubernetes/deployment.yaml
@@ -78,6 +78,16 @@ spec:
         - containerPort: 9102
           hostPort: 9102
           name: http
+        livenessProbe:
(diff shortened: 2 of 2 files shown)


cmd/exporter.go
```diff
@@ -42,6 +42,13 @@ var (
 	enabledEBPFCgroupID = flag.Bool("enable-cgroup-id", true, "whether enable eBPF to collect cgroup id (must have kernel version >= 4.18 and cGroup v2)")
 )
 
+func healthProbe(w http.ResponseWriter, req *http.Request) {
```
jichenjc, line 45: @rootfs @marceloamaral what do you think best way to probe exporter? check with the prometheus endpoint? Thanks

rootfs: healthz looks good to me. If the collector is not working, the metrics won't be there. 
Can you add a component level health status report and embed them in the HTTP response? In this way, we know the health status of attacher, collector, etc

rootfs: The initial PR can get the framework ready, followed by per component level health message construction

marceloamaral: So, is the idea to create a liveness probe configuration later?
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

jichenjc: yes, I am looking for a way to know how to detect it's a health pod 
e.g something is running so we can say it's healthy or not ..

the healthy setup has been done in the manifests already but not sure what to check now

marceloamaral: Just check if the http server replies anything, so it means that the container is running

jichenjc: @rootfs are you looking for something like following?
/healthz

status 200 
module1 : ok
module2 : fail
module3 : ok

rootfs: can we use any additional http status code? https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

e.g. if attacher is up, but collector fail, return 201. If some but not all metrics are available, return 206. All the codes are documented for diagnosis.

jichenjc: um... 

health works like following way

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
```
Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.
```
So I think return 201 or 206 make no difference 


and from other projects I worked before 
https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/main.go#L70
they have a sidecar (but seems kepler no need) and it check some other pods running well or through rpc
looks like we don't have such need as well

so guess the simplest way is to always return ok check here 
if there is something wrong in access http:<host>:9102/healthz , that means pod is not healthy (since <host>:9102/metrics will not work at all)

thoughts?

rootfs: ok, that makes more sense. It returns 4xx when the metrics collector fail, otherwise just return 200.

jichenjc:
remove the log , others same to what we have ,simple HTTP reqeust to /healthz

SamYuan1990 (approved):
LGTM

Issue #254, opened 2022-09-27
CI post action enhance


SamYuan1990:
**Is your feature request related to a problem? Please describe.**
- remove post integration test
- image build from base to kepler

**Describe the solution you'd like**
n/A

**Describe alternatives you've considered**
n/A

**Additional context**
n/A

SamYuan1990:
@marceloamaral , could you please add some reason in comments for remove post integration test checking?

marceloamaral:
I'm not saying that post-integration testing isn't useful, there may be some use cases that will be needed

However, for simplicity's sake for now, I think that our integration test should be run before the merge

SamYuan1990:
> I'm not saying that post-integration testing isn't useful, there may be some use cases that will be needed
> 
> However, for simplicity's sake for now, I think that our integration test should be run before the merge

I am going to create a pr for this and let's see.

SamYuan1990:
close for now as resolved by #267

Issue #192, opened 2022-09-17
CI needs more instrumentation


rootfs:
**Describe the bug**
CI is getting flaky
```
+ ./hack/cluster-deploy.sh
waiting for cluster-clean to finish
+ source cluster-up/common.sh
++ set -e
++ '[' kubernetes = kind ']'
Deploying manifests...
+ CLUSTER_PROVIDER=kubernetes
+ MANIFESTS_OUT_DIR=_output/manifests/kubernetes/generated
+ main pipefail
+ '[' '!' -d _output/manifests/kubernetes/generated ']'
+ echo 'Deploying manifests...'
+ kubectl apply -f _output/manifests/kubernetes/generated
namespace/kepler created
clusterrole.rbac.authorization.k8s.io/kepler-clusterrole created
clusterrolebinding.rbac.authorization.k8s.io/kepler-clusterrole-binding created
serviceaccount/kepler-sa created
daemonset.apps/kepler-exporter created
service/kepler-exporter created
servicemonitor.monitoring.coreos.com/kepler-exporter created
+ kubectl rollout status daemonset kepler-exporter -n kepler --timeout 60s
Waiting for daemon set "kepler-exporter" rollout to finish: 0 of 1 updated pods are available...
error: timed out waiting for the condition
make: *** [Makefile:165: cluster-sync] Error 1
Error: Process completed with exit code 2.
```
**To Reproduce**
Recent CI results [here](https://github.com/sustainable-computing-io/kepler/actions/runs/3071410017/jobs/4962047170#step:8:237)
**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

SamYuan1990:
https://github.com/sustainable-computing-io/kepler/actions/runs/3073381944/jobs/4965474260#step:8:233
I see latest CI passed, @rootfs  is this issue able to reproduce easily?

rootfs:
let's keep an eye on it and re-open this issue if it happens again.

marceloamaral:
Maybe we need a larger timeout?
Let's see if it happen again

rootfs:
https://github.com/sustainable-computing-io/kepler/pull/231#issuecomment-1254422128

marceloamaral:
Thanks @SamYuan1990!

What we need is the artifacts https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

[...]

PR #267, opened 2022-09-29
CI:update integration test with deployment check and push action enha…

SamYuan1990:
CI:update integration test with deployment check and push action enhancement

related with #192
related with #254 

### DIFF
--- a/hack/cluster-deploy.sh
+++ b/hack/cluster-deploy.sh
@@ -31,10 +31,29 @@ function main() {
 
     # Ignore errors because some clusters might not have prometheus operator
     kubectl apply -f ${MANIFESTS_OUT_DIR} || true
-    kubectl rollout status daemonset kepler-exporter -n kepler --timeout 60s
-
-    echo "Check the logs of the kepler-exporter with:"
-    echo "kubectl -n kepler logs daemonset.apps/kepler-exporter"
+    
+    # round for 3 times and each for 60s
+    # check if the rollout status is running
+    deploy_status=1
+    for i in 1 2 3
+    do
+        echo "check deployment status for round $i"
+        kubectl rollout status daemonset kepler-exporter -n kepler --timeout 60s
+        #check rollout status
+        if [ $? -eq 0 ]
+        then
+            deploy_status=0
+            break
+        fi
+    done 
+    # if deployment in error
+    if test $[deploy_status] -eq 1
+    then
+        echo "Check the status of the kepler-exporter"
+        kubectl -n kepler describe daemonset.apps/kepler-exporter
+        echo "Check the logs of the kepler-exporter"
+        kubectl -n kepler logs daemonset.apps/kepler-exporter
+    fi
 }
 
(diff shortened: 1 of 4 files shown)


SamYuan1990:
https://github.com/sustainable-computing-io/kepler/actions/runs/3152440537/jobs/5127657495#step:6:242
CI round test waiting looks good when success.

rootfs (approved):

marceloamaral:
@SamYuan1990 looks like the integration test is failing now, see [here](https://github.com/sustainable-computing-io/kepler/actions/runs/3155912869/jobs/5135076208)

SamYuan1990:
> @SamYuan1990 looks like the integration test is failing now, see [here](https://github.com/sustainable-computing-io/kepler/actions/runs/3155912869/jobs/5135076208)

I will try to have a fix.
the command with exit code 2 breaks the loop and it is not been catched.

hack/cluster-deploy.sh
```diff
@@ -31,10 +31,29 @@ function main() {
...
     kubectl apply -f ${MANIFESTS_OUT_DIR} || true
-    kubectl rollout status daemonset kepler-exporter -n kepler --timeout 60s
-
-    echo "Check the logs of the kepler-exporter with:"
-    echo "kubectl -n kepler logs daemonset.apps/kepler-exporter"
+    
+    # round for 3 times and each for 60s
+    # check if the rollout status is running
+    deploy_status=1
+    for i in 1 2 3
```
marceloamaral, line 38: @SamYuan1990 instead of check this 3 times, why not use a larger timeout?
180s?

hack/cluster-deploy.sh
```diff
@@ -31,10 +31,29 @@ function main() {
...
+        kubectl rollout status daemonset kepler-exporter -n kepler --timeout 60s
+        #check rollout status
+        if [ $? -eq 0 ]
+        then
+            deploy_status=0
+            break
+        fi
+    done 
+    # if deployment in error
+    if test $[deploy_status] -eq 1
```
marceloamaral, line 50: Why don't use the TRAP command?
So, if the error happens in any other place in the script, we will dump the logs

SamYuan1990: may I have a link for TRAP command for details?

hack/cluster-deploy.sh
SamYuan1990: will try to see any difference.

hack/cluster-deploy.sh
marceloamaral: --> [https://medium.com/@dirk.avery/the-bash-trap-trap-ce6083f36700](https://medium.com/@dirk.avery/the-bash-trap-trap-ce6083f36700)

hack/cluster-deploy.sh
marceloamaral: The different will be a more clean code :)

Issue #265, opened 2022-09-29
podName without ns?


jichenjc:
**Describe the bug**

we can have same name pod 
```
# kubectl get pods -A | grep counter
default       counter                                    1/1     Running   1 (47h ago)   2d13h
n1            counter                                    1/1     Running   0             8s

```
but https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/reader.go#L133
seems we only honor the podname , is this on purpose design? 
**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

rootfs:
nice catch! This needs another checking. In an [early implementation](https://github.com/sustainable-computing-io/kepler/blob/3887805f2a85848d4b75ece67e60ace1feab759c/pkg/collector/reader.go#L116-L120), the map is indexed by `podName-namespace`.

rootfs:
@jichenjc you like to stand up a fix and test case?

jichenjc:
yes, I can give a try

marceloamaral:
Nice catch, we need to use maybe the pod ID instead in the pod name in the structure 
Regarding the prometheus metric, it has the pod name and namespace labels

jichenjc:
I tried to add a PR and I think the best fit would be pod ID ,but looks like we are using [1] for containerID but I think it can't be used as we are looking for podID, do we have any reference that I can steadily get a pod ID? @rootfs @marceloamaral 
or I can actually use `namespace/podname` so it will anyway help as well (and maybe better due to smaller footprint than a long pod ID ..?



[1] https://github.com/sustainable-computing-io/kepler/blob/main/pkg/podlister/resolve_container.go#L168

rootfs:
@jichenjc `podName` is the map key, once the cGroupID/PID is resolve and `ppodName` is known, we don't need `containerID`.  If you use `namespace/podname`, that looks a good PodID. There is a length limit of namespace and pod name. In the very unfortunate case of long names, a SHA will be a good to go.

marceloamaral:
We could add the podID in the container info structure:
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/podlister/resolve_container.go#L141
And them, we can have a function to get the podID

jichenjc:
[...]

Issue #264, opened 2022-09-29
no collection of new deployed pod on VM based env?


jichenjc:
**Describe the bug**

I had following k8s , the nginx n1/n2 namespace pods are deployed after kepler daemonset start running 

```
# kubectl get pods -A
NAMESPACE     NAME                                       READY   STATUS    RESTARTS      AGE
....
kepler        kepler-exporter-66zfj                      1/1     Running   0             47m
kube-system   calico-kube-controllers-54c85446d4-pb549   1/1     Running   2 (46h ago)   8d
....
logging       fluentd-es-hlt4c                           1/1     Running   1 (46h ago)   2d12h
logging       kibana-568dfb5bf-bgrt9                     1/1     Running   1 (46h ago)   2d13h
monitoring    alertmanager-main-0                        2/2     Running   2 (46h ago)   8d
monitoring    alertmanager-main-1                        2/2     Running   2 (46h ago)   8d
monitoring    alertmanager-main-2                        2/2     Running   2 (46h ago)   8d
monitoring    blackbox-exporter-6d897d5f66-dtspw         3/3     Running   3 (46h ago)   8d
monitoring    grafana-59664d747b-tj274                   1/1     Running   1 (46h ago)   6d17h
monitoring    kube-state-metrics-776dd9bc5d-rx4bq        3/3     Running   3 (46h ago)   8d
monitoring    node-exporter-thrwp                        2/2     Running   2 (46h ago)   8d
monitoring    prometheus-adapter-867598997b-dphc9        1/1     Running   1 (46h ago)   8d
monitoring    prometheus-adapter-867598997b-kstng        1/1     Running   1 (46h ago)   8d
monitoring    prometheus-k8s-0                           2/2     Running   2 (46h ago)   8d
monitoring    prometheus-k8s-1                           2/2     Running   2 (46h ago)   8d
monitoring    prometheus-operator-65cdf86899-bd4rl       2/2     Running   2 (46h ago)   8d
n1            nginx-deployment-7fb96c846b-8tpmd          1/1     Running   0             13m
n1            nginx-deployment-7fb96c846b-g67pz          1/1     Running   0             13m
n2            nginx-deployment-7fb96c846b-74ctr          1/1     Running   0             13m
n2            nginx-deployment-7fb96c846b-q7hsl          1/1     Running   0             13m

```

note n1 and n2 are nginx I want to do other test.. but I can't see nginx repoted at all

```
[...]

PR #268, opened 2022-09-30
handle inactive pods with inactive threshold

sunya-ch:
This PR is to handle the issue https://github.com/sustainable-computing-io/kepler/issues/264.

Instead of deleting inactive pods at the reading point, we set the`maxInactivePods`. 
If the number of inactive pods is more than this threshold, we check the alive pod from the pod lister and remove the deleted pods from the podEnergy map.

Note that, need to update podName to podID with namespace later (reference: https://github.com/sustainable-computing-io/kepler/issues/265).


pkg/collector/reader.go
```diff
@@ -397,3 +394,16 @@ func getActiveCPUs(cpuTime *[C.CPU_VECTOR_SIZE]uint16) (activeCPUs []int32) {
 	}
 	return
 }
+
+// handleInactivePods
+func handleInactivePods(foundPod map[string]bool) {
+	numOfInactive := len(podEnergy) - len(foundPod)
+	if numOfInactive > maxInactivePods {
```
marceloamaral, line 401: I am wondering if the could:
- have counter, if the pod did not appear for 3-5 consecutive epochs we need to verify if the pod still exist 
- when we verify if the pod exist, the `podLister.ListPods()` could have the pod name label, so that it will minimize the overhead

sunya-ch: I think we do not have to check specific missing pod but rather compute if the number garbage pods is more than the acceptable threshold. When it more than the acceptable threshold, we just check the alive pod from ListPod and remove the deleted ones.

marceloamaral: The problem with that is because we will report information of pods that are not running anymore

This means that if we sum the energy of all reported pods in the prometheus metrics, it will show a value higher than what RAPL is reporting, because we are still reporting the energy of pods that does not exist anymore....

marceloamaral: Hum, but since the current metric is always reset, maybe this might not be a problem.....
Wdyt?

pkg/podlister/resolve_container.go
```diff
@@ -284,3 +283,14 @@ func extractPodContainerIDfromPath(path string) (string, error) {
 	}
 	return systemProcessName, fmt.Errorf("failed to find pod's container id")
 }
+
+// GetAlivePods returns pod list
+func GetAlivePods() map[string]bool {
+	pods, _ := podLister.ListPods()
```
rootfs, line 289: a bit paranoid, if the err actually happens, the `pods` is emtyp, and `alivePods` is empty. Then the caller may get the wrong impression and think all Pods are inactive.

pkg/collector/reader.go
rootfs: this could be a good stress test case to get some conclusions @jichenjc

pkg/podlister/resolve_container.go
sunya-ch: agree. just add error handling change.

rootfs:
tested this fix, it can show the pods that are started after kepler is running but those nginx pods that are running before kepler starts are still missing

marceloamaral:
@rootfs We could add in the podEnergy structure the pods that are running and have not had any updates in the counters, so that we report these pods with energy 0

PR #275, opened 2022-10-03
[WIP][don't merge] Dev:1st impl for integration test

SamYuan1990:
Details for this PR
e2e folder with end to end testing, will run with two models, base on environment `kepler_address`.
1. build kepler during test process and run it targets on local.
2. run test case targets on cluster after cluster port forwarding.

### DIFF
--- a/cmd/exporter.go
+++ b/cmd/exporter.go
@@ -77,10 +77,10 @@ func main() {
 	if err != nil {
 		klog.Fatalf("%s", fmt.Sprintf("failed to create collector: %v", err))
 	}
-	err = newCollector.Attach()
-	if err != nil {
+	_ = newCollector.Attach()
+	/*if err != nil {
 		klog.Fatalf("%s", fmt.Sprintf("failed to attach : %v", err))
-	}
+	}*/
 
 	err = prometheus.Register(newCollector)
 	if err != nil {
--- a/cmd/exporter.go
+++ b/cmd/exporter.go
@@ -28,7 +28,6 @@ import (
 	"github.com/sustainable-computing-io/kepler/pkg/power/rapl"
 
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/version"
 
 	"k8s.io/klog/v2"
@@ -77,7 +76,7 @@ func main() {
 	if err != nil {
 		klog.Fatalf("%s", fmt.Sprintf("failed to create collector: %v", err))
 	}
-	_ = newCollector.Attach()
+	//_ = newCollector.Attach()
 	/*if err != nil {
 		klog.Fatalf("%s", fmt.Sprintf("failed to attach : %v", err))
 	}*/
@@ -87,7 +86,7 @@ func main() {
 		klog.Fatalf("%s", fmt.Sprintf("failed to register collector: %v", err))
 	}
 
-	http.Handle(*metricsPath, promhttp.Handler())
+	//http.Handle(*metricsPath, promhttp.Handler())
 	http.HandleFunc("/healthz", healthProbe)
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		_, err = w.Write([]byte(`<html>
--- a/pkg/attacher/bcc_attacher_stub.go
+++ b/pkg/attacher/bcc_attacher_stub.go
@@ -19,10 +19,6 @@ limitations under the License.
 
 package attacher
 
-import (
-	"fmt"
-)
-
 const (
 	CPUCycleLable       = "cpu_cycles"
 	CPUInstructionLabel = "cpu_instr"
@@ -66,7 +62,7 @@ var (
 )
 
 func AttachBPFAssets() (*BpfModuleTables, error) {
-	return nil, fmt.Errorf("no bcc build tag")
+	return nil, nil
 }
 
 func DetachBPFModules(bpfModules *BpfModuleTables) {
--- a/pkg/collector/reader.go
+++ b/pkg/collector/reader.go
@@ -104,8 +104,12 @@ func (c *Collector) resetCurrValue() {
 
 // resetBPFTables reset BPF module's tables
 func (c *Collector) resetBPFTables() {
-	c.modules.Table.DeleteAll()
-	c.modules.TimeTable.DeleteAll()
+	if c.modules != nil && c.modules.Table != nil {
+		c.modules.Table.DeleteAll()
+	}
+	if c.modules != nil && c.modules.TimeTable != nil {
+		c.modules.TimeTable.DeleteAll()
+	}
 }
 
 // readBPFEvent reads BPF event and maps between pid/cgroupid and container/pod
--- a/pkg/power/acpi/util_linux.go
+++ b/pkg/power/acpi/util_linux.go
@@ -1,0 +1,94 @@
+//go:build linux
+// +build linux
+
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package acpi
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+func getCPUCoreFrequency() map[int32]uint64 {
+	files, err := os.ReadDir(freqPathDir)
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	ch := make(chan []uint64)
+	for i := 0; i < len(files); i++ {
+		go func(i uint64) {
+			path := fmt.Sprintf(freqPath, i)
+			data, err := os.ReadFile(path)
+			if err != nil {
+				return
+			}
+			if freq, err := strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64); err == nil {
+				ch <- []uint64{i, freq}
+				return
+			}
+			ch <- []uint64{i, 0}
+		}(uint64(i))
+	}
+
+	cpuCoreFrequency := map[int32]uint64{}
+	for i := 0; i < len(files); i++ {
+		select {
+		case val, ok := <-ch:
+			if ok {
+				cpuCoreFrequency[int32(val[0])] = val[1]
+			}
+		case <-time.After(1 * time.Minute):
+			klog.V(1).Infoln("timeout reading cpu core frequency files")
+		}
+	}
+
+	return cpuCoreFrequency
+}
+
+func (a *ACPI) IsPowerSupported() bool {
+	file := fmt.Sprintf(powerPath, 1)
+	_, err := os.ReadFile(file)
+	return err == nil
+}
+
+func getPowerFromSensor() (map[string]float64, error) {
+	power := map[string]float64{}
+
+	for i := int32(1); i <= numCPUS; i++ {
+		path := fmt.Sprintf(powerPath, i)
+		data, err := os.ReadFile(path)
+		if err != nil {
+			break
+		}
+		// currPower is in microWatt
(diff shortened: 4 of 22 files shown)


SamYuan1990:
@marceloamaral , if I am correct, in integration test, we will deploy kepler with port 9102 right? could you please help with error https://github.com/sustainable-computing-io/kepler/actions/runs/3174554723/jobs/5171489975#step:8:28

cmd/exporter.go
```diff
@@ -77,10 +77,14 @@ func main() {
 	if err != nil {
 		klog.Fatalf("%s", fmt.Sprintf("failed to create collector: %v", err))
 	}
-	err = newCollector.Attach()
-	if err != nil {
+	_ = newCollector.Attach()
+	/*
+		remove this part of error check for now, as it binds with bcc library
```
SamYuan1990, line 82: @rootfs , could you please suggest/advise?

SamYuan1990:
Sorry, for asking for help on this WIP pr.

cmd/exporter.go
```diff
@@ -77,10 +77,14 @@ func main() {
...
+		if instance wihout bcc, start will fail.
+		if err != nil {
 		klog.Fatalf("%s", fmt.Sprintf("failed to attach : %v", err))
```
rootfs, line 85: why this failed?

cmd/exporter.go
SamYuan1990: https://github.com/sustainable-computing-io/kepler/actions/workflows/unit_test_mac.yml
https://github.com/sustainable-computing-io/kepler/actions/workflows/unit_test.yml
for those two jobs we don't have bcc.

the reason in deep for the failure is about the reader logic.
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/reader.go#L106-L109
and some null pointer return by `pkg/attacher/bcc_attacher_stub.go` if bcc not installed.

I suppose there are two ways, either we make our code robust(for rest api handler part) or we make our code decouple with bcc(if we can't decouple with bcc, we force ourself bind with it.)

rootfs: @SamYuan1990 if this is doable, maybe have yet another bcc attacher stub implementation for integration test and in the stub function, return certain mock data
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/attacher/bcc_attacher_stub.go#L42-L52
e.g. 
```go
func (table *Table) Iter() *TableIterator {
	return &TableIterator{mock_data}
}

func (it *TableIterator) Next() bool {
	return false
}

func (it *TableIterator) Leaf() []byte {
	return mock_data
}
```

pkg/collector/reader.go
```diff
@@ -104,8 +104,12 @@ func (c *Collector) resetCurrValue() {
 
 // resetBPFTables reset BPF module's tables
 func (c *Collector) resetBPFTables() {
-	c.modules.Table.DeleteAll()
-	c.modules.TimeTable.DeleteAll()
+	if c.modules != nil && c.modules.Table != nil {
```
SamYuan1990, line 107: @rootfs , do you know where we init modules and tables?

pkg/power/acpi/util_linux.go
```diff
@@ -0,0 +1,94 @@
+//go:build linux
+// +build linux
```
SamYuan1990, line 2: @rootfs ,those path are linux specific, so we have to split at here for different device?

pkg/attacher/bcc_attacher_stub.go
```diff
@@ -66,7 +62,7 @@ var (
 )
 
 func AttachBPFAssets() (*BpfModuleTables, error) {
-	return nil, fmt.Errorf("no bcc build tag")
+	return nil, nil
```
SamYuan1990, line 65: @rootfs , here is the pre defined error for without bcc tag.

SamYuan1990:
@rootfs , @marceloamaral , it seems we got error at https://github.com/sustainable-computing-io/kepler/actions/runs/3180212481/jobs/5183521776#step:7:17 could you please advise?

SamYuan1990:
topic may related https://github.com/kubernetes-sigs/kind/search?q=failed+to+execute+portforward+in+network+namespace+%22host%22&type=issues

SamYuan1990:
> > topic may related https://github.com/kubernetes-sigs/kind/search?q=failed+to+execute+portforward+in+network+namespace+%22host%22&type=issues
> 
> Could [`ClusterIP: None`](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/kubernetes/deployment.yaml#L126) be causing trouble?

I give a try by a new commit, but seems not the root cause.
from the log, I suppose there some error happened during ip v4 ip v6 transform.

pkg/power/acpi/util_linux.go
rootfs: @marceloamaral any idea? maybe a stub here too?

pkg/collector/reader.go
rootfs: the tables are auto filled; the delete functions remove existing table entries, the ebpf program will fill the tables again.

BenTheElder:
> @rootfs , I am now try to force KIND use IPv4 to avoid:
> https://kind.sigs.k8s.io/docs/user/known-issues/#ipv6-port-forwarding

That issue is only relevant when you are asking kind to forward ports from the host to the node, which you don't appear to be using here.

When you are using the Kubernetes port forward API to tunnel traffic through the API server to a workload, that is completely unrelated.


AFAICT, the kubernetes-api-level port forward is failing to connect to localhost:9102 in the pod namespace with neither ipv4 nor ipv6 connections succeeding. It's not clear why, other than the connection is refused.

> E1004 08:02:47.046060 7990 portforward.go:406] an error occurred forwarding 9102 -> 9102: error forwarding port 9102 to pod 2b02221f9f4b01c499fabccd3d35d62e5591d6b1303d9[18](https://github.com/sustainable-computing-io/kepler/actions/runs/3180383090/jobs/5183888088#step:7:19)294587bd80d23e38f, uid : failed to execute portforward in network namespace "host": failed to connect to localhost:9102 inside namespace "2b022[21](https://github.com/sustainable-computing-io/kepler/actions/runs/3180383090/jobs/5183888088#step:7:22)f9f4b01c499fabccd3d35d62e5591d6b1303d918294587bd80d[23](https://github.com/sustainable-computing-io/kepler/actions/runs/3180383090/jobs/5183888088#step:7:24)e38f", IPv4: dial tcp4 1[27](https://github.com/sustainable-computing-io/kepler/actions/runs/3180383090/jobs/5183888088#step:7:28).0.0.1:9102: connect: connection refused IPv6 dial tcp6 [::1]:9102: connect: connection refused

Are we sure there's an application listening at 127.0.0.1:9102 when we start the port forward?

More discussion in https://github.com/kubernetes-sigs/kind/issues/2953

rootfs:
@BenTheElder Thank you Ben for the help! The application should be running and listening to the port (in [this](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/kubernetes/deployment.yaml#L119) Service).

BenTheElder:
port-forward forwards to pods (not services), it will attempt to open a connection from within the pod's namespace, the tunnel looks roughly like:

kubectl <=> API server <=> kubelet <=> CRI <=> socket opened within target pod. IIRC kubernetes will always do this on `localhost`. The specific implementation in this case will be containerd's.

My KIND co-maintainer aojea previously patched this, because it used to literally use `socat` and supported IPv4 only. The current implementation has "happy eyeballs" IPv4+IPv6 (because Go) IIRC.

I'm noticing that the pod seems to be configured in hostNetwork and looks like the process should be bound to 0.0.0.0 ...

I'm not sure if we / Kubernetes have tests for hostNetwork port forward, it's possible that's broken in KIND (or containerd), otherwise unless the app isn't listening yet or the wrong pod is selected, I'm at a loss ...

rootfs:
Good info @BenTheElder! So based on your feedback, a more appropriate CI deployment is to use pod network. @SamYuan1990 can you give that a try? Thanks!

BenTheElder:
Instead of kubectl port forward, you could try configuring a port forward in kind and use a matching nodePort, similar to https://kind.sigs.k8s.io/docs/user/ingress/

`kubectl port forward` is more of a debugging tool and is not the most robust, though i'd still expect it to work here 🤔 , but as a general rule I wouldn't use it for anything load bearing with any cluster, even if you use the API directly.

Something like:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- extraPortMappings:
  - containerPort: 30001
    hostPort: 30001
    protocol: TCP
```


Or actually since the pod appears(?) to be in hostNetwork and you have a single node:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- extraPortMappings:
  - containerPort: 9102
    hostPort: 9102
    protocol: TCP
```

> if I am right, github agent doesn't support ipv6 for now.
> https://github.com/actions/runner-images/issues/668

We run some of our CI in GitHub actions and lack of IPv6 is not an issue, the ipv6 port forward connect is a red-herring, it's portforward is trying both ipv4 and ipv6 and both are failing. You can use kind / kubernetes without ipv6, if anything optional ipv6 in KIND involved a _lot_ of fixing tests and Kubernetes, and then again in conjunction with the folks working on  dualstack.

SamYuan1990:
> Instead of kubectl port forward, you could try configuring a port forward in kind and use a matching nodePort, similar to https://kind.sigs.k8s.io/docs/user/ingress/
> 
> `kubectl port forward` is more of a debugging tool and is not the most robust, though i'd still expect it to work here 🤔 , but as a general rule I wouldn't use it for anything load bearing with any cluster, even if you use the API directly.
> 
> Something like:
> 
> ```yaml
> kind: Cluster
> apiVersion: kind.x-k8s.io/v1alpha4
> nodes:
> - extraPortMappings:
>   - containerPort: 30001
>     hostPort: 30001
>     protocol: TCP
> ```
> 
> Or actually since the pod appears(?) to be in hostNetwork and you have a single node:
> 
> ```yaml
> kind: Cluster
> apiVersion: kind.x-k8s.io/v1alpha4
> nodes:
> - extraPortMappings:
>   - containerPort: 9102
>     hostPort: 9102
> ```
> 
> I.E. adds `-p 127.0.0.1:9102:9102/TCP` to running the kind node.
> 
> > if I am right, github agent doesn't support ipv6 for now.
> > [actions/runner-images#668](https://github.com/actions/runner-images/issues/668)
> 
> We run some of our CI in GitHub actions and lack of IPv6 is not an issue, the ipv6 port forward connect is a red-herring, it's portforward is trying both ipv4 and ipv6 and both are failing. You can use kind / kubernetes without ipv6, if anything optional ipv6 in KIND involved a _lot_ of fixing tests and Kubernetes, and then again in conjunction with the folks working on dualstack.

let me try

SamYuan1990:
@rootfs , @marceloamaral , take a look at https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:13 pls

SamYuan1990:
some updates:
- tried with https://kind.sigs.k8s.io/docs/user/configuration/#extra-port-mappings but, `The Service "kepler-exporter" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'` should we use `ClusterIP` together with extra-port-mappings?
- fixed one panic error, but why pod in panic passed `kubectl rollout status daemonset kepler-exporter -n kepler --timeout 180s` ?  question here as the panic will make the service down, which breaks livenessProbe. 
- once panic been fixed, the port forward works. @_@

BenTheElder:
> The Service "kepler-exporter" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when type is 'ClusterIP' should we use ClusterIP together with extra-port-mappings?

nodePort can only be set on a service of type [NodePort](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport). shouldn't be related to the node level port mappings.

I think you can skip the service though and just map 9102 from the host to the kind node since the pod appears to be in hostNetwork.

marceloamaral:
I've been using port-forward with no problem on a kind cluster.
So probably the problem here was just related to the panic error.

The command `kubectl rollout status daemonset kepler-exporter -n kepler --timeout 180s` verifies if the container has been started, but will not check if the container failed shortly after starting....

Maybe we should sleep and then check for failed containers....

rootfs:
@SamYuan1990 now just a couple of lint error, ready to merge

SamYuan1990:
> @SamYuan1990 now just a couple of lint error, ready to merge

 no no no.... I am going to bed... and don't merge it for now. please take a look at #283 1st.

SamYuan1990:
close and move to #292 without issue. thanks guys.

Issue #282, opened 2022-10-07
make bcc logic part more robust


SamYuan1990:
we have an issue as we installed bcc at container, but if host vm/instance don't bcc lib installed, and bcc been used as volume mount of pod, the pod will exited as bcc not found.

>
        @rootfs , @marceloamaral , take a look at https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:13 pls

_Originally posted by @SamYuan1990 in https://github.com/sustainable-computing-io/kepler/issues/275#issuecomment-1271050426_

SamYuan1990:
```
Run kubectl logs $(kubectl -n kepler get pods -oname) -n kepler
I1007 03:00:18.[4](https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:5)9110[5](https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:6)       1 config.go:62] config EnabledEBPFCgroupID enabled:  true
I1007 03:00:18.491172       1 config.go:63] config getKernelVersion:  5.15
I1007 03:00:18.491193       1 config.go:[6](https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:7)[7](https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:8)] config set EnabledEBPFCgroupID to  false
modprobe: FATAL: Module kheaders not found in directory /lib/modules/5.15.0-1020-azure
chdir(/lib/modules/5.15.0-1020-azure/build): No such file or directory
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x[8](https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:9) pc=0x81ae56]

goroutine 1 [running]:
github.com/iovisor/gobpf/bcc.(*Module).Load(0x0, {0xd372f4, 0xc}, 0x2?, 0x2?, 0xc000434d60?)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/vendor/github.com/iovisor/gobpf/bcc/module.go:202 +0x36
github.com/iovisor/gobpf/bcc.(*Module).LoadTracepoint(...)
[...]

Issue #162, opened 2022-09-11
Enhance CI with basic integration test


SamYuan1990:
**Is your feature request related to a problem? Please describe.**
So far, we just verify by `until kubectl get svc --all-namespaces  ...`
to make it better, we'd better have a test suite and targeting on running service.

**Describe the solution you'd like**
as above.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rootfs:
The service readiness detection is to be supported in kepler-operator

sallyom:
this issue is folded into https://github.com/sustainable-computing-io/kepler/issues/350, closing here and moving conversation to  #350 tracked on the Project Board

PR #283, opened 2022-10-07
Add panic handling when attaching the bpf program to do not termine kepler

marceloamaral:
#### Why this PR is needed?
Kepler should work in the case that it could not attach a bpf program. We have the logic for this.
However, there are sometimes that it causes a panic and then terminates the program.

The PR #275 having this problem where the VM does not have the `Module kheaders`, details are [here](https://github.com/sustainable-computing-io/kepler/actions/runs/3201921603/jobs/5230372207#step:8:13).

This PR fix the issue #282

#### What this PR does?
Golang [recover](https://gobyexample.com/recover) function make it possible to recover from a panic.
Therefore, we use this when attaching the bpf program and instead of terminate the program we just log the error.

### DIFF
--- a/cmd/exporter.go
+++ b/cmd/exporter.go
@@ -79,7 +79,7 @@ func main() {
 	}
 	err = newCollector.Attach()
 	if err != nil {
-		klog.Fatalf("%s", fmt.Sprintf("failed to attach : %v", err))
+		klog.Infof("%s", fmt.Sprintf("failed to attach : %v", err))
 	}
 
 	err = prometheus.Register(newCollector)
--- a/pkg/attacher/bcc_attacher.go
+++ b/pkg/attacher/bcc_attacher.go
@@ -63,6 +63,11 @@ var (
 )
 
 func loadModule(objProg []byte, options []string) (*bpf.Module, error) {
+	defer func() {
+		if err := recover(); err != nil {
+			klog.Infof("failed to attach the bpf program:", err)
+		}
+	}()
 	m := bpf.NewModule(string(objProg), options)
 	// TODO make all entrypoints yaml-declarable
 	sched_switch, err := m.LoadTracepoint("sched_switch")
--- a/pkg/attacher/bcc_attacher.go
+++ b/pkg/attacher/bcc_attacher.go
@@ -62,13 +62,14 @@ var (
 	EnableCPUFreq = true
 )
 
-func loadModule(objProg []byte, options []string) (*bpf.Module, error) {
+func loadModule(objProg []byte, options []string) (m *bpf.Module, err error) {
 	defer func() {
-		if err := recover(); err != nil {
-			klog.Infof("failed to attach the bpf program:", err)
+		if r := recover(); r != nil {
+			err = fmt.Errorf("failed to attach the bpf program: %v", err)
+			klog.Infoln(err)
 		}
 	}()
-	m := bpf.NewModule(string(objProg), options)
+	m = bpf.NewModule(string(objProg), options)
 	// TODO make all entrypoints yaml-declarable
 	sched_switch, err := m.LoadTracepoint("sched_switch")
 	if err != nil {
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -30,3 +30,9 @@ jobs:
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
             CTR_CMD: docker
+
+      - name: test if kepler is still alive
+        run: |
+          sleep 60
+          kubectl logs $(kubectl -n kepler get pods -oname) -n kepler
+          kubectl get all -n kepler


SamYuan1990:
I suggest add parts below in CI to check pod status, and verify this `recover` works. (logically it should, but just in case.)
```      - name: get pod logs
        run: |
          sleep 60
          kubectl logs $(kubectl -n kepler get pods -oname) -n kepler
          kubectl get all -n kepler
```

marceloamaral:
@rootfs please, do not merge this PR
I added this logic to test if the kepler is still alive but it has a sleep of 60s, which will make the test very slow

After verifying this, I will remove this logic

SamYuan1990:
@marceloamaral , the pod https://github.com/sustainable-computing-io/kepler/actions/runs/3202720302/jobs/5231993693#step:7:28 is CrashLoopBackOff.

SamYuan1990:
Attach function need a recover too?
```
goroutine 1 [running]:
github.com/iovisor/gobpf/bcc.(*Module).TableId.func2(0x2705d70?, 0x9?)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/vendor/github.com/iovisor/gobpf/bcc/module.go:467 +0x19
github.com/iovisor/gobpf/bcc.(*Module).TableId(0x15979c0?, {0xd35109?, 0x0?})
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/vendor/github.com/iovisor/gobpf/bcc/module.go:467 +0x6a
github.com/sustainable-computing-io/kepler/pkg/attacher.AttachBPFAssets()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/attacher/bcc_attacher.go:128 +0x4e5
github.com/sustainable-computing-io/kepler/pkg/collector.(*Collector).Attach(0xc00040e2b0)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/collector/collector.go:104 +0x33
main.main()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/cmd/exporter.go:80 +0x40f
```

SamYuan1990:
@marceloamaral , I noticed one thing. as https://github.com/sustainable-computing-io/kepler/actions/runs/3204057772/jobs/5234890386#step:7:16
please let me know as we installed prometheus operator, will prometheus auto invoke some url for kepler pod?

as I am worried, just catch `Attach()` function will not fix all possible issues.

marceloamaral:
@SamYuan1990 Kepler is being successfully deployed now

The point with `recover()` is that we need to have the err variable defined in the function output, otherwise it will "loose" the content.

Anyway, I fixed that and it is working now! Please have a look, if you agree, I will remove the CI code.

marceloamaral:
> will prometheus auto invoke some url for kepler pod?

Yes, that is how Prometheus works. It will scrape metrics from applications.

SamYuan1990 (approved):
LGTM

SamYuan1990:
> @SamYuan1990 Kepler is being successfully deployed now
> 
> The point with `recover()` is that we need to have the err variable defined in the function output, otherwise it will "loose" the content.
> 
> Anyway, I fixed that and it is working now! Please have a look, if you agree, I will remove the CI code.

I hope you can keep the CI code.

cmd/exporter.go
```diff
@@ -79,7 +79,7 @@ func main() {
 	}
 	err = newCollector.Attach()
 	if err != nil {
-		klog.Fatalf("%s", fmt.Sprintf("failed to attach : %v", err))
+		klog.Infof("%s", fmt.Sprintf("failed to attach : %v", err))
```
SamYuan1990, line 82: @rootfs , but we may need to think twice before merge. 
for this pr, we can accept with info level for now, if bcc libs not found on host instance. for long term considering, 
Q: if host instance/vm without bcc then what should we do?
Q: should we provide any by default approach to ensure restful api works?

marceloamaral: The api will work if bpf program is not running. Which means that the code will execute without performance counters.

SamYuan1990:
I will re-check this pr at about 24 hours ago?
if this pr removed [WIP] without merge, then I will merge it. cc: @rootfs

marceloamaral:
@SamYuan1990 ok, let's keep the CI code.
In the case the test start to be slow, we should verify it again.

pkg/attacher/bcc_attacher.go
```diff
@@ -62,8 +62,14 @@ var (
 	EnableCPUFreq = true
 )
 
-func loadModule(objProg []byte, options []string) (*bpf.Module, error) {
-	m := bpf.NewModule(string(objProg), options)
+func loadModule(objProg []byte, options []string) (m *bpf.Module, err error) {
+	defer func() {
+		if r := recover(); r != nil {
```
rootfs, line 67: there is no more panic or fatalf, recover() won't triggered

cmd/exporter.go
rootfs: the collector loop will not be able to read anything if eBPF is not loaded. That'll bring more to debug. I suggest:
- add build flags for panic()/fatalf and only build with them for CI test
- investigate non-bcc alternatives in a separate stream if necessary

SamYuan1990: > the collector loop will not be able to read anything if eBPF is not loaded. That'll bring more to debug. I suggest:
> 
> * add build flags for panic()/fatalf and only build with `Infof` for CI test
> * investigate non-bcc alternatives in a separate stream if necessary

In my point of view, I suppose this PR is just a bug fix, and we can do further thing through #282 or a new ticket.
I am not sure if #282 , related with #192  and blocks #162 as the pod in CI will not in running status before this fix.

And once again, I hope we can make kepler running in MVC pattern for restapi.
which 
- return number of 0/-1/-NaN instead of a panic as restful response.
- use adopter/strategy pattern adopt with different eBPF libraries and CPU architectures.

or we have an agreement that if kepler running on vm/instance which fails with eBPF loading there should be a panic?
if so, which means for CI we may have to install bcc at ubuntu based github action agent.

pkg/attacher/bcc_attacher.go
SamYuan1990: well, the 1st panic real happens at loadModule function, and the `m := bpf.NewModule(string(objProg), options)` real good as it init variable of `m`. As a hot fix, I do agree to make it as it is.
or maybe we'd better give some comments in line around or put init variable of m in creation/init function when controller been created.

marceloamaral: @rootfs Sorry I did not understand your comment...
We do have a panic happening in the case there is any problem with bpf attach
But Kepler should not stop to run if we don't have bpf

SamYuan1990:
@rootfs are we good enough to merge this PR?

PR #284, opened 2022-10-07
set config values from configmap or environment vars

sunya-ch:
This PR add a function `getConfig` to set configuration values from ConfigMap file or environment variable.

Working in a similar way to this python code in kepler-model-server.

```py
def getConfig(key, default):
    # check configmap path
    file = os.path.join(CONFIG_PATH, key)
    if os.path.exists(file):
        with open(file, "r") as f:
            return f.read()
    # check env
    return os.getenv(key, default)
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 
@@ -44,18 +45,40 @@ type config struct {
 
 var c config
 
+const (
+	defaultMetricValue = ""
+)
+
 var (
 	EnabledEBPFCgroupID = false
 
-	EstimatorModel        = "" // auto-select
-	EstimatorSelectFilter = "" // no filter
-	CoreUsageMetric       = "curr_cpu_cycles"
-	DRAMUsageMetric       = "curr_cache_miss"
-	UncoreUsageMetric     = ""                // no metric (evenly divided)
-	GeneralUsageMetric    = "curr_cpu_cycles" // for uncategorized energy; pkg - core - uncore
+	EstimatorModel        = getConfig("ESTIMATOR_MODEL", defaultMetricValue)         // auto-select
+	EstimatorSelectFilter = getConfig("ESTIMATOR_SELECT_FILTER", defaultMetricValue) // no filter
+	CoreUsageMetric       = getConfig("CORE_USAGE_METRIC", CPUInstruction)
+	DRAMUsageMetric       = getConfig("DRAM_USAGE_METRIC", CacheMiss)
+	UncoreUsageMetric     = getConfig("UNCORE_USAGE_METRIC", defaultMetricValue) // no metric (evenly divided)
+	GeneralUsageMetric    = getConfig("GENERAL_USAGE_METRIC", CPUInstruction)    // for uncategorized energy; pkg - core - uncore
 
 	versionRegex = regexp.MustCompile(`^(\d+)\.(\d+).`)
+
+	configPath = "/etc/config"
 )
+
+func getConfig(configKey, defaultValue string) (result string) {
+	result = string([]byte(defaultValue))
+	key := string([]byte(configKey))
+	configFile := filepath.Join(configPath, key)
+	value, err := os.ReadFile(configFile)
+	if err == nil {
+		result = bytes.NewBuffer(value).String()
+	} else {
+		strValue, present := os.LookupEnv(key)
+		if present {
+			result = strValue
+		}
+	}
+	return
+}
 
 // EnableEBPFCgroupID enables the eBPF code to collect cgroup id if the system has kernel version > 4.18
 func EnableEBPFCgroupID(enabled bool) {
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+const (
+	// counter - attacher package
+	CPUCycle       = "cpu_cycles"
+	CPUInstruction = "cpu_instr"
+	CacheMiss      = "cache_miss"
+	// bpf - attacher package
+	CPUTime = "cpu_time"
+	// cgroup - cgroup package
+	CgroupfsMemory       = "cgroupfs_memory_usage_bytes"
+	CgroupfsKernelMemory = "cgroupfs_kernel_memory_usage_bytes"
+	CgroupfsTCPMemory    = "cgroupfs_tcp_memory_usage_bytes"
+	CgroupfsCPU          = "cgroupfs_cpu_usage_us"
+	CgroupfsSystemCPU    = "cgroupfs_system_cpu_usage_us"
+	CgroupfsUserCPU      = "cgroupfs_user_cpu_usage_us"
+	CgroupfsReadIO       = "cgroupfs_ioread_bytes"
+	CgroupfsWriteIO      = "cgroupfs_iowrite_bytes"
+	BytesReadIO          = "bytes_read"
+	BytesWriteIO         = "bytes_writes"
+	BlockDevicesIO       = "block_devices_used"
+	// kubelet - podlister package
+	KubeletContainerCPU    = "container_cpu_usage_seconds_total"
+	KubeletContainerMemory = "container_memory_working_set_bytes"
+	KubeletNodeCPU         = "node_cpu_usage_seconds_total"
+	KubeletNodeMemory      = "node_memory_working_set_bytes"
+
+	// system
+	CPUFrequency = "avg_cpu_frequency"
+)
--- a/pkg/attacher/bcc_attacher.go
+++ b/pkg/attacher/bcc_attacher.go
@@ -48,9 +48,9 @@ type BpfModuleTables struct {
 }
 
 const (
-	CPUCycleLable       = "cpu_cycles"
-	CPUInstructionLabel = "cpu_instr"
-	CacheMissLabel      = "cache_miss"
+	CPUCycleLable       = config.CPUCycle
+	CPUInstructionLabel = config.CPUInstruction
+	CacheMissLabel      = config.CacheMiss
 )
 
 var (
--- a/pkg/cgroup/stat_reader.go
+++ b/pkg/cgroup/stat_reader.go
@@ -18,6 +18,8 @@ package cgroup
 
 import (
 	"path/filepath"
+
+	"github.com/sustainable-computing-io/kepler/pkg/config"
 )
 
 var MemUsageFiles = []string{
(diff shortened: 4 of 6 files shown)


pkg/config/config.go
```diff
@@ -47,16 +47,31 @@ var c config
...
 
-	EstimatorModel        = "" // auto-select
-	EstimatorSelectFilter = "" // no filter
-	CoreUsageMetric       = "curr_cpu_cycles"
-	DRAMUsageMetric       = "curr_cache_miss"
-	UncoreUsageMetric     = ""                // no metric (evenly divided)
-	GeneralUsageMetric    = "curr_cpu_cycles" // for uncategorized energy; pkg - core - uncore
+	EstimatorModel        = getConfig("EstimatorModel", "")        // auto-select
+	EstimatorSelectFilter = getConfig("EstimatorSelectFilter", "") // no filter
+	CoreUsageMetric       = getConfig("CoreUsageMetric", "cpu_instr")
```
marceloamaral, line 52: The names of core, dram, and instruction metrics are defined in [bcc_attacher.go](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/attacher/bcc_attacher.go#L51).

So I think the default values should be those of the [bcc_attacher.go](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/attacher/bcc_attacher.go#L51) instead of be set again here.

marceloamaral (changes requested):
Looks good. 

I am just suggesting a small update.

pkg/config/config.go
sunya-ch: @marceloamaral  Actually, I am trying not to import any relative module to config module because it might cause the loop import error. attacher will call config module to get the default config. 
We might also separate all variable definitions to another file and let every module call it.

marceloamaral: Right!

What about move the constants to the metric file. 
So that we keep all metric names in only one place.

```
const (
	CPUCycleLable       = "cpu_cycles"
	CPUInstructionLabel = "cpu_instr"
	CacheMissLabel      = "cache_miss"
)
```

pkg/config/config.go
```diff
@@ -47,16 +47,31 @@ var c config
...
+	UncoreUsageMetric     = getConfig("UncoreUsageMetric", "")           // no metric (evenly divided)
+	GeneralUsageMetric    = getConfig("GeneralUsageMetric", "cpu_instr") // for uncategorized energy; pkg - core - uncore
 
 	versionRegex = regexp.MustCompile(`^(\d+)\.(\d+).`)
+
+	ConfigPath = "/etc/config"
 )
 
+func getConfig(key, defaultValue string) string {
+	configFile := fmt.Sprintf("%s/%s", ConfigPath, key)
```
jichenjc, line 63: @sunya-ch @marceloamaral if I read this correctly, we are actually expecting multiple files for multiple metrics
is this work as design? should we create a config file and read KV from it ? I think it's more scale?

pkg/config/config.go
```diff
@@ -47,16 +47,31 @@ var c config
...
+	value, err := os.ReadFile(configFile)
+	if err == nil {
+		return string(value)
+	}
+	strValue, present := os.LookupEnv(key)
```
jichenjc, line 68: is this a best practice that use e.g `DRAMUsageMetric` as ENV VAR instead of something like `DRAM_USAGE_METRIC` ?

pkg/config/config.go
marceloamaral: Good point! I think it should be only one config file

pkg/config/config.go
marceloamaral: I also think that env variables should be `DRAM_USAGE_METRIC`

pkg/config/config.go
sunya-ch: When we set the ConfigMap CR of Kubernetes, 

```yaml
data:
  key1: val1
  key2: val2
```
It will automatically create two files.
<mount_path>/key1, <mount_path>/key2
Content is the value for each key.

If we want them to be in the a single file, we have to create CR like this
```yaml
data: |
   key1=var1
   key2=var2
```
It will create single file with the below content.
```
key1=var1
key2=var2
```

The first type is configmap by design of kubernetes. We can also map only subset of configuration values to the pod by key specification, for example, in case the case that we have a shared configuration between model-server and kepler.

pkg/config/config.go
sunya-ch: I also think that all-capital should be more common for environment variable. 
I just change it to the same format as go variables for simplicity as golint asks me to change the name to camel case. 
Anyway, I will change key string back to all-capital case.

pkg/config/config.go
sunya-ch: @marceloamaral I just added `pkg/config/types.go`. Please kindly confirm.

pkg/config/config.go
sunya-ch: Made an update (to all-capital case) here.

https://github.com/sustainable-computing-io/kepler/pull/284/commits/0c113ad1be69cc6422f419e6ff8c085ebdbdb7e3#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5

pkg/config/config.go
```diff
@@ -47,16 +47,31 @@ var c config
...
+	DRAMUsageMetric       = getConfig("DRAM_USAGE_METRIC", CacheMiss)
+	UncoreUsageMetric     = getConfig("UNCORE_USAGE_METRIC", "")              // no metric (evenly divided)
+	GeneralUsageMetric    = getConfig("GENERAL_USAGE_METRIC", CPUInstruction) // for uncategorized energy; pkg - core - uncore
 
 	versionRegex = regexp.MustCompile(`^(\d+)\.(\d+).`)
+
+	ConfigPath = "/etc/config"
 )
 
+func getConfig(key, defaultValue string) string {
```
marceloamaral, line 62: Could you also try to fix these possible "memory leaking" issues?

pkg/config/config.go
```diff
@@ -47,16 +47,31 @@ var c config
...
+	configFile := fmt.Sprintf("%s/%s", ConfigPath, key)
```
marceloamaral, line 63: Same "memory leaking" issue here

pkg/config/types.go
```diff
@@ -0,0 +1,46 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+const (
+	// counter - attacher package
+	CPUCycle       = "cpu_cycles"
```
marceloamaral, line 21: The names of the cgroup metrics start with `cgroup`, which makes the origin of the metric very transparent.

So, could we add the `kubelet_` prefix to the kubelet metrics and maybe the `hardware_counter_` prefix to its metrics too?
Note that, I prefer hardware_counter_ instead of bpf_ because if we change the method to collect the hardware counter, e.g., using perf, the prefix name will still make sense....

On the other hand, cpu time is specific to bpf, so I think the metric source prefix should be `bpf_`. Note that, I think this metric will be deprecated later since cgroup also has the cpu time...

pkg/config/types.go
```diff
@@ -0,0 +1,46 @@
...
+	CPUInstruction = "cpu_instr"
+	CacheMiss      = "cache_miss"
+	// bpf - attacher package
+	CPUTime = "cpu_time"
+	// cgroup - cgroup package
+	CgroupfsMemory       = "cgroupfs_memory_usage_bytes"
```
marceloamaral, line 27: Can we have the metric source prefix only `cgroup` instead of `cgrpupfs`?
wdyt?

pkg/config/types.go
```diff
@@ -0,0 +1,46 @@
...
+	BytesWriteIO         = "bytes_writes"
+	BlockDevicesIO       = "block_devices_used"
+	// kubelet - podlister package
+	KubeletContainerCPU    = "container_cpu_usage_seconds_total"
+	KubeletContainerMemory = "container_memory_working_set_bytes"
+	KubeletNodeCPU         = "node_cpu_usage_seconds_total"
+	KubeletNodeMemory      = "node_memory_working_set_bytes"
+
+	// system
+	CPUFrequency = "avg_cpu_frequency"
```
marceloamaral, line 45: Not sure what is the best metric source prefix for cpu frequency
But this metric comes from the OS, so maybe just `os_`

pkg/config/types.go
```diff
@@ -0,0 +1,46 @@
...
+	// cgroup - cgroup package
+	CgroupfsMemory       = "cgroupfs_memory_usage_bytes"
+	CgroupfsKernelMemory = "cgroupfs_kernel_memory_usage_bytes"
+	CgroupfsTCPMemory    = "cgroupfs_tcp_memory_usage_bytes"
+	CgroupfsCPU          = "cgroupfs_cpu_usage_us"
+	CgroupfsSystemCPU    = "cgroupfs_system_cpu_usage_us"
+	CgroupfsUserCPU      = "cgroupfs_user_cpu_usage_us"
+	CgroupfsReadIO       = "cgroupfs_ioread_bytes"
+	CgroupfsWriteIO      = "cgroupfs_iowrite_bytes"
+	BytesReadIO          = "bytes_read"
```
marceloamaral, line 35: Aren't the `bytes_read`, `bytes_writes` and `block_devices_used` also from cgroup?

pkg/config/types.go
sunya-ch: @marceloamaral  
Changing the metric name affects the other components ( e.g., online trainer of kepler-model-server ) and also has a different purpose from this PR. 
Let's make renaming list to a single note and do the metric renaming in the different PR.

pkg/config/types.go
sunya-ch: save as above. Let's rename all of these in other PR.

pkg/config/types.go
sunya-ch: save as above. Let's rename all of these in other PR.

pkg/config/types.go
sunya-ch: save as above. Let's rename all of these in other PR.

pkg/config/config.go
sunya-ch: Do you have any guidelines how to fix it?

pkg/config/types.go
marceloamaral: Ok, sounds good!

pkg/config/types.go
marceloamaral: Ok, let's keep tracking this!

pkg/config/config.go
marceloamaral: Not sure, but I think you need to check if the variable is null....

rootfs (approved):

rootfs:
@marceloamaral ready to merge?

pkg/config/config.go
sunya-ch: I think it is the problem of string handling.
https://go101.org/article/memory-leaking.html
Also, there are so many errors in the action log that is not reported...
My guess is that it reported only the first 10 rows. 
Screenshot 2022-10-13 at 09 30 09


Why don't we make it as known issue in the code for now and solve it later when we know the exact solution? 
I made multiple attempts but don't want to spam the github action flow for this anymore :sweat_smile:

marceloamaral: Ok, I investigated the problem a little more.

The scape problem is more related to performance and unnecessary memory allocation.

There are many different contexts where the problem can appear, but here it is related to memory being moved to the heap (scape to heap memory).

This means that after `getConfig` returns, its internal variables are deleted, but the function using `getConfig` will use that variable that no longer exists, and then go will move that value to the heap so that the value is still can be accessed.... More references [here](https://medium.com/a-journey-with-go/go-introduction-to-the-escape-analysis-f7610174e890) and [here](https : //dev.to/human/escape-analysis-in-go-part-1-42bd).

So in practice, most likely the problem is that the `configKey, defaultValue string` values ​​are being used for the `result string` in `getConfig`. However, `configKey, defaultValue` are not being allocated, they are variables or conts.


I think if you create constants it will fix this problem.

```
const (
    	defaultEstimatorModel        =  "ESTIMATOR_MODEL"
	defaultEstimatorSelectFilter = "ESTIMATOR_SELECT_FILTER"
	defaultCoreUsageMetric       = "CORE_USAGE_METRIC"
	defaultDRAMUsageMetric       = "DRAM_USAGE_METRIC"
	defaultUncoreUsageMetric     = "UNCORE_USAGE_METRIC"
	defaultGeneralUsageMetric    = "GENERAL_USAGE_METRIC"
        defaultMetricValue = ""
)

var (
	EnabledEBPFCgroupID = false

	EstimatorModel        = getConfig(defaultEstimatorModel, defaultMetricValue)         // auto-select
	EstimatorSelectFilter = getConfig(defaultEstimatorSelectFilter, defaultMetricValue) // no filter
	CoreUsageMetric       = getConfig(defaultCoreUsageMetric, CPUInstruction)
	DRAMUsageMetric       = getConfig(defaultDRAMUsageMetric, CacheMiss)
	UncoreUsageMetric     = getConfig(defaultUncoreUsageMetric, defaultMetricValue)              // no metric (evenly divided)
	GeneralUsageMetric    = getConfig(defaultGeneralUsageMetric, CPUInstruction) // for uncategorized energy; pkg - core - uncore
)
```

Btw, you can test it locally in your machine and don't need to verify in github...
`make escapes_detect`

sunya-ch: Tried that in local but does not work...
learking param is still there even I defined those all constants.

sunya-ch: One significant I pushed before is to use string(([]byte)(configKey)) assigning new variable in the function. 
It removes the leaking param report as shown here. 
```
Error: pkg/config/config.go:63:16: configKey does not escape
Error: pkg/config/config.go:63:27: defaultValue does not escape
```
https://github.com/sustainable-computing-io/kepler/actions/runs/3238779981

However, I don't think it is the point to do so.

marceloamaral: Right, the problem is because we are using global variable context.....
So it is not creating a local variable which the reference will be use in the `getConfig`

I tried many different solutions without any success and I think your solution is good

```
func getConfig(configKey, defaultValue string) (result string) {
	// making local copies of the global variable parameters to prevent memory leaking
	key := string(([]byte)(configKey))
	result = string(([]byte)(defaultValue))
	configFile := filepath.Join(configPath, key)
	value, err := os.ReadFile(configFile)
	if err == nil {
		result = bytes.NewBuffer(value).String()
	} else {
		strValue, present := os.LookupEnv(key)
		if present {
			result = strValue
		}
	}
	return
}
```

marceloamaral: Note that the real problem is the `leaking param: configKey`. This is not something that we should ignore...

Another solution is using pointers, and then we will have the message `moved to heap`, which is not optimal but is ok...

sunya-ch: Got it. Made the changes.

rootfs:
lgtm
@marceloamaral any more change needed?

marceloamaral (approved):
/lgtm

Issue #286, opened 2022-10-07
Prometheus metrics are not conforming with the Prometheus metrics guideline


marceloamaral:
**Is your feature request related to a problem? Please describe.**
Currently the prometheus metrics are not conforming with the [Prometheus metrics guideline](https://prometheus.io/docs/practices/naming/), and more details [here](https://prometheus.io/docs/instrumenting/writing_exporters/).

The key rules that we are violating are:
- should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers)
- should have a suffix describing the unit, in plural form
- `_total` is a convention for counters, you should use it if you’re using the COUNTER type.
- Prefer exporting a counter of joules, then rate(joules[5m]) gives you power in Watts.
- metric names should allow someone who is familiar with Prometheus but not a particular system to make a good guess as to what a metric means.
- metric names should be avoided to be procedurally generated
- metric names for applications should generally be prefixed by the exporter name, e.g. `kepler_`
- metric should make sense when summed or averaged, e.g. we should not have package energy with core energy
- use labels to differentiate the characteristics, that is, they should not contain values, e.g., cpu instructions

**Describe the solution you'd like**
Replace the old metric names with new metric ones.
Report energy per container instead of per pod. The user can aggregate the container energy in the promQL.
For the sake of compatibility with other modules, we should also keep some deprecated metrics and remove it later.
The new metrics should be:
```
//  new metrics
kepler_node_info{cpu_architecture}
kepler_node_core_joules_total{package, instance, source="rapl"}
kepler_node_uncore_joules_total{package, instance, source="rapl"}
kepler_node_dram_joules_total{package, instance, source="rapl"}
kepler_node_package_joules_total{package, instance, source="rapl"}
kepler_node_platform_joules_total{instance="node1", source="hwmon"}
kepler_node_other_host_components_joules_total{instance="node1"}
kepler_node_gpu_joules_total{index, instance, source="nvidia"}
kepler_node_cpu_scaling_frequency_hertz{instance, core}

kepler_container_core_joules_total{pod_name="", container_name="", container_namespace="", command=""}
kepler_container_uncore_joules_total{pod_name="", container_name="", container_namespace="", command=""}
kepler_container_dram_joules_total{pod_name="", container_name="", container_namespace="", command=""}
kepler_container_package_joules_total{pod_name="", container_name="", container_namespace="", command=""}
[...]

PR #287, opened 2022-10-07
Make kepler metrics conform to the Prometheus metrics guideline

marceloamaral:
**Why this PR is needed?**
Currently it is hard to understand all the prometheus metrics and even know which are the metrics that we are exporting...
The metric naming is complex and does not follow the prometheus metric name guideline.
More details are written in the issue #286

**What this PR does?**
This PR updates the prometheus metrics along with some changes to enable the new metrics.

Additionally, Prometheus suggest to only report metrics in joules instead of watts. Giving that, we don't need to report the current power consumption since it can be calculated using promQL. There are more details about this in the issue #286

For the sake of compatibility with other modules, we keep some deprecated metrics and will remove it later.

**Additional comments**
The changes are carefully separated in different commits to be easier to review.

I will update the Grafana dashboard in another PR. There are already many updates in this PR....


rootfs:
@sustainable-computing-io/reviewer

rootfs:
The metric change will also impact dashboard @williamcaban @Feelas @sallyom

marceloamaral:
Also, note that the goal is to update the metrics again later, we should export container energy consumption and not pod. Like `container_cpu_usage_seconds_total`, but we will have `kepler_container_package_joules_total`...
Did not use the container name right now because it will require much more changes in the code...

marceloamaral:
Moved this to draft, since we should report metrics per container not per pod
And then, the metric will have the pod label, for example the `container_cpu_usage_seconds_total`, but we will have `kepler_container_package_joules_total`

pkg/collector/prometheus.go
```diff
@@ -0,0 +1,583 @@
...
+				)
+			}
+		}
+	}()
+}
+
+// updatePodMetrics send pod metrics to prometheus
+func (c *PrometheusCollector) UpdatePodMetrics(wg *sync.WaitGroup, ch chan<- prometheus.Metric) {
+	for _, pod := range podEnergy {
+		wg.Add(1)
```
SamYuan1990, line 479: We are going to have pod*goroutines below added into wg?
any chance to make wg added by pod?

pkg/collector/prometheus.go
```diff
@@ -0,0 +1,583 @@
...
+		podCPUInstrTotal:              podCPUInstrTotal,
+		podCacheMissTotal:             podCacheMissTotal,
+		podCPUTime:                    podCPUTime,
+		podEnergyStat:                 podEnergyStat, // deprecated
+	}
+}
+
+// Collect implements the prometheus.Collector interface
+func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
+	lock.Lock()
```
SamYuan1990, line 348: do we still need this lock?

it 1st time of request will be blocked by wg until it finished.
so here comes with the 2nd request.
will the 2nd one waiting until 1st return? or it needs a lock for waiting 1st completed?

pkg/collector/util.go
```diff
@@ -0,0 +1,24 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collector
+
+func Min(x, y int) int {
```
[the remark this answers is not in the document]
marceloamaral: I will remove it, but the problem is that the `math.Min` only works with float, and then we need to do a lot of conversions
`int(math.Min(float64(len(container.Command)), 10))`

pkg/collector/prometheus.go
```diff
@@ -0,0 +1,583 @@
...
+
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+)
+
+const (
+	namespace        = "kepler"
+	miliJouleToJoule = 1000
+)
+
+var (
```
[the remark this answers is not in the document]
marceloamaral: I think it is better to keep it here right now
But we can move variables and constants arounds in another PR

pkg/collector/util.go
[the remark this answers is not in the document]
SamYuan1990: we can keep it if we have to do with lots of conversions.
I don't want to see too many stack in our code.
`int(math.Min(float64(len(container.Command)), 10))` seems 5 stacks?

`Min(len(container.Command),10)` is better.

pkg/collector/prometheus.go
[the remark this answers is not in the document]
SamYuan1990: I see

pkg/collector/util.go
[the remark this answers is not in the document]
SamYuan1990: btw, where do we use function Min?

pkg/collector/prometheus.go
```diff
@@ -0,0 +1,583 @@
...
+			}
+		}
+	}()
+}
+
+// updatePodMetrics send pod metrics to prometheus
+func (c *PrometheusCollector) UpdatePodMetrics(wg *sync.WaitGroup, ch chan<- prometheus.Metric) {
+	for _, pod := range podEnergy {
+		wg.Add(1)
+		go func(pod *PodEnergy) {
```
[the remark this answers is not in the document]
marceloamaral: right, the naming looks strange....
But we had in the code this list called `podEnergy` that was a list of `PodEnergy` structures....

pkg/collector/prometheus.go
marceloamaral: We could create one gorotine per pod, or many per pod
Go is very efficient to handle a very large number of gorotines
So, I am creating multiple gorotines per pod to speedup the metric calculation process as much as possible

pkg/collector/prometheus.go
marceloamaral: The `lock` here is to synchronize with the metric collection
Reporting metrics in the middle of the `reader` function will create wrong data....
The wait is to make sure that all the gorotines reported the metrics when a Collect is triggered

pkg/collector/prometheus.go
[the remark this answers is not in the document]
SamYuan1990: ... ok just makes it better.

pkg/collector/prometheus.go
SamYuan1990: may I know where can I know more about how `reader ` and `Collect` works together?
I want to ensure multiple thread in safe and effective.

pkg/collector/prometheus.go
```diff
@@ -0,0 +1,594 @@
...
+		}(container)
+
+		wg.Add(1)
+		go func(container *ContainerMetrics) {
+			defer wg.Done()
+			ch <- prometheus.MustNewConstMetric(
+				c.containerDesc.containerCoreJoulesTotal,
+				prometheus.CounterValue,
+				float64(container.EnergyInCore.Aggr)/miliJouleToJoule,
+				container.PodName, container.ContainerName, container.Namespace, container.Command[:int(math.Min(float64(len(container.Command)), 10))],
```
SamYuan1990, line 527: `container.Command[:int(math.Min(float64(len(container.Command)), 10))` should we only calculate it once?

SamYuan1990: and 10 should be a const?

SamYuan1990:
```
const commandLenLimit=10
var _command := container.Command
if len(_command) > commandLenLimit {
_command = container.Command[:commandLenLimit]
}
```

pkg/collector/prometheus.go
SamYuan1990: https://github.com/sustainable-computing-io/kepler/pull/287/files#r990643209
as what I am worried about, so far with large number of gorotines, some var may calculated in many times as duplicated. I am not sure by one gorotine per pod we can avoid duplicate.

pkg/collector/prometheus.go
marceloamaral: I will use in this explanation the structure names as I refactored in PR

The `containersMetrics` contains all the metrics that the `reader` collects/calculates, e.g. energy consumption and resource utilization from the bpf

The `reader` is a gorotine that updates the `containersMetrics` every `SamplePeriodSec`, which is 3s

The `Collect` function implements the prometheus interface, which reports metrics on an endpoint in `\metrics`.
The prometheus app is typically configured with a default scrape interval and will get metrics from an endpoint according to that interval. But this interval can also be set in `servicemonitor`, in kepler we set this to be 3s, so it will match the `SamplePeriodSec`.....

So when prometheus tries to get data from kepler, the `Collect` function will be called and will access the `containersMetrics` metrics. We use `lock` to synchronize and ensure that Collect will only read the data after the `reader` has finished updating `containersMetrics`.

Note that having internal goroutines in the `Collect` function does not interfere with the `reader` function since the gorotines in the `Collect` function only read the data. So they will never create any race conditions!

pkg/collector/prometheus.go
```diff
@@ -0,0 +1,583 @@
...
+			float64(nodeEnergy.SensorEnergy.Aggr())/miliJouleToJoule,
+			nodeName, "acpi",
+		)
+		ch <- prometheus.MustNewConstMetric(
+			c.nodeDesc.nodeOtherComponentsJoulesTotal,
+			prometheus.CounterValue,
+			float64(nodeEnergy.EnergyInOther.Aggr())/miliJouleToJoule,
+			nodeName,
+		)
+		if config.EnabledGPU {
```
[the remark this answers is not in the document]
marceloamaral: I did not wanted to create a gorotine only for the gpu metric, so I added other metrics together....

pkg/collector/util.go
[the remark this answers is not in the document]
marceloamaral: There was already a logic to set the `command` size up to 10 characters....
Not sure why it was there, probably it was appearing some extreme large strings there....Don't know....

So, to decrease the command to up to 10, we can resize the slice `command[:10]`
But if the slice size is 7 for example, it will panic...
Therefore, the logic gets the `Min(len(command), 10)`

pkg/collector/prometheus.go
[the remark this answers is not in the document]
SamYuan1990: ok, I see.

pkg/collector/prometheus.go
marceloamaral: Which variable will be calculated?
The goroutines will only read data...

pkg/collector/prometheus.go
SamYuan1990: em ...
`reader` update `containersMetrics` in each 3s since 0s.
`Collect ` consume `containersMetrics` in each 3s since 2.x s.

in an edge case, `Collect ` WIP, and `reader` updated data. then ... the result of `Collect` will be two parts, one for before `reader` updates at 3s and the other is `reader` updates after 3s?

is that acceptable?

pkg/collector/util.go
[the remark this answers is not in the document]
SamYuan1990: copy?
```
_command = make([]string,10)
copy(_command, command)
```

I tested with int, if the len of source less than 10, it will fill up with 0 as make([]int,10) default value is 0.

pkg/collector/prometheus.go
SamYuan1990: count `container.Command[:int(math.Min(float64(len(container.Command)), 10))`

marceloamaral: Good catch, I will update that!

pkg/collector/prometheus.go
marceloamaral: The result of `Collect` will be only one part, after the `reader`, never in between....

The `reader` only start when the `Collect` has finished, and vice-versa....
So, in practice, the `Collect` will report the results of the previous `reader` iteration

Since the Collect function is very fast (specially because we have now gorotines) I expect that it will not change the `reader` sampling interval too much...but maybe we should log how long the `reader` was waiting to start

pkg/collector/prometheus.go
marceloamaral: Yes, sounds good

KaiyiLiu1234:
The Model Server is currently using the old prometheus metrics. Will the old ones be kept for legacy?

marceloamaral:
> Will the old ones be kept for legacy?

@KaiyiLiu1234, yes!
That is why I included the deprecated tag in the old metrics.
When the model server gets updated, we can remove the old metrics...

marceloamaral:
I added the node name as the label `instance` also in the container metrics, as suggested [here](https://github.com/sustainable-computing-io/kepler/issues/286#issuecomment-1277799905)

rootfs:
let's merge this one. @marceloamaral would you like to follow up with an option to turn on/off/enable metrics? That'll give us flexibility to meet multiple goals.

rootfs:
@sustainable-computing-io/reviewer any more feedback?

sunya-ch:
I agree to apply these changes. However, I, of course, will affect the currently trained model. 
All models needs to be re-trained and may be re-collected (with the container level data).
The default weight will not be available until all previously steps have re-done (that I cannot put my hand on this work within these three weeks).

marceloamaral:
@sunya-ch you might be referring to the case that Kepler is running on the VM, right?
So that Kepler will not use the ratio model. And the ratio model will not be impacted by pre-trained models....


But in relation to the other context, where the ratio model cannot be used, we will need to recalibrate the models.
But that's all model server logic, right?
Note that this also happens if the model was trained on a different system and needs to be recalibrated again when Kepler is running on another system...

So I don't think it's going to be too harmful to merge this PR now...
Especially because there are some use cases that have recently disabled Kepler to run on VMs, e.g the PR #294.

rootfs:
I did some rough tests and want to have some idea the cpu  load and scrape latency if the metrics go into the conformance direction.

I use node_exporter as a reference, since the amount and the format of metrics match what the new kepler metrics will become in a well populated environment

On my setup, the node_exporter port is 9100.

First of all, there are around 3K metrics exported by node_exporter:
```
# curl http://localhost:9100/metrics |grep -v \# |wc -l
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  215k    0  215k    0     0   189k      0 --:--:--  0:00:01 --:--:--  189k
3234
```

Next, I just keep scrape the node_exporter metrics endpoint every 3 seconds (as used by kepler service monitor)

```bash
while true; do time curl http://localhost:9100/metrics 2>&1 > /dev/null;sleep 3;done
```
```console
real    0m1.102s
user    0m0.004s
sys     0m0.007s
```

During the scraping, node_exporter has high cpu usage:
image

By comparison, current kepler latency is less than 0.02 seconds and cpu usage is less than 5%.

The latency is likely proportionally related to the computation done by node_exporter. If kepler metrics gets expansive, the latency and cpu will be a huge concern.


I suggest:
- Let's keep metrics used by model server, autoscaler, and other special purposes. They are not designed to be used end users Prometheus conformance is less important than efficiency.
- Performance and scalability impact need to be studied in more details.
- The container level metrics needs to be discussed. The benefits can go both ways: although it is nice to have more info but unless there is a real use case for fine grained info, let's prioritize scalability and efficiency first.

sunya-ch:
> @sunya-ch you might be referring to the case that Kepler is running on the VM, right? So that Kepler will not use the ratio model. The ratio model will not be impacted by pre-trained models....
> 
> But in relation to the other context, where the ratio model cannot be used, we will need to recalibrate the models. But that's all model server logic, right? Note that this also happens if the model was trained on a different system and needs to be recalibrated again when Kepler is running on another system...
> 
> So I don't think it's going to be too harmful to merge this PR now... Especially because there are some use cases that have recently disabled Kepler to run on VMs, e.g the PR #294.

Model server used exported metric from Kepler to do the online training for the other system that doesn't have the power measurement (no matter it is the VM or not).
The point is not whether recalibrating due to the system change. The point is the name of features is mismatch and not support right now and the training scope is changed from pod to container. It is not that severe as from VM-scope to pod-scope because it is the same process level and sharing the operating processes. However, there is no experimental result support whether how much that affects the model accuracy yet.

marceloamaral:
@rootfs very good analysis! Ok, let's keep the old metrics, I'll remove the deprecated label from them.

----
@sunya-ch
> The point is that the resource name is incompatible

It's not necessary as we keep the old metrics as they are...
In the end, the request to the Model Server will be just a bigger array with more instances...(per container, not per pod)
And of course, we will need to validate everything, but we are still working on validate everything right?

jichenjc:
>>>very good analysis! Ok, let's keep the old metrics, I'll remove the deprecated label from them.

@marceloamaral should we revert the revert of podID PR?

marceloamaral:
@jichenjc, don't think so, this PR here will fix the issue....

jichenjc:
@marceloamaral got it :)

sunya-ch (approved):
Note that, after merging this PR, there are required updates on [kepler-model-server](kepler-model-server), [kepler-estimator](https://github.com/sustainable-computing-io/kepler-estimator), and this PR https://github.com/sustainable-computing-io/kepler/pull/225.

marceloamaral:
@sunya-ch we will still keep the old metrics
But eventually we must update [kepler-model-server](https://github.com/sustainable-computing-io/kepler/pull/kepler-model-server) and [kepler-estimator](https://github.com/sustainable-computing-io/kepler-estimator)

And we definitely need to update the PR #225.

marceloamaral:
Add also the old metric [pod_cpu_instructions](https://github.com/sustainable-computing-io/clever/blob/main/dashboards/clever-dashboard-w-IPS.json#L329) since the [clever](https://github.com/sustainable-computing-io/clever) dashboard use it

marceloamaral:
Reopening, it was closed by mistake

marceloamaral:
As I was already updating a lot of parts, I also needed to combine this PR with another PR that introduced a design change.
I also took the opportunity to refactor and reorganize some packages to improve code readability.

PR #288, opened 2022-10-08
user podID instead of podName to avoid duplicate

jichenjc:
Fixes: #265 

### DIFF
--- a/pkg/collector/reader.go
+++ b/pkg/collector/reader.go
@@ -84,7 +84,8 @@ func init() {
 	for i := 0; i < len(*pods); i++ {
 		podName := (*pods)[i].Name
 		podNamespace := (*pods)[i].Namespace
-		podEnergy[podName] = NewPodEnergy(podName, podNamespace)
+		podID := string((*pods)[i].ObjectMeta.UID)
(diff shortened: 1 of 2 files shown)


marceloamaral:
While we were discussing using podID, I've created another PR (#287) that changes the logic to report metrics per container and not per pod...

When using containerID, we will not need to use podID....
I'm sorry you spend time working on this....

jichenjc:
ok, not a big deal, your PR title ` Make Prometheus metrics conform with the Prometheus metrics guideline ` seems need update 

anyway, do we see use case that multiple container in one pod should be considered a bunch so that we should report energy about it ? otherwise, we can wait for your PR then close this

marceloamaral:
I updated the title, thanks!

So I think we can simplify as much as possible and avoid duplication by just having the container metrics....

Pod metrics can be aggregated later in Prometheus with promQL, as is done for cpu, e.g., `container_cpu_usage_seconds_total`

rootfs (approved):

marceloamaral:
@rootfs I think this PR was going to be close not merged....

marceloamaral:
I am reverting this for now, but we can merge it again if needed

PR #292, opened 2022-10-11
impl for integration test

SamYuan1990:
Details for this PR
e2e folder with end to end testing, will run with two models, base on environment kepler_address.

build kepler during test process and run it targets on local.
run test case targets on cluster after cluster port forwarding.

### DIFF
--- a/e2e/healthcheck_test.go
+++ b/e2e/healthcheck_test.go
@@ -1,0 +1,30 @@
+package e2e_test
+
+import (
+	"net/http"
+	"os"
+	"os/exec"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("healthz check should pass", func() {
+	Context("start server with health check", func() {
+		It("should work properly", func() {
+			address, ok := os.LookupEnv("kepler_address")
+			if !ok {
+				address = "localhost:8888"
+				cmd := exec.Command(keplerBin)
+				keplerSession, err = gexec.Start(cmd, nil, nil)
+				Expect(err).NotTo(HaveOccurred())
+				time.Sleep(5 * time.Second) // wait for server start up
+			}
+			resp, err := http.Get("http://" + address + "/healthz")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.StatusCode).To(Equal(http.StatusOK))
+		})
+	})
+})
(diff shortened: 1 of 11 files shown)


SamYuan1990:
redo for #275

marceloamaral (changes requested):
I think the tests should be in a folder called `test/`
So, you will have `test/e2e/`

marceloamaral:
> will run with two models, base on environment kepler_address.

Which models? Power models?

e2e/healthcheck_test.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var _ = Describe("healthz check should pass", func() {
+	Context("start server with health check", func() {
+		It("should work properly", func() {
+			address, ok := os.LookupEnv("kepler_address")
+			if !ok {
+				address = "localhost:8888"
+				cmd := exec.Command(keplerBin)
+				keplerSession, err = gexec.Start(cmd, nil, nil)
```
marceloamaral, line 21: Will the test start Kepler alone or with the other components, like the Model Server?

If it is only Kepler, what is the integration test part?

I am not saying that this test here is not important, but it might be not integration test as the title suggest...

marceloamaral: In the case it is not integration test, this is definitely a functional test that we need to have

vendor/modules.txt
```diff
@@ -86,6 +86,8 @@ github.com/onsi/ginkgo/types
 ## explicit; go 1.18
 github.com/onsi/gomega
 github.com/onsi/gomega/format
+github.com/onsi/gomega/gbytes
```
rootfs, line 89: why not change in `go.mod` and `go.sum`?

rootfs:
@SamYuan1990 I just have a question on go mod. 

@sunya-ch sounds good! After this PR, can you create an issue/request? 

@marceloamaral @sunya-ch anything else is needed for this PR? It'll be great to have an initial integration test in place and add more in the future.

marceloamaral (approved):
/lgtm

This is a good start.

Even though I think it is a functional test, so that the workflow file name would be different... I am ok to keep like that for now....

sunya-ch:
> @SamYuan1990 I just have a question on go mod.
> 
> @sunya-ch sounds good! After this PR, can you create an issue/request?
> 
> @marceloamaral @sunya-ch anything else is needed for this PR? It'll be great to have an initial integration test in place and add more in the future.

Sure. I can push the PR for the deployment CI test later after all components is clearly integrated.

PR #294, opened 2022-10-12
Exclude VM node when deploy kepler exporter

ruomengh:
Currently kepler exporter cannot get data successfully on VM node. With this PR change, kepler exporter will not deploy on VM nodes. Only bare metal nodes will be scheduled for kepler deployment.

Signed-off-by: Hao, Ruomeng <ruomeng.hao@intel.com>

### DIFF
--- a/manifests/kubernetes/deployment.yaml
+++ b/manifests/kubernetes/deployment.yaml
@@ -57,6 +57,14 @@ spec:
     spec:
       hostNetwork: true
       serviceAccountName: kepler-sa
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: feature.node.kubernetes.io/cpu-cpuid.HYPERVISOR
+                operator: DoesNotExist
+                values:
       containers:
       - name: kepler-exporter
         image: quay.io/sustainable_computing_io/kepler:latest


rootfs:
assign to @sunya-ch

marceloamaral:
I thought there were some people running Kepler on VMs....

We have many estimate logic to estimate metrics that cannot be collected, for example, the power metrics.
This scenario is very tight related to Kepler running on VMs...

rootfs:
@marceloamaral sounds good, let's take this.

marceloamaral:
@jichenjc will this impact you?

jichenjc:
@marceloamaral yes, and will this impact CI (github action use VM) as well? anyway, I think I can tolerate this a little bit 
as I can modify the manifest, thanks for let me know ~

marceloamaral:
We could keep the VM deployment disabled by default for now but include a logic in the `make cluster-sync` to enable it via
an environment variable

jichenjc:
>>>We could keep the VM deployment disabled by default for now but include a logic in the make cluster-sync to enable it via
an environment variable

ok, I will add some logic here to make it happen , and this might overtime diversify the feature implemented and tested
but we can always improve

jichenjc:
I googled this `feature.node.kubernetes.io/cpu-cpuid.HYPERVISOR` but not find helpful info 

Also, I patched the manifests from this PR but seems not working ,as I am still able to work on VM now

```
# kubectl get daemonset -n kepler -o yaml | grep nodeAffi -A 5
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
              - matchExpressions:
                - key: feature.node.kubernetes.io/cpu-cpuid.HYPERVISOR
                  operator: DoesNotExist

kepler        kepler-exporter-2vw4j                      1/1     Running   0             3m30s
```

jichenjc:
ok, found https://github.com/sustainable-computing-io/kepler/issues/299#issuecomment-1278368776
this explain everything

jichenjc:
I found https://github.com/kubernetes-sigs/node-feature-discovery/issues/554#issuecomment-1004815921 but I doubt whether that's just for node discovery..

jichenjc:
all clear , I deployed https://kubernetes-sigs.github.io/node-feature-discovery/stable/get-started/index.html on my node
and it provided the flag as this PR showed, but I think we should not introduce this dependency just for such minor check, that seems overkill

and I think detect running on hypervisor is also a long way to go as we might run on different platform (windows, linux, Mac) or physical (ARM,x86 , s390x) , this is not the key value of kepler, so 

I propose to revert this PR until we have agreement on what's the best way to go

jichenjc:
@rootfs @ruomengh @marceloamaral

rootfs:
agreed, I think it is better to have a customizable node selector rules in the deployment.  This will be an CRD spec for the operator.

In the mean time, @jichenjc @ruomengh @marceloamaral how about use kustomize + customizable property, instead of a static manifest?

jichenjc:
>>>how about use kustomize + customizable property, instead of a static manifest?

yes, that will be terrific and helpful for future deployment, and I am wondering that will be helpful in this ticket

marceloamaral:
> how about use kustomize + customizable property, instead of a static manifest

@rootfs this is a very good direction! We'll need to document how to use it.

@jichenjc will be awesome if you can work on it

jichenjc:
>>>will be awesome if you can work on it

yes, it's on my todo list

PR #296, opened 2022-10-13
Revert "user podID instead of podName to avoid duplicate"

marceloamaral:
Reverts sustainable-computing-io/kepler#288

The PR #287 introduces the concept of use containerID, so that collecting the podID will not be necessary


jichenjc:
/lgtm

thanks, based on the communication we have , I am ok to revert

rootfs (approved):

PR #297, opened 2022-10-13
Split OpenShift machineconfigs into separate kernel-devel and cgroupv2

Feelas:
From what I understand (@rootfs please correct me) the CGroupsv2 are no longer required to successfully run Kepler. On the other hand, kernel-devel packages seem to be necessary.

Since kernel-devel is necessary and cgroupsv2 aren't, my proposition is to split the manifests into kernel-devel and cgroupv2 added separately so it is possible to install kernel-devel via MachineConfigs separately. I don't have a cluster suitable to verify whether the kustomization.yaml works correctly right now, so it would be good to test it.

@williamcaban @sallyom WDYT? We could somehow make it optional in the OpenShift deploymen scripts to install CGroupsv2 or not, but I don't have time budget to work this out right now.

@sustainable-computing-io/kepler-deployment

### DIFF
--- a/manifests/openshift/cluster-prereqs/50-master-cgroupv2.yaml
+++ b/manifests/openshift/cluster-prereqs/50-master-cgroupv2.yaml
@@ -2,12 +2,10 @@
 apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
-  name: 50-master-kernel-devel-cgroupv2
+  name: 50-master-cgroupv2
   labels:
     machineconfiguration.openshift.io/role: master
 spec:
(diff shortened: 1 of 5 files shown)


rootfs:
@williamcaban @sallyom can you take a look? thanks

marceloamaral:
That makes sense.

Although, it might be better to be tested by the team that is using this....

rootfs (approved):

sallyom:
@williamcaban @Feelas sorry i missed this last week - I will update the OpenShift manifests based on this PR once I confirm things are working as expected. Thanks!

PR #302, opened 2022-10-17
Detect kepler is running inside VM

jichenjc:
followup, we will use kustomization to build different manifest ..


manifests/kubernetes/deployment.yaml
```diff
@@ -81,6 +73,7 @@ spec:
         - 0.0.0.0:9102
         - -enable-gpu=true
         - -enable-cgroup-id=true
+	- -allow-running-inside-vm=true
```
jichenjc, line 76: allow here to make CI happen

follow up, will update CI script to patch the manifest during start of test

rootfs:
@jichenjc We need to sync up on the VM case. The limitations of running Kepler on VM are that: 1) missing perf counter and 2) missing MSR (and RAPL) counter. Let's check them in more details.

For 1),  on certain VM configurations (a comprehensive reference can be found on [this VTUNE guide](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-vtune-amplifier-functionality-on-aws-instances.html)), perf counters are available.

For 2), there are two approaches to estimate in-vm power consumption: the [current implementation](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/rapl/source/estimate.go) using energy consumption coefficients and general models by @sunya-ch described [here](https://github.com/sustainable-computing-io/kepler/discussions/266).

jichenjc:
@rootfs this is actually a `better`(my thought) way to do #294 

for the things you described above, how about create an issue to track those changes needed?
as both PR are actually forbid VM running kepler and this PR provide a `force` way?

marceloamaral:
As mentioned by @rootfs  in PR #294, we can use kustomize + customizable for the deployment
In this case, we will not need this PR here

jichenjc:
@marceloamaral we actually have no way to detect it's a VM or not unless we create https://kubernetes-sigs.github.io/node-feature-discovery/ deployment in our manifest, I think it's overkill 

in the manifest /kustomize, I am going to enable the flag in CI and disable the flag by default of this PR

jichenjc:
@marceloamaral @rootfs please help check update
as mentioned, without this PR, the vm/bm  check doesn't work at all ...

rootfs:
@jichenjc we need it answer what is the purpose of detecting VM in the first place. It makes sense for people not wanting to run kepler on VM nodes as the current model is not strong enough. I am fine with deployment level differentitaion. 

But as long as we are able to get power source and metrics, VM is like the same as physical machines, then what additional benefit does VM detection have?

jichenjc:
>>>But as long as we are able to get power source and metrics, VM is like the same as physical machines, then what additional benefit does VM detection have?

@rootfs we merged https://github.com/sustainable-computing-io/kepler/pull/294, but as I mentioned, it doesn't work due to we never asked for corrsponding node-feature-discovery install and node will not be labeled , this PR only improve that logic.

unless this is not true or we figure out the VM can work well now? 
```
Currently kepler exporter cannot get data successfully on VM node. With this PR change, kepler exporter will not deploy on VM nodes. Only bare metal nodes will be scheduled for kepler deployment.
```

So I Think we need have a set of todo in order to make VM fully functional? anyway, I can work on to improve VM related area as well as I only have VM , no BM :(

rootfs:
@jichenjc there are some hooks in [eBPF attacher](https://github.com/sustainable-computing-io/kepler/blob/92f3815f798eb6a2699fb492035dda3a9983261c/pkg/attacher/bcc_attacher.go#L89-L91) that detect VM.

jichenjc:
@marceloamaral that comment said `// some hypervisors don't expose perf counters` 
is it `all` hypervisor doesn't do so and we can think it's safe to think if we enter the branch you give, it's VM?

rootfs:
perf counter can be exposed through hypervisor cpu bypass. 

@jichenjc Kepler works for both VM and baremetal. This is one of the goals of integrating model server, model estimator and kepler.

marceloamaral:
@jichenjc @rootfs
I am sure if we should test if we are running inside a VM.

The logic should be, if we do not have power and hardware counter metrics, we will use the model server/estimator.

rootfs:
we don't need to distinguish VM vs BM.

PR #311, opened 2022-10-19
Add kustomization of manifests

jichenjc:
-created vm/bm model folder
-patch manifest to BM deployment so it should reject VM env
-I didn't change major manifest usage (as I don't know whether we have other usage other than CI so don't want to break them), that's use a symlink to keep original things still working
-updated the build manifest so new manifest file will be generated in _output folder


ruomengh:
> @jichenjc looks cool! Can you also update [README](https://github.com/sustainable-computing-io/kepler/blob/main/README.md) and [installation guide](https://github.com/sustainable-computing-io/kepler/tree/main/manifests#installing-kepler-on-kubernetes)?

+1. It would be helpful to update installation guide as well.

marceloamaral (changes requested):
Seems pretty good!

Can you also update the [cluster-deploy.sh](https://github.com/sustainable-computing-io/kepler/blob/main/hack/cluster-deploy.sh#L25) file?
This PR changes the generated manifest deployment folder so we need to update it

We could also include an environment variable to select the correct manifest, like:
`BARE_METAL_NODE=True`

We make it True by default

marceloamaral:
Or `BARE_METAL_NODE_ONLY=True` since there are scenarios that mix them

manifests/README.md
```diff
@@ -10,6 +10,10 @@ The operating system must provide:
 Consult the documentation of your Linux distribution on details for enabling these prerequisites.
 ## Installing Kepler on Kubernetes
 
+Simple option:
+
+For people who just want a quick try:
+
 ```bash
 # NOTE: These manifest take care of creating the namespace and Kepler exporter
 kubectl apply -f kubernetes/deployment.yaml
```
rootfs, line 19: is this path still valid?

jichenjc: technicaly yes, as there is a symlink to base (deployment.yaml -> base/deployment.yaml) and  base/deployment.yaml
still useable for now

the reason for doing this is to make things simple(so first time user has less step to perform), but I think sooner or later we might need run kustomize inevitable

jichenjc:
@rootfs @marceloamaral can you help take another look? Thanks

rootfs (approved):

rootfs:
lgtm

marceloamaral (approved):
Looks great, thanks!
/lgtm

PR #313, opened 2022-10-20
Create ephemeral cluster with k3d/k3s

akyriako:
##What
Added an option for k3d/k3s clusters 

##Why
An alternative flavor to kind

##How
Based on the existing cluster/cluster-up/kind/common.sh script logic, with the necessary minor changes to facilitate k3d 

##Tested
On Linux ubuntu2214 5.15.0-48-generic

##Anything else
The original common.sh script creates an artefact in [manifests/kubernetes/deployment.yaml , line 124] by adding an additional trailing "-host" every time the script is being executed. Eventually, it ends up with a pretty long "proc-host-host-host-etc-etc" value for hostPath.path (mitigated for k3d common.sh)


SamYuan1990:
thanks for contribution.
em... may I know any change on CI as adding github action with k3s as k8s provider?

cluster-up/cluster/k3d/common.sh
```diff
@@ -0,0 +1,216 @@
...
+
+function _load_prometheus_operator_images_to_local_registry {
+    for img in $(_get_prometheus_operator_images); do
+        $CTR_CMD pull $img
+        $CTR_CMD tag $img "localhost:${REGISTRY_PORT}/${img}"
+        $CTR_CMD push "localhost:${REGISTRY_PORT}/${img}"
+    done
+} 
+
+function _deploy_prometheus_operator {
```
SamYuan1990, line 83: ... en I suppose for any k8s, not matter kind or k3s, we should use same steps to deploy prometheus right?

cluster-up/common.sh
```diff
@@ -22,4 +22,8 @@ set -e
 # the cluster kind is a kubernetes cluster
 if [ ${CLUSTER_PROVIDER} = "kind" ]; then
     CLUSTER_PROVIDER="kubernetes"
+fi
```
SamYuan1990, line 25: @marceloamaral , maybe not in this PR, but we always change the CLUSTER_PROVIDER to kubernetes? and this common.sh just do a validation?

manifests/kubernetes/deployment.yaml
```diff
@@ -121,7 +121,7 @@ spec:
           type: Directory
       - name: proc
         hostPath:
-          path: /proc
```
SamYuan1990, line 124: I suppose this change breaks our CI on kind for now.

SamYuan1990 (changes requested):
please adjust deployment.yaml to keep current CI works.

cluster-up/cluster/k3d/manifests/k3d.yml
```diff
@@ -0,0 +1,67 @@
...
+kubeAPI: # same as `--api-port myhost.my.domain:6445` (where the name would resolve to 127.0.0.1)
+  # host: "kepler.k3d.local" # important for the `server` setting in the kubeconfig
+  hostIP: "127.0.0.1" # where the Kubernetes API will be listening on
+  hostPort: "6445" # where the Kubernetes API listening port will be mapped to on your host system
+# image: rancher/k3s:v1.20.4-k3s1 # same as `--image rancher/k3s:v1.20.4-k3s1`
+network: _cluster_network # same as `--network my-custom-net`
+# subnet: "172.28.0.0/16" # same as `--subnet 172.28.0.0/16`
+# token: superSecretToken # same as `--token superSecretToken`
+volumes: # repeatable flags are represented as YAML lists
+  - volume: /proc:/proc-host # same as `--volume '/my/host/path:/path/in/node@server:0;agent:*'`
```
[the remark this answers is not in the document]
rootfs: for reference `/proc` is used by kepler on cgroupv1 environment.

manifests/kubernetes/deployment.yaml
rootfs: btw, the integration CI on kind timed out twice. Need some investigation.

manifests/kubernetes/deployment.yaml-e
```diff
@@ -0,0 +1,143 @@
...
+        hostPath:
+          path: /lib/modules
+          type: Directory
+      - name: tracing
+        hostPath:
+          path: /sys
+          type: Directory
+      - name: proc
+        hostPath:
+          path: /proc-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host-host
```
rootfs, line 124: this line looks odd

rootfs:
@akyriako thank you for the contribution! Can you break the PR into two? One for the k3d cluster, that is independent and can get merged quickly. The second is to add the k3d to the CI.

manifests/kubernetes/deployment.yaml-e
akyriako: @rootfs that's what I meant in "#Anything else" comment of the PR, I have mitigated it for k3d; for kind it still creates an never ending chain of "-host" every time you run it.

manifests/kubernetes/deployment.yaml-e
```diff
@@ -0,0 +1,143 @@
+apiVersion: v1
+kind: Namespace
```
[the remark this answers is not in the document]
akyriako: Hi @SamYuan1990 that's autogenerated file, in my opinion (in the extend I understood the scripts) this is not necessary

cluster-up/cluster/k3d/common.sh
akyriako: @SamYuan1990 i have made no change in the prometheus and grafana provisioning related functions, there are identical with the ones from kind

SamYuan1990:
enabled CI for this PR.

.github/workflows/integration_test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        kube_provider: [kind]
+        kube_provider: [kind, k3d]
```
marceloamaral, line 11: I don't think we need to test in both tools... in the end they both create a similar k8s cluster.
Just one type of tool in CI should be enough...
Testing creating cluster with both tools is just an additional overhead that might not be necessary...

Support for more cluster types is helpful for the developer.

cluster-up/common.sh
marceloamaral: Right, this is because `make cluster-deploy` access the `kubernetes` folder to deploy de manifests....
And the folder path is created using the `CLUSTER_PROVIDER`....

We could improve that in somehow in the future....

marceloamaral (changes requested):
I agree with @rootfs that this PR should not configure the CI....

I made some arguments about this in a previous comment.

marceloamaral:
Both kind and k3d creates kuberntes cluster. So we should test only one.
Microshift on the other hand, creates OpenShift cluster and this should be a useful test.

marceloamaral:
Is there any advantage of creating a k3d clusters instead of a kind cluster?

jiere:
> Is there any advantage of creating a k3d clusters instead of a kind cluster?

I also curious about this topic. 
I just noticed this thread since @jichenjc mentioned k3s alternative in #711  [comments](https://github.com/sustainable-computing-io/kepler/pull/711#discussion_r1209178251)

jichenjc:
I am saying `kind` is container based and if we want to get real test data that we want to compare the `real data` instead of data from container like kind, I was told the data from container/VM is not well supported (yet)

SamYuan1990:
@akyriako , @marceloamaral , @rootfs 
as we moving infrastructure script into local-dev-cluster, hence I am going to move this ticket to
https://github.com/sustainable-computing-io/local-dev-cluster/issues/22

Issue #253, opened 2022-09-27
CI commit message check


SamYuan1990:
**Is your feature request related to a problem? Please describe.**
Try to add commit message check for each pr.

**Describe the solution you'd like**
https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines

- Try to keep the subject line to 50 characters or less; do not exceed 72 characters
- Providing additional context if I am right means formatting in topic: something
- The first word in the commit message subject should be capitalized unless it starts with a lowercase symbol or other identifier
- Do not end the commit message subject with a period
- Add a single blank line before the commit message body

**Describe alternatives you've considered**
n/A

**Additional context**
        @SamYuan1990, could you try following the Kubernetes commit message guideline?
https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines

It's good to provide meaningful messages with context!

_Originally posted by @marceloamaral in https://github.com/sustainable-computing-io/kepler/issues/215#issuecomment-1253419614_

SamYuan1990:
https://github.com/GsActions/commit-message-checker

PR #324, opened 2022-10-22
CI: enable commit msg check

SamYuan1990:
this pr use https://github.com/GsActions/commit-message-checker with regex as implementation.

### DIFF
--- a/.github/workflows/commitMsg.yml
+++ b/.github/workflows/commitMsg.yml
@@ -1,0 +1,37 @@
+name: 'Commit Message Check'
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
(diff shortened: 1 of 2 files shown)


SamYuan1990:
resolve #253

rootfs:
What are the magic format look like? Should the format be documented?

SamYuan1990:
> What are the magic format look like? 

https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines
it's not any magic format, as it discussed in #253 , pre discussed with @marceloamaral before, we want to have some commit message check to ensure commit message is meaningful.

> Should the format be documented?
good idea.

rootfs:
lgtm

@marceloamaral any feedback?

marceloamaral (approved):
Looks good!
/lgtm

> The first word in the commit message subject should be capitalized unless it starts with a lowercase symbol or other identifier

I was not aware of that :)

PR #336, opened 2022-10-25
refactory determineHostByteOrder function

jichenjc:
2 places define this function , but 1 place are not used ..
so remove both of them and move defintion to utils 


marceloamaral (approved):
Looks good, thanks!
 /lgtm

PR #339, opened 2022-10-27
bug: Fix build manifest

jichenjc:
in order to make cluster-sync happy, modified some make build-manifest  in previous PR

but this makes the image used for general deployment (not CI) using CI image (we sed it to devel )
and it lead to image not found issue

with this PR, check whether it's CI, if it's , sed the iamge
if it's not ,keep using latest image

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -29,6 +29,8 @@ jobs:
         run: make cluster-sync
         env:
             CLUSTER_PROVIDER: ${{matrix.kube_provider}}
+            IMAGE_REPO: "localhost:5001/kepler"
+            IMAGE_TAG: "devel"
             CTR_CMD: docker
 
       - name: test if kepler is still alive
--- a/hack/build-manifest.sh
+++ b/hack/build-manifest.sh
@@ -27,7 +27,7 @@ then
 fi
 
 CLUSTER_PROVIDER=${CLUSTER_PROVIDER:-kubernetes}
-IMAGE_TAG=${IMAGE_TAG:-devel}
+IMAGE_TAG=${IMAGE_TAG:-latest}
 IMAGE_REPO=${IMAGE_REPO:-quay.io/sustainable_computing_io/kepler}
 
 MANIFESTS_OUT_DIR=${MANIFESTS_OUT_DIR:-"_output/manifests/${CLUSTER_PROVIDER}/generated"}
--- a/cluster-up/README.md
+++ b/cluster-up/README.md
@@ -30,6 +30,6 @@ This destroys the whole cluster.
 
 ```bash
 export CLUSTER_PROVIDER=`kind`
-make cluster-sync
+IMAGE_REPO="localhost:5001/kepler" IMAGE_TAG="devel" make cluster-sync
 ```
 This removes a running kepler deployment, creates a new docker image, pushes it to the local cluster registry, deploys kepler with the newly created image, and waits for the kepler container to be in the running state.
--- a/dev/README.md
+++ b/dev/README.md
@@ -20,6 +20,12 @@ export IMAGE_REPO=index.docker.io/myrepo
 export IMAGE_TAG=mybuild
 ```
 
+By default, please use:
+```bash
+export IMAGE_REPO="localhost:5001/kepler"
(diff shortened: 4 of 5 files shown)


jichenjc:
@marceloamaral can you help take a quick look? Thanks

hack/build-manifest.sh
```diff
@@ -46,7 +47,7 @@ kubectl kustomize ${MANIFESTS_OUT_DIR}/vm > ${MANIFESTS_OUT_DIR}/vm/deployment.y
 
 if [[ $CLUSTER_PROVIDER == "openshift" ]]; then
     cat manifests/${CLUSTER_PROVIDER}/kepler/01-kepler-install.yaml | sed "s|image:.*|image: $IMAGE_REPO:$IMAGE_TAG|" > ${MANIFESTS_OUT_DIR}/kepler/01-kepler-install.yaml
-else
+elif [[ $RUN_K8S_CI == "true" ]]; then
     sed -i "s|image:.*|image: $IMAGE_REPO:$IMAGE_TAG|" ${MANIFESTS_OUT_DIR}/bm/deployment.yaml
```
marceloamaral, line 51: When we deploy in CI (or local cluster) we need to provide a custom $IMAGE_REPO:$IMAGE_TAG 
Otherwise it will use the default ones, so that it should not really change the image name....

The default values should work fine for not CI environment....
```bash
IMAGE_TAG=${IMAGE_TAG:-devel}
IMAGE_REPO=${IMAGE_REPO:-quay.io/sustainable_computing_io/kepler}
```
Maybe the image tag is wrong?

marceloamaral: I think that the problem is the image tag should be `latest`
So, just changing that should be enough
```bash
IMAGE_TAG=${IMAGE_TAG:-latest}
```

jichenjc: we use image: localhost:5001/kepler:devel 
in the CI (make cluster-sync)

marceloamaral: Actually I am not finding where we define the new `IMAGE_TAG` and `IMAGE_REPO` env variable
```bash
IMAGE_TAG="devel"
IMAGE_REPO ="localhost:5001/kepler"
```
I should be here:
https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/integration_test.yml#L32


And when it is not CI, it should be the default ones

marceloamaral: For example, In my local environment I do
```bash
IMAGE_REPO ="localhost:5001/kepler" IMAGE_TAG="devel" make cluster-sync
```

marceloamaral: Passing the correct repo and tag, we don't need a new variable to say it's CI
Also, because the same approach will be used for testing on a local cluster, not just CI

jichenjc: so your suggestion is to update the CI script && Test env setup to force localhost:5001/kepler:devel?

marceloamaral: Yes, that should be the natural way! We should use the IMAGE_REPO and IMAGE_TAG only to define the kepler image...

Some other projects do in this way.... for example [kubevirt](https://github.com/kubevirt/kubevirt)

jichenjc: @marceloamaral update. thanks

.github/workflows/integration_test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             GRAFANA_ENABLE: "false"
 
       - name: simple test - deploy kepler
-        run: make cluster-sync
+        run: IMAGE_REPO="localhost:5001/kepler" IMAGE_TAG="devel" make cluster-sync
         env:
```
marceloamaral, line 30: We can define in the env field
```yaml
- name: simple test - deploy kepler
        run: make cluster-sync
        env:
            IMAGE_REPO: localhost:5001/kepler
            IMAGE_TAG: devel 
            CLUSTER_PROVIDER: ${{matrix.kube_provider}}
            CTR_CMD: docker
```

marceloamaral (changes requested):
Looks good, just did a small comment.

marceloamaral (approved):
/lgtm

Issue #346, opened 2022-10-31
Unit test is failing in macOS after enabling nvidia GPU metrics...


marceloamaral:
**Describe the bug**
[Unit test](https://github.com/sustainable-computing-io/kepler/actions/runs/3361335455/jobs/5571601524#step:4:53) is failing in macOS after enabling nvidia GPU metrics...

The problem is that Mac OS/BSD does not support RTLD_DEEPBIND

**To Reproduce**
make test-mac-verbose in macOS

**Additional context**
Do we really need to run the test in macOS environment?

I mean, we can run the tests in containers running on top of macOS...
Actually, that is the way I do. We don't need to run the tests directly in the macOS...

This bug is impacting the PR #345

marceloamaral:
Closing this PR because the macOS use case is important to show the support of cross-building.

```
Some systems have compatibility issues with the nvidia library. 
See https://github.com/sustainable-computing-io/kepler/issues/184
Therefore, we can disable files that use the NVIDIA library using the "+build gpu" tag. 
This means that the compiler will only include these files if the compilation has the gpu tag.
```

So, we disable gpu in the macOS test.

PR #345, opened 2022-10-31
Add GPU metric in the same flow as the other metrics

marceloamaral:
**What is the problem?**
The current GPU power consumption follows a very different path from the power calculation of the other components.
For example, CPU and DRAM metric uses the model package to calculate power consumption, but GPU power consumption is calculated in the power package which should only be responsible for collecting information...

Additionally, the current GPU metrics collector is only collecting GPU memory utilization, with no information about GPU compute utilization.

Finally, the collector package converts the containerMetric and nodeMetric structure into many collections (ie arrays) to send data to the model package. Not only does this duplicate the data, it complicates the logic as the metrics and container information must follow a specific order and the metric names are hidden, and many collections must be updated.

**What does this PR do?**
This PR:
- adds SM GPU utilization metric
- adds GPU metric in the same flow as other metrics
- calculates GPU power consumption similarly to other components calculation
- fixes test slowness (was trying to connect to a non-existing model server)
- reduces the number of collections by sharing the containerMetric and nodeMetric structure with the model package. This not only improves the code, it also simplifies the code to be easier to understand.

Btw, this was a TODO comment in the code.

This PR is related to the issues #310 and #320.

**Comments to the reviewers**
I have split the changes into multiple commits for making the review easier.

Issue #184, opened 2022-09-15
Build s390x image via buildx cross-platform


huoqifeng:
Leverage docker buildx cross-platform to build s390x image.

- [ ] ~~builder~~
- [ ] base https://github.com/sustainable-computing-io/kepler/pull/194, https://github.com/sustainable-computing-io/kepler/pull/202
- [ ] kepler https://github.com/sustainable-computing-io/kepler/pull/203
- [ ] Build scripts for multi-arch for local  https://github.com/sustainable-computing-io/kepler/pull/241

cooktheryan:
this could be a cool tool to leverage for this work https://github.com/docker/build-push-action

Example https://github.com/migtools/gitops-primer/blob/main/.github/workflows/validate-primer.yaml#L361-L368

huoqifeng:
I noticed package `nvidia-driver-NVML-515.48.07-1.el8.x86_64.rpm` here in base image https://github.com/sustainable-computing-io/kepler/blob/main/build/Dockerfile.base#L6, looks only has x86 arch supported?

huoqifeng:
@mengfj @SamYuan1990 @jichenjc I don't think LinuxONE has nvidia card installed, so it's reasonable that package `nvidia-driver-NVML` does not support s390x. Is this package key to kepler?

SamYuan1990:
> @mengfj @SamYuan1990 @jichenjc I don't think LinuxONE has nvidia card installed, so it's reasonable that package `nvidia-driver-NVML` does not support s390x. Is this package key to kepler?

Personally,  I suppose we need to have a MVP on s390x.

Specifically for automation point of view, the MVP is to build s390x based golang binary and running it in a s390x based fedora image?
As far as I know:
- there lots of golang build tag in current build, sooner or later we should have one for s390x. (as so far kepler build logic is basing on build tag for different hardware.)
- then for docker image, one thing hard is for base image, others is looking for different libs, as nvidia one found today.
- we can use matric for x86/arm/s390x to control the build.
ideally, the make script will look like either
```
make ${arch}
```
or
``` 
# in make file
if $uname ... -eq s390x 
GO_BUILD_TAGS = $GO_BUILD_TAGS + s390_tag
fi
```

personally, consider we may have different docker file for different arch, I prefer below
```
make ${arch}
```

@rootfs , some questions:
1. for any vm without nvidia hardware? what will be happen? what build tag should we use?
2. any other comments?

rootfs:
[...]

PR #347, opened 2022-10-31
Remove unit test in macOS

marceloamaral:
Fix the issue #346

We don't need to run the test in macOS environment,  since we can run the tests in container while using a macOS.


SamYuan1990:
https://github.com/sustainable-computing-io/kepler/pull/345/files#r1009406770
I don't suppose it is an issue for MacOS.

SamYuan1990:
or say, as our GPU code relays on `NVIDIA` package.
Hence we have issue with
- no GPU case but has to install NVIDIA package?
- no NVIDIA GPU has to install NVIDIA GPU package?

hence, just remove MacOS test case code is not targeting on root cause of #346

SamYuan1990:
> @SamYuan1990
> 
> I didn't understand what you think the problem is... Could you explain more?
> 
> I'm having the same problem on my macOS, but I can compile the code in containers....

will NVIDIA GPU package is installed on containers but not MacOS. as line 

https://github.com/sustainable-computing-io/kepler/blob/bf17c6bfa1062e4d557d0aeb1ce2ffc67ade803e/build/Dockerfile.base#L21

but on MacOS, by default, we don't install this package, hence when a golang code invokes `	"github.com/NVIDIA/go-nvml/pkg/nvml"` on MacOS, or **any instance without this library been installed or mount ?** the error will happen as 
`Error: vendor/github.com/NVIDIA/go-nvml/pkg/dl/dl.go:34:18: could not determine kind of name for C.RTLD_DEEPBIND`

as 

https://github.com/sustainable-computing-io/kepler/actions/runs/3361335455/jobs/5571601524#step:4:53

as 

https://github.com/NVIDIA/go-nvml/blob/main/pkg/dl/dl.go#L25

need NVIDIA C library

SamYuan1990:
@marceloamaral , 
https://github.com/sustainable-computing-io/kepler/pull/347#issuecomment-1297090880
is the explain for why there is a c library issue on mac as you found.

https://github.com/sustainable-computing-io/kepler/pull/347#issuecomment-1297082575
is a general question as how should we handle this c library missing case.
in my point of view, we should follow build tags.

it's also been found at https://github.com/sustainable-computing-io/kepler/issues/184#issuecomment-1249013531

SamYuan1990 (changes requested):
please think twice before merge.

as we have different build tag.
macOS Unit test plays role of no bcc, no GPU unit test.
and bcc unit test with container plays role of bcc and GPU build tag unit test.

do we force relays on BCC and GPU otherwise make kepler may `panic` is acceptable?

marceloamaral:
@SamYuan1990 converted to draft to investigate it more

marceloamaral:
Closing this PR because the macOS use case is important to show the support of cross-building.

```
Some systems have compatibility issues with the nvidia library. 
See https://github.com/sustainable-computing-io/kepler/issues/184
Therefore, we can disable files that use the NVIDIA library using the "+build gpu" tag. 
This means that the compiler will only include these files if the compilation has the gpu tag.
```

PR #351, opened 2022-11-01
Enhancements-process: template and folder 

sallyom:
how about this template? it might be more than what's needed but not all sections have to be filled out - I followed [this template from openshift/enhancements](https://github.com/openshift/enhancements/blob/master/guidelines/enhancement_template.md)

/cc
@rootfs
@husky-parul 
@wangchen615


rootfs:
@sallyom thanks! this is a quite comprehensive template.

rootfs (approved):

Issue #350, opened 2022-11-01
Improve integration test suite to test basic metric query


sallyom:
[From #162](https://github.com/sustainable-computing-io/kepler/issues/162) @SamYuan1990 

**Is your feature request related to a problem? Please describe.**
So far, we just verify by `until kubectl get svc --all-namespaces  ...`
to make it better, we'd better have a test suite and targeting on running service.

**Describe the solution you'd like**
as above.

rootfs:
@sallyom sounds good, please list the previously opened issues for this task.

sallyom:
previous work:
- https://github.com/sustainable-computing-io/kepler/pull/283
- https://github.com/sustainable-computing-io/kepler/pull/292

sallyom:
From #162 @rootfs & @SamYuan1990 

> The service readiness detection is to be supported in kepler-operator

> > so what's our scope of integration test?

Will open a design doc to discuss integration test goals

SamYuan1990:
is there any thing we left in this issue?

sallyom:
@SamYuan1990 @rootfs this merged last week, https://github.com/sustainable-computing-io/kepler/pull/357  so I believe this issue is completed! :partying_face: :tada:

sallyom:
https://github.com/sustainable-computing-io/kepler/issues/349 will add more in-depth integration tests. The current suite satisfies this issue.

PR #357, opened 2022-11-03
Test: fix for #350 impl for endpoints test case

SamYuan1990:
- keep design as build and run server before test for local test.
- keep design as when specific env given use target and skip build.
- use table to test path as test metrics to avoid copy paste myself.
- adjust code accordingly


SamYuan1990:
@rootfs , do we have any other api endpoint need to check with?
cc: @sallyom

rootfs (approved):

rootfs:
Thanks @SamYuan1990 !

We'll add model server to check the endpoint next

PR #376, opened 2022-11-09
update: openshift manifests

sallyom:
update: openshift manifests


rootfs (approved):

rootfs:
@sallyom CI requires the commit subject line less than 50 characters

manifests/microshift/model-server-deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+          privileged: true
+        ports:
+        - containerPort: 8100
+        volumeMounts:
+          - name: cfm
+            mountPath: /etc/config
+            readOnly: true
+          - mountPath: /models
+            name: models
+        command: ["python3.8",  "model_server.py"]
```
williamcaban, line 66: @rootfs FYI. I noticed we are using `python script_name.py` as commands in the pods definitions. This will buffer output by default and the stdout logs will not be displayed as they happen. To achieve unbuffered stdout on a container executing python directly as command we should be using the `-u` option like `python -u script_name.py` for unbuffered stdout.

rootfs: Thanks for the note, cc @sunya-ch @KaiyiLiu1234 on model server side.

sallyom: @williamcaban with recent updated the model-server deployment is no longer necessary - also, the MicroShift deployment no longer requires special consideration, so no longer including separate MicroShift manifests.

manifests/openshift/cluster-prereqs/kustomization.yaml
```diff
@@ -2,5 +2,5 @@ commonLabels:
   sustainable-computing.io/app: kepler
 
 resources:
-- 51-master-kernel-devel.yaml
-- 51-worker-kernel-devel.yaml
+- 50-master-cgroupv2.yaml
```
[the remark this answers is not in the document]
sallyom: commented and documented

sallyom:
@williamcaban @rootfs finally found the correct incantation for commit msgs :smile_cat:  this is ready for review, thank you

rootfs (approved):

rootfs:
@williamcaban wdyt?

rootfs:
thank you @dagrayvid for confirmation!

Issue #354, opened 2022-11-02
Improve unit test coverage to 50+%


SamYuan1990:
- [x] #430
- [x] #431
- [x] #432
- [x] #433
- [x] #434
- [x] #435
- [x] #436
- [x] #437

SamYuan1990:
maybe we need to clarify #411

SamYuan1990:
ref to https://github.com/sustainable-computing-io/kepler/actions/runs/3574669171/jobs/6010219583

SamYuan1990:
coverage: 0.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/bpfassets	0.068s
coverage: 0.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/bpfassets/attacher	0.059s
coverage: 51.1% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/cgroup	0.116s
coverage: 39.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/collector	0.096s
coverage: 51.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/collector/metric	0.079s
coverage: 60.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/config	0.052s
coverage: 9.3% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/kubelet	0.066s
coverage: 0.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/manager	0.064s
coverage: 55.4% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/model	0.272s
coverage: 0.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/model/estimator/local	0.062s
coverage: 0.0% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/model/estimator/sidecar	0.043s
?   	github.com/sustainable-computing-io/kepler/pkg/model/types	[no test files]
coverage: 58.6% of statements
ok  	github.com/sustainable-computing-io/kepler/pkg/utils	0.054s
coverage: [no statements]
ok  	github.com/sustainable-computing-io/kepler/pkg/version	0.064s

SamYuan1990:
- [ ]  github.com/sustainable-computing-io/kepler/pkg/bpfassets
- [ ] github.com/sustainable-computing-io/kepler/pkg/bpfassets/attacher
[...]

PR #380, opened 2022-11-11
Test: add test case for package utils

SamYuan1990:
related to #354 

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -15,11 +15,14 @@ func CreateTempFile(contents string) (filename string, reterr error) {
 		return "", err
 	}
 	defer func() {
-		if err := f.Close(); err != nil {
+		if err = f.Close(); err != nil {
 			return
 		}
 	}()
-	_, _ = f.WriteString(contents)
+	_, err = f.WriteString(contents)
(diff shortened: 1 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -36,18 +39,10 @@ func DetermineHostByteOrder() binary.ByteOrder {
 }
 
 const (
-	systemProcessName      string = "system_processes"
-	systemProcessNamespace string = "system"
+	SystemProcessName      string = "system_processes"
+	SystemProcessNamespace string = "system"
 )
 
-func GetSystemProcessName() string {
```
rootfs, line 43: lets keep the access function to protect the variable from being changed outside of the pkg scope.

SamYuan1990: > lets keep the access function to protect the variable from being changed outside of the pkg scope.

do we have this kind of case?
as you can see I removed this Get function and meanwhile make the const able to be access from other packages.
in my point of view, is there any case to modify a const?

rootfs: right, it is already a const, no mutable possible outside the scope.

rootfs (approved):

PR #381, opened 2022-11-11
testing: enable cpu profiling

rootfs:
will use this to investigate performance and scalability issues

Issue #365, opened 2022-11-07
Kepler general energy metrics vs energy tracing metrics


williamcaban:
**Is your feature request related to a problem? Please describe.**

The recent changes in the kind of metrics exposed increase the level of visibility to trace how much energy is consumed by which elements on a process/Pod/container. This kind of metrics is valuable for applications owners looking to optimize the energy consumption by their application or services stack. At the same time, the granularity of these metrics make it extremely difficult for functionalities that require a quick understanding of the system (e.g. schedulers, multi-cluster level dashboards, etc) and considerable increase the load on the Prometheus stack.

**Describe the solution you'd like**

I like to propose for Kepler to have a flag that can be use to influence the type of metrics that it exposes on the metric endpoint. Basically, enable Kepler to be used by general use cases (e.g. report the top-5 energy consumption by namespace, pods or apps; report the software+OS-view energy consumption by nodes, etc) and another mode for developers that want to do low level tracing of energy consumption of their apps.

The mode for the general use cases should provide a simple single metric per container or Pod, that can be used for user facing reporting and executive types of reports.

**Describe alternatives you've considered**
With the current granular metrics:

- We have seen a significant impact on Prometheus utilization (way too many metrics in very short time even on micro environments)
- It has become very challenging using the metrics of the general use cases as the query formula gets very long in not time, and the system takes several seconds to process it once
- For comparison, when loading the metrics on DataFrames to run analytics in a Jupyter notebooks for an 8 nodes clusters with ~500 Pods, loading a 2hr dataset with 1 minutes resolution: The current granular metrics takes 2min 17s (as reported by `%timeit`) to load. With the previous aggregated metrics, the same cluster setup with the same datasets and resolution was loading in less than 1 second.

**Additional context**
The general use cases should also cover https://github.com/sustainable-computing-io/kepler/issues/301

rootfs:
thank you @williamcaban for the investigation and proposal

> - We have seen a significant impact on Prometheus utilization (way too many metrics in very short time even on micro environments)
> - For comparison, when loading the metrics on DataFrames to run analytics in a Jupyter notebooks for an 8 nodes clusters with ~500 Pods, loading a 2hr dataset with 1 minutes resolution: The current granular metrics takes 2min 17s (as reported by %timeit) to load. With the previous aggregated metrics, the same cluster setup with the same datasets and resolution was loading in less than 1 second.


[...]

PR #383, opened 2022-11-13
pod cache: cache system process

rootfs:
per profiling stack in #381, the pod cache update causes the latency. As previously in [release 0.3](https://github.com/sustainable-computing-io/kepler/blob/release-0.3/pkg/podlister/resolve_container.go#L125), when the pod is not found, the system process name is used in the cache.

### DIFF
--- a/pkg/cgroup/resolve_container.go
+++ b/pkg/cgroup/resolve_container.go
@@ -122,15 +122,11 @@ func getContainerInfo(cGroupID, pid uint64) (*ContainerInfo, error) {
 		return cinfo, nil
 	}
 
-	// in the case the containerID is not a kubernetes container, the updateListPodCache will not add it to the cache
 	if _, ok := containerIDToContainerInfo[containerID]; !ok {
 		// some system process might have container ID, but we need to replace it if the container is not a kubernetes container
-		if info.ContainerName == utils.SystemProcessName {
-			containerID = utils.SystemProcessName
-		}
+		containerIDToContainerInfo[containerID] = info
 	}
 
-	containerIDToContainerInfo[containerID] = info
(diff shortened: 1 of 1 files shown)


rootfs:
This partially fixes the high cpu usage issue in #365

pkg/cgroup/resolve_container.go
```diff
@@ -125,12 +125,9 @@ func getContainerInfo(cGroupID, pid uint64) (*ContainerInfo, error) {
 	// in the case the containerID is not a kubernetes container, the updateListPodCache will not add it to the cache
 	if _, ok := containerIDToContainerInfo[containerID]; !ok {
 		// some system process might have container ID, but we need to replace it if the container is not a kubernetes container
```
jichenjc, line 127: this comment line not valid any more?

rootfs: this line is still valid

rootfs: setting `containerID` to `utils.SystemProcessName` doesn't add the system processes to the cache and prevent the next refresh. That's why we use `containerID` as it is in the cache.

marceloamaral (approved):
/lgtm

this might improve the ListPod issue described in #391

pkg/cgroup/resolve_container.go
marceloamaral: @rootfs I am wondering if having some system process with real containerID will impact system process metric aggregation in somehow.

marceloamaral: Maybe we need to update the containerID after adding it to the cache...

Sorry, I merged this PR, but we might need to update it...

PR #392, opened 2022-11-15
Add support to enable mem profiling

marceloamaral:
**What does this PR do**
The PR #381 introduced the support for CPU profiling, this PR enables MEM profiling.

**Additional Comments**
To get the profiling of kepler running in the k8s cluster, you should edit the manifest to enable it
```
        command:
        - /usr/bin/kepler
        - -address
        - 0.0.0.0:9102
        - -enable-gpu=true
        - -enable-cgroup-id=true
        - -v=1
        - -cpuprofile=cpu.prof
        - -memprofile=mem.prof
```

By default, the profiling will execute for 60s, but you change it with `-profile-duration=seconds`.

To copy the files from the Kepler container:
```
pod=$(kubectl get pod -n kepler | grep kepler-exporter | awk '{print $1}')
kubectl cp -n kepler $pod:cpu.prof cpu.prof
kubectl cp -n kepler $pod:mem.prof mem.prof
```

Example of using the profile:
```
go tool pprof cpu.prof
go tool pprof -alloc_objects mem.prof
```

### DIFF
--- a/cmd/exporter.go
+++ b/cmd/exporter.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"runtime"
 	"runtime/pprof"
 	"syscall"
 	"time"
@@ -56,6 +57,8 @@ var (
 	enabledEBPFCgroupID          = flag.Bool("enable-cgroup-id", true, "whether enable eBPF to collect cgroup id (must have kernel version >= 4.18 and cGroup v2)")
 	exposeHardwareCounterMetrics = flag.Bool("expose-hardware-counter-metrics", true, "whether expose hardware counter as prometheus metrics")
 	cpuProfile                   = flag.String("cpuprofile", "", "dump cpu profile to a file")
+	memProfile                   = flag.String("memprofile", "", "dump mem profile to a file")
+	profileDuration              = flag.Int("profile-duration", 60, "duration in seconds")
 )
 
 func healthProbe(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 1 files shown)


cmd/exporter.go
```diff
@@ -79,21 +82,52 @@ func main() {
...
+			klog.Fatal("could not start CPU profile: ", err)
+		}
+		klog.Infof("Started CPU profiling")
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		go func() {
+			defer f.Close()
+			select {
+			case <-sigs:
+				break
```
rootfs, line 101: if SIGTERM raised, also exit kepler program

cmd/exporter.go
```diff
@@ -56,6 +57,8 @@ var (
 	enabledEBPFCgroupID          = flag.Bool("enable-cgroup-id", true, "whether enable eBPF to collect cgroup id (must have kernel version >= 4.18 and cGroup v2)")
 	exposeHardwareCounterMetrics = flag.Bool("expose-hardware-counter-metrics", true, "whether expose hardware counter as prometheus metrics")
 	cpuProfile                   = flag.String("cpuprofile", "", "dump cpu profile to a file")
+	memProfile                   = flag.String("memprofile", "", "dump mem profile to a file")
```
jichenjc, line 60: maybe we can update this settings by default to CI later on so by default we are able to obtain that data?

marceloamaral: @jichenjc this is a good idea, and we need to think which test we should do that.
Maybe the integration test? Or a new test? This would require not only run the profile but also dump in the logs the information.

Btw, to enable it I was think to use some environment variable to use with `make cluster-sync`. For example:
`ENABLE_CPU_PROFILING=true`

Let's do it in another PR.

jichenjc: yes, the environment variable works for me 

and agree we can do this in another PR..

cmd/exporter.go
marceloamaral: thanks @rootfs, I updated it

PR #395, opened 2022-11-15
e2e: ensure metrics exist

rootfs:


rootfs:
it looks VMs still have missing metrics

e2e/e2e_metric_test.go
```diff
@@ -0,0 +1,77 @@
...
+		"kepler_container_cpu_cpu_time_us",
+		"kepler_container_dram_joules_total",
+		"kepler_container_gpu_joules_total",
+		"kepler_container_joules_total",
+		"kepler_container_other_host_components_joules_total",
+		"kepler_container_package_joules_total",
+		"kepler_container_uncore_joules_total",
+		"kepler_exporter_build_info",
+		"kepler_node_core_joules_total",
+		"kepler_node_cpu_scaling_frequency_hertz",
```
jichenjc, line 43: "kepler_container_core_joules_total"
"kepler_container_dram_joules_total"
"kepler_container_gpu_joules_total"
"kepler_container_joules_total"
"kepler_container_other_host_components_joules_total"
"kepler_container_package_joules_total"
"kepler_container_uncore_joules_total"
"kepler_exporter_build_info"
"kepler_node_energy_stat"
"kepler_node_nodeInfo"
"kepler_node_other_host_components_joules_total"
"kepler_node_platform_joules_total"
"kepler_pod_energy_stat"


from my local VM env
something like `kepler_node_dram_joules_total` (not only this but should be simliar) are missing

jichenjc: I am guessing 

https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/node_energy_collector.go#L47
return empty nodeComponentsEnergy due to both conditions above not satisifed ..

jichenjc: see above ,that's my output ...

jichenjc: @rootfs 

this might because the conf by default not provided so that the LinearRegressor init failed for node ...
and in turn https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/node_energy_collector.go#L47
returns empty
```
I1116 02:24:17.933746       1 lr.go:166] LR Model (AbsModelWeight): no config
I1116 02:24:17.933751       1 lr.go:166] LR Model (AbsComponentModelWeight): no config
I1116 02:24:17.933755       1 lr.go:166] LR Model (DynModelWeight): no config
I1116 02:24:17.933758       1 lr.go:157] getWeightFromServer failed, using InitModelURL:https://raw.githubusercontent.com/sustainable-computing-io/kepler-model-server/main/tests/test_models/DynComponentModelWeight/CgroupOnly/ScikitMixed/ScikitMixed.json to load
```

SamYuan1990:
please fix for unit test error and lint issues.

e2e/e2e_metric_test.go
```diff
@@ -0,0 +1,77 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		req.Header.Set("Accept", "application/json")
+		resp, err := http.DefaultClient.Do(req)
+		Expect(err).NotTo(HaveOccurred())
+		defer resp.Body.Close()
+		Expect(resp.StatusCode).To(Equal(http.StatusOK))
+		body, err := io.ReadAll(resp.Body)
+		Expect(err).NotTo(HaveOccurred())
+		content := string(body)
+		for _, metric := range kepler_metrics {
```
SamYuan1990, line 67: just a suggestion
currently the test logic will get the http response and looping array ...
for any issue during round robin, it will exist with error as `Fail`

here is the suggestion:
get the http response in the Before() and use DescribeTable() to go through with each metric
as sample as e2e/restapicheck_test.go
hence 1 single error will not break all test case?

SamYuan1990:
> https://github.com/sustainable-computing-io/kepler/actions/runs/3474246799/jobs/5807190524#step:8:39
> 
> ```
> metric kepler_container_core_joules_total not found
> ```
> 
> do we consider to have all metrics as const and listed in a go lang file?

+ @marceloamaral in loop, do you think missing metric is related to we running integration test in github agent as ubuntu vm? which cause bcc running in different module?
btw, in current CI process, we didn't install BCC at host vm for kind cluster.

Hence, I am not sure how kepler running without bcc module load from host vm related with metrics missing issue as we found on CI.

rootfs:
depend on #397

e2e/e2e_metric_test.go
```diff
@@ -0,0 +1,77 @@
...
+			reStr := metric + "{.*} \\d*\\.?\\d*" // metric{.*}\d*\.?\d*
```
SamYuan1990, line 68: May I have a result for this regex?
I suppose we just need to check type definition for metrics. hence we don't need `.d` ?

Makefile
```diff
@@ -142,7 +142,7 @@ test: ginkgo-set tidy-vendor
 	@go test -tags $(GO_BUILD_TAGS) ./... --race --bench=. -cover --count=1 --vet=all
 
 test-verbose: ginkgo-set tidy-vendor
-	@go test -tags $(GO_BUILD_TAGS) -covermode=atomic -coverprofile=coverage.out -v ./... --race --bench=. -cover --count=1 --vet=all
+	@go test -tags $(GO_BUILD_TAGS) -covermode=atomic -coverprofile=coverage.out -v ./pkg/... --race --bench=. -cover --count=1 --vet=all
```
rootfs, line 145: @SamYuan1990 do you know why unit_test_with_bcc still kicks off e2e test?

rootfs: never mind, the failed test is in integration

e2e/e2e_metric_test.go
rootfs: let's check both metrics name, format, and counter number to ensure they exist and are non-zero (next step)

e2e/e2e_metric_test.go
rootfs: Let's wait till all the node and container models are supported. I am not sure what will happen to all the metrics when kepler runs on VMs, maybe they will be all supported and hence no error should be expected.

rootfs:
ubuntu runner doesn't have kernel source
```
modprobe: FATAL: Module kheaders not found in directory /lib/modules/5.15.0-1022-azure
chdir(/lib/modules/5.15.0-1022-azure/build): No such file or directory
```

rootfs:
so this is what I did to make the tests pass:
- break the e2e and unit test in Makefile, so mac unit tests don't have to run e2e
- for kind cluster on ubuntu runner:
   - install kernel source on ubuntu
   - bind mount kernel directory to kind container
- e2e metric test:
  - fail only if metrics missing on cgroup v2

manifests/kubernetes/base/deployment.yaml
```diff
@@ -115,6 +117,10 @@ spec:
         hostPath:
           path: /proc
           type: Directory
+      - hostPath:
```
rootfs, line 120: this is only needed for kind cluster running on ubuntu runner
@jichenjc can you help to make this into a kustomization?

Makefile
SamYuan1990, line 145: e2e test cases without build tag, hence it will be tested once covered in test scope.

e2e/e2e_metric_test.go
SamYuan1990: ok, for enhancement, we can in following pr.

.github/workflows/integration_test.yml
```diff
@@ -42,6 +47,8 @@ jobs:
       - name: run integration_test
         run: |
           kubectl port-forward $(kubectl -n kepler get pods -oname) 9102:9102 -n kepler -v7 &
-          go test ./e2e/... --race --bench=. -cover --count=1 --vet=all
+          kubectl logs -n kepler daemonset/kepler-exporter
```
SamYuan1990, line 50: do we need those checking?

Makefile
SamYuan1990: @rootfs , I suppose we should add e2e folder into general test case. For detail, you can ref https://github.com/sustainable-computing-io/kepler/blob/main/e2e/e2e_suite_test.go#L27

SamYuan1990 (changes requested):

Makefile
SamYuan1990: or, may I know the reason why we can't run this new added case at locally?
We have checked with restapi with url path locally, https://github.com/sustainable-computing-io/kepler/blob/main/e2e/restapicheck_test.go#L24 and for this pr, we added detail checking for URL path `/metric`
hence, I want to check the metric with details locally as well.

SamYuan1990 (approved):

rootfs:
odd, the commit history messed up after rebase, closing and opening a clean one

Issue #396, opened 2022-11-15
some cGroup available metrics are missing


sunya-ch:
**Describe the bug**

There are two problems with previous cgroup logic.
1. there is no need to find slice before scope if the walk logic is correctly.
https://github.com/sustainable-computing-io/kepler/blob/fa0b3fa5b673c27bcd32898cea939c7770d1df97/pkg/cgroup/slice_handler.go#L164-L179
This will introduce unnecessary operations. 

2. in some system, both cpu.stat (cpu.stat) and individual files for each stat exists and in cpu.stat, there is no summary stat inside. The logic here should be skipped.
https://github.com/sustainable-computing-io/kepler/blob/fa0b3fa5b673c27bcd32898cea939c7770d1df97/pkg/cgroup/stat_reader.go#L102-L107

With the second problem, only memory stat is read and cpu stat always return 0 and cause inaccurate inference. 
```bash
Available cgroup metrics from cgroup:
 [cgroupfs_memory_usage_bytes cgroupfs_kernel_memory_usage_bytes cgroupfs_tcp_memory_usage_bytes]
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Related issues**
https://github.com/sustainable-computing-io/kepler/issues/365
https://github.com/sustainable-computing-io/kepler/pull/377
https://github.com/sustainable-computing-io/kepler/issues/352

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

jichenjc:
>>>there is no need to find slice before scope if the walk logic is correctly.

PR #398, opened 2022-11-16
log: add log for node components assignment

jichenjc:


jichenjc:
related to #396

pkg/collector/node_energy_collector.go
```diff
@@ -46,9 +48,13 @@ func (c *Collector) updatePlatformEnergy() {
 func (c *Collector) updateNodeComponentsEnergy() {
 	nodeComponentsEnergy := map[int]source.NodeComponentsEnergy{}
 	if components.IsSystemCollectionSupported() {
+		klog.V(5).Info("support components.IsSystemCollectionSupported")
```
marceloamaral, line 51: Can you update the message to a sentence instead of a code?
Otherwise it seems to be a debug message not a log.
For example:
`System energy collection is supported`

jichenjc: done~

PR #387, opened 2022-11-14
rapl: remove MSR

rootfs:
fix #373

Issue #373, opened 2022-11-09
Cannot start up exporter with Kind


vincent-pli:
**Describe the bug**
Try `kepler` with Kind, but the exporter cannot startup, get below errors in logs:
not sure if the Kind is not in the support list.
```
[root@experimental kepler]# k -n kepler logs -f kepler-exporter-kgvdl
panic: runtime error: index out of range [16] with length 16

goroutine 1 [running]:
github.com/sustainable-computing-io/kepler/pkg/power/rapl/source.mapPackageAndCore()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/power/rapl/source/msr_util.go:88 +0x234
github.com/sustainable-computing-io/kepler/pkg/power/rapl/source.InitUnits()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/power/rapl/source/msr_util.go:147 +0x1d
github.com/sustainable-computing-io/kepler/pkg/power/rapl/source.(*PowerMSR).IsSupported(...)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/power/rapl/source/msr.go:22
github.com/sustainable-computing-io/kepler/pkg/power/rapl.init.0()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/power/rapl/power.go:53 +0xdb
```

**To Reproduce**

**Additional context**

jichenjc:
it is 
I am guessing kind env report different data lead to code handling not well
technically it should report error or return empty map etc

assume you are using latest version?

rootfs:
@vincent-pli 

The offending line is [here](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/rapl/source/msr_util.go#L88)

Based on the error, it looks you have 16 cores (indexed from 0 to 15) and somehow CPU core to pkg mapping finds core 16. 

Can you get the following output?

```bash
cat /sys/devices/system/cpu/cpu*/topology/physical_package_id
```

rootfs:
btw, MSR is turned off, oddly go build doesn't take the [hint](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/rapl/power.go#L53)

vincent-pli:
Checked the code and VM, it's weird...
my core number is 16 and 
`cat /sys/devices/system/cpu/cpu15/topology/physical_package_id`
get 30

@rootfs  @jichenjc

[...]

PR #400, opened 2022-11-16
Experimental code for weird case of physical_package_id

vincent-pli:
For #373 
feel free to close it.

### DIFF
--- a/pkg/power/components/source/rapl_msr_util.go
+++ b/pkg/power/components/source/rapl_msr_util.go
@@ -45,10 +45,12 @@ const (
 )
 
 var (
-	fds        []int
-	byteOrder  binary.ByteOrder
-	packageMap []int
-	maxPackage = -1
+	fds       []int
+	byteOrder binary.ByteOrder
+
+	// package - core
+	packageMap  map[int]int
+	packageList []int
 
 	powerUnits, timeUnits           float64
 	cpuEnergyUnits, dramEnergyUnits []float64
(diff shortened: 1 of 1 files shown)


pkg/power/components/source/rapl_msr_util.go
```diff
@@ -85,21 +83,20 @@ func mapPackageAndCore() error {
 		if err != nil {
 			return err
 		}
-		if packageMap[id] == -1 {
-			packageMap[id] = i
-		}
-		if maxPackage < id {
-			maxPackage = id
-		}
+
```
marceloamaral, line 86: Using the package id instead of the array position looks like a good approach.

I don't know the details, but the original code might have some reason to do not update the packageMap[i] when it is not null, i.e., `!=-1`.

So, I would keep this logic.

marceloamaral:
@vincent-pli 
Can you also update the commit message. We follow the kubernetes guideline for commit message.
You need to define the context of the code and use imperative sentences, for example:
`power: rapl: msr: update the package id logic`

rootfs:
@vincent-pli thank you! please follow Marcelo's review and I'll close #387

pkg/power/components/source/rapl_msr_util.go
vincent-pli: Well, I don't know too much detail either, but from code perspective, the original `packageMap` is an array and adopt index as content of `physical_package_id`, but in my env, the `physical_package_id` is not as its expectation(equal to cpu core index, 0 - 15), so I adopt `map`, I guess I cannot keep the original logic, I mean '!=-1'.
based on the original logic: 
```
if packageMap[id] == -1 {
    packageMap[id] = i
}
```
seems, it try to avoid reset, does that means the `physical_package_id` may has duplicated value?

vincent-pli:
*_*, have no idea, I alreay use:
`git commit --amend` and `git push --force`

rootfs:
@vincent-pli the regex looks for a space next to `:`.

PR #401, opened 2022-11-16
Replace the containerID of system process with a common name for aggregation

marceloamaral:
The PR #383 fixed the problem of not adding the system containerID to the cache.

However, it removed the logic  of using common name `system_process` to aggregate all the system process resource utilization. Without this logic, we will have the system processes resource utilization divided in different "container" instances because of using different containerIDs.

### DIFF
--- a/pkg/cgroup/resolve_container.go
+++ b/pkg/cgroup/resolve_container.go
@@ -123,10 +123,16 @@ func getContainerInfo(cGroupID, pid uint64, withCGroupID bool) (*ContainerInfo, 
 	}
 
 	if _, ok := containerIDToContainerInfo[containerID]; !ok {
-		// some system process might have container ID, but we need to replace it if the container is not a kubernetes container
 		containerIDToContainerInfo[containerID] = info
+		// some system process might have container ID, but we need to replace it if the container is not a kubernetes container
+		if info.ContainerName == utils.SystemProcessName {
+			containerID = utils.SystemProcessName
+			// in addition to the system container ID, add also the system process name to the cache
+			if _, ok := containerIDToContainerInfo[containerID]; !ok {
+				containerIDToContainerInfo[containerID] = info
+			}
+		}
 	}
-
 	return containerIDToContainerInfo[containerID], nil
 }
 


pkg/cgroup/resolve_container.go
```diff
@@ -123,10 +123,16 @@ func getContainerInfo(cGroupID, pid uint64, withCGroupID bool) (*ContainerInfo,
 	}
 
 	if _, ok := containerIDToContainerInfo[containerID]; !ok {
-		// some system process might have container ID, but we need to replace it if the container is not a kubernetes container
 		containerIDToContainerInfo[containerID] = info
+		// some system process might have container ID, but we need to replace it if the container is not a kubernetes container
```
rootfs, line 127: containerID is a sysfs path, e.g. `/sys/fs/cgroup/system.slice/NetworkManager.service`, why convert it to SystemProcessName?

marceloamaral: When exporting the prometheus metric, all system processes are aggregated under the `system_process` container name.

If we use different containerID for different system processes, the `system_process` will not have all metric values aggregated. We will have different instances in the containerMetrics map.

rootfs: that would work. But at the moment, all containerID are file path, they don't give more info.

marceloamaral: Not sure what you mean by the containerID is the file path.
The containerID is actually the ID used by the container runtime.
We have the function [extractPodContainerIDfromPath](https://github.com/sustainable-computing-io/kepler/blob/c8f0528980944d3952f001ac09cd4e285cebe73e/pkg/cgroup/resolve_container.go#L291) which extracts the container ID from the cgroup file.
The function gets the container ID when the file has a line with the string `crio`, `docker` or `containerd`.

However, some system processes have the string `containerd` and return a containerID. So when we try to get the name of the container, it has `system_process` as its name. So for containers with `system_process` name we should use `system_process` as ID as well as due to aggregation reasons.

marceloamaral: Right, that is why we need to also include the `containerID` in the cache before updating it

marceloamaral:
@rootfs can we merge this PR?

pkg/cgroup/resolve_container.go
marceloamaral: For example, without replacing the containerID of system we have:
```
I1129 08:53:33.624567       1 container_hc_collector.go:122] 83e7046898d0316eb15b766e89127d298755a2245e974889086972e17937b3c4, comm:migration/125, inst:51677
I1129 08:53:33.624672       1 container_hc_collector.go:122] system_processes, comm:kworker/104:6, inst:15588
I1129 08:53:33.624714       1 container_hc_collector.go:122] system_processes, comm:containerd-shim, inst:13799693
I1129 08:53:33.624748       1 container_hc_collector.go:122] 99d57222fb764b18824c9cdac61e59a0650d4e26cf921bb90b6b0aaddf3766e0, comm:migration/121, inst:61072
I1129 08:53:33.624844       1 container_hc_collector.go:122] system_processes, comm:kworker/85:1, inst:75046
I1129 08:53:33.624886       1 container_hc_collector.go:122] system_processes, comm:ksoftirqd/105, inst:200038
I1129 08:53:33.624923       1 container_hc_collector.go:122] system_processes, comm:kube-controller, inst:82086859
I1129 08:53:33.624964       1 container_hc_collector.go:122] system_processes, comm:in:imklog, inst:457003
I1129 08:53:33.625003       1 container_hc_collector.go:122] system_processes, comm:kworker/136:1, inst:16339
I1129 08:53:33.625044       1 container_hc_collector.go:122] system_processes, comm:kworker/100:1H, inst:194341
```

We can see that two `comm:migration*` has containerID and then the instructions will not be aggregated in the `system_processes`....

rootfs (approved):

Issue #405, opened 2022-11-17
ModelConfig is not set by configmap/environment variables


sunya-ch:
**Describe the bug**

The current ModelConfig in model module is hard-coded and the set up in configmap is not applied. 

For example:
https://github.com/sustainable-computing-io/kepler/blob/0fa7d82084ea65b50f8ee9c631c7b1d66a0c5d52/pkg/model/container_power.go#L35-L37

In addition, with the current logic, the initEstimateFunction will try initiating local LR if the estimator sidecar failed to connect which will cause unexpected behaviour. Firstly, it is silent malfunction. Secondary, the url is linked to the zip file for sidecar estimator and cannot be interpreted as weight for local LR estimator.

https://github.com/sustainable-computing-io/kepler/blob/0fa7d82084ea65b50f8ee9c631c7b1d66a0c5d52/pkg/model/model.go#L59-L78

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

PR #406, opened 2022-11-17
set ModelConfig from config

sunya-ch:
This PR is to fix the issue https://github.com/sustainable-computing-io/kepler/issues/405 by setting the ModelConfig by config map or environment variable according to the discussion in https://github.com/sustainable-computing-io/kepler/discussions/393 and changing the `initEstimateFunction` to absolute if-else.


rootfs (approved):

rootfs:
@sunya-ch once CI issue fixed, this is lgtm

sunya-ch:
@rootfs Sorry, I forgot to add some files in the previous commits.Btw, I don't know how to deal with the Unit test error. It seems like there is a change in test file without bcc. 

This following error comes again. I think we already omitted the bcc library load before with the build tag but it seems we change to other logic now. I'm not sure how it works.

```bash
Error: vendor/github.com/iovisor/gobpf/bcc/module.go:32:10: fatal error: bcc/bcc_common.h: No such file or directory
[11](https://github.com/sustainable-computing-io/kepler/actions/runs/3493032475/jobs/5847408424#step:5:12)
   32 | #include <bcc/bcc_common.h>
~~~~~~~~~~~~~~~~~
```

rootfs:
unit test is being fixed in #409

PR #407, opened 2022-11-17
Test: Add test case

SamYuan1990:
related #354 


SamYuan1990:
too late, fell sleep.

SamYuan1990:
lesson and learn from this pr. for BCC tag, as it different with no BCC tag and some value will come up in randomly orderer.... we should be care of that when adding test case....

rootfs (approved):

Issue #349, opened 2022-11-01
end-to-end integration with model server, estimator, and kepler


wangchen615:
- [x] e2e integration
- [x] test cases for validation

rootfs:
Currently developing the e2e integration and validation through https://github.com/sustainable-computing-io/kepler-e2e-test

rootfs:
bump up this issue

SamYuan1990:
are we good enough for now?
we have test case as golang code, check both metrics and value of each metric.

rootfs:
#449 is a step to that. It gets estimator into the e2e test

rootfs:
Kepler, estimator and model server are integrated via #449. Kepler e2e test is now available.

Next step is to brainstorm on the test cases that cover all the components and are runnable on CI.

rootfs:
Track next step in #456

PR #425, opened 2022-11-28
CI: test case enhancement for e2e metric

SamYuan1990:
Following https://github.com/prometheus/prometheus/blob/main/model/textparse/promparse_test.go as prometheus handler.
Following https://onsi.github.io/ginkgo/#table-specs refactor test case.


SamYuan1990:
related with #349

SamYuan1990:
try to add bcc lib at host agent.

rootfs:
looks good, let's see what the CI's decision

SamYuan1990:
@rootfs , if I am right with those metrics, all of them should larger than 0 once bcc lib there? and could you please clarify how cgroup version works?

rootfs:
the current offline metrics based model uses cgroup v2 stats, that's why we check the kepler metrics only when cgroup is v2.

SamYuan1990:
> the current offline metrics based model uses cgroup v2 stats, that's why we check the kepler metrics only when cgroup is v2.

but if fact, those metrics are there... no matter cgroup v2 or not. for example, even if we disabled gpu there for CI. `kepler_container_gpu_joules_total` still appears.

SamYuan1990:
> the current offline metrics based model uses cgroup v2 stats, that's why we check the kepler metrics only when cgroup is v2.

so which means, cgroup version has no impact with metrics appears or not, but just impacts the value?

SamYuan1990:
> the current offline metrics based model uses cgroup v2 stats, that's why we check the kepler metrics only when cgroup is v2.

by leaving value check works as todo item in another pr make this pr ready for review.

rootfs (approved):

PR #447, opened 2022-12-06
fix: add disks handling processing

jichenjc:


pkg/collector/metric/container_metric.go
```diff
@@ -185,6 +185,8 @@ func (c *ContainerMetrics) extractUIntCurrAggr(metric string) (curr, aggr uint64
 		return c.CPUTime.Curr, c.CPUTime.Aggr, nil
 	// hardcode cgroup metrics
 	// TO-DO: merge to cgroup stat
+	case config.BlockDevicesIO:
```
marceloamaral, line 188: Can we have all the metric names defines in only one place? e.g, in the config package?
For example, `config.BlockDevicesIO`, `config.ByteRead`., etc.

jichenjc:
I can do a follow up PR on this

PR #457, opened 2022-12-09
gpu: enable gpu if the device is initialized properly

rootfs:


pkg/power/accelerator/source/gpu_nvml.go
```diff
@@ -57,7 +57,7 @@ func (n *GPUNvml) Init() (err error) {
 		err = fmt.Errorf("failed to get nvml device count: %v", nvml.ErrorString(ret))
 		return err
 	}
-	klog.V(1).Infof("found %d gpu devices\n", count)
+	fmt.Printf("found %d gpu devices\n", count)
 	devices = make([]interface{}, count)
```
jichenjc, line 61: we agreed to switch from fmt to klog
why need this change back?

rootfs: this is called in `init` function, klog is not initialized yet so klog Infof doesn't print anything at this time.

jichenjc: ok, thanks for point that out

PR #465, opened 2022-12-13
gpu power: add GPU features

rootfs:
The node total GPU usage is always 0, and thus forcing GPU energy evenly attributed on all Pods. Since GPU process utilization already accounts GPU resource usage, this PR uses GPU utilization, that is already set in the container metrics, to account GPU energy ratio.

### DIFF
--- a/pkg/model/estimator/local/ratio.go
+++ b/pkg/model/estimator/local/ratio.go
@@ -104,10 +104,12 @@ func UpdateContainerEnergyByRatioPowerModel(containersMetrics map[string]*collec
 			containerResUsage = float64(container.CounterStats[config.GpuUsageMetric].Curr)
 			nodeTotalResUsage = nodeMetrics.GetNodeResUsagePerResType(config.GpuUsageMetric)
 			nodeResEnergyUtilization = float64(nodeMetrics.GetEnergyValue(collector_metric.GPU))
-			klog.V(5).Infof("gpu power: containerID %v containerResUsage: %f, nodeTotalResUsage: %f, nodeResEnergyUtilization: %f, containerNumber: %f\n", containerID, containerResUsage, nodeTotalResUsage, nodeResEnergyUtilization, containerNumber)
 			containerGPUEnergy := getEnergyRatio(containerResUsage, nodeTotalResUsage, nodeResEnergyUtilization, containerNumber)
 			if err := containersMetrics[containerID].EnergyInGPU.AddNewCurr(containerGPUEnergy); err != nil {
 				klog.Infoln(err)
+			} else {
+				klog.V(5).Infof("gpu power ratio: containerID %v containerResUsage: %f, nodeTotalResUsage: %f, nodeResEnergyUtilization: %f, containerNumber: %f containerGPUEnergy: %v",
+					containerID, containerResUsage, nodeTotalResUsage, nodeResEnergyUtilization, containerNumber, containersMetrics[containerID].EnergyInGPU.Curr)
 			}
 		}
 
--- a/pkg/collector/metric/utils.go
+++ b/pkg/collector/metric/utils.go
@@ -18,6 +18,8 @@ package metric
 
 import (
 	"github.com/sustainable-computing-io/kepler/pkg/bpfassets/attacher"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+	"github.com/sustainable-computing-io/kepler/pkg/power/accelerator"
 
 	"k8s.io/klog/v2"
 )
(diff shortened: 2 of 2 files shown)


dagrayvid:
Thanks @rootfs, I'm seeing more reasonable numbers now on the baremetal OCP cluster with the GPU operator installed, with a simple test Pod running matrix multiplication. I am seeing between 20 and 50 Watts for the workload Pod while under load. From DCGM I see 20 Watts at idle and 70 Watts at full load, so between 20-50 seems "right".

rootfs:
thank you @dagrayvid for confirmation!

pkg/model/estimator/local/ratio.go
```diff
@@ -102,12 +102,15 @@ func UpdateContainerEnergyByRatioPowerModel(containersMetrics map[string]*collec
 		// calculate the container gpu energy consumption
 		if accelerator.IsGPUCollectionSupported() {
 			containerResUsage = float64(container.CounterStats[config.GpuUsageMetric].Curr)
-			nodeTotalResUsage = nodeMetrics.GetNodeResUsagePerResType(config.GpuUsageMetric)
+			// the utilization based accounting already takes into account the container usage at the node level
+			nodeTotalResUsage = containerResUsage
```
sunya-ch, line 107: Would does this mean the ratio (containerResUsage/nodeTotalResUsage) always 1? 
Every container will have the same GPU energy = nodeResEnergyUtilization.

pkg/model/estimator/local/ratio.go
```diff
@@ -102,12 +102,15 @@ func UpdateContainerEnergyByRatioPowerModel(containersMetrics map[string]*collec
 		// calculate the container gpu energy consumption
 		if accelerator.IsGPUCollectionSupported() {
 			containerResUsage = float64(container.CounterStats[config.GpuUsageMetric].Curr)
-			nodeTotalResUsage = nodeMetrics.GetNodeResUsagePerResType(config.GpuUsageMetric)
+			// the utilization based accounting already takes into account the container usage at the node level
```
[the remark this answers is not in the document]
rootfs: sounds good, comment added

pkg/model/estimator/local/ratio.go
sunya-ch: I think instead of avoiding gpu metric in usage = 0.
We need to add the `config.GPUMemUtilization` in the feature list in the following function same as the other metric. https://github.com/sustainable-computing-io/kepler/blob/34473a4ba7d193af039e873915337ca2059a689e/pkg/collector/metric/utils.go#L25.

https://github.com/sustainable-computing-io/kepler/blob/34473a4ba7d193af039e873915337ca2059a689e/pkg/collector/metric/utils.go#L53

https://github.com/sustainable-computing-io/kepler/blob/f5a971fa2894f1fbe7de1d245ceea2d4701282ab/pkg/collector/metric/node_metric.go#L165-L175

sunya-ch (commented):
I have a concern on setting `nodeTotalResUsage = containerResUsage`.

pkg/model/estimator/local/ratio.go
rootfs: > Would does this mean the ratio (containerResUsage/nodeTotalResUsage) always 1?

@sunya-ch when the container doesn't use gpu, the containerResUsage is 0, so nodeTotalResUsage is 0 and this container only gets the idle power. When the container that uses GPU is counted, `containerResUsage/nodeTotalResUsage` is 1, then this container gets the dynamic power.

sunya-ch: I see. I think it looks good to me now. 
Also, even with the shared GPU case, things might be more complex with the utilization and cannot just treat as the same as the other metric.

sunya-ch (approved):

marceloamaral (changes requested):
After the @sunya-ch comment, I think we should fix the bug instead of remove the sum of containers.

pkg/model/estimator/local/ratio.go
rootfs: yes, GPU sharing will be a totally different solution, we probably will use similar [ML](http://users.ece.utexas.edu/~derek/Papers/HPCA2015_GPUPowerModel.pdf) based approach.

rootfs:
@marceloamaral can you elaborate?

pkg/model/estimator/local/ratio.go
marceloamaral: Of course we can have different power models (ratio, ML, LR, etc).
But if the GPU partitioning can fully isolate applications, this might be vary similar to the container use case.

rootfs:
@sunya-ch second thought, your fix is actually a better one. Reverted mine and used yours

marceloamaral (approved):
/lgtm

PR #471, opened 2022-12-15
Add support to find acpi sensor power without hwmon

marceloamaral:
My test environment doesn't have the power_average hwmon file, but it does have the acpi node power metric.

This PR finds the original power_average acpi file when hwmon is not available.

By the way, even though I calibrated my system, hwmon still didn't report the file.

After this PR, I can see the node platform energy consumption.


rootfs:
there are some reported issues on kernel acpi power meter. Can you add an option so this search can be turned off on unsupported kernel and HW?

marceloamaral:
> there are some reported issues on kernel acpi power meter. Can you add an option so this search can be turned off on unsupported kernel and HW?

@rootfs Which issues? The code try to find the paths, if they do not exist, no node power metric will be collected.
So, if there are some annoying messages, we can try to avoid them.
Which messages are you seen?

marceloamaral:
@rootfs, I saw you message describing the problem. Thanks.
https://www.suse.com/support/kb/doc/?id=000017865

marceloamaral:
@rootfs I am handling the error and stop to collect acpi node power if we cannot read the file because of the kernel bug.

PR #475, opened 2022-12-16
Create Release v0.4 GH Workflow

husky-parul:
Signed-off-by: Parul <parsingh@redhat.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,82 @@
+name: Release
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      tag:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+    - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
```
rootfs, line 5: so this action only starts when a tag is created?

husky-parul:
I am adding create tag steps and push image steps

rootfs (approved):

PR #481, opened 2023-01-03
fix: switch from dummy to estimated method

jichenjc:
estimate need files 

"/var/lib/kepler/data/normalized_cpu_arch.csv"   ==> from docker file 
"/var/lib/kepler/data/power_data.csv" ==> from docker file 
"/proc/meminfo" ==> host level, same to other /proc usage in kepler

### DIFF
--- a/pkg/bpfassets/perf_event_bindata.go
+++ b/pkg/bpfassets/perf_event_bindata.go
@@ -367,11 +367,13 @@ var _bindata = map[string]func() (*asset, error){
 // directory embedded in the file by go-bindata.
 // For example if you run go-bindata on data/... and data contains the
 // following hierarchy:
-//     data/
-//       foo.txt
-//       img/
-//         a.png
-//         b.png
+//
+//	data/
+//	  foo.txt
+//	  img/
+//	    a.png
+//	    b.png
+//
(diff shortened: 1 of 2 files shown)


rootfs (approved):

rootfs:
@jichenjc looks good. Now we don't need [kepler/data](https://github.com/sustainable-computing-io/kepler/tree/main/data) and [their container image copy](https://github.com/sustainable-computing-io/kepler/blob/main/build/Dockerfile#L36-L39), can you remove them too?

rootfs:
@jichenjc please also remove https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/components/source/estimate.go

jichenjc:
>>>@jichenjc please move [GetCPUArchitecture](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/components/source/estimate.go#L62) from [estimate.go](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/components/source/estimate.go) and remove the rest of the file.

um... I switch from dummy to estimated .. then why remove this file? maybe I misunderstand your comment?

rootfs:
the estimate method is not used any more, can you follow up with a new PR to remove this method? Thanks.

PR #492, opened 2023-01-10
exporter: run exporter as non root user

### LINKED ISSUE #491: Allow Kepler to be deployed in managed Kubernetes clusters without root privilege
Now Kepler uses EBPF, which needs root privilege to access the worker node to get those counters. That does not work for managed cluster services, such as IKS, AKS, etc.

**Describe the solution you'd like**
Here is a possible solution to get round the root privilege requirements.

```Unprivileged users inside the container need CAP_SYS_ADMIN permission already assigned to the container to run a malicious eBPF program. For Linux kernels 5.8 and above, a new permission, CAP_BPF, is added to allow users to run eBPF programs. CAP_BPF is a subset of CAP_SYS_ADMIN.```

https://www.tigera.io/blog/cve-2021-31440-kubernetes-container-escape-using-ebpf/

**Describe alternatives you've considered**
e.g. using CAP_SYS_ADMIN

**Additional context**

### PR DESCRIPTION
**Pull request by rootfs**
fix #491

### DIFF
--- a/manifests/config/exporter/exporter.yaml
+++ b/manifests/config/exporter/exporter.yaml
@@ -52,7 +52,13 @@ spec:
             memory: "400Mi"
             cpu: "100m"
         securityContext:
-          privileged: true
+           runAsUser: 2000
+           privileged: true
+           capabilities:
+             add:
+             # - BPF is only available in kernel 5.8+
+             - NET_ADMIN
+             - SYS_ADMIN
         command:
         - /bin/sh
         - -c
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -7,7 +7,7 @@ ARG MAKE_TARGET=cross-build-linux-$ARCH
 ARG BIN_TIMESTAMP
 ARG SOURCE_GIT_TAG
 
-USER root
+#USER root
 
 LABEL name=kepler-builder
 
--- a/build/Dockerfile.builder
+++ b/build/Dockerfile.builder
@@ -1,6 +1,6 @@
 FROM quay.io/sustainable_computing_io/kepler_base:ubi-8.6-bcc-0.24 as builder
 
-USER root
+#USER root
 
 LABEL name=kepler-builder
 

### CODE REVIEW

**Comment by rootfs** [comment]
this is the result
```console
# kubectl exec -ti -n kepler daemonset/kepler-exporter -c kepler-exporter id
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
uid=2000(2000) gid=2000 groups=2000
```

rootfs:
It looks `kind` doesn't behave the same

rootfs:
integration test failed with this kepler log msg
```
could not open bpf map: cpu_freq_array, error: Operation not permitted
```

This is related to the memory size limit to privileged users, as explained in https://github.com/iovisor/bcc/pull/279

rootfs:
it looks GH Action runner cannot resize memlock (reconfiguring docker doesn't help). I need a new CI kustomization to use `root` user.

manifests/config/exporter/exporter.yaml
```diff
@@ -52,7 +52,17 @@ spec:
             memory: "400Mi"
             cpu: "100m"
         securityContext:
-          privileged: true
+           runAsUser: 2000
```
jichenjc, line 55: I followed update here and seems I can setup it locally 

```
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
uid=2000(2000) gid=2000 groups=2000
```

so the TODO for this is to make our CI able to run with non-root user ,but you encounter some problem ,correct?

rootfs: yes, non-privileged root needs to have 1M memlock (ulimit -l 10240) to run ebpf. But Github Action cannot set it. So we have to have a kustomize for non root user. @jichenjc can you take this into a new PR?

PR #494, opened 2023-01-11
ebpf: fix cpu time and HW stats for running processes

### LINKED ISSUE #487: Incorrect per-Pod power usage for test benchmark
**Describe the bug**
I'm running the stress-ng benchmark for a 10 minute runtime, configured to stress ~50% of the CPUs available in my system, and seeing <5 watts attributed to the workload Pod from the grafana dashboard. 

Under load vs at idle I can see a delta of 200 Watts from the iDRAC. From prometheus I see a similar delta for irate(kepler_node_package_joules_total[1m]).  As a sanity check, I also verified that there is >50% cpu usage on the node, and and 60 cores are used according to `pod:container_cpu_usage:sum` in Prometheus.

This is running on one worker node in an OpenShift cluster with 3 control-plane and 2 worker-nodes, all with Intel Icelake CPUs.

**To Reproduce**
I'm running stress-ng with the [benchmark-operator](https://github.com/cloud-bulldozer/benchmark-operator). Under the hood, this runs stress-ng in a container with this job file:
```
run parallel  # run stressors sequentially
verbose	      # verbose output
metrics-brief # show metrics at end of each run
timeout 480   # stop each stressor after 480 seconds
cpu 64        # cpu stressors
cpu-load 100  # percentage to which the cpu gets loaded 
```

**Expected behavior**
>100 Watts attributed to the stress-ng workload Pod.

**Screenshots**
Kepler dashboard:
image

Kepler dashboard all namespaces:
image

iDRAC:
image

Total Node power usage according to kepler per-node metrics:
image

k8s hardware resource usage during run:
image

/cc @rootfs @sunya-ch

rootfs:
fix #487

### DIFF
--- a/bpfassets/perf_event/perf_event.c
+++ b/bpfassets/perf_event/perf_event.c
@@ -46,7 +46,7 @@ typedef struct pid_time_t
 } pid_time_t;
 
 // processes and pid time
-BPF_HASH(processes, u32, process_metrics_t);
+BPF_HASH(processes, u64, process_metrics_t);
 BPF_HASH(pid_time, pid_time_t);
 
 // perf counters
@@ -188,9 +188,7 @@ static inline u64 get_on_cpu_avg_freq(u32 *cpu_id, u64 on_cpu_cycles_delta, u64 
 // int kprobe__finish_task_switch(switch_args *ctx)
 int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 {
-    // u64 cur_pid = bpf_get_current_pid_tgid() & 0xffffff;
-    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
-    u32 cur_pid = task->tgid;
+    u64 cur_pid = bpf_get_current_pid_tgid() >> 32;
 #ifdef SET_GROUP_ID
     u64 cgroup_id = bpf_get_current_cgroup_id();
 #else
@@ -199,8 +197,8 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 
     u64 cur_ts = bpf_ktime_get_ns();
     u32 cpu_id = bpf_get_smp_processor_id();
-    u32 prev_tgid = prev->tgid;
-    u64 on_cpu_time_delta = get_on_cpu_time(cur_pid, prev_tgid, cpu_id, cur_ts);
+    u64 prev_pid = prev->pid;
+    u64 on_cpu_time_delta = get_on_cpu_time(cur_pid, prev_pid, cpu_id, cur_ts);
     u64 on_cpu_cycles_delta = get_on_cpu_cycles(&cpu_id);
     u64 on_cpu_ref_cycles_delta = get_on_cpu_ref_cycles(&cpu_id);
     u64 on_cpu_instr_delta = get_on_cpu_instr(&cpu_id);
@@ -209,7 +207,7 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 
     // store process metrics
     struct process_metrics_t *process_metrics;
-    process_metrics = processes.lookup(&prev_tgid);
+    process_metrics = processes.lookup(&prev_pid);
     if (process_metrics == 0)
     {
         process_metrics_t new_process = {};
--- a/bpfassets/perf_event/perf_event.c
+++ b/bpfassets/perf_event/perf_event.c
@@ -208,21 +208,7 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
     // store process metrics
     struct process_metrics_t *process_metrics;
     process_metrics = processes.lookup(&prev_pid);
-    if (process_metrics == 0)
-    {
-        process_metrics_t new_process = {};
-        new_process.pid = cur_pid;
-        new_process.cgroup_id = cgroup_id;
-        new_process.process_run_time = on_cpu_time_delta;
-        bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
-
-        new_process.cpu_cycles = on_cpu_cycles_delta;
-        new_process.cpu_instr = on_cpu_instr_delta;
-        new_process.cache_miss = on_cpu_cache_miss_delta;
-
-        processes.update(&cur_pid, &new_process);
-    }
-    else
+    if (process_metrics != 0)
     {
         // update process time
         process_metrics->process_run_time += on_cpu_time_delta;
@@ -230,6 +216,16 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
         process_metrics->cpu_cycles += on_cpu_cycles_delta;
         process_metrics->cpu_instr += on_cpu_instr_delta;
         process_metrics->cache_miss += on_cpu_cache_miss_delta;
+    }
+
+    process_metrics = processes.lookup(&cur_pid);
+    if (process_metrics == 0)
+    {
+        process_metrics_t new_process = {};
+        new_process.pid = cur_pid;
+        new_process.cgroup_id = cgroup_id;
+        bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
+        processes.update(&cur_pid, &new_process);
     }
 
     return 0;
--- a/bpfassets/perf_event/perf_event.c
+++ b/bpfassets/perf_event/perf_event.c
@@ -46,7 +46,7 @@ typedef struct pid_time_t
 } pid_time_t;
 
 // processes and pid time
-BPF_HASH(processes, u64, process_metrics_t);
+BPF_HASH(processes, u32, process_metrics_t);
 BPF_HASH(pid_time, pid_time_t);
 
 // perf counters
@@ -65,12 +65,12 @@ BPF_ARRAY(cache_miss, u64, NUM_CPUS);
 // cpu freq counters
 BPF_ARRAY(cpu_freq_array, u32, NUM_CPUS);
 
-static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
+static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u32 cpu_id, u64 cur_ts)
 {
     u64 cpu_time = 0;
 
     // get pid time
-    pid_time_t prev_pid_key = {.pid = prev_pid};
+    pid_time_t prev_pid_key = {.pid = prev_pid, .cpu = cpu_id};
     u64 *prev_ts = pid_time.lookup(&prev_pid_key);
     if (prev_ts != 0)
     {
@@ -83,7 +83,7 @@ static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
             pid_time.delete(&prev_pid_key);
         }
     }
-    pid_time_t new_pid_key = {.pid = cur_pid};
+    pid_time_t new_pid_key = {.pid = cur_pid, .cpu = cpu_id};
     pid_time.update(&new_pid_key, &cur_ts);
 
     return cpu_time;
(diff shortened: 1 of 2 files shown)


bpfassets/perf_event/perf_event.c
```diff
@@ -206,7 +209,7 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 
     // store process metrics
     struct process_metrics_t *process_metrics;
-    process_metrics = processes.lookup(&cur_pid);
+    process_metrics = processes.lookup(&prev_tgid);
```
rootfs, line 212: per [tutorial](https://www.brendangregg.com/offcpuanalysis.html), 
```
finish_task_switch() function is called in the context of the next-running thread. 
```

and reference the oncpu analysis code [here](https://github.com/iovisor/bcc/blob/master/tools/cpudist.py)

rootfs:
Running kepler in console shows it is able to capture the stressng better. The following shows that in two metric queries within 5 seconds, the stressng power is the 2nd most heavy Pod, its power is (1241.147-1205.418)J/5sec=7W, while the system process power is (4251.894-4156.351)J/5sec=19W
```console
# while true; do curl localhost:8888/metrics 2>&1 |grep kepler_container_package_joules_total |sort -k 2 -g |tail -2; sleep 5;echo "-------";done
kepler_container_package_joules_total{command="stress-ng",container_name="stressng",container_namespace="benchmark-operator",pod_name="stressng-workload-e36f1e7a-zvc5n"} 1205.418
kepler_container_package_joules_total{command="",container_name="system_processes",container_namespace="system",pod_name="system_processes"} 4156.351
-------
kepler_container_package_joules_total{command="stress-ng",container_name="stressng",container_namespace="benchmark-operator",pod_name="stressng-workload-e36f1e7a-zvc5n"} 1241.147
kepler_container_package_joules_total{command="",container_name="system_processes",container_namespace="system",pod_name="system_processes"} 4251.894
```

bpfassets/perf_event/perf_event.c
```diff
@@ -197,7 +199,8 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 
     u64 cur_ts = bpf_ktime_get_ns();
     u32 cpu_id = bpf_get_smp_processor_id();
-    u64 on_cpu_time_delta = get_on_cpu_time(cur_pid, prev->pid, cur_ts);
+    u32 prev_tgid = prev->tgid;
```
marceloamaral, line 202: `u32 prev_pid = prev->pid;`

bpfassets/perf_event/perf_event.c
```diff
@@ -197,7 +199,8 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
...
+    u64 on_cpu_time_delta = get_on_cpu_time(cur_pid, prev_tgid, cpu_id, cur_ts);
```
marceloamaral, line 203: `get_on_cpu_time(cur_pid, prev_pid, cpu_id, cur_ts);`

bpfassets/perf_event/perf_event.c
marceloamaral, line 212: process_metrics = processes.lookup(&prev_pid);

marceloamaral (changes requested):
Thanks @rootfs, this makes a lot of sense and fixed some issues in my local experiments.

We should use `pid` instead of `tgid` to improve the accuracy of the measurements. Using `tgid` to identify tasks within the same group would lead to incorrect calculations of CPU time, as it would use the same start time for multiple tasks with different start times. (It can probably also impact other metrics)

rootfs:
sounds good @marceloamaral Let's use pid then

bpfassets/perf_event/perf_event.c
```diff
@@ -188,7 +188,9 @@ static inline u64 get_on_cpu_avg_freq(u32 *cpu_id, u64 on_cpu_cycles_delta, u64
 // int kprobe__finish_task_switch(switch_args *ctx)
 int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 {
-    u64 cur_pid = bpf_get_current_pid_tgid() >> 32;
+    // u64 cur_pid = bpf_get_current_pid_tgid() & 0xffffff;
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    u32 cur_pid = task->tgid;
 #ifdef SET_GROUP_ID
     u64 cgroup_id = bpf_get_current_cgroup_id();
```
marceloamaral, line 195: We probably cannot use the current cgroup id, but we should also use the previous cgroup id!
This would require a ebpf hash to track the previous cgroup id in the given cpu.

marceloamaral: This is also the case of the command (wrongly reported). However, since we aggregate all system processes, we probably do not need the command. So we could remove it to minimize overheads.

rootfs: Based on the info below, the function runs in the new thread context, so `cgroup_id = bpf_get_current_cgroup_id();` should get the new pid's cgroup.

```The finish_task_switch() function is called in the context of the next-running thread. An eBPF program can instrument this function and argument using kprobes, fetch the current PID (via bpf_get_current_pid_tgid()), and also fetch a high resolution timestamp (bpf_ktime_get_ns())```

File: bpfassets/perf_event/perf_event.c [hunk 5]
```diff
@@ -206,7 +207,7 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
 
     // store process metrics
     struct process_metrics_t *process_metrics;
-    process_metrics = processes.lookup(&cur_pid);
+    process_metrics = processes.lookup(&prev_pid);
```
marceloamaral, line 210: the `new_process.pid = cur_pid;` should be:
`new_process.pid = prev_pid;`

after this change we can merge this PR

bpfassets/perf_event/perf_event.c
marceloamaral: Right, since "`the finish_task_switch() function is called in the context of the next-running thread` ", and the collected metrics are for the previous thread, we need to use the pid and cgroup id of the previous thread.

rootfs:
tested `stressng`, the spike of stressng pod matches the activity 
image

bpfassets/perf_event/perf_event.c
rootfs: updated. Now the `new_process.pid` is only created for cur_pid.

bpfassets/perf_event/perf_event.c
```diff
@@ -206,30 +207,26 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
...
+    {
+        // update process time
+        process_metrics->process_run_time += on_cpu_time_delta;
+
+        process_metrics->cpu_cycles += on_cpu_cycles_delta;
+        process_metrics->cpu_instr += on_cpu_instr_delta;
+        process_metrics->cache_miss += on_cpu_cache_miss_delta;
+    }
+
     process_metrics = processes.lookup(&cur_pid);
```
marceloamaral, line 221: I believe there is no need to update this. The only necessary update was to change the process ID to use the previous task. It's important to note that we should only update the metrics for the context of the previous task, as we were only collecting metrics for that task. There will be no data from the current task, so there is no need to add it to the hash at this time. Note that adding it to the hash would only cause the metrics to accumulate for the next context switch when it becomes the previous task.

Given this information, it is worth noting that we may encounter a loss of data with the new logic as it does not incorporate the current data of the previous task into the hash. As we have done previously, a new instance should be created within the hash.

rootfs: The new logic doesn't change the way `prev_pid` gets updated: as long as the `prev_pid` is there, its HW counters and cpu time are updated. 

The change only applies for `new_pid` (i.e. not previously seen before). In this case, since the new_pid has not started, only the pid and cgroup are recorded.

bpfassets/perf_event/perf_event.c
```diff
@@ -206,30 +207,26 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
...
     if (process_metrics == 0)
     {
         process_metrics_t new_process = {};
         new_process.pid = cur_pid;
```
marceloamaral, line 225: the should be prev_pid
`new_process.pid = prev_pid;`

bpfassets/perf_event/perf_event.c
```diff
@@ -206,30 +207,26 @@ int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
...
         new_process.cgroup_id = cgroup_id;
-        new_process.process_run_time = on_cpu_time_delta;
         bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
-
-        new_process.cpu_cycles = on_cpu_cycles_delta;
-        new_process.cpu_instr = on_cpu_instr_delta;
-        new_process.cache_miss = on_cpu_cache_miss_delta;
-
         processes.update(&cur_pid, &new_process);
```
marceloamaral, line 228: This also should be the `prev_pid` and we should add the initial values
```
        new_process.process_run_time = on_cpu_time_delta;
        new_process.cpu_cycles = on_cpu_cycles_delta;
        new_process.cpu_instr = on_cpu_instr_delta;
        new_process.cache_miss = on_cpu_cache_miss_delta;
        processes.update(&prev_pid, &new_process);
```

bpfassets/perf_event/perf_event.c
marceloamaral: We should remove line 221
`process_metrics = processes.lookup(&cur_pid);`

bpfassets/perf_event/perf_event.c
marceloamaral: If we do not add the initial values, we will loose it

marceloamaral: The full example of the code:
```c
int kprobe__finish_task_switch(struct pt_regs *ctx, struct task_struct *prev)
{
    // u32 cur_pid = bpf_get_current_pid_tgid() >> 32;
    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
    u32 cur_pid = task->pid;    
#ifdef SET_GROUP_ID
    u64 cgroup_id = bpf_get_current_cgroup_id();
#else
    u64 cgroup_id = 0;
#endif
    u32 prev_pid = prev->pid;

    u64 cur_ts = bpf_ktime_get_ns();
    u32 cpu_id = bpf_get_smp_processor_id();
    u64 on_cpu_time_delta = get_on_cpu_time(cur_pid, prev_pid, cur_ts);
    u64 on_cpu_cycles_delta = get_on_cpu_cycles(&cpu_id);
    u64 on_cpu_ref_cycles_delta = get_on_cpu_ref_cycles(&cpu_id);
    u64 on_cpu_instr_delta = get_on_cpu_instr(&cpu_id);    
    u64 on_cpu_cache_miss_delta = get_on_cpu_cache_miss(&cpu_id);
    u64 on_cpu_avg_freq = get_on_cpu_avg_freq(&cpu_id, on_cpu_cycles_delta, on_cpu_ref_cycles_delta);

    // store process metrics
    struct process_metrics_t *process_metrics;
    process_metrics = processes.lookup(&prev_pid);
    if (process_metrics == 0)
    {
        process_metrics_t new_process = {};
        new_process.pid = prev_pid;
        new_process.cgroup_id = cgroup_id;
        bpf_probe_read_kernel(&new_process.comm, sizeof(new_process.comm), &prev->comm);

        new_process.process_run_time = on_cpu_time_delta;
        new_process.cpu_cycles = on_cpu_cycles_delta;
        new_process.cpu_instr = on_cpu_instr_delta;
        new_process.cache_miss = on_cpu_cache_miss_delta;

        processes.update(&prev_pid, &new_process);
    }
    else
    {
        // update process time
        process_metrics->process_run_time += on_cpu_time_delta;
        process_metrics->cpu_cycles += on_cpu_cycles_delta;
        process_metrics->cpu_instr += on_cpu_instr_delta;
        process_metrics->cache_miss += on_cpu_cache_miss_delta;
    }

    return 0;
}
```

rootfs:
```c
struct process_metrics_t *process_metrics;
    process_metrics = processes.lookup(&prev_pid);
    if (process_metrics != 0)
    {
        // update process time
        process_metrics->process_run_time += on_cpu_time_delta;

        process_metrics->cpu_cycles += on_cpu_cycles_delta;
        process_metrics->cpu_instr += on_cpu_instr_delta;
        process_metrics->cache_miss += on_cpu_cache_miss_delta;
    }

    process_metrics = processes.lookup(&cur_pid);
    if (process_metrics == 0)
    {
        process_metrics_t new_process = {};
        new_process.pid = cur_pid;
        new_process.cgroup_id = cgroup_id;
        bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
        processes.update(&cur_pid, &new_process);
    }
```

marceloamaral: Ok! I believe your current approach is more effective as it correctly set the cgroup id.

marceloamaral (approved):
/lgtm

PR #498, opened 2023-01-17
feature: rootless deployment

jichenjc:
Fixes: #491

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ ARCH :=$(shell uname -m |sed -e "s/x86_64/amd64/" |sed -e "s/aarch64/arm64/")
 GIT_VERSION     := $(shell git describe --dirty --tags --always --match='v*')
 VERSION         ?= $(GIT_VERSION)
 LDFLAGS         := "-w -s -X 'github.com/sustainable-computing-io/kepler/pkg/version.Version=$(VERSION)'"
+ROOTLESS	?= false
 
 ifdef IMAGE_REPO
 	IMAGE_REPO := $(IMAGE_REPO)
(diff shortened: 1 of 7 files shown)


jichenjc:
need update https://sustainable-computing.io/installation/kepler/ later on, basically we can 

default is `false` which means we use root in CI, then if you need non-root, use following 
```
ROOTLESS=true make
```

and generate manifest is 
```
OPTS="ROOTLESS" make build-manifest
```

then we have 

```
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
            - SYS_ADMIN
            - SYS_RESOURCE
          privileged: true
          runAsGroup: 2000
          runAsUser: 2000
          seLinuxOptions:
            level: s0

```
in `_output/generated-manifest/deployment.yaml`

jichenjc:
@rootfs FYI

jichenjc:
currently we are using 
```
- NET_ADMIN
            - SYS_ADMIN
            - SYS_RESOURCE
```

maybe we should also consider check whether they are using 5.8+ kernel then use BPF priv ...

Makefile
```diff
@@ -90,14 +91,22 @@ cross-build: cross-build-linux-amd64 cross-build-linux-arm64
 _build_containerized: genbpfassets tidy-vendor format
 	@if [ -z '$(CTR_CMD)' ] ; then echo '!! ERROR: containerized builds require podman||docker CLI, none found $$PATH' >&2 && exit 1; fi
 	echo BIN_TIMESTAMP==$(BIN_TIMESTAMP)
+
+	rm -f $(SRC_ROOT)/build/Dockerfile.tmp
+	cp $(SRC_ROOT)/build/Dockerfile $(SRC_ROOT)/build/Dockerfile.tmp
+
+	@if [ '$(ROOTLESS)' = 'false' ] ; then echo "user root" >> $(SRC_ROOT)/build/Dockerfile.tmp; fi
```
rootfs, line 98: we don't need to add `user root`. If no user is defined, the default is then root.

jichenjc: ok, updated

rootfs:
@jichenjc thanks, can you also update the doc [here](https://github.com/sustainable-computing-io/kepler-doc/blob/main/docs/installation/kepler.md#build-manifests)?

jichenjc:
yes, will update , thanks~

PR #508, opened 2023-01-24
Add support of Idle and Dynamic energy calculation

marceloamaral:
Measuring idle power consumption is important for getting an accurate picture of overall energy usage. 

It's important to measure the idle and dynamic energy separately so that the energy consumption of resource-intensive processes can be evaluated fairly. For example, It's not fair to assign all idle energy to a process just considering the resource utilization since the idle power does not vary with the utilization.

Therefore, this PR adds the support of Idle and Dynamic energy calculation for a container.
The Idle energy is calculated by using the minimal observed energy. In my experiments, few seconds are enough to get the idle state.

Currently the model server only calculates the dynamic energy, so that the overall metrics will more transparent.

For the prometheus metrics, this PR introduces a new label called `mode` which shows if the energy is idle or dynamic:
```
# HELP kepler_node_package_joules_total Aggregated RAPL value in package (socket) in joules
# TYPE kepler_node_package_joules_total counter
kepler_node_package_joules_total{instance="node1",mode="dynamic",package="0",source="rapl"} 0.005
kepler_node_package_joules_total{instance="node1",mode="idle",package="0",source="rapl"} 0.005
```

### DIFF
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -34,12 +34,14 @@ import (
 )
 
 const (
-	CORE   = "core"
-	DRAM   = "dram"
-	UNCORE = "uncore"
-	PKG    = "pkg"
-	GPU    = "gpu"
-	OTHER  = "other"
+	CORE      = "core"
+	DRAM      = "dram"
+	UNCORE    = "uncore"
+	PKG       = "pkg"
+	GPU       = "gpu"
+	OTHER     = "other"
+	PLATFORM  = "platform"
+	FREQUENCY = "frequency"
 )
 
 var (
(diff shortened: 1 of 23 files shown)


pkg/collector/metric/node_metric.go
```diff
@@ -197,54 +278,146 @@ func (ne *NodeMetrics) AddNodeComponentsEnergy(componentsEnergy map[int]source.N
...
-	case UNCORE:
-		val = ne.EnergyInUncore.Curr()
-	case PKG:
-		val = ne.EnergyInPkg.Curr()
-	case GPU:
-		val = ne.EnergyInGPU.Curr()
-	case OTHER:
-		val = ne.EnergyInOther.Curr()
+func (ne *NodeMetrics) UpdateIdleEnergy() {
+	for pkgID := range ne.TotalEnergyInPkg.Stat {
```
rootfs, line 286: can you add a comment her detailing the idea of how to set the idle power?

rootfs:
@marceloamaral Is the idle power a variable metrics? It is intuitive to think that idle power is constant. Do you happen to have a sample of idle power over time? It would be great if you can also add a note there explaining what makes idle power change so end users have some idea how to use it in practice.

marceloamaral:
Thanks @rootfs for the comments.

> Is the idle power a variable metrics? It is intuitive to think that idle power is constant. 

The idle power is the minimal energy consumption that the system present at idle state. Therefore we search for the minimal idle power when the cpu utilization is also minimal. 

In fact, the idle power of the system (for example CPU package (socket) idle power) will be close to constant. Although Idle power could potentially vary with the node temperature, we are not covering it now, because in practice it is hard to happen.

However, the container idle power will vary because it will be defined by the number of running containers in the cluster. So a cluster with more containers, each container will have lower idle power. 

>Do you happen to have a sample of idle power over time? 

Sure, consider the following example, where I keep the cluster idle for 2 minutes and run 3 cpu intensive workloads. You can see that after we discover the idle power, it does not change.
image

>It would be great if you can also add a note there explaining what makes idle power change so end users have some idea how to use it in practice.

Right! We need to update the documentation! I will add information about it.

Regarding the end user, I think the common use case will be using the aggregation, dynamic+idle.

rootfs (approved):

rootfs:
@sunya-ch do you like to take a look? Thanks

sunya-ch:
I take a quick look on the changes. The changes look good to me. 
I may add other modes in addition to "dynamic" when we apply different isolation approach which I will elaborate about that later.

sunya-ch (approved):
/lgtm

PR #509, opened 2023-01-25
dev: local env set up

SamYuan1990:


rootfs:
@SamYuan1990 it would be better if we can move the `cluster-up and sync` to its own repo (I am going to do that). The cluster repo will have its own release cycle so github action can get the artifact from there. And kepler can use that repo to get the scripts.

rootfs:
@SamYuan1990 the repo is here https://github.com/sustainable-computing-io/local-dev-cluster

SamYuan1990:
> https://github.com/sustainable-computing-io/local-dev-cluster

wait a min

.github/workflows/local_env.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        kube_provider: [kind]
+    steps:
+      - uses: actions/checkout@main
+      - name: kubectl
+        run: curl -LO https://dl.k8s.io/release/v1.25.4/bin/linux/amd64/kubectl
+      - name: local cluster set up
+        run: make cluster-up
```
SamYuan1990, line 17: @marceloamaral here you are.

rootfs (approved):

PR #513, opened 2023-01-27
collector: add kepler_process metrics

### LINKED ISSUE #512: add kepler_process metrics
The Kepler container metrics are aggregated at container level. A process level aggregation is needed for non-containerized workload. 
**Describe the solution you'd like**
Add a kepler_process metric prefix, if process level aggregation is enabled.

**Describe alternatives you've considered**

**Additional context**
As part of #511 effort

rootfs:
fix #512 

In Kepler, all processes that don't have container IDs are classified as `system process` container. So we can use this special container to calculate process based energy.

In this PR, those processes that do not have container IDs are allocated into `kepler_process` metrics. These processes' aggregate energy consumption is based on `system process` container.

### DIFF
--- a/pkg/collector/metric/process_metric.go
+++ b/pkg/collector/metric/process_metric.go
@@ -44,7 +44,7 @@ type ProcessMetrics struct {
 	CounterStats map[string]*UInt64Stat
 	// ebpf metrics
 	CPUTime           *UInt64Stat
-	SoftIQRCount      []UInt64Stat
+	SoftIRQCount      []UInt64Stat
 	GPUStats          map[string]*UInt64Stat
 	DynEnergyInCore   *UInt64Stat
 	DynEnergyInDRAM   *UInt64Stat
@@ -68,7 +68,7 @@ func NewProcessMetrics(pid uint64, command string) *ProcessMetrics {
 		Command:            command,
 		CPUTime:            &UInt64Stat{},
 		CounterStats:       make(map[string]*UInt64Stat),
-		SoftIQRCount:       make([]UInt64Stat, config.MaxIRQ),
+		SoftIRQCount:       make([]UInt64Stat, config.MaxIRQ),
 		DynEnergyInCore:    &UInt64Stat{},
 		DynEnergyInDRAM:    &UInt64Stat{},
 		DynEnergyInUncore:  &UInt64Stat{},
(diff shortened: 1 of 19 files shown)


rootfs:
This is a sample output of process energy metrics
```
kepler_process_core_joules_total{command="dbus-daemo",mode="dynamic",pid="1120"} 20.489
kepler_process_core_joules_total{command="prometheus",mode="dynamic",pid="3094917"} 23.065
kepler_process_core_joules_total{command="top",mode="dynamic",pid="1190893"} 25.347
kepler_process_core_joules_total{command="node",mode="dynamic",pid="2424966"} 43.318
kepler_process_core_joules_total{command="swapper/5",mode="dynamic",pid="0"} 4964.944
```

rootfs:
Once `kepler_process` metrics are turned on, scrapping puts quite some pressure on CPU
image

However, without scrapping, kepler's cpu usage is not as high
image

This puts the [Prometheus client scaling discussion](https://github.com/sustainable-computing-io/kepler/discussions/439) to the front line again. Before we have a scalable Prometheus client, we can only export a subset of process metrics

pkg/collector/metric/process_metric.go
```diff
@@ -0,0 +1,210 @@
...
+
+type ProcessMetrics struct {
+	PID          uint64
+	Command      string
+	CounterStats map[string]*UInt64Stat
+	// ebpf metrics
+	CPUTime           *UInt64Stat
+	SoftIQRCount      []UInt64Stat
+	GPUStats          map[string]*UInt64Stat
+	DynEnergyInCore   *UInt64Stat
```
KaiyiLiu1234, line 49: If power metrics are measured in int, would that mean the model server when returning a prediction should have the prediction it rounded to int (currently model server returns predictions in float)?

pkg/collector/metric/process_metric.go
```diff
@@ -0,0 +1,210 @@
...
+	ProcessFeaturesNames []string
+)
+
+type ProcessMetrics struct {
+	PID          uint64
+	Command      string
+	CounterStats map[string]*UInt64Stat
+	// ebpf metrics
+	CPUTime           *UInt64Stat
+	SoftIQRCount      []UInt64Stat
```
KaiyiLiu1234, line 47: Shouldn't this be SoftIRQ instead of SoftIQR

pkg/collector/metric/process_metric.go
rootfs: yes, they are rounded to int

pkg/collector/metric/process_metric.go
rootfs: nice catch!

pkg/collector/prometheus_process_collector.go
```diff
@@ -0,0 +1,285 @@
...
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/sustainable-computing-io/kepler/pkg/bpfassets/attacher"
+	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+)
+
+type processDesc struct {
```
KaiyiLiu1234, line 30: My only other concern is that like other containers, system process container looks like it also records Cgroup stats. However this is not being exported in the general prometheus collector for new metrics or in the prometheus_process_collector.go as referenced. In a future PR, can we consider including CGroup related metrics instead of only the Hardware Counters.

rootfs: It is a debatable use case. On environment no cgroups are configured, the cgroup metrics may not be available for processes.

KaiyiLiu1234: I see. In that case soft irq metrics and counter metrics are the way to go. The reason I ask is because we are exploring setting up package power models for counter metrics, cgroup features, and irq metrics.

marceloamaral:
@rootfs could you please explain it better?

Is the sum of all values in `kepler_container_pkg_joules_total` is the total RAPL package? Or some processes are not being reported in this metric anymore?

The `kepler_process_core_joules_total` is reporting fine-grained processes info right? Including the label of process command right?

rootfs:
@marceloamaral yes, `kepler_container_pkg_joules_total` doesn't change, it adds up to the total RAPL package energy. The `kepler_process_core_joules_total` is a fine grained reporting of the `system_process` energy, process command and pid labels are in this metric.

Issue #518, opened 2023-02-03
warnings reported in kepler logs


jichenjc:
**Describe the bug**

logs has this 
```
/virtual/main.c:29:9: warning: 'HZ' macro redefined [-Wmacro-redefined]
#define HZ 1000
        ^
include/asm-generic/param.h:8:10: note: previous definition is here
# define HZ             CONFIG_HZ
         ^
1 warning generated.

```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #519, opened 2023-02-03
fix: avoid warnings about marco redefine

jichenjc:
Fixes: #518


bpfassets/perf_event/perf_event.c
```diff
@@ -26,8 +26,6 @@ limitations under the License.
 #define CPU_REF_FREQ 2500
 #endif
 
-#define HZ 1000
```
rootfs, line 29: how about

```c
#ifndef HZ
#define HZ 1000
#endif
```

rootfs (approved):

PR #521, opened 2023-02-05
Test: Add benchmark testing and refactor

SamYuan1990:
add bench mark testing at code level for function UpdateContainerEnergyByRatioPowerModel
during implementation found we can do refactor for some values and functions to avoid calculate those in k-v loop to save cpu time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,6 +78,14 @@ Back to common checking process:
 - Deploy Kepler to a local kind cluster with image stored at local image registry.
 - Use kubectl to check that the change was applied as intended.
   
+## Benchmark
+steps:
+```
+go test -cpuprofile cpu.prof -memprofile mem.prof -bench .
+pprof -http=":8091" ./cpu.prof
+```
+ref https://dave.cheney.net/2014/06/07/five-things-that-make-go-fast
+
 ## Sign Commits
 
 Please sign and commit your changes with `git commit -s`. More information on how to sign commits can be found [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
--- a/pkg/model/estimator/local/benchmark_test.go
+++ b/pkg/model/estimator/local/benchmark_test.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package local_test
+
+import (
+	"strconv"
+	"testing"
+
+	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+	"github.com/sustainable-computing-io/kepler/pkg/model/estimator/local"
+	"github.com/sustainable-computing-io/kepler/pkg/power/components/source"
+)
+
+func benchmarkNtesting(b *testing.B, continerNumber int) {
+	containersMetrics := map[string]*collector_metric.ContainerMetrics{}
+	nodeMetrics := collector_metric.NewNodeMetrics()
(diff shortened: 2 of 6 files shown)


SamYuan1990:
https://github.com/sustainable-computing-io/kepler/actions/runs/4095708814/jobs/7062816909#step:6:393
sample benchmark here.

CONTRIBUTING.md
```diff
@@ -78,6 +78,14 @@ Back to common checking process:
 - Deploy Kepler to a local kind cluster with image stored at local image registry.
 - Use kubectl to check that the change was applied as intended.
   
+## Benchmark
```
SamYuan1990, line 81: added usage base on my personal blog and https://blog.csdn.net/oe1019/article/details/112308242 relate reference.
I suppose we are able to generate prof and flame graph by golang benchmark.
cc: @rootfs , @marceloamaral

pkg/model/estimator/local/benchmark_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+	"github.com/sustainable-computing-io/kepler/pkg/model/estimator/local"
+	"github.com/sustainable-computing-io/kepler/pkg/power/components/source"
+)
+
+func benchmarkNtesting(b *testing.B, continerNumber int) {
+	containersMetrics := map[string]*collector_metric.ContainerMetrics{}
+	nodeMetrics := collector_metric.NewNodeMetrics()
+	collector_metric.ContainerMetricNames = []string{config.CoreUsageMetric}
+	nodeMetrics.AddNodeResUsageFromContainerResUsage(containersMetrics)
```
jichenjc, line 33: is this needed? seems containersMetrics  is empty at this point?

pkg/model/estimator/local/benchmark_test.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2021.
```
jichenjc, line 2: 2023 :) ?

pkg/model/estimator/local/benchmark_test.go
```diff
@@ -0,0 +1,98 @@
...
+	collector_metric.ContainerMetricNames = []string{config.CoreUsageMetric}
+	nodeMetrics.AddNodeResUsageFromContainerResUsage(containersMetrics)
+
+	nodePlatformEnergy := map[string]float64{}
+	// initialize the node energy with aggregated energy, which will be used to calculate delta energy
+	nodePlatformEnergy["sensor0"] = 10
+	nodeMetrics.SetLastestPlatformEnergy(nodePlatformEnergy)
+	nodeMetrics.UpdateIdleEnergy()
+	// the second node energy will represent the idle and dynamic power
+	nodePlatformEnergy["sensor0"] = 20 // 5J idle, 5J dynamic power
```
jichenjc, line 41: didn't get = 20 then // 5J idle, 5J dynamic, what's the relationship between them? how it's calculated?

rootfs (approved):

rootfs:
@SamYuan1990 this is a good start. Critical paths like this really need benchmarking.

pkg/model/estimator/local/benchmark_test.go
SamYuan1990: well, in fact those code parts is reuse from the unit test. and I didn't clean up the comments.
if the unit test fail the test case will broken, hence didn't re check the value in benchmark testing.
I will adjust those part in coming patch, but don't have plan to refactor test code as common function for now.

jichenjc: ok, not a big deal , though hard for me to understand the logic behind it :) 
so can you help check above comments then we can merge it?

pkg/model/estimator/local/benchmark_test.go
SamYuan1990: updated

pkg/model/estimator/local/benchmark_test.go
SamYuan1990: comments removed.

pkg/model/estimator/local/benchmark_test.go
SamYuan1990: adjust this part after data creation.

pkg/model/estimator/local/benchmark_test.go
SamYuan1990: @jichenjc , you are able to check the test logic at file ratio_test.go

jichenjc: got it , thanks

PR #523, opened 2023-02-06
Update the approach of checking the CPUArchitecture

### LINKED ISSUE #522: The issue of getCPUArchitecture
**Describe the bug**
Current cpu_model doesn't support the latest Intel Sapphire Rapids CPU. When I wanted to add it, I found kepler use [**archspec**,](https://github.com/archspec/archspec) a python package to retrieve the cpu information.

So I read the source code of archspec project, it uses cpu flags to judge the CPU Architecture, which is not very accurate and always need to maintain that repo manually for the new CPU.

I think it's better to use **cpuid** to get the correct CPU Architecture.

kepler_node_nodeInfo{cpu_architecture="unknown"}

**Expected behavior**
kepler_node_nodeInfo{cpu_architecture="Sapphire Rapids"}

LeiZhou-97:
fix #522 

1. Use cpuid instead of archspec to retrieve correct CPUArchitecture information. 
2. Add Sapphire Rapids cpu model.

Signed-off-by: Zhou,Lei <lei.zhou@intel.com>

### DIFF
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -17,16 +17,10 @@ limitations under the License.
 package metric
 
 import (
-	"encoding/csv"
 	"fmt"
-	"io"
 	"math"
-	"os"
-	"os/exec"
 	"strconv"
-	"strings"
 
-	"github.com/jszwec/csvutil"
 	"k8s.io/klog/v2"
 
 	"github.com/sustainable-computing-io/kepler/pkg/config"
@@ -45,8 +39,6 @@ const (
 )
 
 var (
-	CPUModelDataPath = "/var/lib/kepler/data/normalized_cpu_arch.csv"
-
 	NodeName            = getNodeName()
 	NodeCPUArchitecture = getCPUArch()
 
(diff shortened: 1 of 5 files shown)


rootfs:
thank you for the `cpuid` idea @LeiZhou-97!

pkg/collector/metric/node_metric.go
```diff
@@ -106,7 +123,7 @@ func getCPUArchitecture() (string, error) {
 		if err := dec.Decode(&p); err == io.EOF {
 			break
 		}
-		if strings.HasPrefix(myCPUModel, p.Name) {
+		if strings.Contains(myCPUModel, p.Architecture) {
```
rootfs, line 126: if we can get uniform cpu architecture names, we don't need to read `normalized_cpu_arch.csv` any more; the file is there to help resolve different naming conventions (temporary workaround)

pkg/collector/metric/node_metric.go
```diff
@@ -85,11 +84,29 @@ func getCPUArchitecture() (string, error) {
 		klog.V(2).Infof("cpu arch override: %v\n", cpuArchOverride)
 		return cpuArchOverride, nil
 	}
-	output, err := exec.Command("archspec", "cpu").Output()
+
+	// use cpuid to get CPUArchitecture
+	grep := exec.Command("grep", "uarch")
+	output := exec.Command("cpuid", "-1")
```
rootfs, line 90: per `cpuid` [manpage](https://man7.org/linux/man-pages/man4/cpuid.4.html), it is a x86 tool.  Can you refactor the code to externalize the `getCPUArchitecture` call in separate files and conditionally compile it to use `cpuid` on x86 and `archspec` on arm64?

LeiZhou-97: Updated. Move all unrelated node_metric func to utils.go and distinguish x86/arm64. (if you think it's better to create a new file for cpu architecture, I can edit it again).

rootfs (approved):

PR #525, opened 2023-02-07
nit: More logs to makre it readable

jichenjc:
make logs from 
model.go:85] Model Config CONTAINER_TOTAL: {false}
to
model.go:85] Model Config CONTAINER_TOTAL: {UseEstimatorSidecar:false SelectedModel: SelectFilter: InitModelURL:}


SamYuan1990 (approved):
LGTM

Issue #543, opened 2023-02-20
porting kepler to IBM s390x platform 


jichenjc:
The goal is to summarize the IBM s390x porting 

- get energy from IBM Z platform (e.g LPAR, CEC level) , this is input from LAPR HMC API directly 
- port data model and eBPF code (processing part) 
  - we can have PERF_COUNT_HW_CPU_CYCLES, PERF_COUNT_HW_INSTRUCTIONS
  - we can have soft_irq
  - we can't have other x86 eBPF metrics => need either skip or find alternative 
- current energy res model
  - PKG, CORE, UNCORE, DRAM , OTHER, GPU, PLATFORM exist in x86
  - propose to consider PKG, CORE, DRAM for now??   
- current obtain model ,seems not suitable for z ,need find new one (processing part) 
  - getCPUArchitecture function need update on 1) detect s390x, 2) model for s390x  
  - at beginning the ratio might be used ,basically it will read from HMC API first, then follow [ratio model](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/container_power.go#L81) to calculate the dynamic energy, 
        -  validate whether idle energy model is valid on z 
        -  validate dynamic energy difference
- show (output)
   - should be similar to general, may need minor fix

jichenjc:
first, install bcc on RHEL8.4+ , then try some command , it works fine 
```
yum install libbpf
yum install bcc-tools

[root@m5404019 ~]# /usr/share/bcc/tools/vfsstat
TIME         READ/s  WRITE/s  FSYNC/s   OPEN/s CREATE/s
08:22:03:       169      130        0        7        0
08:22:04:       379      202        0       17        0
08:22:05:       244      175        0       10        0
08:22:06:        58       71        0        1        0
08:22:07:       296      123        0       80        0
08:22:08:       154      155        0        2        0
[root@m5404019 ~]# uname -a
Linux m5404019 4.18.0-305.el8.s390x #1 SMP Thu Apr 29 09:06:01 EDT 2021 s390x s390x s390x GNU/Linux
[root@m5404019 ~]#
```

jichenjc:
need some steps:

cp /usr/bin/gcc /usr/bin/s390x-linux-gnu-gcc  ==> so we can compile the c file 
CGO_ENABLED=1 make cross-build-linux-s390x  ==> so go can compile c lang

stuck at no bcc/bcc_common.h  .. seems only Ubuntu might be ok .. 
[...]

PR #540, opened 2023-02-15
Kepler Stand-Alone Binary Packaging

husky-parul:
This PR will create Kepler stand-alone binary for  `arch x86_64`.

PR #549, opened 2023-02-27
enhance: build bpfdata in standalone build and add s390x support

jichenjc:


jichenjc:
related to https://github.com/sustainable-computing-io/kepler/issues/543

SamYuan1990:
I think the major problem is about buildx support s390x in github action?

rootfs (approved):

rootfs:
@SamYuan1990 there is a rpm PR #540. The rpm will be built by ondemand or scheduled GHA (not there yet)

jichenjc:
>>>I think the major problem for this PR is about buildx support s390x in github action?

this PR only provide a way to build locally
how to do this in github action is a follow up thing ..

PR #556, opened 2023-03-02
fix: add s390x CPU arch

jichenjc:


pkg/collector/metric/utils.go
```diff
@@ -155,6 +155,35 @@ func getArm64Architecture() (string, error) {
 	return strings.TrimSuffix(string(output), "\n"), nil
 }
 
+func getS390xArchitecture() (string, error) {
+	// use cpuid to get CPUArchitecture
```
rootfs, line 159: is this comment valid?

jichenjc: :( 

updated

PR #557, opened 2023-03-03
fix: set HardwareCountersEnabled to true when perf enabled

jichenjc:


pkg/bpfassets/attacher/bcc_attacher.go
```diff
@@ -129,6 +129,8 @@ func AttachBPFAssets() (*BpfModuleTables, error) {
 	bpfModules.Table = table
 	bpfModules.CPUFreqTable = cpuFreqTable
 
+	HardwareCountersEnabled = true
```
rootfs, line 132: turn it off if perf counters are not enabled [here](https://github.com/sustainable-computing-io/kepler/blob/f7dc88578aa311b8751350a035d6ab4fd3b84248/pkg/bpfassets/attacher/bcc_attacher.go#L99)

jichenjc: @rootfs we have multiple counters
https://github.com/sustainable-computing-io/kepler/blob/f7dc88578aa311b8751350a035d6ab4fd3b84248/pkg/bpfassets/attacher/bcc_attacher.go#L89

so which one should I use for `HardwareCountersEnabled ` ? 
or should I set to false if any of the counter in the list is not enabled?

rootfs: set it to false as long as one of the counters is not enabled

rootfs (approved):

PR #566, opened 2023-03-08
change config path from generic /etc/config to /etc/kepler/kepler.conf

husky-parul:
- change config path from generic /etc/config to /etc/kepler/kepler.conf
- update kepler.spec to install /etc/kepler/kepler.conf
- provide default /etc/kepler/kepler.conf as ENABLE_PROCESS_METRICS=true
Signed-off-by: Parul <parsingh@redhat.com>


husky-parul:
@rootfs Please TAL

rootfs (approved):

PR #567, opened 2023-03-09
enhance: add process irq metrics to prometheus

jichenjc:


rootfs (approved):

rootfs:
a note on the process metrics. It currently overloads the prometheus client, so every time the metrics endpoint is scraped, the cpu load just spikes. We'll revisit this topic later.

jichenjc:
>>>a note on the process metrics. It currently overloads the prometheus client, so every time the metrics endpoint is scraped, the cpu load just spikes. We'll revisit this topic later.

do we have any issue tracking this now? this sounds like something we want to avoid ..

PR #571, opened 2023-03-10
config: parse config yaml

rootfs:
currently every config key has its own config file. This PR is to use yaml for all config keys.

This is tested with this config file `/etc/kepler/kepler.config`:
```
ENABLE_PROCESS_METRICS: true
MODEL_CONFIG: |
  PROCESS_COMPONENTS_INIT_URL=https://raw.githubusercontent.com/sustainable-computing-io/kepler-model-server/main/tests/test_models/DynComponentModelWeight/BPFOnly/ScikitMixed
/ScikitMixed.json
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -19,11 +19,13 @@ package config
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
+
+	"gopkg.in/yaml.v2"
 
 	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


husky-parul (approved):

sunya-ch:
Sorry for delay. But I just notice that you are trying to change for per-file config data to a single yaml.
In this case, you need to change the way to mount the config map to the container. 

With the current definition, it is still create separate file for each.

```
# ls /etc/kepler/kepler.config
BIND_ADDRESS  CGROUP_METRICS  CPU_ARCH_OVERRIDE  ENABLE_EBPF_CGROUPID  ENABLE_GPU  KEPLER_LOG_LEVEL  KEPLER_NAMESPACE  METRIC_PATH  MODEL_CONFIG
```

sunya-ch:
Btw, it could be more complicated to change to a single file when considering MODEL_CONFIG.
The way reading the MODEL_CONFIG also needed to be change. Now, it parses the `=` config. 

```
# cat /etc/kepler/kepler.config/MODEL_CONFIG
CONTAINER_COMPONENTS_ESTIMATOR=false
CONTAINER_COMPONENTS_INIT_URL=https://raw.githubusercontent.com/sustainable-computing-io/kepler-model-server/main/tests/test_models/DynComponentModelWeight/CgroupOnly/ScikitMixed/ScikitMixed.json
```

PR #572, opened 2023-03-11
CI: try to refine build scripts in docker and make to get ready for buildx

SamYuan1990:
1. refine make for tidy.
1.a update lint and fmt same with CI
1.b update make script into ### env define ### ### Default ### ### build container ### ### build binary ### ### toolkit ### 5 parts.
2. refine docker to get ready with buildx.
3. use buildx in GHA to build multiple CPU arch images.

will use https://github.com/docker/build-push-action this GHA for image build and push

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   integration_test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         kube_provider: [kind]
@@ -34,13 +34,13 @@ jobs:
           sleep 60
           kubectl logs $(kubectl -n kepler get pods -oname) -n kepler
           kubectl get all -n kepler
-
+      
       - name: run integration_test
         run: |
           docker ps -a
           mkdir -p /tmp/.kube
           kind get kubeconfig --name=kind > /tmp/.kube/config
-          kubectl port-forward $(kubectl -n kepler get pods -oname) 9102:9102 -n kepler -v7 &
+          kubectl port-forward --address localhost $(kubectl -n kepler get pods -oname) 9102:9102 -n kepler -v7 &
           kubectl logs -n kepler daemonset/kepler-exporter
           kubectl get pods -n kepler -o yaml
           go test -tags=bcc ./e2e/... --race --bench=. -cover --count=1 --vet=all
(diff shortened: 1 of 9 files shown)


build/Dockerfile
```diff
@@ -2,8 +2,6 @@ ARG ARCH=amd64
 
 FROM quay.io/sustainable_computing_io/kepler_builder:ubi-8.6-bcc-0.24-go1.18 as builder
 
-
-ARG MAKE_TARGET=cross-build-linux-$ARCH
```
SamYuan1990, line 6: @rootfs, I don't like to specific a build command, as it makes our make file too big and different branches there. for example we can use goos but there different parameters as goos and os there... I will go to sleep for today, and take this pr later.
if you have time, could you please check with other maintainers for their feed back on how to samplify ?

Makefile
```diff
@@ -60,89 +56,105 @@ ifneq ($(shell command -v dpkg),)
...
-		--platform="linux/$(ARCH)" \
+		--platform="linux/$(GOARCH)" \
 		.
 
-	$(CTR_CMD) tag $(IMAGE_REPO)/kepler:$(SOURCE_GIT_TAG)-linux-$(ARCH) $(IMAGE_REPO)/kepler:$(IMAGE_TAG)
+	$(CTR_CMD) tag $(IMAGE_REPO)/kepler:$(SOURCE_GIT_TAG)-linux-$(GOARCH) $(IMAGE_REPO)/kepler:$(IMAGE_TAG)
 
 .PHONY: _build_containerized
 
-build-containerized-cross-build-linux-amd64:
```
SamYuan1990, line 112: I don't get the reason of the design here....
if we use buildx, then for each platform it able to get OS/CPU arch by GOENV, which means `make _build_containerized` works for all platforms. Furthermore, without buildx support, we can't assume use CPU in arch A to build arch B as the base image are different. Right?

.github/workflows/integration_test.yml
```diff
@@ -34,13 +39,18 @@ jobs:
...
           kubectl get all -n kepler
-
+      
       - name: run integration_test
         run: |
           docker ps -a
           mkdir -p /tmp/.kube
           kind get kubeconfig --name=kind > /tmp/.kube/config
-          kubectl port-forward $(kubectl -n kepler get pods -oname) 9102:9102 -n kepler -v7 &
+          docker network ls
```
rootfs, line 48: can you check cni network too?

SamYuan1990: I don't have the approach to hack into kind and check cni for now.
so far, it seems that kind network defined in docker is not same with the cni from the uuid.

SamYuan1990: I suppose some thing we can do is that to upgrade kind and kubectl to see any changes happen?

SamYuan1990: related with https://github.com/moby/moby/issues/17666

SamYuan1990:
@rootfs please merge this PR clarify , it may has breaking change, as push action can't covered in PR phase.

rootfs (approved):

PR #583, opened 2023-03-15
refactory: containerMetrics => systemProcessMetrics

jichenjc:
took me a while until I understand why containerMetrics need used in the process code path.
it's acutally a special container => system container

### DIFF
--- a/pkg/model/estimator/local/ratio_process.go
+++ b/pkg/model/estimator/local/ratio_process.go
@@ -31,22 +31,22 @@ import (
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
 )
 
-// UpdateProcessEnergyByRatioPowerModel calculates the process energy consumption based on the energy consumption of the container that contains all the processes
-func UpdateProcessEnergyByRatioPowerModel(processMetrics map[uint64]*collector_metric.ProcessMetrics, containerMetrics *collector_metric.ContainerMetrics) {
-	pkgDynPower := float64(containerMetrics.DynEnergyInPkg.Delta)
-	coreDynPower := float64(containerMetrics.DynEnergyInCore.Delta)
-	uncoreDynPower := float64(containerMetrics.DynEnergyInUncore.Delta)
-	dramDynPower := float64(containerMetrics.DynEnergyInDRAM.Delta)
-	otherDynPower := float64(containerMetrics.DynEnergyInOther.Delta)
-	gpuDynPower := float64(containerMetrics.DynEnergyInGPU.Delta)
+// UpdateProcessEnergyByRatioPowerModel calculates the process energy consumption based on the energy consumption of the container that contains all the processes (which is a special container metrics for all system processes)
+func UpdateProcessEnergyByRatioPowerModel(processMetrics map[uint64]*collector_metric.ProcessMetrics, systemProcessMetrics *collector_metric.ContainerMetrics) {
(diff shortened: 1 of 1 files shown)


rootfs:
@jichenjc let's keep the container metric naming. If we decide to expose processes in each container in the future, container metric is more appropriate.

marceloamaral:
I also prefer to use `containerMetrics` or `processesMetrics` (plural as it could be aggregation of processes), because `systemProcessMetrics` can be misleading as they are not necessarily user processes.

Btw, are you also changing the key of this map? I mean instead of use containerID?

rootfs:
@jichenjc the process metrics are extracted from each container. Currently only the system processes are exported. It is possible in the future, all the processes in all containers are exported too, if the prometheus client scaling issue is resolved.

jichenjc:
>>>all the processes in all containers are exported too, if the prometheus client scaling issue is resolved.

ok, what's the use case of expose process in container if we do have container metrics ? 
curious because I Think the goal is to expose container metrics through container method and process through process method only?

marceloamaral:
As all processes have a cgroup ID and some cgroups can have container id (a docker container has only one cgroup).
Libvirt domain also have a cgroup ID.

One solution would be changing the logic to cgroup metrics. Wdyt?

jichenjc:
>>>One solution would be changing the logic to cgroup metrics.

you mean instead of each process metric, we expose cgroup instead? 
@rootfs what's your suggestion?

marceloamaral:
@jichenjc I had another idea for this.

We can use a generic name, `metricsMap` and choose the key based on the level of aggregation needed/configured. 
The options for the key can be `podID`, `containerID`, `cgroupID` and `processID`. Different level of aggregation will have different overhead and purposes. Wdyt?

marceloamaral:
@jichenjc and @rootfs, after reviewing of the code, I noticed that there are now two maps: `c.ContainersMetrics` and `c.ProcessMetrics`, which seem to contain duplicated information.

It would be better to have a single map to reduce overhead and simplify the code. 
As suggested before, the map could be configured with different keys depending on whether Kepler will report per-process or per-container metrics.

jichenjc:
>>>As suggested before, the map could be configured with different keys depending on whether Kepler will report per-process or per-container metrics.

that's a good idea, I will check what we can do on this ,guess it's due to we add container first then process
so should consider refactory if possible

jichenjc:
we have following logic for now:

container metrics => use container ID which is a string / and no process ID 
process metrics => system ID (all process share same string) / process ID 

so how about we do a `fmt("%s-%d", <system container ID >, PID) ` as key for processes
for each process then we can combine them together as a new variable like `metrics` ?

and for process we can ignore all the container related stuff like `PodName`, `Namespace` etc?

marceloamaral:
@jichenjc sounds interesting.

>and for process we can ignore all the container related stuff like PodName, Namespace etc?

That is right!

Btw, do you think that PID from different container will collide?
If not, why not only use the PID as the key? 
So, we don't need to create a key with `containerID-PID` when per-process metrics is enabled.
Therefore, when  per-process metrics is enabled we use the PID as the key, otherwise we can use the containerID.

In my point of view, when per-process metrics is enabled there's no need to aggregate the metrics per container in the metrics map. We should report the energy/metrics per-process also in the Prometheus metrics. That is we will need to add a new label to the Prometheus metrics that includes the process name and ID. This way, users can aggregate the container metrics using a Prometheus query.

Issue #560, opened 2023-03-06
CPURefCycleLabel not used, remove it?


jichenjc:
we have those conf defined:

```
Counters = map[string]perfCounter{
                CPUCycleLabel:       {unix.PERF_TYPE_HARDWARE, unix.PERF_COUNT_HW_CPU_CYCLES, true},
                CPURefCycleLabel:    {unix.PERF_TYPE_HARDWARE, unix.PERF_COUNT_HW_REF_CPU_CYCLES, true},
                CPUInstructionLabel: {unix.PERF_TYPE_HARDWARE, unix.PERF_COUNT_HW_INSTRUCTIONS, true},
                CacheMissLabel:      {unix.PERF_TYPE_HARDWARE, unix.PERF_COUNT_HW_CACHE_MISSES, true},
        }
```

but the bpf reader has:

```
// TODO in sync with bpf program
type ProcessBPFMetrics struct {
        CGroupID       uint64
        PID            uint64
        ProcessRunTime uint64
        CPUCycles      uint64
        CPUInstr       uint64
        CacheMisses    uint64
        VecNR          [config.MaxIRQ]uint16 // irq counter, 10 is the max number of irq vectors
        Command        [16]byte
}
```

so should remove this `CPURefCycleLabel` since it's not used ?

marceloamaral:
No, we cannot remove the `CPURefCycleLabel` since it is part of the logic to expose the `CPURefCycle` hardware counter to the ebpf code. This is metric is used to calculate the CPU frequency within the ebpf code, but is not needed to be used in the userspace (golang code).

PR #589, opened 2023-03-17
remove: remove CPURefCycleLabel as it's not used, no need to collect

jichenjc:
related to https://github.com/sustainable-computing-io/kepler/issues/560

### DIFF
--- a/pkg/bpfassets/attacher/attacher_types.go
+++ b/pkg/bpfassets/attacher/attacher_types.go
@@ -22,7 +22,6 @@ import (
 
 const (
 	CPUCycleLabel       = config.CPUCycle
-	CPURefCycleLabel    = config.CPURefCycle
 	CPUInstructionLabel = config.CPUInstruction
 	CacheMissLabel      = config.CacheMiss
 
--- a/pkg/bpfassets/attacher/bcc_attacher.go
+++ b/pkg/bpfassets/attacher/bcc_attacher.go
@@ -49,7 +49,6 @@ type perfCounter struct {
 var (
 	Counters = map[string]perfCounter{
 		CPUCycleLabel:       {unix.PERF_TYPE_HARDWARE, unix.PERF_COUNT_HW_CPU_CYCLES, true},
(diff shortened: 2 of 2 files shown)


marceloamaral:
CPURefCycle is used to calculate the CPU frequency.

marceloamaral (changes requested):
CPURefCycle is used to calculate the CPU frequency.

@rootfs unless if we want to remove the CPU frequency we should keep this metric.

rootfs:
@marceloamaral we need to keep the cpu frequency metrics. @jichenjc finds no reference of `CPURefCycle` in calculating CPU frequency. @marceloamaral can you double check?

jichenjc:
there is CPUFreq https://github.com/sustainable-computing-io/kepler/blob/main/pkg/bpfassets/attacher/bcc_attacher.go#L129
I think it's not realated to the Refcycle (at least didn't find usage in the code) ..

marceloamaral:
The calculation is done in the bpf code, please see [here](https://github.com/sustainable-computing-io/kepler/blob/main/bpfassets/perf_event/perf_event.c#L181).

The bpf code needs to have access to the hardware counter `CPURefCycle`.

jichenjc:
yes, but https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/container_hc_collector.go#L38
didn't use it at all ..
and not handling here as well
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/container_hc_collector.go#L107

marceloamaral:
@jichenjc yes, because we do not use this metric in the `userspace` (golang code).

The `CPURefCycle` is only necessary in the ebpf code and we cannot disable it if we need to calculate the CPU frequency.

Note that, the code that this PR is trying to remove enables ebpf to collect the `CPURefCycle`. Without this code, this metrics cannot be collect and then the CPU frequency cannot be calculated.

jichenjc:
ok, thanks for the clarification , learn something :)

PR #596, opened 2023-03-23
nit: remove an unnessary TODO

jichenjc:


jichenjc:
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/model.go#L49
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/model.go#L53

plan to fix this TODO but found it's used separately ,so remove the TODO

SamYuan1990 (approved):
LGTM

Issue #591, opened 2023-03-17
RAPL code issues


jiere:
**Describe the bug**
There are some code issues on pkg/power module RAPL side.
1. In pkg/power/component/power.go, for platforms which support RAPL sysfs, there is no chance to reach the RAPL msr side, not sure if the rapl_msr*.go are dead code or not?
2. In pkg/power/component/source/rapl_sysfs_util.go, current code to get cpuNum and cpuPackage is low efficient, could be refined.
3. In pkg/power/component/source/rapl_msr_util.go, current code for CPU package and core mapping has logical error.
See function "mapPackageAndCore" in rapl_msr_util.go:
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/components/source/rapl_msr_util.go#L81
The "id" here is the specific core's physical_package_id, i.e. On two-sockets platforms, the "id" should be either 0 or 1, so the "packageMap" concept is wrong after the for loop.

**Expected behavior**
1. See function "getNumPackage" and "getNumCPUs" in rapl_sysfs_util.go.
For Server CPUs like Intel 4th Xeon SP and AMD EYPC Genoa, etc, there might be hundreds of cores per package, the more efficient way to get cpu number and package number is leveraging the output of "lscpu“ command:

Ubuntu system example:
$ lscpu |grep -E "CPU\(s\)|Socket\(s\)"
CPU(s):                          128
On-line CPU(s) list:             0-127
Socket(s):                       2
NUMA node0 CPU(s):               0-31,64-95
NUMA node1 CPU(s):               32-63,96-127

RHEL system example:
$ lscpu |grep -E "CPU\(s\)|Socket\(s\)"
CPU(s):              224
On-line CPU(s) list: 0-223
Socket(s):           2
NUMA node0 CPU(s):   0-55,112-167
NUMA node1 CPU(s):   56-111,168-223

For CPU number, maybe need some more code to fetch the exact number; for package number, just "lscpu |grep Socket" is enough.

2. If the RAPL MSR code is not dead code, we need to fix the issue mentioned above.

jiere:
> @jiere On certain environment (hyper V if my memory is right), the RAPL sysfs files are missing, but msr are accessible. That gave rise to the msr code. If you find that is not the case any more, we can remove the msr code.

Thanks @rootfs for your quick reply, I will check later and get you tuned.

jichenjc:
[...]

PR #599, opened 2023-03-25
RAPL code refactoring

jiere:
Due to https://github.com/sustainable-computing-io/kepler/issues/591, made below changes:
1. Correct some typo on code.
2. Simplify code logic and improve code efficiency in RAPL Sysfs.
3. Correct code logic error in RAPL MSR.
4. Align with the latest Intel SDM.

Signed-off-by: Jie Ren <jie.ren@intel.com>


marceloamaral (changes requested):
Since you are improving this code, could you please check the logic here https://web.eece.maine.edu/~vweaver/projects/rapl/rapl-read.c

pkg/power/components/source/rapl_sysfs_util.go
```diff
@@ -19,44 +19,20 @@ package source
...
-
-		if id, err := strconv.Atoi(strings.TrimSpace(string(data))); err == nil {
-			if _, exist := pkgMap[id]; !exist {
-				numPackage++
-				pkgMap[id] = true
-			}
-		}
-	}
-	return numPackage
+	out, _ := exec.Command("bash", "-c", "lscpu |grep Socket").Output()
```
marceloamaral, line 28: We should avoid run bash command from the code since it is not a good design practice.

marceloamaral: Here is an alternative implementation

```golang
func getNumPackage() int {
  sockets := make(map[int]bool)
  file, err := os.Open("/proc/cpuinfo")
  if err != nil {
      fmt.Println(err)
      return 0
  }
  defer file.Close()

  scanner := bufio.NewScanner(file)
  for scanner.Scan() {
      line := scanner.Text()
      if strings.Contains(line, "physical id") {
          substrings := strings.Split(line, ":")
          socket, err := strconv.Atoi(strings.TrimSpace(substrings[1]))
          if err != nil {
              fmt.Println("Error:", err)
              return 0
          }
          sockets[socket] = true
      }
  }

  if err := scanner.Err(); err != nil {
      fmt.Println(err)
      return 0
  }
  
  return len(sockets)
}
```

marceloamaral:
Btw, thanks @jiere for contributing to improve the Kepler code!

pkg/power/components/source/rapl_sysfs_util.go
jiere: Thanks @marceloamaral  for pointing out his, yes, this alternative implementation scans single file instead of opens N topology files, the for loop times are same. If we choose this way, maybe we could use single function to both count core number and package number.

jichenjc:
>>>

jichenjc:
>>>do we kepler need to take care both client-platform CPUs and server-platform CPUs? I once have an assumption that it is meaningful to deal with Data Center power efficiency and carbon awareness.

when you say `client`, is it the kepler running on platform? I think the focus is the measurement target, so mostly `server` side you mention?

marceloamaral:
Indeed, we are aware that some platforms may not have `uncore` (PP1) or `psys`. 
However, this is not an issue if a power source does not exist. They will just have 0 values.

Note that differentiating between `client` and `server` platforms is not necessary. What matters is that in the MSR logic, we need to take into account that "The DRAM units is different for different CPU models," as can be seen in the code that I shared.

Thank you for catching the issue with the `sysfs` logic!
It should be a simple fix by checking if the string `package-0` contains the substring `package`.

pkg/power/components/source/rapl_sysfs_util.go
marceloamaral: Yes, we should count both in only one loop.

jiere:
> What matters is that in the MSR logic, we need to take into account that "The DRAM units is different for different CPU models," as can be seen in the code that I shared.

Double checked on ark.intel.com, all those models(Haswell EP/Broadwell EP/Skylake X/Xeon Phi) are in "Discontinued" marketing status, which means EOL. So we are safe to simply regard that the DRAM energy unit always equals to CPU energy unit.

Thank you for your careful and nice comments, let me kick off a new diff later.

marceloamaral:
@jiere please check my comments in the PR ##601

jiere:
The reason for `power` pkg to maintain a "pkg-to-corelist" map is that current `powerInterface` methods are package level organized, and each cpu core has its "/dev/cpu/*/msr".
Ideally when we need CPU package level power consumption metrics, we can loop the package's each core's MSR to get specific data.
Up till now, Intel could only support package level power meter, so all cpu cores' MSR power data are same if they belong to the same package.
But I did heard that AMD has a patch on core level power meter, which is perf event based.

jiere:
> Right, I just want to ensure that we avoid redundant works and conduct topology discovery only once and through a single method to have consistency in the code.

I cannot agree more :-)

jiere:
@marceloamaral and all other reviewers, I have rebased my diff and kicked off the update, please continue reviewing, thanks.
For the "bash -c" code style issue, I have found a solution, referenced code [here](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/metric/utils.go#L157-L184).

I do agree with @marceloamaral 's point that we need to maintain as little as possible redundant code in a project, currently we have so many `utils.go` among various of pkgs, maybe someday we can consolidate them to a common place or project scope pkg to maintain them as one piece. But currently I choose not, the `config` pkg may not be very proper to act as the whole project's common API/lib repository.

pkg/power/components/source/rapl_sysfs_util.go
```diff
@@ -19,44 +19,39 @@ package source
...
 	"strconv"
 	"strings"
-
-	"k8s.io/klog/v2"
 )
 
-func getNumCPUs() int {
-	data, err := os.ReadFile(cpuInfoPath)
+func getNumPackage() (int, error) {
+	grep := exec.Command("grep", "Socket")
```
marceloamaral, line 28: Can we avoid the use of bash commands? It's better to use libraries instead of bash commands since library results in more transparency, efficiency, and better code design.

Furthermore, bash commands are not the most efficient approach. In this particular case, you are running two commands: `lscpu` and `grep`. The `grep` command is traversing a file as the approach that I suggested before, so we can simply open the file and read it directly.

marceloamaral: Also, we should save the num package in a global variable so that we don't need to traverse a file every time that we call this function.

marceloamaral (changes requested):
Thank you @jiere for the good work!

I just have only one comment regarding using the [bash command](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/metric/utils.go#L157-L184). We should update this code that you found....

In general, it is better to stick to a single language and use Golang for all aspects of the project. This can help ensure that your code is consistent, easy to understand, and maintainable over time.

pkg/power/components/source/rapl_sysfs_util.go
jiere: OK,  @marceloamaral, I used to understand your point that the code style like using "bash -c" in Golang is bad because it has an assumption that the system has installed `bash`, but `lscpu` and `grep` should be included in any Linux distro.

Now I think I totally got your points! The bad news is that not only the `getS390xArchitecture()` directly called `exec.Command`, but also for `getArm64Architecture()` and `getX86Architecture()`... Maybe more other places...
It seems to be a tradeoff between code independency and code simplicity.

Another question is where should be the best place for us to save such project-wide global variables and APIs? As I mentioned above, `config` pkg might not be very appropriate to act as such role, since they are configuration related...

Maybe this topic is a little bit leaving the current PR's title:) Shall we close current one and open a new issue to trace similar refactoring efforts throughout the whole project code?

marceloamaral: @jiere, let's fix the RAPL issues with this PR first and then we refactor the other codes later. But we should replace the existing `getNumPackage` function with a code that reads the `/proc/cpuinfo` file. So we move a step torwards this direction.

In a separate PR, we can create a new package called `runtime` at the same level as `config`. The  `runtime` package will be responsible for providing information about the runtime system, including the CPU topology, model, cgroup version, etc. We can also fix the other codes this new PR, if possible.

marceloamaral: Regarding the Library to read the runtime system information, we can use https://github.com/jaypipes/ghw#cpu

pkg/power/components/source/rapl_msr_util.go
```diff
@@ -47,57 +43,36 @@ const (
...
 
 var (
+	cpu       *ghw.CPUInfo
 	fds       []int
 	byteOrder binary.ByteOrder
 
-	// package - core
-	packageMap  map[int]int
-	packageList []int
+	numPackages, numCores int
```
marceloamaral, line 50: I am not finding where you set the values in this variables. Shouldn't it be in the `init()` ?

marceloamaral (changes requested):
Looks great, I have only one comment.

pkg/power/components/source/rapl_msr_util.go
jiere: Sorry, my bad, let me rebase a single diff.

marceloamaral:
@jiere did you run `make tidy-vendor` before pushing the PR?

jiere:
> @jiere did you run `make tidy-vendor` before pushing the PR?

Yes, I did. And below are the current status of my local workspace:
~/dev/kepler$ git status
On branch rapl-enhance
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        vendor/github.com/StackExchange/
        vendor/github.com/ghodss/
        vendor/github.com/go-ole/
        vendor/github.com/jaypipes/
        vendor/github.com/mitchellh/
        vendor/howett.net/

nothing added to commit but untracked files present (use "git add" to track)

Shall I include these vendor/xxx untracked files also in current PR?

Issue #570, opened 2023-03-10
per core level CPU time accounting in eBPF


rootfs:
**Is your feature request related to a problem? Please describe.**
Per conversation with @sunya-ch, early eBPF code has per core level CPU time. This detail helps train and use models with high accuracy. However, on large setup with many cores, the eBPF map may run out of memory, so this detailed CPU time accounting was converted to aggregate CPU time across all the cores. As a result, the model accuracy degraded. 

**Describe the solution you'd like**
The per core CPU time info should be the default format in eBPF. The granularity can change only if the eBPF memory usage is out of control. 

**Describe alternatives you've considered**

**Additional context**

sunya-ch:
Thank you for posting this.
Here is corresponding issue in kepler-model-server project: https://github.com/sustainable-computing-io/kepler-model-server/issues/78

marceloamaral:
We don't need that old implementation with memory issues.

To get per-core CPU time, we can just change the key in the eBPF map. There are some hints on how to do that in the code.

The new key will be based on the process PID and CPU ID, so we can group the data by core. But we'll have to make sure how it impacts the code that aggregates the metrics.

sunya-ch:
> We don't need that old implementation with memory issues.
> 
> To get per-core CPU time, we can just change the key in the eBPF map. There are some hints on how to do that in the code.
> 
> The new key will be based on the process PID and CPU ID, so we can group the data by core. But we'll have to make sure how it impacts the code that aggregates the metrics.

Thank you, Marcelo. I just check the bpf code according to your point. If I understand correctly, we can just add cpu_id to process_metrics_t 
https://github.com/sustainable-computing-io/kepler/blob/7942754e9324f675ca99572e45a0db8dcb1a5ff0/pkg/bpfassets/perf_event_bindata.go#L91-L101

and update 
https://github.com/sustainable-computing-io/kepler/blob/7942754e9324f675ca99572e45a0db8dcb1a5ff0/pkg/collector/container_hc_collector.go#L38-L48

right?

sunya-ch:
> We don't need that old implementation with memory issues.
> 
> To get per-core CPU time, we can just change the key in the eBPF map. There are some hints on how to do that in the code.
> 
> The new key will be based on the process PID and CPU ID, so we can group the data by core. But we'll have to make sure how it impacts the code that aggregates the metrics.

[...]

PR #601, opened 2023-03-27
add process_count metric for mapping container to CPU/Package

sunya-ch:
Regarding https://github.com/sustainable-computing-io/kepler/issues/570 (specifically the idea in https://github.com/sustainable-computing-io/kepler/issues/570#issuecomment-1484535723),

**work included in this PR:**
- add function to getCPUPackageMap mapping from CPU ID to Package ID 
- refactor updateBPFMetrics function to prevent potential complexity test failed

**undo work (postponed to other PR):**
- expose pid_time BPFHash to ProcessCPUTimestamp Table in BpfModuleTables 
- reset the table together with main table (c.bpfHCMeter.Table) 
- readProcessCPUTimestamp and assign to c.ContainersMetrics/c.ProcessMetrics CPUIDs as a count of processes (pids) that use that CPU of considering container (for process case will be always 1)
- export CPUIDs to container metric `kepler_container_process_count_current` with label of CPU ID and Package ID

```bash
# HELP kepler_container_process_count_current Latest number of processes that have executed on a specific CPU
# TYPE kepler_container_process_count_current gauge
kepler_container_process_count_current{container_name="containerA",container_namespace="test",cpu="1",package="1",pod_name="podA"} 1
```
- add simple test case for exported metric (undo, postponed to other PR)

### DIFF
--- a/pkg/collector/container_hc_collector.go
+++ b/pkg/collector/container_hc_collector.go
@@ -51,6 +51,56 @@ func (c *Collector) resetBPFTables() {
 	c.bpfHCMeter.Table.DeleteAll()
 }
 
+// updateBasicBPF
+func (c *Collector) updateBasicBPF(containerID string, ct *ProcessBPFMetrics, isSystemProcess bool) {
+	// update ebpf metrics
+	// first update CPU time
+	if err := c.ContainersMetrics[containerID].CPUTime.AddNewDelta(ct.ProcessRunTime); err != nil {
+		klog.V(5).Infoln(err)
+	}
+	// update IRQ vector
+	for i := 0; i < config.MaxIRQ; i++ {
+		if err := c.ContainersMetrics[containerID].SoftIRQCount[i].AddNewDelta(uint64(ct.VecNR[i])); err != nil {
+			klog.V(5).Infoln(err)
+		}
+	}
+	// track system process metrics
+	if isSystemProcess && config.EnableProcessMetrics {
+		for i := 0; i < config.MaxIRQ; i++ {
+			if err := c.ProcessMetrics[ct.PID].SoftIRQCount[i].AddNewDelta(uint64(ct.VecNR[i])); err != nil {
+				klog.V(5).Infoln(err)
+			}
+		}
+	}
+}
+
+// updateHWCounters
+func (c *Collector) updateHWCounters(containerID string, ct *ProcessBPFMetrics, isSystemProcess bool) {
+	// update HW counters
+	for _, counterKey := range collector_metric.AvailableHWCounters {
+		var val uint64
+		switch counterKey {
+		case attacher.CPUCycleLabel:
+			val = ct.CPUCycles
+		case attacher.CPUInstructionLabel:
+			val = ct.CPUInstr
+		case attacher.CacheMissLabel:
+			val = ct.CacheMisses
+		default:
+			val = 0
+		}
+		if err := c.ContainersMetrics[containerID].CounterStats[counterKey].AddNewDelta(val); err != nil {
+			klog.V(5).Infoln(err)
+		}
+		// track system process metrics
+		if isSystemProcess && config.EnableProcessMetrics {
+			if err := c.ProcessMetrics[ct.PID].CounterStats[counterKey].AddNewDelta(val); err != nil {
+				klog.V(5).Infoln(err)
+			}
+		}
+	}
+}
+
 // updateBPFMetrics reads the BPF tables with process/pid/cgroupid metrics (CPU time, available HW counters)
 func (c *Collector) updateBPFMetrics() {
 	if c.bpfHCMeter == nil {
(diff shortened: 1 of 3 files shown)


sunya-ch:
Need to refactor function for golint test.
https://github.com/sustainable-computing-io/kepler/compare/c18f4e9b795394d9a8720b30169e4c52fcd34ef4..f40480427bf4ec61ebf8139be7e1d04d78e560a9

move logic to updateBasicBPF and updateHWCounters separating from `updateBPFMetrics`.
There is one point that I am not quite sure I did it right is that the previous code put the logic to update SoftIRQCount inside the Hardware counter loop (collector_metric.AvailableHWCounters). 
I don't think it should be there. So, I separate it to call only once as seen in updateBasicBPF.

rootfs:
@sunya-ch thank! The refactor looks good to me.

pkg/collector/container_hc_collector.go
```diff
@@ -49,15 +55,89 @@ type ProcessBPFMetrics struct {
...
+		}
+		if isSystemProcess {
+			if err := c.ProcessMetrics[ct.PID].CounterStats[counterKey].AddNewDelta(val); err != nil {
+				klog.V(5).Infoln(err)
+			}
+		}
+	}
+}
+
+// updateCPUIDs
```
jichenjc, line 131: this is useless?

pkg/collector/container_hc_collector.go
```diff
@@ -66,67 +146,44 @@ func (c *Collector) updateBPFMetrics() {
...
-				for i := 0; i < config.MaxIRQ; i++ {
-					if err = c.ProcessMetrics[ct.PID].SoftIRQCount[i].AddNewDelta(uint64(ct.VecNR[i])); err != nil {
-						klog.V(5).Infoln(err)
+		if isSystemProcess {
+			// always 1
+			c.ProcessMetrics[ct.PID].CPUIDs = cpuIDs
+		} else {
+			if _, found := foundContainer[containerID]; !found {
+				// newly set to 1
+				c.ContainersMetrics[containerID].CPUIDs = cpuIDs
```
jichenjc, line 181: CPUID means CPU counters or ID ? the name seems means CPU ID but actually it's number/count that run the process ?

pkg/collector/container_hc_collector.go
```diff
@@ -66,67 +146,44 @@ func (c *Collector) updateBPFMetrics() {
...
+			} else {
+				// add value for container (i.e., multiple PIDs)
+				for cpuID := range cpuIDs {
```
jichenjc, line 184: is it always one item per line 173? 
```
cpuIDs = make(map[int32]int)
```

pkg/collector/container_hc_collector.go
sunya-ch: Sorry. Need to remove it.

marceloamaral (changes requested):
If I remember correctly there is a related discussion about this in another PR.

Instead of generating a new table, why not use of the existing table containing process information along with CPU time, and utilize a key with of both the PID and CPU ID instead of just the PID? Then, we can iterate the table and get the cpu id and time per process.

sunya-ch:
> If I remember correctly there is a related discussion about this in another PR.
> 
> Instead of generating a new table, why not use of the existing table containing process information along with CPU time, and utilize a key with of both the PID and CPU ID instead of just the PID? Then, we can iterate the table and get the cpu id and time per process.

@marceloamaral 
To clarify, there are multiple BPFHash tables but the focus for this PR will be these two.
```
// processes and pid time
BPF_HASH(processes, u64, process_metrics_t);
BPF_HASH(pid_time, pid_time_t);
```
`processes` uses pid as a key and it aggregates all stats associated to pid.
Meanwhile, `pid_time` uses (pid, cpuid) as a key and keep the timestamp for computing on CPU-time.

The `processes` is loaded to userspace with go struct `ProcessBPFMetrics`. 

Firstly, I also think to just add CPU id to the hash key but it is not trivial because we have to change from aggregating group by (pid) to group by (pid,cpuid). It will multiply the number of entry up to the number of cores per pid which could be a big tradeoff in big server. 
That is why I decided to just load the pid_time table to userspace.

pkg/collector/container_hc_collector.go
sunya-ch: Changed the name to `processCountOnCPU`. 
Firstly, I just thought about exposing the CPU IDs but now it more like pushing the ID to label and have the process count as a value instead. 
Thanks.

pkg/collector/container_hc_collector.go
sunya-ch: My mistake. Made the changes. It will be ones if pid found in the BPFHash table (which is expected). If not, it will be zeros.

sunya-ch:
Update on naming and fix some bugs:
- fix naming CPUIDs --> ProcessCountOnCPU 
- not delete cpu_time table (because it should be all handled by BPF program)
- fix bugs on setting ContainersMetrics not ProcessMetrics for multiple PIDs case

marceloamaral:
Having the `(pid,cpuid)` to the `processes` table will also help to calculate the energy consumption per socket. We are now naively using the aggregation of all sockets.

In fact, when I create the `(pid,cpuid)` to the `pid_time`, I had in mind to extend it to the `processes` table later.

We will have a more general solution and concentrate the data in only one table in the userspace.

sunya-ch:
> Having the `(pid,cpuid)` to the `processes` table will also help to calculate the energy consumption per socket. We are now naively using the aggregation of all sockets.
> 
> In fact, when I create the `(pid,cpuid)` to the `pid_time`, I had in mind to extend it to the `processes` table later.
> 
> We will have a more general solution and concentrate the data in only one table in the userspace.

At least we should make that optional to be turned off in a big server. Otherwise, we might meet the limit of BPFHash size. I don't think we should not include that to this PR yet. Need more discussion and overhead test in scale.

marceloamaral:
Right, but If we don't separate hardware counters per CPU, how can we separate it per socket?

Also, It would be more advisable to implement a solution that minimizes the need for code changes in the future.

Note that the `pid_time` table is already limited by the number of CPUs, and there may be a data mismatch when it reaches its size limit since the `processes` table will have more data than the `pid_time` table.

I do agree that we can improve/investigate scalability for large systems in another PR, such as changing to `BPF_PERCPU_HASH`, for now, we could only use `BPF_HASH` with a key of `(pid, cpuid)` to address the issue.

sunya-ch:
> Right, but If we don't separate hardware counters per CPU, how can we separate it per socket?
> 
> Also, It would be more advisable to implement a solution that minimizes the need for code changes in the future.
> 
> Note that the `pid_time` table is already limited by the number of CPUs, and there may be a data mismatch when it reaches its size limit since the `processes` table will have more data than the `pid_time` table.
> 
> I do agree that we can improve/investigate scalability for large systems in another PR, such as changing to `BPF_PERCPU_HASH`, for now, we could only use `BPF_HASH` with a key of `(pid, cpuid)` to address the issue.

This PR provides a in-the-middle solution to at least tell which cores and which packages (socket) the process is using during the capture. If the load is steady over that period of time, we can divide the usage by the number of CPU in that package.

marceloamaral:
So you are saying that we should make the `(pid, cpuid)` to `pid_time` optional?

Using the `(pid, cpuid)` to `pid_time` has the same possible limitation in huge system as for the `processes` table.

Btw, to have a safe solution for now, we can increase the hash bucket size to [65536](https://github.com/iovisor/bcc/blob/d1b1b61ceaee16e9d68e3c66e3ebddd55ba36d0f/libbpf-tools/tcpsynbl.bpf.c#L12). I tested it with no problem.

```c
#ifndef MAX_THREADS
#define MAX_THREADS 65536
#endif

BPF_HASH(processes, pid_time_t, process_metrics_t, MAX_THREADS);
```

marceloamaral:
My suggestion is to concentrate all data in the `processes` table instead of having multiple tables for consistency. This has the advantage of a clear logic in the userspace code. Also, we should have both tables `processes` and `pid_time` using the same keys for consistency.



The link you shared is very interesting as they created a hash with 10240000 buckets, which is huge. 

For our purposes, we can assume that the largest system we have seen has 256 cores, including hyper-threading, and is typically composed of four sockets with 64 cores per socket. Giving that the Linux Kernel is NUMA-aware by default, so on average, a task rarely context-switches between sockets. Then, in the worst-case scenario, if a task does context-switch in the half of all cores within a socket within a 3-second interval, we would have 32 entries for a task. To support this scenario, we could use a hash size of 65536 (the highest value that I saw in the bcc tools example, but still a very low value compared to the link you shared), allowing us to record the data of 2048 tasks within a 3-second interval, which is still a sizable number of tasks since the system only have 256 cores.

However, given that very large systems may also have a large amount of DRAM, we could consider making it optional to increase the hash size in the future.

sunya-ch:
> My suggestion is to concentrate all data in the `processes` table instead of having multiple tables for consistency. This has the advantage of a clear logic in the userspace code. Also, we should have both tables `processes` and `pid_time` using the same keys for consistency.
> 
> The link you shared is very interesting as they created a hash with 10240000 buckets, which is huge.
> 
> For our purposes, we can assume that the largest system we have seen has 256 cores, including hyper-threading, and is typically composed of four sockets with 64 cores per socket. Giving that the Linux Kernel is NUMA-aware by default, so on average, a task rarely context-switches between sockets. Then, in the worst-case scenario, if a task does context-switch in the half of all cores within a socket within a 3-second interval, we would have 32 entries for a task. To support this scenario, we could use a hash size of 65536 (the highest value that I saw in the bcc tools example, but still a very low value compared to the link you shared), allowing us to record the data of 2048 tasks within a 3-second interval, which is still a sizable number of tasks since the system only have 256 cores.
> 
> However, given that very large systems may also have a large amount of DRAM, we could consider making it optional to increase the hash size in the future.

It seems like a big change because it will change the logic not only in kernel space as well as in user space too to keep per-cpu stat. Could that be a future work in another PR? 

Also, I am still worried about kernel memory concern when number of running process (like thousands in one of our system) as well as the number of cores (like hundreds). Each entry will contain at least 70 bytes for the current collecting metric. If we increase a single metric (u64), it will multiply 8 by the number of cores x pids. I think we should not put that the load to the kernel if not necessary (e.g., when you have only one NUMA). How's about having the method that extrapolates the socket usage from the count of context switch on CPU first?

marceloamaral:
> Could that be a future work in another PR (you can take over it)?

The problem with that is a significant amount of code would need to be undone later in order to change the information per core. Also, using the `(pid, cpuid)` to `pid_time` has the same possible limitation in huge system as for the `processes` table, so we should remove it right? Additionally, it may be difficult find someone who can work on this later and you are already doing this :)

Alternatively, as you suggested, a potential solution would be to report values per socket instead of per core. Would this work for you or is it crucial to have the information per core?

sunya-ch:
> > Could that be a future work in another PR (you can take over it)?
> 
> The problem with that is a significant amount of code would need to be undone later in order to change the information per core. Also, using the `(pid, cpuid)` to `pid_time` has the same possible limitation in huge system as for the `processes` table, so we should remove it right? Additionally, it may be difficult find someone who can work on this later and you are already doing this :)
> 
> Alternatively, as you suggested, a potential solution would be to report values per socket instead of per core. Would this work for you or is it crucial to have the information per core?

For me, it is totally fine to use per-package. However, I think you must have a way to implement that in an efficient way. I prefer undoing the code in this PR for the part that export the process_count and read the pid_time table. For this PR, I will keep only the refactor of container_hc due to complexity alert and the part that I provide a mapping between CPU to package number because these two parts are still necessary no matter which approach we are going to make the per-package/per-CPU happen. After merge, you or others who will continue working on this can utilize them. Is this okay for you?

marceloamaral (changes requested):
I am ok if you want to work in the metrics per socket in another PR.

pkg/collector/metric/utils.go
```diff
@@ -232,3 +232,20 @@ func getCPUArchitecture() (string, error) {
 
 	return "", fmt.Errorf("no CPU power model found for architecture %s", myCPUModel)
 }
+
+func getCPUPackageMap() (cpuPackageMap map[int32]string) {
```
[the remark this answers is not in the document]
sunya-ch: Sorry, I didn't track that one but yes this will read the file at the beginning because the topology is not supposed to change over time. The ReadFile function is efficient enough to quickly read a small file and close it. 

Parsing `/proc/cpuinfo` is also another choice but we need to make sure that there will be no formatting compatibility in the future. Also the larger cores the more you need to run through a big file with irrelevant information. That's why I think using the raw file (`/sys/devices/system/cpu/cpu%d/topology/physical_package_id`) is better.

sunya-ch:
> I am ok if you want to work in the metrics per socket in another PR.

@marceloamaral  I will leave it open for anyone to lend a hand on it (including myself if I managed to schedule the time for it). 

I made the changes by undoing the process_time reading part. Details are updated in the PR description.

marceloamaral (approved):
Thanks @sunya-ch!
/lgtm

@jiere please try to reuse the function `getCPUPackageMap()` to initialize a global map in the `config` package with the CPU topology. You might move this function as well as.

PR #603, opened 2023-03-28
add test case and clean up for some code

SamYuan1990:
we should add test case before merge code.


SamYuan1990:
by this PR, raise up test coverage to 49%

rootfs (approved):

Issue #606, opened 2023-03-28
Kubelet update is blocked by network IO?


SamYuan1990:
When I try to add benchmark for `updateKubeletMetrics` I found it seems a blocked network IO?

Here is the tracing stack:

https://github.com/sustainable-computing-io/kepler/blob/3f117b3c8b198806a7f26cc0e46bb18830bb0789/pkg/collector/container_cgroup_collector.go#L45

```
func GetContainerMetrics() (containerCPU, containerMem map[string]float64, nodeCPU, nodeMem float64, retErr error) {
	return podLister.ListMetrics()
}
```

```
// ListMetrics accesses Kubelet's metrics and obtain pods and node metrics
func (k *KubeletPodLister) ListMetrics() (containerCPU, containerMem map[string]float64, nodeCPU, nodeMem float64, retErr error) {
	resp, err := httpGet(metricsURL)
...
```

hence each time for `updateKubeletMetrics`, before it goes to loop containers/pods, there is a IO at networking to get and convert kubelet. 
Usually we thought IO is blocking, as the major time cost, and meaning less for benchmarking based on container.
Suggestion as decouple this network IO from code stack.

and btw, why we rebuild constants below for each container?
```
	cpuMetricName := collector_metric.AvailableKubeletMetrics[0]
	memMetricName := collector_metric.AvailableKubeletMetrics[1]
```

marceloamaral:
Ok, this is a good opportunity to discuss the need for both kubelet and cgroup metrics. 
In my opinion, it seems that the kubelet metrics may duplicate the cgroup metrics, so we only need the cgroup metrics.

@sunya-ch @rootfs ↑↑

SamYuan1990:
> Ok, this is a good opportunity to discuss the need for both kubelet and cgroup metrics. In my opinion, it seems that the kubelet metrics may duplicate the cgroup metrics, so we only need the cgroup metrics.
> 
> @sunya-ch @rootfs ↑↑

I suppose we need a scope of both kubelet features and cgroup features. Can anyone help?
CC: @jichenjc

SamYuan1990:
link with https://github.com/sustainable-computing-io/kepler/issues/605

rootfs:
The following is my mental map. For cgroup v1, kubelet metrics is the way to go.

Models | Features | Use Case
-- | -- | --
CounterOnly | HW Perf Counters | Baremetal servers
[...]

Issue #605, opened 2023-03-28
Refactor for code seems?


SamYuan1990:
https://github.com/sustainable-computing-io/kepler/blob/3f117b3c8b198806a7f26cc0e46bb18830bb0789/pkg/collector/container_cgroup_collector.go#L30-L39

In recently I am checking code blocks above, and here are some thing I request refactor:
Overall speaking, in code blocks above, 
we generate stat Readers for each container, and ... considering that `updateCgroupMetrics` is invoked every 3 seconds, hence as there is 
- [ ] No cache hence each time rebuild for stat reader
Which means for a running container/pod, we rebuild it's statReaders for each time, with out cache.
in my point of view, the statReaders just contains as Path for files, hence for a running container in Cgroup, those path are static, and we don't need rebuild.
- [ ] Data handle couple with fetch data
and then the next item comes, as `func convertToStandard(stats map[string]interface{}) map[string]interface{}` .
which means each time we construct data handler with data.(fetch data with construct data handler at same time)
- [ ] Force type convert 
further more, as we have to use `readVal := cgroupFSValue.(uint64)` as force type convert to set up value.

marceloamaral:
@SamYuan1990 good catch!

I do agree that we need to improve this.

Btw, what is the status of the investigation to use an external library to collect the cgroup metrics?
For example, as kublet use cadvisor, can we also use it?
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/cadvisor_linux.go#L36

SamYuan1990:
> https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/cadvisor_linux.go#L36

em ? https://github.com/sustainable-computing-io/kepler/discussions/558#discussioncomment-5275875
does anyone work on it?

jichenjc:
I assume this will be combined with #606（remove kubelet and use cgroup instead) because of we anyway need update code to adopt cgroupv1/v2 ? @marceloamaral @SamYuan1990 
otherwise I can spend some time on it

SamYuan1990:
> I assume this will be combined with #606（remove kubelet and use cgroup instead) because of we anyway need update code to adopt cgroupv1/v2 ? @marceloamaral @SamYuan1990 otherwise I can spend some time on #588( switch to containerd/cgroup)

I think the final outcome for this issue is same with #606

jichenjc:
[...]

PR #609, opened 2023-03-31
Add external lib should support both cgroup v1 and v2

marceloamaral:
The current code for collecting cgroup status metrics is only compatible with cgroup v2, and is somewhat intricate and challenging to modify. The machine with cgroup v1 always has the following message:
```
I0404 04:11:50.148797       1 slice_handler.go:179] Not able to find any valid .scope file in /sys/fs/cgroup/cpu/system.slice, this likely cause all cgroup metrics to be 0
```

With this pull request, we propose incorporating an external library that will enable the collection of metrics for both cgroup v1 and v2, streamlining the process and enhancing maintainability. The lib also support collecting metrics from different operating system and container runtimes.

This PR also attempts to consolidate the cgroup metrics in only one place, but I did not removed the kubelet metrics yet.

Since macOS [(darwin) OS doesn't support cgroups](https://github.com/containerd/cgroups/issues/55), this PR deactivates cgroup metric collection for such OS. Note that while golang build tags could be used to remove the relevant code during compilation, but the [vendoring process doesn't currently have this capability](https://github.com/golang/go/issues/48859). Consequently, I'm temporarily removing the file with library dependencies during macOS tests, and then restoring it afterwards. This process can be seen in the Makefile.

As part of this PR, the `BytesRead` and `BytesWrite` metrics have been removed from the `containerMetrics` map and are now solely included in the `CgroupFSStats` metrics map to prevent duplication.

After this PR we can see cgroup metrics in a system with cgroup v1:
```
I0331 14:19:59.203637       1 container_cgroup_collector.go:41] energy from pod/container (1 active processes): name: prometheus-k8s-0/prometheus namespace: monitoring
	cgrouppid: 0 pid: [3521699] comm: prometheus
	Dyn ePkg (mJ): 0 (0) (eCore: 0 (0) eDram: 0 (0) eUncore: 0 (0)) eGPU (mJ): 0 (0) eOther (mJ): 0 (10719)
	Idle ePkg (mJ): 0 (110280) (eCore: 0 (0) eDram: 0 (80557) eUncore: 0 (0)) eGPU (mJ): 0 (0) eOther (mJ): 0 (1166917)
	CPUTime:  0 (0)
	NetTX IRQ: 0 (0)
	NetRX IRQ: 5 (48)
	Block IRQ: 0 (0)
	counters: map[cache_miss:0 (0) cpu_cycles:0 (0) cpu_instr:0 (0) cpu_ref_cycles:0 (0)]
	cgroupfs: map[block_devices_used:0 (12) cgroupfs_cpu_usage_us:5547 (310939430) cgroupfs_ioread_bytes:0 (0) cgroupfs_iowrite_bytes:0 (43868160) cgroupfs_kernel_memory_usage_bytes:0 (16101376) cgroupfs_memory_usage_bytes:0 (109453312) cgroupfs_system_cpu_usage_us:10000 (104750000) cgroupfs_tcp_memory_usage_bytes:0 (0) cgroupfs_user_cpu_usage_us:0 (197460000)]
	kubelets: map[container_cpu_usage_seconds_total:0 (310) container_memory_working_set_bytes:0 (97976320)]
```

This PR is related to #606 #605.


pkg/cgroup/cgroup_stats.go
```diff
@@ -17,84 +17,110 @@ limitations under the License.
 package cgroup
```
SamYuan1990, line 17: we'd better have one ore more interfaces for this package and follow https://github.com/sustainable-computing-io/kepler/blob/main/CONTRIBUTING.md#macos to make mac unit test passed.

pkg/cgroup/cgroup_stats.go
```diff
@@ -17,84 +17,110 @@ limitations under the License.
...
+	statsv2 "github.com/containerd/cgroups/v3/cgroup2/stats"
+
 	"github.com/sustainable-computing-io/kepler/pkg/config"
 )
 
-const (
-	ioStatFile = "io.stat"
-	reIOStat   = `(\d+):(\d+).rbytes=(\d+).wbytes=(\d+)` // 8:16 rbytes=58032128 wbytes=0 rios=120 wios=0 dbytes=0 dios=0
-)
+func getcGroupStat(pid int) (interface{}, error) {
```
SamYuan1990, line 31: are we able to avoid using `interface{}` but use an `type interface` instead?
even if I don't want to do to deep optimism on cpu stack....
but an `interface{}` means we need at least one or two additional steps at CPU stack.
one for 
```
if cgroup==1
else
```
one for `(*statsv1.Metrics).CPU.Usage.Total` ?
considering that we will do this stack for each `PID` and memory,cpu,IO as PID * 3 times, some how it is wasted.

pkg/cgroup/cgroup_stats.go
```diff
@@ -17,84 +17,110 @@ limitations under the License.
...
+	}
 }
 
-func ReadCgroupIOStat(cGroupID, pid uint64) (rBytes, wBytes uint64, disks int, err error) {
-	var path string
-	if config.EnabledEBPFCgroupID {
-		path, err = getPathFromcGroupID(cGroupID)
-	} else {
-		path, err = getPathFromPID(procPath, pid)
+func GetAvailableCgroupMetrics() []string {
```
SamYuan1990, line 58: const? instead of function?

pkg/cgroup/cgroup_stats.go
```diff
@@ -17,84 +17,110 @@ limitations under the License.
...
-		return readIOStat(p)
+
+	readCgroupMemoryStat(statsHandler, statsMap)
+	readCgroupCPUStat(statsHandler, statsMap)
+	readCgroupIOStat(statsHandler, statsMap)
+	return statsMap, nil
+}
+
+func readCgroupMemoryStat(statsHandler interface{}, statsMap map[string]uint64) {
+	if config.GetCGroupVersion() == 1 {
```
SamYuan1990, line 83: cpu stack as above....

pkg/collector/container_cgroup_collector.go
```diff
@@ -25,15 +25,17 @@ import (
...
-	klog.V(5).Infof("overall cgroup stats %v", cgroup.SliceHandlerInstance)
-
-	for containerID := range c.ContainersMetrics {
-		cgroup.TryInitStatReaders(containerID)
-		cgroupFSStandardStats := cgroup.GetStandardStat(containerID)
-		for cgroupFSKey, cgroupFSValue := range cgroupFSStandardStats {
-			readVal := cgroupFSValue.(uint64)
-			if _, ok := c.ContainersMetrics[containerID].CgroupFSStats[cgroupFSKey]; ok {
-				c.ContainersMetrics[containerID].CgroupFSStats[cgroupFSKey].SetAggrStat(containerID, readVal)
+	for key := range c.ContainersMetrics {
```
SamYuan1990, line 28: I was wondering that:
for each ContainersMetrics has a cgroup `control`, and for each time we do updateCgroupMetrics()
we just get data from `ContainersMetrics.control.stat()` and refresh data there.

current implementation still 
- reload control for each time from disk.(wasted?)
- and a temp memory make as `cGroupMetrics`(wasted?)

pkg/collector/container_hc_collector.go
```diff
@@ -125,31 +130,21 @@ func (c *Collector) updateBPFMetrics() {
...
 			c.createProcessMetricsIfNotExist(ct.PID, comm)
-			if err := c.ProcessMetrics[ct.PID].CPUTime.AddNewDelta(ct.ProcessRunTime); err != nil {
+			err := c.ProcessMetrics[ct.PID].CPUTime.AddNewDelta(ct.ProcessRunTime)
+			if err != nil {
 				klog.V(5).Infoln(err)
 			}
 		}
 
 		c.ContainersMetrics[containerID].CurrProcesses++
-		// system process should not include container event
```
SamYuan1990, line 141: are we able to remove this part?

pkg/cgroup/cgroup_stats_test.go
```diff
@@ -1,58 +0,0 @@
-package cgroup
```
SamYuan1990, line 1: we should consider with test for linux

SamYuan1990 (changes requested):

pkg/cgroup/cgroup_stats.go
marceloamaral: We can return the content in a function, similar to the other functions employed [here](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/metric/stats.go#L41). Typically, it's suggested to employ Get and Set to access values from other packages. However, I agree that  we don't need to create a new variable for each function call, even if it's called just once. I will update this.

pkg/cgroup/cgroup_stats.go
marceloamaral: It makes sense. I will update it to use `type cCgroupStat interface{}`

pkg/collector/container_hc_collector.go
marceloamaral: Yes, because we read those metrics in the `updateCgroupMetrics()`

marceloamaral: @sunya-ch is it ok only keep the `BytesRead` and `BytesWrite` metrics only in the `CgroupFSStats` metrics?

SamYuan1990:
@marceloamaral , I suppose we can use this PR to fix #606 and another one for #605 ?

pkg/collector/container_cgroup_collector.go
marceloamaral: @SamYuan1990 I changed the logic, now we open the stat only once for the lifetime of the container.

pkg/collector/container_hc_collector.go
```diff
@@ -55,19 +55,22 @@ func (c *Collector) resetBPFTables() {
 func (c *Collector) updateBasicBPF(containerID string, ct *ProcessBPFMetrics, isSystemProcess bool) {
 	// update ebpf metrics
 	// first update CPU time
-	if err := c.ContainersMetrics[containerID].CPUTime.AddNewDelta(ct.ProcessRunTime); err != nil {
+	err := c.ContainersMetrics[containerID].CPUTime.AddNewDelta(ct.ProcessRunTime)
```
marceloamaral, line 58: I changed logic to handle the err because of `golint` issues.

pkg/cgroup/cgroup_stats.go
marceloamaral: Fixed that. Thanks.

pkg/cgroup/cgroup_stats_test.go
marceloamaral: I added the test.

marceloamaral:
@SamYuan1990 this PR is ready to review again!

> I suppose we can use this PR to fix https://github.com/sustainable-computing-io/kepler/issues/606 and another one for https://github.com/sustainable-computing-io/kepler/issues/605 ?

Yes!

pkg/cgroup/cgroup_stats_darwin.go
```diff
@@ -0,0 +1,40 @@
+//go:build darwin
+// +build darwin
+
+/*
+Copyright 2021.
```
jichenjc, line 5: nit: 2023

pkg/cgroup/cgroup_stats_test.go
```diff
@@ -2,57 +2,124 @@ package cgroup
...
+	"github.com/containerd/cgroups/v3/cgroup1"
+	"github.com/opencontainers/runtime-spec/specs-go"
+
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+)
 
-	"github.com/sustainable-computing-io/kepler/pkg/utils"
+const (
+	defaultDirPerm = 0o755
```
jichenjc, line 18: weird, I am having this error when executing `make` but I am able to run `make test` 
```
# make
make: command: Command not found
make: command: Command not found
GO111MODULE=off go get -d github.com/go-bindata/go-bindata/...
./hack/bindata.sh
go mod tidy -v
go mod vendor
./automation/presubmit-tests/gofmt.sh
Checking go format
pkg/cgroup/cgroup_stats_test.go:18:20: expected ';', found 'IDENT' o755
pkg/cgroup/cgroup_stats_test.go:26:1: expected 'IDENT', found 'var'
pkg/cgroup/cgroup_stats_test.go:28:1: expected 'IDENT', found 'type'
pkg/cgroup/cgroup_stats_test.go:28:1: missing constant value
pkg/cgroup/cgroup_stats_test.go:28:17: expected ';', found 'struct'

```

marceloamaral: This is very strange, I can run `./automation/presubmit-tests/gofmt.sh` with no problem....
It also ran in the CI with no problem....

@jichenjc did you do a new `git clone` or revert a old code?
Also, did you try to see those lines in `pkg/cgroup/cgroup_stats_test.go` that the go format is complaining?

jichenjc: I will check more on this , this is an existing repo I do 
```
git fetch upstream pull/609/head:cgroupv1v2
```

to Pull your PR and do some test locally..

marceloamaral: Humm, I see. It might be worth trying to do a new `git clone` of kepler and then do `git fetch upstream pull/609/head:cgroupv1v2`. But I am not sure what is the problem....

pkg/cgroup/cgroup_stats_linux.go
```diff
@@ -0,0 +1,145 @@
...
+		manager, err := cgroups.Load(cgroups.V1, cgroups.StaticPath(path))
+		if err != nil {
+			return &hander, err
+		}
+		v1StatHandler, err := manager.Stat(cgroups.IgnoreNotExist)
+		if err != nil {
+			return &hander, err
+		}
+		hander.cgroupV1StatHandler = v1StatHandler
+	} else {
```
jichenjc, line 65: not sure worthy a quick comment on what we are doing here ..

marceloamaral: @jichenjc I included a comment in the function.

jichenjc:
/lgtm

I compared the metrics in original cgroup/stat_reader.go and new cgroup metrics
seems we still keep those in the new implmentation so think we are good to do go for more test by merge it
```
config.CgroupfsMemory: {
                {Name: "memory.current", Converter: DefaultConverter},
                {Name: "memory.usage_in_bytes", Converter: DefaultConverter},
        },
        config.CgroupfsKernelMemory: {
                {Name: "memory.kmem.usage_in_bytes", Converter: DefaultConverter},
        },
        config.CgroupfsTCPMemory: {
                {Name: "memory.kmem.tcp.usage_in_bytes", Converter: DefaultConverter},
        },
        config.CgroupfsCPU: {
                {Name: "cpuacct.usage", Converter: NanoToMicroConverter},
                {Name: "usage_usec", Converter: DefaultConverter},
        },
        config.CgroupfsSystemCPU: {
                {Name: "cpuacct.usage_sys", Converter: NanoToMicroConverter},
                {Name: "system_usec", Converter: DefaultConverter},
        },
        config.CgroupfsUserCPU: {
                {Name: "cpuacct.usage_user", Converter: NanoToMicroConverter},
                {Name: "user_usec", Converter: DefaultConverter},
        },
        config.CgroupfsReadIO: {
                {Name: "rbytes", Converter: DefaultConverter},
        },
        config.CgroupfsWriteIO: {
                {Name: "wbytes", Converter: DefaultConverter},
        },

```

pkg/cgroup/cgroup_stats_darwin.go
marceloamaral: fixed!

pkg/collector/container_cgroup_collector.go
```diff
@@ -25,15 +25,34 @@ import (
...
+		if c.ContainersMetrics[key].PID == 0 {
+			klog.V(3).Infof("PID 0 does not have cgroup")
+			return
```
SamYuan1990, line 31: return or continue?

pkg/collector/container_cgroup_collector.go
```diff
@@ -25,15 +25,34 @@ import (
...
+			handler, err := cgroup.NewCGroupStatHandler(int(c.ContainersMetrics[key].PID))
+			if err != nil {
+				klog.V(3).Infoln("Error: could not start cgroup stat handler for PID:", c.ContainersMetrics[key].PID)
+				return
+			}
+			c.ContainersMetrics[key].CgroupStatHandler = handler
+		}
+
+		// we need to check again if CgroupStatHandler is not nil because on darwin OS the handler will always be nil
+		if c.ContainersMetrics[key].CgroupStatHandler != nil {
```
SamYuan1990, line 43: maybe we can have another PR to update CgroupStatMap from GetCGroupStat without cGroupMetrics as a by pass value to reduce memory alloc and GC.

pkg/collector/container_cgroup_collector.go
SamYuan1990: @marceloamaral , this one as blocking with PR merge from my point of view.

pkg/collector/container_hc_collector.go
SamYuan1990: @sunya-ch , could you please confirm this? once @sunya-ch looks good, we can go ahead with merge.

SamYuan1990 (approved):
Considering it as a huge PR and great work. I would like leave comments in other pr for implementation. But, I hope we double confirm `return` for PID 0 before merge it.
@rootfs , @sunya-ch , @marceloamaral , @jichenjc

.github/workflows/integration_test.yml
```diff
@@ -43,6 +43,6 @@ jobs:
           kubectl port-forward --address localhost $(kubectl -n kepler get pods -oname) 9102:9102 -n kepler -v7 &
           kubectl logs -n kepler daemonset/kepler-exporter
           kubectl get pods -n kepler -o yaml
-          go test -tags=bcc ./e2e/... --race --bench=. -cover --count=1 --vet=all
+          go test -tags='bcc linux' ./e2e/... --race --bench=. -cover --count=1 --vet=all
```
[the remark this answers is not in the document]
marceloamaral: It needs the linux to test the cgroup test

Makefile
```diff
@@ -205,9 +208,14 @@ test: ginkgo-set tidy-vendor
 
 test-verbose: ginkgo-set tidy-vendor
 	@go test -tags $(GO_BUILD_TAGS) -covermode=atomic -coverprofile=coverage.out -v $$(go list ./... | grep pkg | grep -v bpfassets) --race --bench=. -cover --count=1 --vet=all
-
-test-mac-verbose: ginkgo-set tidy-vendor
-	@go test ./... --race --bench=. -cover --count=1 --vet=all
+	
+test-mac-verbose: ginkgo-set
```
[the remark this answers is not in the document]
marceloamaral: Unfortunately I cannot do it in another PR because the mac test in this PR will fail...

pkg/collector/container_cgroup_collector.go
marceloamaral: @SamYuan1990 actually, the message is misleading I will update it.
We cannot get cgroup metrics for the pid 0. Basically because there is no `/proc/0/cgroup`. 

So, i will update the message to `klog.V(3).Infof("PID 0 does not have cgroup metrics")`

marceloamaral: Good catch, it should be a `continue` not a `return`.

SamYuan1990: my focus is on the `return` here, if the PID is 0 and the function is returned. Hence other PID will be skipped?
in other word, do we have a case which `c.ContainersMetrics[0].PID==0` and `c.ContainersMetrics[1].PID==1` in same c.ContainersMetrics array?

marceloamaral: I have actually updated it to `continue`

Issue #301, opened 2022-10-16
aggregate energy consumption from Pod level to higher level


rootfs:
**Is your feature request related to a problem? Please describe.**
Pod level energy reporting is useful on a per node basis. At the cluster level, high level aggregation is more useful.

**Describe the solution you'd like**
When calling [`ListPods`](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/podlister/resolve_container.go#L131),  `OwnerReferences` in the Pod metadata should be parsed to get the Pods' owner. 

**Describe alternatives you've considered**

**Additional context**

jichenjc:
maybe `per application (OCP concept) or deployment level` as user might pay attention to the workload 
and daemonset consumption per cluster is a good measure to side effect from power perspective?

marceloamaral:
This is very useful, specially to report `Job` energy consumption

rootfs:
For Deployment Pods, there will be steps I think:
- at kubelet level, we can get the Pod owner. This works for certain API objects such as DaemonSet/ReplicaSet, not to Deployment,etc.  This query should be isolated to individual nodes, will not stress the API server.
- If we want to go the eventual owner of the ReplicaSet, we need to query API server and go through the ReplicaSet's OwnerReference to find the Deployment info. This query goes to the API server but since at the ReplicaSet level, the traffic is less frequent.

jichenjc:
>>>at kubelet level, we can get the Pod owner. This works for certain API objects such as DaemonSet/ReplicaSet, not to Deployment,etc. This query should be isolated to individual nodes, will not stress the API server.
    If we want to go the eventual owner of the ReplicaSet, we need to query API server and go through the ReplicaSet's OwnerReference to find the Deployment info. This query goes to the API server but since at the ReplicaSet level, the traffic is less frequent.

ok, this seems reasonable, maybe we can go from Daemonset first as it should be simpler

rootfs:
@jichenjc sounds good, PR welcome

marceloamaral:
Performing get operations for the API server can be expensive on a large scale....

How about creating a watcher with the node label as an event filter....
This should have very low overhead....

So we do the top-down approach instead of the bottom-up approach.

----
I like that this direction with watcher as this will also later support the challenge of identifying if a resource has been deleted

----
An example to watch pods, but the idea is to watch the other resources
```golang

import (
        k8sv1 "k8s.io/api/core/v1"
	"k8s.io/client-go/tools/cache"
[...]

PR #611, opened 2023-04-04
Update the collection of the container and pod info using container runtime

marceloamaral:
The current code retrieves a list of pods from the Kubelet API and searches through the entire list to gather container and pod details, such as name and ID. This approach can have performance and scalability issues if the node has a large number of pod and containers.

Moreover, there has been an issue (#606) with the Kubelet update being blocked by network, which highlights the need to eliminate dependence on the Kubelet API.

Additionally, in my experiments, there have been containers where the information obtained from the Kubelet API did not include the container ID, resulting in the container being classified as a `system_process`. This was due to a delay in the Kubelet's update of the container information, making it difficult to retrieve the information immediately after the container creation. Note that this issue is somewhat challenging to reproduce. 

This PR address these concerns by using the Kubelet library to access the container runtime directly instead of relying on the Kubelet API as a proxy. This approach offers several benefits, including (1) the ability to obtain information for a given container ID without traversing a list, (2) avoid the Kubelet API latency issues, and (3) reduced overhead on the Kubelet.

After this PR and the PR #609, we can probably completely remove the dependency on the Kubelet API.


SamYuan1990:
waiting for @rootfs 's feedback on security point of view about mount container runtime socket.

marceloamaral:
@rootfs has pointed the docker socket issue https://rules.sonarsource.com/kubernetes/RSPEC-6429. Basically we should avoid to expose the docker socket.

But what about cAdvisor? It exposes the docker socket
https://github.com/google/cadvisor/blob/a52ec5d60cf70b22f8b6d204780aec7a222cf6bb/deploy/kubernetes/base/daemonset.yaml

marceloamaral:
@rootfs, @SamYuan1990, @jichenjc, and @sunya-ch, it's worth noting that enabling access to the Kubelet API can also introduce security vulnerabilities. This article highlights some of the issues: https://faun.pub/attacking-kubernetes-clusters-using-the-kubelet-api-abafc36126ca.

As the container runtime security issue, it might also be possible to execute commands in any container using  the Kubelet API with a command like this: `https://${IP_ADDRESS}:10250/run/<namespace>/<pod>/<container> -d "cmd=<command-to-run>"`

However, if the cluster admin disables unauthenticated access to the Kubelet API, Kepler may not be able to resolve pod and container names. Perhaps it would be worth reconsidering the approach of watching for pods by filtering with the node name, as I previously suggested in this comment: https://github.com/sustainable-computing-io/kepler/issues/301#issuecomment-1281705775.

Should I close this PR and implement another version using `Watcher`?

SamYuan1990:
> @rootfs, @SamYuan1990, @jichenjc, and @sunya-ch, it's worth noting that enabling access to the Kubelet API can also introduce security vulnerabilities. This article highlights some of the issues: https://faun.pub/attacking-kubernetes-clusters-using-the-kubelet-api-abafc36126ca.
> 
> As the container runtime security issue, it might also be possible to execute commands in any container using the Kubelet API with a command like this: `https://${IP_ADDRESS}:10250/run/<namespace>/<pod>/<container> -d "cmd=<command-to-run>"`
> 
> However, if the cluster admin disables unauthenticated access to the Kubelet API, Kepler may not be able to resolve pod and container names. Perhaps it would be worth reconsidering the approach of watching for pods by filtering with the node name, as I previously suggested in this comment: [#301 (comment)](https://github.com/sustainable-computing-io/kepler/issues/301#issuecomment-1281705775).
> 
> Should I close this PR and implement another version using `Watcher`?

My understanding as:
- we need pod and container informations.
- we should get those in a secure way as much as possible. (vulnerabilities risks)

hence it sounds like a balance/trade off here, between security and feature.
I am not sure that we can reuse some codes from k8s, as kubelet/kubectl to get pod info?
as k8s is our dependency ... we'd better reuse k8s libraries/code.

marceloamaral:
@SamYuan1990 let's discuss this here https://github.com/sustainable-computing-io/kepler/discussions/615

PR #614, opened 2023-04-05
try to reduce memory usage by value set directly

SamYuan1990:
In my point of view, we can use GetCGroupStat to update metrics value without a map in the middle as value by pass.


SamYuan1990:
> @SamYuan1990 yes, we can do that!
> 
> I can review the PR again after you fix the errors.

please help test this PR on linux.

pkg/collector/metricdefine/define.go
```diff
@@ -0,0 +1,139 @@
+package metricdefine
```
marceloamaral, line 1: This is a type definition. Instead of use the `metricdefine` name, can we just use a `type` package? So we will have `pkg/collector/metric/type/type.go`

pkg/cgroup/cgroup_stats_linux.go
```diff
@@ -80,64 +81,40 @@ func NewCGroupStatHandler(pid int) (CCgroupStatHandler, error) {
...
-func readCgroupV2CPUStat(handler *statsv2.Metrics, statsMap map[string]uint64) {
-	statsMap[config.CgroupfsCPU] = handler.CPU.UsageUsec
-	statsMap[config.CgroupfsSystemCPU] = handler.CPU.SystemUsec
-	statsMap[config.CgroupfsUserCPU] = handler.CPU.UserUsec
-}
-
-func readCgroupV1IOStat(handler *statsv1.Metrics, statsMap map[string]uint64) {
-	// Each ioEntry is an io device.
-	for _, ioEntry := range handler.Blkio.IoServiceBytesRecursive {
+func (handler CCgroupV1StatHandler) GetCGroupStat(containerID string, cgroupStatMap map[string]*metricdefine.UInt64StatCollection) {
```
marceloamaral, line 84: This is not `Get` anymore. Can you change the name to `SetCGroupStat`?

pkg/cgroup/cgroup_stats_linux.go
```diff
@@ -80,64 +81,40 @@ func NewCGroupStatHandler(pid int) (CCgroupStatHandler, error) {
...
 	}
 }
 
-func readCgroupV2IOStat(handler *statsv2.Metrics, statsMap map[string]uint64) {
-	// Each ioEntry is an io device.
-	for _, ioEntry := range handler.Io.GetUsage() {
-		statsMap[config.CgroupfsReadIO] += ioEntry.Rbytes
-		statsMap[config.CgroupfsWriteIO] += ioEntry.Wbytes
-		statsMap[config.BlockDevicesIO] += 1
+func (handler CCgroupV2StatHandler) GetCGroupStat(containerID string, cgroupStatMap map[string]*metricdefine.UInt64StatCollection) {
```
marceloamaral, line 105: Same here.

marceloamaral (changes requested):
Looks good. I just have few comments.

pkg/collector/metricdefine/define.go
SamYuan1990: em... we can't name a package as type.

marceloamaral: Sorry, it should be `types` in plural.

We already have this in our code and it is commonly used in other projects.
See here: https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/types/types.go

SamYuan1990: updated

PR #616, opened 2023-04-06
Fix cgroup stat collector and build tags

marceloamaral:
There is a bug in the current code that collects cgroup metrics.
We are calling the stat function only once, so that we do not update the cgroup metrics and are only using old values.

This PR fix this.

### DIFF
--- a/pkg/cgroup/cgroup_interface.go
+++ b/pkg/cgroup/cgroup_interface.go
@@ -19,5 +19,5 @@ package cgroup
 import "github.com/sustainable-computing-io/kepler/pkg/collector/metric/types"
 
 type CCgroupStatHandler interface {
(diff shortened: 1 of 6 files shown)


jichenjc:
not sure why I can't trigger merge.. @marceloamaral maybe you can merge by yourself

marceloamaral:
@jichenjc 
humm, do you have the right to approve a PR? 
If yes, I need an approval before merging this PR.

jichenjc (approved):

jichenjc:
ok, I thought we use /approve /lgtm to auto merge :) 
but seems we don't yet , but anyway, glad it merged

SamYuan1990 (commented):
LGTM

Issue #610, opened 2023-04-04
Timely occur invalid memory address or nil pointer in updateAcceleratorMetrics


sunya-ch:
**Describe the bug**

There is timely occurring bug during collecting GPU metric. It seems some of the following component is not properly initialized by `createContainersMetricsIfNotExist`.

https://github.com/sustainable-computing-io/kepler/blob/4bdddb1e2b5829f142f8fb2beb7db54c8aee0538/pkg/collector/container_accelerator_collector.go#L57

https://github.com/sustainable-computing-io/kepler/blob/4bdddb1e2b5829f142f8fb2beb7db54c8aee0538/pkg/collector/container_accelerator_collector.go#L51

**To Reproduce**

Run Kepler on node with GPU enable for a while. We still cannot identify the trigger point.

**Expected behavior**

**Screenshots**

```bash
found 8 gpu devices
Using nvml to obtain gpu power
I0330 01:56:03.586089       1 exporter.go:150] Kepler running on version: 2b59dbd
I0330 01:56:03.586260       1 config.go:169] using gCgroup ID in the BPF program: true
I0330 01:56:03.586322       1 config.go:170] kernel version: 4.18
I0330 01:56:03.646782       1 rapl_msr_util.go:151] input/output error
I0330 01:56:03.646852       1 power.go:64] Not able to obtain power, use estimate method
I0330 01:56:03.695517       1 exporter.go:170] Initializing the GPU collector
perf_event_open: No such file or directory
I0330 01:56:04.678864       1 bcc_attacher.go:98] failed to attach perf event cpu_cycles_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0330 01:56:04.678916       1 bcc_attacher.go:98] failed to attach perf event cpu_ref_cycles_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0330 01:56:04.678956       1 bcc_attacher.go:98] failed to attach perf event cpu_instr_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0330 01:56:04.678998       1 bcc_attacher.go:98] failed to attach perf event cache_miss_hc_reader: failed to open bpf perf event: no such file or directory
I0330 01:56:04.679009       1 bcc_attacher.go:135] Successfully load eBPF module with option: [-DNUM_CPUS=80 -DSET_GROUP_ID]
[...]

PR #618, opened 2023-04-06
Fix null pointer issue in gpu metrics

marceloamaral:
According to issue #610, null pointer access can occur while collecting GPU metrics.

This PR fixes the null pointer error and also includes an attempt to re-initialize the GPU driver in case it is no longer available.

This PR also fix the problem that the GPU error message was nil.

### DIFF
--- a/pkg/collector/container_accelerator_collector.go
+++ b/pkg/collector/container_accelerator_collector.go
@@ -38,7 +38,7 @@ func (c *Collector) updateAcceleratorMetrics() {
 	// calculate the gpu's processes energy consumption for each gpu
 	for _, device := range accelerator.GetGpus() {
 		if processesUtilization, err = accelerator.GetProcessResourceUtilizationPerDevice(device, time.Since(lastUtilizationTimestamp)); err != nil {
-			klog.V(2).Infoln(err)
+			continue
 		}
 
 		var containerID string
--- a/pkg/power/accelerator/gpu.go
+++ b/pkg/power/accelerator/gpu.go
@@ -20,8 +20,6 @@ limitations under the License.
 package accelerator
 
 import (
-	"fmt"
-
 	accelerator_source "github.com/sustainable-computing-io/kepler/pkg/power/accelerator/source"
 	"k8s.io/klog/v2"
 )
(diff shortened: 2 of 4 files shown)


pkg/collector/container_accelerator_collector.go
```diff
@@ -38,7 +38,7 @@ func (c *Collector) updateAcceleratorMetrics() {
 	// calculate the gpu's processes energy consumption for each gpu
 	for _, device := range accelerator.GetGpus() {
 		if processesUtilization, err = accelerator.GetProcessResourceUtilizationPerDevice(device, time.Since(lastUtilizationTimestamp)); err != nil {
-			klog.V(2).Infoln(err)
```
sunya-ch, line 41: I think we should keep the info log for the error.

pkg/power/accelerator/gpu.go
```diff
@@ -37,12 +36,13 @@ Then, we use gpu.go file to initialize the acceleratorImpl from power.go when gp
 // init initialize the acceleratorImpl and start it
 func init() {
 	acceleratorImpl = &accelerator_source.GPUNvml{}
-	if err = acceleratorImpl.Init(); err == nil {
-		klog.V(1).Infoln("Using nvml to obtain gpu power")
-		fmt.Println("Using nvml to obtain gpu power")
+	err := acceleratorImpl.Init()
+	fmt.Println(err)
```
sunya-ch, line 40: change from printing to logger (klog)

pkg/power/accelerator/gpu.go
```diff
@@ -37,12 +36,13 @@ Then, we use gpu.go file to initialize the acceleratorImpl from power.go when gp
...
+	if err == nil {
+		klog.Infoln("Using nvml to obtain gpu power")
 		return
 	}
-	fmt.Printf("Failed to init nvml: %v, using dummy source to obtain gpu power\n", err)
+	klog.Infof("Failed to init nvml: %v, using dummy source to obtain gpu power\n", err)
```
sunya-ch, line 45: may add explicit V

pkg/collector/container_accelerator_collector.go
sunya-ch: Is there any chance that the err = nil but processUtilization is also nil?
Should we also add the condition to check `processUtilization  != nil`?
Also, why do we put the error log to V2. That may make us miss the error log if there was there in the PR https://github.com/sustainable-computing-io/kepler/issues/610.

sunya-ch (commented):
small changes may be considered especially on consistency on logging.

pkg/power/accelerator/power.go
```diff
@@ -22,11 +22,12 @@ import (
 
 	"github.com/sustainable-computing-io/kepler/pkg/config"
 	accelerator_source "github.com/sustainable-computing-io/kepler/pkg/power/accelerator/source"
+	"k8s.io/klog/v2"
 )
 
 var (
 	acceleratorImpl acceleratorInterface
-	err             = fmt.Errorf("could not start accelerator collector")
+	libErr          = fmt.Errorf("could not start accelerator collector")
```
sunya-ch, line 30: due to glint check: ST1012: error var libErr should have name of the form errFoo (stylecheck)

may change to errLib?

pkg/power/accelerator/gpu.go
marceloamaral: It was for debug, I will remove

pkg/collector/container_accelerator_collector.go
marceloamaral: processUtilization is nil because we create a nil map and always return the error...

Issue #594, opened 2023-03-20
Cannot see the metrics of pods created after Kepler deployment


andersonandrei:
**Describe the bug**

I'm running a serverless platform on top of AKS, so, for each function that I execute in the platform, a new pod is created. Besides, Kepler does not export metrics for new pods. It just exports metrics for the pods that were already in the platform before its own deployment (Kepler deployment). 
For instance, let us consider the following actions: 1) to deploy kepler, 2) to create a new pod, and 3) to delete kepler and deploy it again. If I do 1) and 2), Kepler does not export such pod metrics, so I need to do 1), 2) and 3), and still, the problem persists with new pods created after 3).

I’m using the Kepler interface through Prometheus, but I try the queries with the Kepler metric endpoint with curl from time to time as well. In both cases, I can just see new pods’ metrics after steps 1), 2), and 3).

**To Reproduce**

Deployed Kepler manually, using one of the manifest available, with slight modifications:

```
kind: Namespace
metadata:
  labels:
    sustainable-computing.io/app: kepler
  name: kepler
---
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    sustainable-computing.io/app: kepler
  name: kepler-sa
  namespace: monitoring
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    app.kubernetes.io/component: prometheus
    app.kubernetes.io/instance: k8s
    app.kubernetes.io/name: prometheus
    sustainable-computing.io/app: kepler
  name: prometheus-k8s
  namespace: monitoring
rules:
- apiGroups:
  - ""
  resources:
  - services
  - endpoints
  - pods
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - extensions
  resources:
  - ingresses
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - networking.k8s.io
  resources:
  - ingresses
  verbs:
  - get
  - list
  - watch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    sustainable-computing.io/app: kepler
  name: kepler-clusterrole
rules:
- apiGroups:
  - ""
  resources:
  - nodes/metrics
  - nodes/proxy
  - nodes/stats
  verbs:
  - get
  - watch
  - list
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  labels:
[...]

PR #631, opened 2023-04-14
nit fix to reduce unused memory allocate...

SamYuan1990:
for each time we created ContainerInfo 3 times as duplication.
this PR is a nit fix to reduce creation works as memory allocation.

we don't need to get kubelet IO for each time, hence which makes cache miss or cache ineffective.

remove two useless return value.


SamYuan1990:
not sure if this can help with #594 ?

SamYuan1990:
@marceloamaral  or @jichenjc may I know if this PR is still validated?

marceloamaral:
Can you rebase? This should be good to have.

SamYuan1990:
> Can you rebase? This should be good to have.

rebased. please help with review process. :-)

rootfs (approved):

Issue #633, opened 2023-04-17
Kepler fails to close files leading to frequent pod restarts due to "too many open files" error.


mikeyehua:
It seems Kepler fails to close open files, causing the number of open files to continuously increase until it reaches the limit of open files. This results in the error "too many open files" and frequent pod restarts.

```yaml
2023/04/17 04:38:54 http: Accept error: accept tcp [::]:9102: accept4: too many open files; retrying in 1s
2023/04/17 04:38:55 http: Accept error: accept tcp [::]:9102: accept4: too many open files; retrying in 1s
2023/04/17 04:38:56 http: Accept error: accept tcp [::]:9102: accept4: too many open files; retrying in 1s
I0417 04:38:56.790906       1 rapl_msr_util.go:139] failed to read topology /sys/devices/system/cpu/cpu0/topology/physical_package_id: open /sys/devices/system/cpu/cpu0/topology/physical_package_id: too many open files
I0417 04:38:56.791384       1 rapl_msr_util.go:139] failed to read topology /sys/devices/system/cpu/cpu0/topology/physical_package_id: open /sys/devices/system/cpu/cpu0/topology/physical_package_id: too many open files
```
The limitation of open files is quite large:
```
kubectl exec kepler-exporter-dpwj5 -n kepler -- ulimit -n
1048576
```

The export of the pod description: 

```yaml
Started:      Mon, 17 Apr 2023 12:39:00 +0800
Last State:     Terminated
Reason:       Error
Exit Code:    2
Started:      Mon, 17 Apr 2023 05:47:59 +0800
Finished:     Mon, 17 Apr 2023 12:38:59 +0800
Ready:          True
Restart Count:  80
```

```yaml
Events:
  Type     Reason     Age                    From     Message
  ----     ------     ----                   ----     -------
  Warning  Unhealthy  4m55s (x102 over 14d)  kubelet  Liveness probe failed: Get "http://172.16.15.190:9102/healthz": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
  Normal   Killing    4m55s (x80 over 14d)   kubelet  Container kepler-exporter failed liveness probe, will be restarted
  Normal   Pulled     4m52s (x81 over 14d)   kubelet  Container image "quay.io/sustainable_computing_io/kepler:latest" already present on machine
  Normal   Created    4m52s (x81 over 14d)   kubelet  Created container kepler-exporter
[...]

Issue #632, opened 2023-04-17
bug: keep open files for MSR


jichenjc:
**Describe the bug**

we have this call 
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/container_power.go#L80 for each update phase
but in MSR implementation https://github.com/sustainable-computing-io/kepler/blob/a6dce3b27561c0fb24c9bd33c254682813441424/pkg/power/components/source/rapl_msr.go#L21

at last, we will keep open files 
https://github.com/sustainable-computing-io/kepler/blob/a6dce3b27561c0fb24c9bd33c254682813441424/pkg/power/components/source/rapl_msr_util.go#L78

a sample test shows it will use `DIFFERENT` fd and lead to fd exhausted at last

```
(base) root@jitest19:~/go/src/c# cat h.go
package main

import (
        "syscall"
        "fmt"
)

func main() {
        fd, _ := syscall.Open("./a", syscall.O_RDONLY, 777)
        fmt.Println("----1", fd)
        fd, _ = syscall.Open("./a", syscall.O_RDONLY, 777)
        fmt.Println("----2", fd)
}
(base) root@jitest19:~/go/src/c# ./h
----1 3
----2 4

```
**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #634, opened 2023-04-17
fix: avoid continue open file in MSR model

jichenjc:
#632 #633


marceloamaral (approved):
Makes sense to test it only once.
/lgtm

Issue #364, opened 2022-11-07
Remove BPF dependency so that Kepler can run without BPF


marceloamaral:
**Is your feature request related to a problem? Please describe.**

The bpf collects metrics by processes and maps process ids to container ids.
The structure that contains the container information is `ContainersMetrics`. See [here](https://github.com/sustainable-computing-io/kepler/blob/bf17c6bfa1062e4d557d0aeb1ce2ffc67ade803e/pkg/collector/container_energy_collector.go#L33).

However, currently only the bpf code updates the `ContainersMetrics` map.
So if bpf is not running, the `ContainersMetrics` map will not be updated.

**Describe alternatives you've considered**

We have different approaches to fixing this.
Fix the bpf dependency issue. So that the bpf will always runs. Issue #323 starts this discussion.

Or find other ways to identify the running containers on the system without bpf.

rootfs:
Need to identify the use case that doesn't support/need eBPF and then follow up on the container finding case. Check alternative implementations on hardware counters vs e.g. cpu time.

marceloamaral:
The https://github.com/sustainable-computing-io/kepler/pull/635 will fix this issue

rootfs:
this is fixed now with preinstalled kernel source

PR #635, opened 2023-04-18
Add Apiserver apiserver client to add and remove containers

marceloamaral:
This PR fix #615.

This implementation involves registering Kepler with the apiserver to receive notifications about pod creation and deletion. Additionally, it filters events by node name while monitoring pods, which ensures the code's performance and scalability.

It is important to note that filtering by node label is not possible for jobs and deployments, therefore we only watch for pod updates.


pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,200 @@
...
+func (w *ObjListWatcher) Stop() {
+	close(w.stopChannel)
+}
+
+func (w *ObjListWatcher) handleUpdate(obj interface{}) {
+	switch w.ResourceKind {
+	case "pods":
+		pod, ok := obj.(*k8sv1.Pod)
+		if !ok {
+			klog.Infof("Could not convert obj: %v", w.ResourceKind)
```
jichenjc, line 140: I think we got code this from somewhere ? seems not able to convert should not continue following code?

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func NewObjListWatcher() *ObjListWatcher {
+	w := &ObjListWatcher{
+		stopChannel:  make(chan struct{}),
+		k8sCli:       newK8sClient(),
+		ResourceKind: podResourceType,
+	}
+	if w.k8sCli == nil {
+		return w
```
jichenjc, line 90: should we fatal and error out here?

pkg/kubernetes/watcher.go
marceloamaral: Yes, good catch!

pkg/kubernetes/watcher.go
marceloamaral: No since we can run kepler without the watcher as the original code.
For example, for testing cases that kepler is not running in kubernetes.

marceloamaral:
The issue #594 might be related to this PR.

sunya-ch (commented):
Nice. I leave some comments that might help improve scalability when we have a number of managed pods (auto-annotated and labeled by operators in the cluster).  

The rest looks good to me (should work but might consider add e2e test for another PR). 

Another thing to confirm is that whether [the previous comment about scalability issue](https://github.com/sustainable-computing-io/kepler/pull/89#issuecomment-1175012593) will be compromised since this is also getting info from API server. @rootfs 
In my opinion, as long as we cannot solve the missing info from local runtime API, relying on k8s API server is more preferable.

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (w *ObjListWatcher) handleUpdate(obj interface{}) {
+	switch w.ResourceKind {
+	case "pods":
+		pod, ok := obj.(*k8sv1.Pod)
+		if !ok {
+			klog.Infof("Could not convert obj: %v", w.ResourceKind)
+			return
+		}
+		for _, condition := range pod.Status.Conditions {
```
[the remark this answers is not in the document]
marceloamaral: Thanks @sunya-ch for the comments.

Indeed, we can incorporate a list to check whether the pod information has already been added and skip it if it has. I have updated the code for this.

The code you proposed checks the status of all containers, but we don't need to iterate through all containers for this verification. Instead, we can rely on the `ContainersReady` status of the pod, which indicates that all containers are ready.

marceloamaral:
@rootfs and @sunya-ch It is important to note that by filtering events with the node name, as the PR is doing, we can minimize the scalability issue. Also, the Kubernetes scalability community has stated that the API server now includes several optimizations for watching pod objects.

rootfs:
@dgrisonnet I think kube-state-metrics uses similar field selector to filter watches. Can you share your view? Thanks

marceloamaral:
Thanks @rootfs for the feedbask.

1. identify the use cases where kube-apiserver is the must-have solution.
-- **No overhead to kubelet.** Right now we introduce a very big overhead to kubelet. As do a list operation for every new pid that we find in the system. Additionally, for each list we need to traverse all the elements. In fact, the overhead to apiserver is potentially smaller than the overhead in the kubelet.
-- **Work without access to kubelet**. System admin can disable the access to kubelet for security reasons. So Kepler will not be able to resolve containers.
-- **Work without access to BPF**. Discover containers in a different flow than the BPF code, as detailed in #364. With the apiserver, Kepler will also be able to run without BPF when needed, as BPF requires special security capabilities and the kernel libraries be installed in the host node.

2. quantify the overhead on kube-apiserver. 
-- very interesting reference [here](https://www.sobyte.net/post/2022-06/k8s-reliability-list-data/#13-apiserveretcd-list-overhead). Thanks @rootfs. The article says that one-time LIST operation with filtering criteria is very efficient and we are doing that in this PR.
>But if it’s just a one-time LIST operation with filtering criteria, like the nodename filtering pod example mentioned earlier, then obviously we should let apiserver filter out the data for us by setting label or field filters. LIST 10w pods takes a few tens of seconds (most of the time is spent on data transfer and also takes up a lot of CPU/BW/IO on apiserver), while if only the pods on the local machine are needed, LIST may only take 0.05s to return the results after setting nodeName=node1. It is also very important not to forget to include resourceVersion=0 in the request. If you are using client-go’s ListWatch/informer interface, then it already has ResourceVersion=0 set by default.

>the amount of data involved in the LIST operation the amount of data (in json format) to be processed by apiserver: the above list pods, for example, corresponds to the listed-pods file, **totaling 2GB**. 
the amount of data received by the agent (since the agent may have specified **label/field filters**): in the case of the list pods above, corresponding to the listed-pods-filtered file, **totaling 526K** . Source [here](https://www.sobyte.net/post/2022-06/k8s-reliability-list-data/).

We should investigate later the support for [Watch bookmark feature](https://www.alibabacloud.com/blog/how-does-alibaba-ensure-the-performance-of-system-components-in-a-10000-node-kubernetes-cluster_595469) and [here](https://www.alibabacloud.com/blog/how-does-alibaba-ensure-the-performance-of-system-components-in-a-10000-node-kubernetes-cluster_595469), which can be used to optimize the pod list when the node reboots.

3. other alternatives. In #301
-- As I mentioned, only pod watch is efficient. We should not include other type of resource.

dgrisonnet:
I am at KubeCon right now, I'll try to have a look once I am back.

rootfs:
I've asked OpenShift performance team to share their opinion on this, stay tuned.

pkg/collector/metric/utils.go
```diff
@@ -99,6 +99,9 @@ func isCounterStatEnabled(label string) bool {
 }
 
 func getNodeName() string {
+	if nodeName := os.Getenv("NODE_NAME"); nodeName != "" {
```
SamYuan1990, line 102: ? is it necessary in this PR?

pkg/collector/utils.go
```diff
@@ -20,30 +20,40 @@ import (
...
 	if _, ok := c.ContainersMetrics[containerID]; !ok {
-		var err error
-		var podName, containerName string
-
-		// the acceletator package does not get the cgroup ID, then we need to verify if we can call the cgroup with cGroupID or not
-		podName, _ = cgroup.GetPodName(cGroupID, pid, withCGroupID)
-		containerName, _ = cgroup.GetContainerName(cGroupID, pid, withCGroupID)
-
+		// We feel the info with generic values because the watcher will eventually update it.
+		podName := c.systemProcessName
```
SamYuan1990, line 31: we don't need to alloc a new var named as `podName` in each loop right?

pkg/kubernetes/watcher.go
SamYuan1990: For e2e test case, we can export k8s config and run the test if possible https://github.com/sustainable-computing-io/kepler/blob/main/e2e/e2e_metric_test.go#L50 just for example. 
For unit test case, I am not sure ...

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,213 @@
...
+		stopChannel:  make(chan struct{}),
+		k8sCli:       newK8sClient(),
+		ResourceKind: podResourceType,
+	}
+	if w.k8sCli == nil {
+		return w
+	}
+
+	optionsModifier := func(options *metav1.ListOptions) {
+		options.FieldSelector = fmt.Sprintf("spec.nodeName=%s", collector_metric.NodeName) // to filter events per node
```
SamYuan1990, line 96: what if node name is different with host name? any risk?

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,213 @@
...
+	})
+	IsWatcherEnabled = true
+	return w
+}
+
+func (w *ObjListWatcher) Run() {
+	if !IsWatcherEnabled {
+		return
+	}
+	go w.informer.Run(w.stopChannel)
```
SamYuan1990, line 123: this implementation is referenced from official sdk, exits in safety?

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+}
+
+func (w *ObjListWatcher) Stop() {
+	close(w.stopChannel)
+}
+
+func (w *ObjListWatcher) handleUpdate(obj interface{}) {
+	switch w.ResourceKind {
+	case "pods":
```
SamYuan1990, line 140: "pods" -> podResourceType

pkg/manager/manager.go
```diff
@@ -45,6 +49,11 @@ func New() *CollectorManager {
 	manager.PrometheusCollector.ContainersMetrics = &manager.MetricCollector.ContainersMetrics
 	manager.PrometheusCollector.ProcessMetrics = &manager.MetricCollector.ProcessMetrics
 	manager.PrometheusCollector.SamplePeriodSec = SamplePeriodSec
+	// configure the wather
+	manager.Watcher = kubernetes.NewObjListWatcher()
+	manager.Watcher.Mx = &manager.PrometheusCollector.Mx
```
SamYuan1990, line 54: create Watcher with Mx instead of overwrite.

pkg/collector/metric/utils.go
marceloamaral: Yes, because we filter the events based on the node name due to performance reasons

pkg/kubernetes/watcher.go
marceloamaral: We use the node name that kubernetes uses, so it will never be different.

pkg/kubernetes/watcher.go
marceloamaral: Sorry, can you explain more about this?

pkg/manager/manager.go
marceloamaral: It does not overwrite. It uses the shared locker across other packages.

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,213 @@
...
+	regexReplaceContainerIDPrefix = regexp.MustCompile(`.*//`)
+	IsWatcherEnabled              = false
+	managedPods                   = make(map[string]bool)
+)
+
+type ObjListWatcher struct {
+	// Lock to syncronize the collector update with the watcher
+	Mx *sync.Mutex
+
+	k8sCli       *kubernetes.Clientset
```
dgrisonnet, line 52: you only need the clientset to intiliaze the kubernetes list watcher so there shouldn't be any needs to keep it in the struct

dgrisonnet (commented):
The field selection looks good to me. This is generally how you want to do it with the Kubernetes client.

> ListWatch filtering requires additional cpu/memory on kube-apiserver, this will make things worse in large clusters or with volatile workload. There is a good analysis on LIST in [this](https://www.sobyte.net/post/2022-06/k8s-reliability-list-data/#13-apiserveretcd-list-overhead) post with discussion on node name filter.

That is correct. The thing to keep in mind is that fieldSelector is for the most part a feature that benefits client-side code. As mentioned in the blog post, the kube-apiserver still has to watch everything. However, if there are already controllers watching all the pod resources in the cluster, the overhead from the kube-apiserver perspective will not be very significant.

What tend to be dangerous when you are not using a fieldSelector is to have too many resource in your client cache. Which is one of the reason why we introduced sharding via fieldSelector in kube-state-metrics: https://github.com/kubernetes/kube-state-metrics/pull/1864

pkg/collector/utils.go
marceloamaral: We replace the content when there is no watcher....
I am keeping this for now, but we can improve it later...

pkg/kubernetes/watcher.go
marceloamaral: thanks

pkg/kubernetes/watcher.go
marceloamaral: Not now, but we might use it later to run benchmarks for model training.

marceloamaral:
Thanks @dgrisonnet for the feedback. 
In fact, in this PR we are sharding via fieldSelector as the default approach.

caiocfer:
I was testing this PR because it was mentioned on this issue https://github.com/sustainable-computing-io/kepler/issues/594, I noticed that metrics of pods created after Kepler is deployed started to appear, but after sometime they don't appear anymore, only pods deployed before Kepler were available.
Using some logs I noticed that prometheus_collector only sends the new pods' metrics a couple of times.

pkg/kubernetes/watcher.go
SamYuan1990: I want to ensure the go routine safety here.

marceloamaral:
@caiocfer 

> Using some logs I noticed that prometheus_collector only sends the new pods' metrics a couple of times.

Are the pods currently executing any tasks or are they idle? If a pod is idle, Kepler does not report any metrics and recognizes it as having 0 activity.

Also, can you please run the following command just after executing the workload?
```bash
kubectl exec -it -n kepler daemonset/kepler-exporter \
   -- bash -c "curl localhost:9102/metrics|grep kepler_container" | grep $workloadname
```
In case there are multiple nodes, substitute `daemonset/kepler-exporter` with `pod/$podname`, where `$podname` should be replaced with the name of the Kepler pod running on the node where the workload was executed.

caiocfer:
I was using some stress pods, oddly enough that command returns empty and grafana has 3 metrics of the pod and then stopped receiving.

pkg/kubernetes/watcher.go
```diff
@@ -0,0 +1,213 @@
...
+		pod, ok := obj.(*k8sv1.Pod)
+		if !ok {
+			klog.Infof("Could not convert obj: %v", w.ResourceKind)
+			return
+		}
+		podID := string(pod.GetUID())
+		// Pod object can have many updates such as change in the annotations and labels.
+		// We only add the pod information when all containers are ready, then when the
+		// pod is in our managed list we can skip the informantion update.
+		if _, exist := managedPods[podID]; exist {
```
sunya-ch, line 150: There is an issue here for the new-coming pod. 

For the early read,  even if you have line `condition.Type != k8sv1.ContainersReady && condition.Status != k8sv1.ConditionTrue ` to check but it turns out that container ID has not been assigned yet. 
I have to remove this check to have my new pod added by the watcher.

```
# I added a following log line to see an update 
# klog.Infof("Watcher add new containerID: %s (pod: %s)", containerID, pod.Name)

I0511 08:53:31.124701       1 watcher.go:181] Watcher add new containerID:  (pod: coremark-cpeh-3068050960-k5qgr)
I0511 08:53:33.844029       1 watcher.go:181] Watcher add new containerID: 9aab3a72841724834469a04dd9c770f7b001bb3c38d6ad046ee13097e17db8ed (pod: coremark-cpeh-3068050960-k5qgr)
```

You can notice that the first assign of managedPods containerID is empty. You may try adding empty Check in addition to the above condition.

sunya-ch (changes requested):

pkg/kubernetes/watcher.go
marceloamaral: Thanks @sunya-ch for finding this issue! 
I updated the PR, could you please test it again?

sunya-ch: Still there is missing system process detection. I'm still under debugging.

```bash
# I added the following debugging log
#		if isSystemProcess {
#			klog.V(5).Infof("System process command: %s (cGroup ID %v)", comm, ct.CGroupID)
#		}
kubectl logs kepler-exporter-vdjlc -n kepler|grep "System process command"|grep coremark
I0511 09:27:00.438820       1 container_hc_collector.go:143] System process command: coremark-16thre (cGroup ID 0)
I0511 09:27:03.455934       1 container_hc_collector.go:143] System process command: coremark-16thre (cGroup ID 0)
```

marceloamaral: @sunya-ch it might be related to the function `extractPodContainerIDfromPath` in the file pkg/cgroup/resolve_container.go:L297

sunya-ch (approved):
I am in the middle of debugging for the other part as some processes are still missing. 
I haven't rebased and tested this updated code yet. However, it looks good to me.

pkg/kubernetes/watcher.go
sunya-ch: My current problem is that some processes is still missing (not even be processed in the `updateBPFMetrics` function). For the above issue, I think you updated changes look good to me.

sunya-ch (commented):
Until I found the root cause of current issue, I would like to suspend my approval.

marceloamaral:
@sunya-ch I converted to a draft so no one will accidentally merge this PR until it is fully approved.

caiocfer:
> I was testing this PR because it was mentioned on this issue #594, I noticed that metrics of pods created after Kepler is deployed started to appear, but after sometime they don't appear anymore, only pods deployed before Kepler were available. Using some logs I noticed that prometheus_collector only sends the new pods' metrics a couple of times.

After the last rebase, in my tests, kepler started listing new pods.

sunya-ch (changes requested):
With API server watcher, we can avoid uncertain logic to delete inactive pod at two points below (because we already have logic to remove at delete event). 

https://github.com/sustainable-computing-io/kepler/blob/839bc2a6f810535cebe02ed8e1ebc733e5af9c20/pkg/collector/container_cgroup_collector.go#L47-L49

and 

https://github.com/sustainable-computing-io/kepler/blob/839bc2a6f810535cebe02ed8e1ebc733e5af9c20/pkg/collector/container_hc_collector.go#L175

In my case, a valid entry ContainersMetrics has been removed due to UpdateCgroupMetrics error. And that cause, all stat goes to system process. 

I think we should add the condition`!kubernetes.IsWatcherEnabled` to limit the above deletion logic to only when no watcher.

FYI, 
I am working on debugging code on minikube.
It turns out that `UpdateCgroupMetrics` returns file ( /sys/fs/cgroup/pids/pids.current) not found error. 
Here is my temporary logging.
```golang
		if err != nil {
			klog.V(1).Infof("Cannot UpdateCgroupMetrics %s: %v)", c.ContainersMetrics[key].ContainerName, err)
		}
```

Then I got,
```
Cannot UpdateCgroupMetrics alertmanager: open /sys/fs/cgroup/pids/pids.current: no such file or directory)
```

The pids.current is really not in that folder but inside `kubepod` folder and inside each cgroup folder.

marceloamaral:
Thanks @sunya-ch for this investigation!

It is very curious why minikube changes a lot of cgroup paths……
But it is nice that we are finding some differences between minikube and other k8s clusters…

Lets create a function like k8s library has to ignore some errors. Such as `IgnoreCGroupPidsError(err)`. 
Then we parse the current error and ignore if it did not find the `pids.current`.

rootfs:
here is the kepler container image with this PR quay.io/sustainable_computing_io/kepler:pr-635

andersonandrei:
Hello,

I tried to use image quay.io/sustainable_computing_io/kepler:pr-635 and it seems with a few errors. Also,  the API is not working anymore.

`
> kubectl logs kepler-exporter-xgxjs -n kepler
I0524 12:40:43.422578       1 gpu.go:46] Failed to init nvml, err: could not init nvml: error opening libnvidia-ml.so.1: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
I0524 12:40:43.432658       1 exporter.go:149] Kepler running on version: v0.2-1109-g45d101f-dirty
I0524 12:40:43.432691       1 config.go:175] using gCgroup ID in the BPF program: true
I0524 12:40:43.432874       1 config.go:177] kernel version: 5.4
I0524 12:40:43.432902       1 exporter.go:163] EnabledBPFBatchDelete: true
I0524 12:40:43.432943       1 rapl_msr_util.go:129] failed to open path /dev/cpu/0/msr: no such file or directory
I0524 12:40:43.432991       1 power.go:77] Not able to obtain power, use estimate method
I0524 12:40:43.521129       1 exporter.go:176] Initializing the GPU collector
I0524 12:40:43.523464       1 watcher.go:67] Using in cluster k8s config
W0524 12:40:43.596800       1 reflector.go:424] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
E0524 12:40:43.596898       1 reflector.go:140] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: Failed to watch <unspecified>: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
W0524 12:40:44.889366       1 reflector.go:424] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
E0524 12:40:44.889424       1 reflector.go:140] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: Failed to watch <unspecified>: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
W0524 12:40:48.000074       1 reflector.go:424] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
E0524 12:40:48.000117       1 reflector.go:140] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: Failed to watch <unspecified>: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
W0524 12:40:53.340249       1 reflector.go:424] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
E0524 12:40:53.340301       1 reflector.go:140] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: Failed to watch <unspecified>: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
W0524 12:41:02.553610       1 reflector.go:424] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
E0524 12:41:02.553674       1 reflector.go:140] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: Failed to watch <unspecified>: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
W0524 12:41:20.877488       1 reflector.go:424] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
E0524 12:41:20.877544       1 reflector.go:140] github.com/sustainable-computing-io/kepler/pkg/kubernetes/watcher.go:123: Failed to watch <unspecified>: failed to list <unspecified>: pods is forbidden: User "system:serviceaccount:kepler:kepler-sa" cannot list resource "pods" in API group "" at the cluster scope: Azure does not have opinion for this user.
`

`
> kubectl exec -ti -n kepler daemonset/kepler-exporter -- bash  -c "curl localhost:9102/metrics" | grep kepler_container_core_joules_total
command terminated with exit code 7

> kubectl exec -ti -n kepler daemonset/kepler-exporter -- bash  -c "curl localhost:9102/metrics"                                          
curl: (7) Failed to connect to localhost port 9102: Connection refused
command terminated with exit code 7
`

marceloamaral:
@andersonandrei you need to update the Cluster Role, please see the file `manifests/config/rbac/k8s_clusterrole.yaml` in this PR

andersonandrei:
> @andersonandrei you need to update the Cluster Role, please see the file `manifests/config/rbac/k8s_clusterrole.yaml` in this PR

Yes, I was aware of that and updated the Cluster Role before deploying Kepler

marceloamaral:
Thanks @andersonandrei for letting me know that.

So, in AKS the user might not have the correct assignment for the security group to list/watch all pods.

There are some instructions to enable it here: https://learn.microsoft.com/en-us/troubleshoot/azure/azure-kubernetes/user-cannot-get-cluster-resources

Alternatively you can also configure Azure RBAC: https://samcogan.com/get-to-grips-with-azure-ad-and-aks/

Btw,  am wondering if the kubelet access was working before since it requires the same access permission as apiserver...

andersonandrei:
@marceloamaral, thanks for the details.

But before, with the standard version of Kepler, I could use the API and retrieve the metrics. Also, I tried with another recent PR ([pr-665-may-3](http://quay.io/sustainable_computing_io/kepler:pr-665-may-3)), with the same Cluster Role, and it seems to work (I mean, not solving the issue of this current PR, but at least I don't see the errors above here in the logs, and I can retrieve metrics from the API):
```
> kubectl logs kepler-exporter-cxcb5 -n kepler
I0524 12:57:55.859962       1 gpu.go:46] Failed to init nvml, err: could not init nvml: error opening libnvidia-ml.so.1: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
I0524 12:57:55.863483       1 exporter.go:150] Kepler running on version: v0.4.11-136-g7bae827
I0524 12:57:55.863506       1 config.go:172] using gCgroup ID in the BPF program: true
I0524 12:57:55.863521       1 config.go:174] kernel version:
I0524 12:57:55.863537       1 exporter.go:163] EnabledBPFBatchDelete: true
I0524 12:57:55.863572       1 rapl_msr_util.go:129] failed to open path /dev/cpu/0/msr: no such file or directory
I0524 12:57:55.863606       1 power.go:77] Not able to obtain power, use estimate method
I0524 12:57:55.941736       1 exporter.go:176] Initializing the GPU collector
perf_event_open: No such file or directory
I0524 12:57:57.476493       1 bcc_attacher.go:104] failed to attach perf event cpu_cycles_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0524 12:57:57.476536       1 bcc_attacher.go:104] failed to attach perf event cpu_ref_cycles_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0524 12:57:57.476576       1 bcc_attacher.go:104] failed to attach perf event cpu_instr_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0524 12:57:57.476608       1 bcc_attacher.go:104] failed to attach perf event cache_miss_hc_reader: failed to open bpf perf event: no such file or directory
I0524 12:57:57.476617       1 bcc_attacher.go:143] Successfully load eBPF module with option: [-DNUM_CPUS=2]
I0524 12:57:57.492220       1 exporter.go:218] Started Kepler in 1.628770692s
I0524 12:58:00.492568       1 container_hc_collector.go:127] failed to get bpf table elemets, err: failed to batch get: invalid argument
I0524 12:58:00.523919       1 resolve_container.go:143] container ID /system.slice/containerd.service not found in kubelet query. Adding it to system process
```

API:
```
> kubectl exec -ti -n kepler daemonset/kepler-exporter -- bash  -c "curl localhost:9102/metrics" | grep kepler_container_core_joules_total
# HELP kepler_container_core_joules_total Aggregated RAPL value in core in joules
# TYPE kepler_container_core_joules_total counter
kepler_container_core_joules_total{command="",container_id="",container_name="ryax-runner",container_namespace="ryaxns",mode="dynamic",pod_name="ryax-runner-74998594cf-sks6f"} 190.24
kepler_container_core_joules_total{command="",container_id="",container_name="ryax-runner",container_namespace="ryaxns",mode="idle",pod_name="ryax-runner-74998594cf-sks6f"} 0
kepler_container_core_joules_total{command="",container_id="2a2d64945e3d11775569f012c84c659b2b44e12383fec2f9c67e74faec9152c5",container_name="ryax-authorization",container_namespace="ryaxns",mode="dynamic",pod_name="ryax-authorization-fbcbcdbb7-mptcn"} 190.24
kepler_container_core_joules_total{command="",container_id="2a2d64945e3d11775569f012c84c659b2b44e12383fec2f9c67e74faec9152c5",container_name="ryax-authorization",container_namespace="ryaxns",mode="idle",pod_name="ryax-authorization-fbcbcdbb7-mptcn"} 0
kepler_container_core_joules_total{command="",container_id="40dbbaf075005654290208022a523b4dffd344600268bae31d05caf0f5f60733",container_name="azuredisk",container_namespace="kube-system",mode="dynamic",pod_name="csi-azuredisk-node-m5gjk"} 190.24
kepler_container_core_joules_total{command="",container_id="40dbbaf075005654290208022a523b4dffd344600268bae31d05caf0f5f60733",container_name="azuredisk",container_namespace="kube-system",mode="idle",pod_name="csi-azuredisk-node-m5gjk"} 0
kepler_container_core_joules_total{command="",container_id="441b39106a0cf9a0a6c92951fed3243ad607c8e5a7cf5c3b634a2a68d8b5e13f",container_name="metrics-server",container_namespace="kube-system",mode="dynamic",pod_name="metrics-server-db666f966-pskkn"} 190.24
kepler_container_core_joules_total{command="",container_id="441b39106a0cf9a0a6c92951fed3243ad607c8e5a7cf5c3b634a2a68d8b5e13f",container_name="metrics-server",container_namespace="kube-system",mode="idle",pod_name="metrics-server-db666f966-pskkn"} 0
kepler_container_core_joules_total{command="",container_id="4b7ada85205886d24871e5fffe6184640555342efee8418af62e443973c89cf9",container_name="ryax-authorization-database-migration",container_namespace="ryaxns",mode="dynamic",pod_name="ryax-authorization-fbcbcdbb7-mptcn"} 190.24
kepler_container_core_joules_total{command="",container_id="4b7ada85205886d24871e5fffe6184640555342efee8418af62e443973c89cf9",container_name="ryax-authorization-database-migration",container_namespace="ryaxns",mode="idle",pod_name="ryax-authorization-fbcbcdbb7-mptcn"} 0
kepler_container_core_joules_total{command="",container_id="4bf1d50463b68b42a3e5fbd8f9ecc61c7e6a78ca3c52d08639ea1efd2f006139",container_name="cert-manager",container_namespace="cert-manager",mode="dynamic",pod_name="cert-manager-webhook-997b5dd88-g575d"} 190.24
kepler_container_core_joules_total{command="",container_id="4bf1d50463b68b42a3e5fbd8f9ecc61c7e6a78ca3c52d08639ea1efd2f006139",container_name="cert-manager",container_namespace="cert-manager",mode="idle",pod_name="cert-manager-webhook-997b5dd88-g575d"} 0
kepler_container_core_joules_total{command="",container_id="50e42a404108b102f7d0cb8da5eb9b7ec61f06ffa2625d4f429d450f33aa21ed",container_name="node-driver-registrar",container_namespace="kube-system",mode="dynamic",pod_name="csi-azuredisk-node-m5gjk"} 190.24
kepler_container_core_joules_total{command="",container_id="50e42a404108b102f7d0cb8da5eb9b7ec61f06ffa2625d4f429d450f33aa21ed",container_name="node-driver-registrar",container_namespace="kube-system",mode="idle",pod_name="csi-azuredisk-node-m5gjk"} 0
kepler_container_core_joules_total{command="",container_id="513b9fa44e459fc7826c94d2ca948becf2cf14064c3c387dd0959111c4542347",container_name="metrics-server-vpa",container_namespace="kube-system",mode="dynamic",pod_name="metrics-server-db666f966-pskkn"} 190.24
kepler_container_core_joules_total{command="",container_id="513b9fa44e459fc7826c94d2ca948becf2cf14064c3c387dd0959111c4542347",container_name="metrics-server-vpa",container_namespace="kube-system",mode="idle",pod_name="metrics-server-db666f966-pskkn"} 0
kepler_container_core_joules_total{command="",container_id="56c07d9f117132bf93dd191a5418bbd6d3fab59c89045d4c3af337480be1fb5b",container_name="cert-manager",container_namespace="cert-manager",mode="dynamic",pod_name="cert-manager-cainjector-56bc5f744c-fjswt"} 190.24
kepler_container_core_joules_total{command="",container_id="56c07d9f117132bf93dd191a5418bbd6d3fab59c89045d4c3af337480be1fb5b",container_name="cert-manager",container_namespace="cert-manager",mode="idle",pod_name="cert-manager-cainjector-56bc5f744c-fjswt"} 0
kepler_container_core_joules_total{command="",container_id="5a979e836e69dc2e36812205b165c7f46497bb9a983bb6ff432d4b98865c2e24",container_name="konnectivity-agent",container_namespace="kube-system",mode="dynamic",pod_name="konnectivity-agent-6fcc478f7d-slj4s"} 190.24
kepler_container_core_joules_total{command="",container_id="5a979e836e69dc2e36812205b165c7f46497bb9a983bb6ff432d4b98865c2e24",container_name="konnectivity-agent",container_namespace="kube-system",mode="idle",pod_name="konnectivity-agent-6fcc478f7d-slj4s"} 0
kepler_container_core_joules_total{command="",container_id="5cc2d821c686b05b72d791de18f3dd5f96efafe7c50f4ddfab00118c2723c0c9",container_name="node-driver-registrar",container_namespace="kube-system",mode="dynamic",pod_name="csi-azurefile-node-hp9d9"} 190.24
```

andersonandrei:
Hello,

I'm sorry, but for some reason the Cluster-Role modifications were not applied before. 
Now I'm sure they were made. With that, the errors in the logs of Kepler are gone, but I still have the same initial problem of not seeing new pods through Kepler's API.

Cluster-role:
```
> kubectl describe clusterrole kepler-clusterrole 
Name:         kepler-clusterrole
Labels:       sustainable-computing.io/app=kepler
Annotations:  <none>
PolicyRule:
  Resources      Non-Resource URLs  Resource Names  Verbs
  ---------      -----------------  --------------  -----
  nodes/metrics  []                 []              [get watch list]
  nodes/proxy    []                 []              [get watch list]
  nodes/stats    []                 []              [get watch list]
  pods           []                 []              [get watch list]
```




Kepler logs:
```
> kubectl logs kepler-exporter-87bqt -n kepler
I0525 09:20:49.694747       1 gpu.go:46] Failed to init nvml, err: could not init nvml: error opening libnvidia-ml.so.1: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
I0525 09:20:49.705131       1 exporter.go:149] Kepler running on version: v0.2-1109-g45d101f-dirty
I0525 09:20:49.705159       1 config.go:175] using gCgroup ID in the BPF program: true
I0525 09:20:49.705381       1 config.go:177] kernel version: 5.4
I0525 09:20:49.705682       1 exporter.go:163] EnabledBPFBatchDelete: true
I0525 09:20:49.705768       1 rapl_msr_util.go:129] failed to open path /dev/cpu/0/msr: no such file or directory
I0525 09:20:49.705822       1 power.go:77] Not able to obtain power, use estimate method
I0525 09:20:49.776321       1 exporter.go:176] Initializing the GPU collector
I0525 09:20:49.778549       1 watcher.go:67] Using in cluster k8s config
perf_event_open: No such file or directory
I0525 09:20:51.908668       1 bcc_attacher.go:106] failed to attach perf event cpu_cycles_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0525 09:20:51.908727       1 bcc_attacher.go:106] failed to attach perf event cpu_ref_cycles_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0525 09:20:51.908769       1 bcc_attacher.go:106] failed to attach perf event cpu_instr_hc_reader: failed to open bpf perf event: no such file or directory
perf_event_open: No such file or directory
I0525 09:20:51.908804       1 bcc_attacher.go:106] failed to attach perf event cache_miss_hc_reader: failed to open bpf perf event: no such file or directory
I0525 09:20:51.908816       1 bcc_attacher.go:153] Successfully load eBPF module with option: [-DNUM_CPUS=2]
I0525 09:20:51.921106       1 exporter.go:220] Started Kepler in 2.216006568s
I0525 09:20:54.953912       1 container_hc_collector.go:172] resetting EnabledBPFBatchDelete to false
[2:52](https://sustainableco-ese6814.slack.com/archives/D059G6BMA3V/p1685019121386789)
```

Pods running:
```
> kubectl get pods -n openwhisk -w -o wide        
NAME                                     READY   STATUS    RESTARTS   AGE     IP            NODE                            NOMINATED NODE   READINESS GATES
owdev-alarmprovider-556597d875-f4f49     1/1     Running   0          5h14m   10.244.1.20   aks-intra-99364876-vmss000000   <none>           <none>
owdev-apigateway-84b944bbf-xs5db         1/1     Running   0          5h14m   10.244.1.8    aks-intra-99364876-vmss000000   <none>           <none>
owdev-controller-0                       1/1     Running   0          5h13m   10.244.2.4    aks-intra-99364876-vmss000001   <none>           <none>
owdev-couchdb-74f58b45f6-t7l2q           1/1     Running   0          5h14m   10.244.1.21   aks-intra-99364876-vmss000000   <none>           <none>
owdev-invoker-0                          1/1     Running   0          5h10m   10.244.1.17   aks-intra-99364876-vmss000000   <none>           <none>
owdev-kafka-0                            1/1     Running   0          5h13m   10.244.2.13   aks-intra-99364876-vmss000001   <none>           <none>
owdev-kafkaprovider-7ff5f789cf-hskpz     1/1     Running   0          5h14m   10.244.1.12   aks-intra-99364876-vmss000000   <none>           <none>
owdev-nginx-78d7b67fc6-9srqr             1/1     Running   0          5h14m   10.244.1.14   aks-intra-99364876-vmss000000   <none>           <none>
owdev-redis-54bdcbfd47-9jdjs             1/1     Running   0          5h14m   10.244.1.18   aks-intra-99364876-vmss000000   <none>           <none>
owdev-zookeeper-0                        1/1     Running   0          5h13m   10.244.2.15   aks-intra-99364876-vmss000001   <none>           <none>
wskowdev-invoker-00-1-prewarm-nodejs10   1/1     Running   0          5h5m    10.244.3.3    aks-intra-99364876-vmss000004   <none>           <none>
wskowdev-invoker-00-2-prewarm-nodejs10   1/1     Running   0          5h5m    10.244.3.4    aks-intra-99364876-vmss000004   <none>           <none>
wskowdev-invoker-00-6-guest-linpack      1/1     Running   0          15s     10.244.3.8    aks-intra-99364876-vmss000004   <none>           <none>
```

Kepler queries:
```
[root@kepler-exporter-87bqt /]# curl localhost:9102/metrics | grep kepler_container_core_joules_total | grep wskowdev
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0kepler_container_core_joules_total{command="",container_id="4ea1e725f562d65256edfc45df90ce531fc39b3ce1a579ecdb3617a814eee817",container_name="user-action",container_namespace="openwhisk",mode="dynamic",pod_name="wskowdev-invoker-00-1-prewarm-nodejs10"} 40178.688
kepler_container_core_joules_total{command="",container_id="4ea1e725f562d65256edfc45df90ce531fc39b3ce1a579ecdb3617a814eee817",container_name="user-action",container_namespace="openwhisk",mode="idle",pod_name="wskowdev-invoker-00-1-prewarm-nodejs10"} 0
kepler_container_core_joules_total{command="",container_id="a93d627d8b98446a0501bf74cbf0f86aafd31db4be4ba1639b224b69c61cb317",container_name="user-action",container_namespace="openwhisk",mode="dynamic",pod_name="wskowdev-invoker-00-2-prewarm-nodejs10"} 40178.688
kepler_container_core_joules_total{command="",container_id="a93d627d8b98446a0501bf74cbf0f86aafd31db4be4ba1639b224b69c61cb317",container_name="user-action",container_namespace="openwhisk",mode="idle",pod_name="wskowdev-invoker-00-2-prewarm-nodejs10"} 0
100  122k    0  122k    0     0  24.0M      0 --:--:-- --:--:-- --:--:-- 24.0M

[root@kepler-exporter-d4lv8 /]# curl localhost:9102/metrics | grep kepler_container_core_joules_total | grep wskowdev
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  191k    0  191k    0     0  14.4M      0 --:--:-- --:--:-- --:--:-- 14.4M
(edited)

[root@kepler-exporter-dndr4 /]# curl localhost:9102/metrics | grep kepler_container_core_joules_total | grep wskowdev
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  278k    0  278k    0     0  27.2M      0 --:--:-- --:--:-- --:--:-- 27.2M
```

jichenjc:
>>>not seeing new pods through Kepler's API.

so you still seeing similar error #703 ?

marceloamaral:
The problem might be related to the PID/CgroupID to ContainerID resolution. Some systems have different cgroup path and then Kepler cannot extract the container ID, as in PR #684.
I am talking to Anderson to try to debug it.

rootfs:
@marceloamaral Maybe the openwhisk pre-warm pod could have a name change into the new pods. If the pre-warm pods are cached and api server watcher doesn't catch the pod change event, the newly named pod may not be reported.

cmd/exporter.go
```diff
@@ -151,6 +152,7 @@ func main() {
 	config.SetEnabledHardwareCounterMetrics(*exposeHardwareCounterMetrics)
 	config.SetEnabledGPU(*enableGPU)
 	config.EnabledMSR = *enabledMSR
+	config.SetKubeConfig(*kubeconfig)
```
rootfs, line 155: please add in-cluster config support too

pkg/collector/utils.go
```diff
@@ -20,34 +20,44 @@ import (
...
+		// In case the pod watcher is not enabled, we need to retrieve the information about the
+		// pod and container from the kubelet API. However, we prefer to use the watcher approach
+		// as accessing the kubelet API might be restricted in certain systems.
+		// Additionally, the code that fetches the information from the kubelet API utilizes cache
+		// for performance reasons. Therefore, if the kubelet API delay the information of the
+		// containerID (which occasionally occurs), the container will be wrongly identified for its entire lifetime.
+		if !kubernetes.IsWatcherEnabled {
+			podName, _ = cgroup.GetPodName(cGroupID, pid, withCGroupID)
+			containerName, _ = cgroup.GetContainerName(cGroupID, pid, withCGroupID)
+			if containerName == c.systemProcessName {
```
rootfs, line 44: i suspect there still could be a chance of split brain. If kubelet cannot detect pod name at this time, apiserver won't be able to do it either.

cmd/exporter.go
marceloamaral: if kubeconfig is empty we use the in-cluster configuration

pkg/collector/utils.go
marceloamaral: In scenarios where it is necessary, please note that you can disable the apiserver by using the flag `-apiserver=false`.

rootfs (approved):

sunya-ch:
> @sunya-ch any request needed for this PR?

Sorry for delays.
I think this PR has already satisfied for the purpose of introducing the API server for pod event detection.
It looks good to me.

marceloamaral:
@rootfs ^^

sunya-ch (approved):

PR #626, opened 2023-04-13
Update bcc map key deletion method

marceloamaral:
This PR fix the issue #624

PR #641, opened 2023-04-24
Add kernel headers to run bcc unit test

marceloamaral:
The unit test failed with a bcc error due to the inability to locate the kernel headers. Despite the kernel libs folder being mounted in the docker container, it appears that the headers were not installed during VM setup.

To address this issue, I included a new step in the github workflow that installs the kernel headers prior to running the test.

This might fix the unit test error in the PR #626

### DIFF
--- a/.github/workflows/unit_test_bcc.yml
+++ b/.github/workflows/unit_test_bcc.yml
@@ -8,5 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - name: install linux headers
+      run: sudo apt-get install linux-headers-$(uname -r)
     - name: test bcc in docker container
       run: sudo podman run --privileged -v $(pwd):/home -v /sys:/sys -v /lib/modules:/lib/modules -v /usr/src:/usr/src quay.io/sustainable_computing_io/kepler_builder:go1.18 bash -c "cd /home; export PATH=/usr/local/go/bin:$PATH; make test-verbose"


.github/workflows/unit_test_bcc.yml
```diff
@@ -5,8 +5,10 @@ on:
 
 jobs:
   unit_test_with_bcc:
-    runs-on: ubuntu-latest
+    runs-on: c
```
SamYuan1990, line 8: I suppose github action don't provide this type of agent, hence, it should be ubuntu-latest

.github/workflows/unit_test_bcc.yml
```diff
@@ -5,8 +5,10 @@ on:
...
     steps:
     - uses: actions/checkout@v3
+    - name: install bcc
```
SamYuan1990, line 11: if you want to install bcc,  you can ref https://github.com/sustainable-computing-io/kepler/blob/37996282db229b8ce7c0433a856721b2c1a56400/.github/workflows/unit_test.yml#L29-L33
as apt repo's bcc is out of date.
or you just want header?

SamYuan1990 (changes requested):

SamYuan1990:
em... I am thinking, should we remove duplicate as unit_test with bcc there?
for unit test we have bcc installed already. @rootfs hence for today's CI, it seems we don't need container level testing?

.github/workflows/unit_test_bcc.yml
marceloamaral: right, we probably don't need the bcc.

@SamYuan1990 I updated the PR, can you review again?

SamYuan1990: I am not sure for either #642 or this one works.
if #642 works, which means we can remove one test pipeline as duplication. @rootfs

.github/workflows/unit_test_bcc.yml
```diff
@@ -8,5 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - name: install linux headers
```
SamYuan1990, line 11: @marceloamaral  by this test case, which means the unit test will run without bcc build tag for golang.

marceloamaral:
Closing this since @SamYuan1990 create another PR #642

PR #642, opened 2023-04-24
try to provide bcc and headers in CI for unit test on linux

SamYuan1990:
try to provide bcc and headers in CI for unit test on linux


SamYuan1990:
@marceloamaral , here you are.

PR #666, opened 2023-05-02
Fix container GPU power

marceloamaral:
There exist clusters that lacks CPU power metrics, but exposes GPU power information instead (we are doing experiments in a cluster like that). When CPU power metrics are not available, the Kepler power typically relies on the estimator approach. However, in cases where GPU power metrics are present, it is preferable to use the estimator approach for CPU metrics and the ratio approach for GPU metrics.

This PR allows for the use of two different power models, addressing the aforementioned issue. Additionally, the pull request fixes a bug related to updating GPU metrics for deleted containers where a null pointer error was encountered.

### DIFF
--- a/pkg/collector/container_accelerator_collector.go
+++ b/pkg/collector/container_accelerator_collector.go
@@ -38,6 +38,7 @@ func (c *Collector) updateAcceleratorMetrics() {
 	// calculate the gpu's processes energy consumption for each gpu
 	for _, device := range accelerator.GetGpus() {
 		if processesUtilization, err = accelerator.GetProcessResourceUtilizationPerDevice(device, time.Since(lastUtilizationTimestamp)); err != nil {
+			klog.Infoln(err)
 			continue
 		}
 
@@ -53,7 +54,14 @@ func (c *Collector) updateAcceleratorMetrics() {
 				c.createProcessMetricsIfNotExist(uint64(pid), "")
 			}
 
-			// update container metrics
+			// When a container or pod is deleted, the associated process data may still exist for some time. However, the createContainersMetricsIfNotExist function
+			// will not create an entry in the c.ContainersMetrics map if the container or pod no longer exists. Therefore, even if the function attempted to initialize
+			// the map with the deleted containerID, it may not have been added to the map due to the absence of container information. As a result, these processes
+			// may be classified as system processes rather than being attributed to the deleted container or pod.
+			if _, exist := c.ContainersMetrics[containerID]; !exist {
+				continue
+			}
+			klog.Infoln("DEBUG: pod", c.ContainersMetrics[containerID].PodName, "processUtilizationSample", processUtilization)
 			if err = c.ContainersMetrics[containerID].CounterStats[config.GPUSMUtilization].AddNewDelta(uint64(processUtilization.SmUtil)); err != nil {
 				klog.V(5).Infoln(err)
 			}
--- a/pkg/collector/metric/container_metric.go
+++ b/pkg/collector/metric/container_metric.go
@@ -66,21 +66,23 @@ func NewContainerMetrics(containerName, podName, podNamespace, containerID strin
 		Namespace:     podNamespace,
 		ContainerID:   containerID,
 		ProcessMetrics: ProcessMetrics{
-			CPUTime:            &types.UInt64Stat{},
-			CounterStats:       make(map[string]*types.UInt64Stat),
-			SoftIRQCount:       make([]types.UInt64Stat, config.MaxIRQ),
-			DynEnergyInCore:    &types.UInt64Stat{},
-			DynEnergyInDRAM:    &types.UInt64Stat{},
-			DynEnergyInUncore:  &types.UInt64Stat{},
-			DynEnergyInPkg:     &types.UInt64Stat{},
-			DynEnergyInOther:   &types.UInt64Stat{},
-			DynEnergyInGPU:     &types.UInt64Stat{},
-			IdleEnergyInCore:   &types.UInt64Stat{},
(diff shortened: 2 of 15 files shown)


marceloamaral:
Because of golint error I removed a method that is not being used anywhere....the `GetBasicValues()`...

pkg/collector/container_accelerator_collector.go
```diff
@@ -53,7 +54,14 @@ func (c *Collector) updateAcceleratorMetrics() {
...
 
-			// update container metrics
+			// When a container or pod is deleted, the associated process data may still exist for some time. However, the createContainersMetricsIfNotExist function
+			// will not create an entry in the c.ContainersMetrics map if the container or pod no longer exists. Therefore, even if the function attempted to initialize
+			// the map with the deleted containerID, it may not have been added to the map due to the absence of container information. As a result, these processes
+			// may be classified as system processes rather than being attributed to the deleted container or pod.
+			if _, exist := c.ContainersMetrics[containerID]; !exist {
+				continue
+			}
+			klog.Infoln("DEBUG: pod", c.ContainersMetrics[containerID].PodName, "processUtilizationSample", processUtilization)
```
rootfs, line 64: can this be moved to V(5)?

rootfs:
@marceloamaral looks good! Please fix the logging and feel free to merge it.

jichenjc (approved):

pkg/collector/container_accelerator_collector.go
SamYuan1990: why Infoln with debug msg?

pkg/collector/metric/container_metric.go
```diff
@@ -277,3 +272,47 @@ func (c *ContainerMetrics) UpdateCgroupMetrics() error {
 	}
 	return err
 }
+
+func (c *ContainerMetrics) GetDynEnergyStat(component string) (energyStat *types.UInt64Stat) {
```
SamYuan1990, line 276: component string > const?

pkg/collector/metric/container_metric.go
```diff
@@ -277,3 +272,47 @@ func (c *ContainerMetrics) UpdateCgroupMetrics() error {
...
+		return c.DynEnergyInDRAM
+	case UNCORE:
+		return c.DynEnergyInUncore
+	case GPU:
+		return c.DynEnergyInGPU
+	case OTHER:
+		return c.DynEnergyInOther
+	case PLATFORM:
+		return c.DynEnergyInPlatform
+	default:
```
SamYuan1990, line 292: is this line been used?

pkg/collector/prometheus_collector_test.go
```diff
@@ -127,7 +130,10 @@ var _ = Describe("Test Prometheus Collector Unit", func() {
 		exporter.NodeMetrics.SetNodeComponentsEnergy(componentsEnergies)
 		exporter.NodeMetrics.UpdateIdleEnergy()
 		exporter.NodeMetrics.UpdateDynEnergy()
-		model.UpdateContainerEnergyByRatioPowerModel(*exporter.ContainersMetrics, exporter.NodeMetrics)
+		var wg sync.WaitGroup
+		wg.Add(1)
```
SamYuan1990, line 134: add 1 and wait? what's benefit as we move to go routine?

pkg/model/container_power.go
```diff
@@ -63,13 +66,30 @@ func InitContainerPowerEstimator(usageMetrics, systemFeatures, systemValues []st
 
 // UpdateContainerEnergy returns container energy consumption for each node component
 func UpdateContainerEnergy(containersMetrics map[string]*collector_metric.ContainerMetrics, nodeMetrics *collector_metric.NodeMetrics) {
+	var wg sync.WaitGroup
 	// If the node can expose power measurement per component, we can use the RATIO power model
 	// Otherwise, we estimate it from trained power model
 	if components.IsSystemCollectionSupported() {
-		local.UpdateContainerEnergyByRatioPowerModel(containersMetrics, nodeMetrics)
+		wg.Add(6)
+		go local.UpdateContainerComponentEnergyByRatioPowerModel(containersMetrics, nodeMetrics, collector_metric.PKG, config.CoreUsageMetric, &wg)
```
SamYuan1990, line 74: @rootfs I am not sure if we use go routine for GPU, why not for others?

pkg/model/estimator/local/benchmark_test.go
```diff
@@ -76,22 +77,25 @@ func benchmarkNtesting(b *testing.B, continerNumber int) {
 	}
 	nodeMetrics.AddNodeResUsageFromContainerResUsage(containersMetrics)
 	b.ResetTimer()
-	local.UpdateContainerEnergyByRatioPowerModel(containersMetrics, nodeMetrics)
+	var wg sync.WaitGroup
```
SamYuan1990, line 80: no need wait group in test case?

SamYuan1990 (commented):
do we need go routine?

pkg/collector/container_accelerator_collector.go
marceloamaral: Sorry this was for my own debugging...and should be removed

pkg/collector/prometheus_collector_test.go
marceloamaral: Because in the test we are only updating the energy of only one source.
But the code updates more.

pkg/model/estimator/local/benchmark_test.go
marceloamaral: Just to follow the same logic of the main code.

marceloamaral:
@SamYuan1990 the gorotines are just to speedup and improve scalability of the code since it will calculate for each container.

PR #674, opened 2023-05-08
bpf: read cpu cores from /proc/cpuinfo to get all the cores

rootfs:
Tested on setup with 64 cores, 32 are reserved by applications, kepler can only see 28 of them. This fix is to allow kepler to detect all the cores.

### DIFF
--- a/pkg/bpfassets/attacher/bcc_attacher.go
+++ b/pkg/bpfassets/attacher/bcc_attacher.go
@@ -31,6 +31,8 @@ import (
 
 	bpf "github.com/iovisor/gobpf/bcc"
 
+	"github.com/jaypipes/ghw"
+
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


marceloamaral:
what about using the lib `"github.com/jaypipes/ghw"`?
We use it to discover the cpu info in the package power `pkg/power/components/source/rapl_msr_util.go`

rootfs:
@marceloamaral good point! ghw indeed uses /proc/cpuinfo. I'll rewrite this PR.

pkg/bpfassets/attacher/bcc_attacher.go
```diff
@@ -118,8 +120,16 @@ func AttachBPFAssets() (*BpfModuleTables, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to get program %q: %v", program, err)
 	}
+	// builtin runtime.NumCPU() returns the number of logical CPUs usable by the current process
+	cores := runtime.NumCPU()
+	if cpu, err := ghw.CPU(); err == nil {
+		// we need to get the number of all CPUs,
+		// so if /proc/cpuinfo is available, we can get the number of all CPUs
```
jichenjc, line 127: can we add a log to indicate such difference?

marceloamaral (approved):
/lgtm

PR #665, opened 2023-05-01
[WIP] ebpf related fixes for rt kernels

rootfs:
fix #657

(this PR is on top of #656)

The batch lookup and delete API is currently implemented in a bcc gobuf fork, will be ported after test

PR #678, opened 2023-05-09
bpf: add batch table get

rootfs:
this is ported from #665

### DIFF
--- a/pkg/collector/container_hc_collector.go
+++ b/pkg/collector/container_hc_collector.go
@@ -129,6 +129,7 @@ func (c *Collector) updateBPFMetrics() {
 		if batchGetErr != nil {
 			klog.V(1).Infof("failed to get bpf table elemets, err: %v", batchGetErr)
 			ebpfBatchGet = false
+			// if batch get is not supported we disable it for the next time
 			ebpfBatchGetAndDelete = false
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/collector/container_hc_collector.go
```diff
@@ -109,23 +116,52 @@ func (c *Collector) updateBPFMetrics() {
...
 
-		// if ebpf map batch deletion operation is supported we add the key to the list otherwise we delete the key
-		if config.EnabledBPFBatchDelete {
-			keysToDelete = append(keysToDelete, key)
-		} else {
-			err = c.bpfHCMeter.Table.Delete(key) // deleting the element to reset the counter values
-			if err != nil {
-				klog.Infof("could not delete bpf table elemets, err: %v", err)
+		// if not deleted during get, delete it now
+		if !ebpfBatchGetAndDelete {
```
jichenjc, line 156: if this is false, when do we do the delete operation? if will not append to the delete list or delete directly ...

marceloamaral: @rootfs should this line be `if ebpfBatchGetAndDelete {`

The line 203 is the `!ebpfBatchGetAndDelete` case, right?

marceloamaral (commented):
Looks good to me. Only have one comment.

pkg/collector/container_hc_collector.go
rootfs: this line is right, if `ebpfBatchGetAndDelete == true`, all the keys are already deleted in the batch get

rootfs:
@jichenjc @marceloamaral updated, PTAL thanks

jichenjc:
/lgtm

I assume the original result posted in the origianl issue still valid after the above fix?

Issue #679, opened 2023-05-11
Segmentation violation when ENABLE_PROCESS_METRICS is True and EXPOSE_HW_COUNTER_METRICS is False


alysondeives:
**Describe the bug**

The following SEGV error occurs when ENABLE_PROCESS_METRICS is set to True and EXPOSE_HW_COUNTER_METRICS is set to False:

```
2023-05-11T14:50:07.247197018Z stderr F I0511 14:50:07.247156       1 node_energy_collector.go:57] System energy collection is supported
2023-05-11T14:50:07.290440491Z stderr F panic: runtime error: invalid memory address or nil pointer dereference
2023-05-11T14:50:07.29046993Z stderr F [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xc71f39]
2023-05-11T14:50:07.290472916Z stderr F 
2023-05-11T14:50:07.290475758Z stderr F goroutine 91 [running]:
2023-05-11T14:50:07.290478791Z stderr F github.com/sustainable-computing-io/kepler/pkg/model/estimator/local.UpdateProcessComponentEnergyByRatioPowerModel(0xc00039fb00, 0xc0005ea000, {0xe78e33, 0x8}, {0xc00003a360, 0x9}, 0x756c6f762f736e69?)
2023-05-11T14:50:07.290546201Z stderr F         /opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/model/estimator/local/ratio_process.go:45 +0x1b9
2023-05-11T14:50:07.290550838Z stderr F created by github.com/sustainable-computing-io/kepler/pkg/model.UpdateProcessEnergy
2023-05-11T14:50:07.29055398Z stderr F  /opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/model/process_power.go:78 +0x27a
```

jichenjc:
I didn't see this in latest image ...
```
# kubectl logs ds/kepler-exporter -n kepler  | head
....
I0512 06:21:56.175756       1 config.go:115] ENABLE_PROCESS_METRICS: true
I0512 06:21:56.175759       1 config.go:116] EXPOSE_HW_COUNTER_METRICS: false
```

jichenjc:
Reviewed the code location and I guess it's related to HW counter not enabled when access the array variable..
submit a PR for this

PR #682, opened 2023-05-12
fix: avoid use metrics if HW counter not open

jichenjc:
Fixes: #679

### DIFF
--- a/pkg/model/estimator/local/ratio_process.go
+++ b/pkg/model/estimator/local/ratio_process.go
@@ -27,7 +27,48 @@ import (
 	"k8s.io/klog/v2"
 
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+	"github.com/sustainable-computing-io/kepler/pkg/collector/metric/types"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
 )
+
+// getDataFromAlternativeMetric is used when there is no Hardware counter so we have to use cgroup or eBPF (CPUTime)
+func getDataFromAlternativeMetric(cputime *types.UInt64Stat, usageMetrics string) float64 {
+	// Given there is no HW counter, we have to use cgroup data, and Only CPUTime is available today
+	// So busy CPU is more likely to be accessing memory. Although CPU utilization (CPUTime) does not
+	// directly represent memory access, it remains the only viable proxy available to approximate such information reliably.
+	switch usageMetrics {
+	case config.CoreUsageMetric:
+	case config.DRAMUsageMetric:
+		return float64(cputime.Delta)
+	case config.GpuUsageMetric:
+		// FIXME: find cgroup data in cgroup
+		return 0
+	default:
+		return 0
+	}
+	return 0
+}
+
+func getNodeTotalProcessResourceUsage(containerMetrics *collector_metric.ContainerMetrics, usageMetric string) float64 {
+	var nodeTotalResourceUsage float64
+	if _, ok := containerMetrics.CounterStats[usageMetric]; ok {
+		nodeTotalResourceUsage = float64(containerMetrics.CounterStats[usageMetric].Delta)
+	} else {
+		// this will be system container (all processes's CPU Time)
+		nodeTotalResourceUsage = getDataFromAlternativeMetric(containerMetrics.CPUTime, usageMetric)
+	}
+	return nodeTotalResourceUsage
+}
+
+func getProcessResUsage(process *collector_metric.ProcessMetrics, usageMetric string) float64 {
+	var processResUsage float64
+	if _, ok := process.CounterStats[usageMetric]; ok {
+		processResUsage = float64(process.CounterStats[usageMetric].Delta)
+	} else {
+		processResUsage = getDataFromAlternativeMetric(process.CPUTime, usageMetric)
+	}
+	return processResUsage
+}
 
 // TODO: we should not calculate the process power based on the container power. Instead of using the container metrics we should use the system metrics and we do in the container power model.
 // UpdateProcessComponentEnergyByRatioPowerModel calculates the process energy consumption based on the energy consumption of the container that contains all the processes
@@ -42,19 +83,17 @@ func UpdateProcessComponentEnergyByRatioPowerModel(processMetrics map[uint64]*co
 
 	// if usageMetric exist, divide the power using the ratio. Otherwise, evenly divide the power.
 	if usageMetric != "" {
-		nodeTotalResourceUsage = float64(containerMetrics.CounterStats[usageMetric].Delta)
+		nodeTotalResourceUsage = getNodeTotalProcessResourceUsage(containerMetrics, usageMetric)
 	}
 
 	for pid, process := range processMetrics {
-		if _, ok := process.CounterStats[usageMetric]; ok {
-			processResUsage := float64(process.CounterStats[usageMetric].Delta)
-			processPkgEnergy := getEnergyRatio(processResUsage, nodeTotalResourceUsage, totalDynPower, processesNumber)
-			if err := processMetrics[pid].GetDynEnergyStat(component).AddNewDelta(processPkgEnergy); err != nil {
-				klog.Infoln(err)
-			}
-			if err := processMetrics[pid].GetIdleEnergyStat(component).AddNewDelta(idlePowerPerProcess); err != nil {
-				klog.Infoln(err)
-			}
+		processResUsage := getProcessResUsage(process, usageMetric)
+		processPkgEnergy := getEnergyRatio(processResUsage, nodeTotalResourceUsage, totalDynPower, processesNumber)
+		if err := processMetrics[pid].GetDynEnergyStat(component).AddNewDelta(processPkgEnergy); err != nil {
+			klog.Infoln(err)
+		}
+		if err := processMetrics[pid].GetIdleEnergyStat(component).AddNewDelta(idlePowerPerProcess); err != nil {
+			klog.Infoln(err)
 		}
 	}
 }
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -311,12 +311,12 @@ func (ne *NodeMetrics) SetNodeOtherComponentsEnergy() {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


pkg/model/estimator/local/ratio_process.go
```diff
@@ -42,7 +42,10 @@ func UpdateProcessComponentEnergyByRatioPowerModel(processMetrics map[uint64]*co
 
 	// if usageMetric exist, divide the power using the ratio. Otherwise, evenly divide the power.
 	if usageMetric != "" {
```
marceloamaral, line 44: Is cgroup metrics enabled?

I think if the hardware counters are disabled, we should use cpu time from cgroup metric instead.

jichenjc: ok, add a check if cgroup is enabled, use it , I Think we have cgroup cpu time some where already
but
- how to calcualte the total usage based on cpu time, refer to estimate method?
- if no cgroup, no HW counter, we use -1 to avoid nil?

marceloamaral: I think that we should update the usage variables when HCs are disables during the Kepler initialization.

pkg/config/config.go
```
CoreUsageMetric       = getConfig("CORE_USAGE_METRIC", CPUInstruction)
DRAMUsageMetric       = getConfig("DRAM_USAGE_METRIC", CacheMiss)
UncoreUsageMetric     = getConfig("UNCORE_USAGE_METRIC", defaultMetricValue) // no metric (evenly divided)
GeneralUsageMetric    = getConfig("GENERAL_USAGE_METRIC", CPUInstruction)    // for uncategorized energy; pkg - core - uncore
```

The verification should be triggered in cmd/exporter.go

However, the cgroup metrics are not in the `CounterStats`, so we will still need to verify if the HW is not available and use the `CgroupStatMap` for containers and `SoftIRQCount` for processes.

jichenjc: ok, let me update the whole logic

jichenjc: ok, something like following ? 

```
if process

if HW COUNTER = true:
      systemcontainerMetrics.CounterStats[usageMetric]   ==> usageMetric = HW related, e.g cpu_cycles,  cpu_instr
else if CGROPU = true
      systemcontainerMetrics.Statmap[usageMetric] ==> usageMetric = cgroup related, e.g SOFTIRQ COUNT
else:
      report error
```

jichenjc: if container:

if HW COUNTER = true:
      containerMetrics.CounterStats[usageMetric]   ==> usageMetric = HW related, e.g cpu_cycles,  cpu_instr
else if CGROPU = true:
      containerMetrics.CgroupStatMap[config.CgroupfsCPU] ==> use config.CgroupfsCPU
else 
      do nothing and log

jichenjc: @marceloamaral why `CPUTime` not used for process ? why use SOFTIRQCOUNT ?

marceloamaral: > why CPUTime not used for process ? why use SOFTIRQCOUNT ?

Oh right! Let's use `CPUTime`. It will be a better metric.

jichenjc:
@marceloamaral please let me know whether the proposed way is a good direction... thanks

pkg/model/estimator/local/ratio_process.go
```diff
@@ -29,6 +29,27 @@ import (
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
 )
 
+func getNodeTotalResourceUsage(containerMetrics *collector_metric.ContainerMetrics, usageMetric string) float64 {
+	var nodeTotalResourceUsage float64
+	if _, ok := containerMetrics.CounterStats[usageMetric]; ok {
+		nodeTotalResourceUsage = float64(containerMetrics.CounterStats[usageMetric].Delta)
+	} else {
+		// this will be system container (all processes's CPU Time
```
marceloamaral, line 37: Should the parenthesis be closed?

marceloamaral (changes requested):
Looks good.

Can you also update with similar logic the `pkg/model/estimator/local/ratio.go`?

pkg/model/estimator/local/ratio_process.go
```diff
@@ -29,6 +29,27 @@ import (
...
+	}
+	return nodeTotalResourceUsage
+}
+
+func getProcessResUsage(process *collector_metric.ProcessMetrics, usageMetric string) float64 {
+	var processResUsage float64
+	if _, ok := process.CounterStats[usageMetric]; ok {
+		processResUsage = float64(process.CounterStats[usageMetric].Delta)
+	} else {
+		processResUsage = float64(process.CPUTime.Delta)
```
jichenjc, line 48: figure out some issue,... config.DRAMUsageMetric can't be used here as we need use different value than CPUTime..

jichenjc:
@marceloamaral I updated the code based on some findings.. can you help check whether the new code is reasonable？

will add lr.go change after confirm this is the right direction.. thanks

pkg/model/estimator/local/ratio_process.go
marceloamaral: Cgroup metrics do not include specific measurements related to memory access. While CPU utilization (i.e. CPUTime) is not directly indicative of memory access, it is currently the only available proxy for such information that we can rely on.

Can you please include a comment about it?

pkg/model/estimator/local/ratio_process.go
```diff
@@ -27,8 +27,51 @@ import (
...
+// getTransferredDataFromMetric is used when there is Hardware counter so we have to use cgroup or eBPF (CPUTime)
+// directly, currently for CoreUsageMetric we use cputime and for DRAM, use cgroup dram info
+func getTransferredDataFromMetric(containerMetrics *collector_metric.ContainerMetrics, usageMetrics string) float64 {
+	switch usageMetrics {
+	case config.CoreUsageMetric:
+		// the containerMetrics will be system container which contains all process metrics,
+		// now, we use the containerMetrics.CPUTime.Delta so we use the all CPUTime in this period
+		// as total CPUTime so we can divide by each process CPUTime
+		return float64(containerMetrics.CPUTime.Delta)
+	case config.DRAMUsageMetric:
```
jichenjc, line 42: @marceloamaral I switch to this way, not sure this is right direction?

pkg/model/estimator/local/ratio_process.go
```diff
@@ -27,8 +27,51 @@ import (
 	"k8s.io/klog/v2"
 
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
 )
 
+// getTransferredDataFromMetric is used when there is Hardware counter so we have to use cgroup or eBPF (CPUTime)
+// directly, currently for CoreUsageMetric we use cputime and for DRAM, use cgroup dram info
+func getTransferredDataFromMetric(containerMetrics *collector_metric.ContainerMetrics, usageMetrics string) float64 {
```
marceloamaral, line 35: Why the name `TransferredData`?

pkg/model/estimator/local/ratio_process.go
```diff
@@ -27,8 +27,51 @@ import (
...
+	switch usageMetrics {
+	case config.CoreUsageMetric:
+		// the containerMetrics will be system container which contains all process metrics,
+		// now, we use the containerMetrics.CPUTime.Delta so we use the all CPUTime in this period
+		// as total CPUTime so we can divide by each process CPUTime
+		return float64(containerMetrics.CPUTime.Delta)
+	case config.DRAMUsageMetric:
+		// same to CPUTime, use Cgroup's memory value and sum all Delta value as summary thus
+		// we can divide from each process memory as well
+		return float64(containerMetrics.CgroupStatMap[config.CgroupfsMemory].SumAllDeltaValues())
```
marceloamaral, line 45: return float64(containerMetrics.CPUTime.Delta), right?

jichenjc: no, for DRAM, I use the cgroup sum all metrics, not use CPUTime ..
I can certainly go back to CPUTime but want to know use CPUTime to represent DRAM is a good idea

marceloamaral: CPU time has better correlation to memory access than memory allocation (config.CgroupfsMemory)...

pkg/model/estimator/local/ratio_process.go
jichenjc: I can make it another name, like getCgroupDataFromMetrics

pkg/model/estimator/local/ratio_process.go
marceloamaral: Memory allocation alone does not provide insights into memory access. However, a busy CPU is more likely to be accessing memory. Although CPU utilization (CPUTime) does not directly represent memory access, it remains the only viable proxy available to approximate such information reliably.

jichenjc: ok, will switch back to original implementation, thanks for the insight~

pkg/model/estimator/local/ratio_process.go
marceloamaral: what about: `getDataFromAlternativeMetric()`

pkg/model/estimator/local/ratio_process.go
```diff
@@ -27,8 +27,51 @@ import (
 	"k8s.io/klog/v2"
 
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
 )
 
+// getTransferredDataFromMetric is used when there is Hardware counter so we have to use cgroup or eBPF (CPUTime)
```
marceloamaral, line 33: `getTransferredDataFromMetric is used when there is Hardware counter so we have to use cgroup or eBPF (CPUTime)`
->
`getTransferredDataFromMetric is used when there is no Hardware counter so we have to use cgroup or eBPF (CPUTime)`

PR #686, opened 2023-05-15
Ignore cgroup error related to minikube cluster

marceloamaral:
As described in this GitHub pull PR comment (https://github.com/sustainable-computing-io/kepler/pull/635#pullrequestreview-1425906295), Minikube clusters may have varying cgroup paths, leading to errors when attempting to read cgroup stats.

Rather than assuming that the cgroup error signifies the container's deletion and deleting containerID from the map, we should handle such errors by ignoring them.

### DIFF
--- a/pkg/collector/container_cgroup_collector.go
+++ b/pkg/collector/container_cgroup_collector.go
@@ -17,6 +17,8 @@ limitations under the License.
 package collector
 
 import (
+	"strings"
+
 	"github.com/sustainable-computing-io/kepler/pkg/cgroup"
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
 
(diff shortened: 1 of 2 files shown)


sunya-ch:
What is the rest error we expect from cgroup that should delete the container? 

FYI, I'm working on making the current cgroup working on minikube. Will push the PR later after confirming that it works.
Also, it is not only the problem minikube. Actually, it is the problem of the way we use the cgroup library. Previously, we did see the same cgroup file hierarchy before and we was aware of it with previously implemented stat reader. 
We just need to implement some interface on cgroup library to support that.

marceloamaral:
> What is the rest error we expect from cgroup that should delete the container?

If the stat was initialized and we could read before (that is the cgroup path existed) but we suddenly cannot read it again, the cgroup was most probably deleted because the container was deleted.

marceloamaral:
This PR now uses the way that the containerd library suggest to collect cgroup v1 stats, ignoring missing files.

Before, we were not ignoring missing files and we could not collect metrics in some systems.

pkg/collector/container_cgroup_collector.go
```diff
@@ -44,7 +46,7 @@ func (c *Collector) updateCgroupMetrics() {
 		}
 		err := c.ContainersMetrics[key].UpdateCgroupMetrics()
 		// if the cgroup metrics of a container does not exist, it means that the container was deleted
-		if key != c.systemProcessName && err != nil {
+		if key != c.systemProcessName && strings.Contains(err.Error(), "cgroup deleted") {
```
sunya-ch, line 49: How's about cgroups.ErrCgroupDeleted.Error() == err.Error()?
Again, shall we really just ignore the cgroup error if the cgroup metric is enabled? I think we should at least log the error that we ignore.

sunya-ch: Got it. Could you check why bcc error in the test case? Would it be fixed with rerunning the test.
The rest change looks fine to me.

rootfs:
Failed tests seem to be related to a network issue, not related to this PR

rootfs (approved):

pkg/collector/container_cgroup_collector.go
rootfs:
```go
if key != c.systemProcessName && err != nil && strings.Contains(err.Error(), "cgroup deleted") 
```

otherwise this will panic when `err == nil`

marceloamaral:
I am forcing to rerun the tests.

Issue #687, opened 2023-05-15
cgroup_stats doesn't support some cgroup hierarchy (e.g., minikube on kvm) - cause container removal


sunya-ch:
**Describe the bug**

This PR is a follow-up debugging from the PR https://github.com/sustainable-computing-io/kepler/pull/635.
However, this is not a bug due to the API server but instead on the way we call containerd library for getting cgroup stat.
In our previous implementation, we was aware of the different cgroup hierarchy that cgroup file list non-slice cgrouppath (such as running minikube on kvm with cgroupV1). 

```
$ cat /proc/$(pidof kepler)/cgroup
11:cpuset:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
10:freezer:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
9:devices:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
8:memory:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
7:hugetlb:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
6:blkio:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
5:pids:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
4:cpu,cpuacct:/kubepods/burstable/podb3a17320-c345-46a3-91a3-8ecea61a3752/ad4432c70a254c0c0cc7c70563c41297f653eaf9d00ce3864881124ddf31723d
[...]

PR #688, opened 2023-05-15
Fix issue 687 by filling empty path

sunya-ch:
This PR fixes the issue https://github.com/sustainable-computing-io/kepler/issues/687.
The idea is when there is no empty subsystem to get a unified path, we just use the most common subsystem that is pids to get the path. 

### DIFF
--- a/pkg/cgroup/cgroup_stats_linux.go
+++ b/pkg/cgroup/cgroup_stats_linux.go
@@ -43,11 +43,14 @@ type CCgroupV12StatManager struct {
 // See comments here: https://github.com/sustainable-computing-io/kepler/pull/609#discussion_r1155043868
 func NewCGroupStatManager(pid int) (CCgroupStatHandler, error) {
 	p := fmt.Sprintf(procPath, pid)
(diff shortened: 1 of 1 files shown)


pkg/cgroup/cgroup_stats_linux.go
```diff
@@ -43,11 +43,14 @@ type CCgroupV12StatManager struct {
...
 func NewCGroupStatManager(pid int) (CCgroupStatHandler, error) {
 	p := fmt.Sprintf(procPath, pid)
-	_, path, err := cgroups.ParseCgroupFileUnified(p)
+	cgroupMap, path, err := cgroups.ParseCgroupFileUnified(p)
 	if err != nil {
 		return nil, err
 	}
-
+	if path == "" {
+		// if there is no None subsystem (<controller>::<cgrouppath>), use common path from pids subsystem
```
marceloamaral, line 51: There is a typo:
`if there is no None subsystem`
->
`if there is no subsystem`

marceloamaral (changes requested):
Looks good, after fixing the typo we can merge this PR.

rootfs (approved):

PR #690, opened 2023-05-16
nit: remove GPUStats to avoid confusion as it's not used at all

jichenjc:


jichenjc:
found this during other dev item ..

marceloamaral:
I believe we should not remove the GPU statistics since processes can utilize the GPU and it contributes to the overall power consumption.

jichenjc:
>>>I believe we should not remove the GPU statistics since processes can utilize the GPU and it contributes to the overall power consumption.

it is not used anywhere, I try to find GPU metrics and found this, then search but seems no where update this value or set it

marceloamaral (approved):
Right, the `GPUStats` is never used.

The GPU metrics are updated in `CounterStats` as in pkg/collector/container_accelerator_collector.go:L65

PR #692, opened 2023-05-17
Update adding the estimated value to gauge logic

marceloamaral:
The pre-trained Power Model provides node component power estimates, which return gauge metrics in contrast to the counter metrics used by RAPL.

Consequently, updating the estimated node component as counters would be considered a bug, as it would not appropriately aggregate the gauge values.

To address this issue, this PR includes logic to determine whether a metric should be treated as a gauge or a counter, ensuring proper handling of gauge values.


rootfs:
@marceloamaral does this change impact the dashboard?

rootfs (approved):

Issue #640, opened 2023-04-22
Container image release/versioning


yellowhat:
First of all. Thank you very much for this amazing project.

**Is your feature request related to a problem? Please describe.**

It is very difficult to incrementally update `kepler`:

- github releases do not have a corresponding container image tag, i.e there is no `v0.4.10`
- `latest` container tag points to a non github release, currently:
  - `latest` container tag is 3 days old while the latest github release is 1 month old
  - `release-0.4` container tag is 4 months old,  while the `release-0.4` release is 1 month old

therefore it is not guarantee to always use a "released" container image.

**Describe the solution you'd like**

- Create container image with `rpm_version` and `rpm_release` tag.
- Only the `release` job can create container image with `latest` tag
- The `image` job use commit sha tag and/or branch name, not `latest`

**Describe alternatives you've considered**
-

**Additional context**
-

jichenjc:
>>>latest container tag is 3 days old while the latest github release is 1 month old

latest is always last merged PR built image .. so I see no issue on this 

>>>release-0.4 container tag is 4 months old, while the release-0.4 release is 1 month old

0.5 is being planned, not sure we want to create 0.4.x ... @SamYuan1990 @rootfs FYI

yellowhat:
I am sorry, but this is not what I am pointing out.

let me clarify:

* the `latest` tag should point to the latest **release** (something known that should work) not a PR; a PR will have its own tag (ie `<branch>`)
* I was not pointing out how long it takes to release a new version, but the fact that there is a discrepancy between what `release-0.4` container tag is and what `release-0.4` github release is by looking at the creation date.
* If I would like to rollback to a previous release (ie 0.4.10 > 0.4.9) there is no container image after the creation of a new release.

Does this clarify my points?

Thanks

SamYuan1990:
it seems just updated with latest but not release-0.4 https://github.com/sustainable-computing-io/kepler/actions/runs/4437344205/jobs/7786948391#step:6:549 @rootfs

rootfs:
@SamYuan1990 can you make PR tagged build images? That'll be very helpful to share and validate fixes before merge.

SamYuan1990:
> @SamYuan1990 can you make PR tagged build images? That'll be very helpful to share and validate fixes before merge.

Let me think and try.
for PR level, 
- I don't want to build pre PR in automatic hence there will be lots of them and some thing as nit change( e.x. log typo) seems useless.
[...]

PR #693, opened 2023-05-17
CI add specific build job

SamYuan1990:
ref https://github.com/actions/checkout/tree/v3/#checkout-pull-request-head-commit-instead-of-merge-commit try to use github specific to check out code and meanwhile for container image tag.

### DIFF
--- a/.github/workflows/image_pr.yml
+++ b/.github/workflows/image_pr.yml
@@ -1,0 +1,38 @@
+name: imagePRlevel
+
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      commitSHA:
+        description: "pr commit SHA"
+        required: true
+
+jobs:
+  image_build_for_specific_commit:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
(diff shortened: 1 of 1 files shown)


SamYuan1990:
as a try for #640

SamYuan1990:
@rootfs , so far it require a full git commit SHA for example https://github.com/SamYuan1990/kepler/actions/runs/5004169317/jobs/8966392722 
https://github.com/SamYuan1990/kepler/actions/runs/5004190998/jobs/8966440203 (feature branch commit hash as `adf1106abc5d3dfb4bf9a227da7196c170c90fe9`)
any other suggestions?

.github/workflows/image_pr.yml
```diff
@@ -0,0 +1,38 @@
+name: imagePRlevel
+
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      commitSHA:
+        description: "pr commit SHA"
+        required: true
```
rootfs, line 9: do we have to provide sha manually or it can be set to the last commit?

SamYuan1990: manually

rootfs (approved):

Issue #645, opened 2023-04-24
Provide network/storage/accelerator power attribution


rootfs:
- [ ] collect network/storage/accelerator utilization metrics
- [ ] collect network/storage/accelerator power metrics
- [ ] create models to estimate power usage by workloads based on their utilization metrics

jichenjc:
what will be add in addition to our existing model (e.g the network card, the SAN or the GPU/DPU ) etc??

marceloamaral:
There are two approaches that we can take to tackle this issue.

First, we can conduct a fine-grain estimation of the power consumption of each process for a given device (i.e., CPU, DRAM, and GPU). Alternatively, we can estimate the total power consumption of the node using multiple metrics, leading to a coarse-grain estimation (i.e., Platform power).

The Kepler code already has the coarse-grain (i.e., Platform power) and fine-grain (i.e., CPU, DRAM, and GPU) estimations.

Therefore, If we can collect network and storage power consumption data, we can develop a power model to estimate the power consumption per process.

Furthermore, we can improve the current Platform power model by including more metrics that better reflect the utilization of other resources. However, we must ensure that the collected Platform power includes the power consumption of the target resource.

jichenjc:
>>> network and storage power consumption data

I assume the network = NIC /DPU, yes, this seems ok, so either coarse-grain or fine-grain should be able to provide such data
but storage usually SAN/SDS will be used which is distributed, so maybe not that easy..

jiere:
Due to accelerator power attribution topic, I think there are two kinds of definition of the word "accelerator":
1. Stand-alone HW accelerator, such as GPU. DPU also can be regarded as such accelerator since it offloads specific computing from CPU.
2. CPU built-in accelerator, such as Intel Xeon SP's bunch of in-die/in-package accelerators.
The work mechanism should be greatly different between the two. So let's think about how to organize the current code hierarchy of pkg/power and define appropriate framework to support different kinds of accelerators.

marceloamaral:
Sounds good.

But note that CPU built-in accelerator on Intel Processors is part of the RAPL UNCORE power.

jiere:
> But note that CPU built-in accelerator on Intel Processors is part of the RAPL UNCORE power.
[...]

PR #711, opened 2023-05-28
Platform Validation enhancement proposal

jiere:


rootfs:
this idea looks good, @SamYuan1990 wdyt?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Step 3: Run test cases and get test result.
+
+Step 4: Generate test report and save to somewhere.
+
+Step 5: Delete the local cluster and recover the orignal environment on the test node.
+
+### Implementation Details/Notes/Constraints
+
+Our validation test is local-dev based, in another word, [Kind](https://kind.sigs.k8s.io/) based.
```
jichenjc, line 90: as a lot of hardware related things here, kind might not be (at least not today)  should we consider move this from kind to other e.g k3s or other solutions that eligible to do so?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Step 2: Deploy Kepler in the cluster.
+
+There are three kinds of deployment available in Kepler:
+
+Deploy using Helm Chart. Follow guide [here](https://github.com/sustainable-computing-io/kepler-helm-chart/blob/main/README.md).
+
+Deploy using Kepler Operator. Follow guide [here](https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/README.md).
+
+Deploy from Kepler Source Code. Follow guide [here](https://sustainable-computing.io/installation/kepler/).
```
jichenjc, line 78: assume this will be primary use case?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Currently, Kepler's test cases focus on code logic verification, lack of actual platform validation.
+
+Kepler's [data](https://github.com/sustainable-computing-io/kepler/tree/main/data) files majorly contain static information, or lack of data sync mechanism.
+
+Test cases and test report are not available, which could be the evidence or check criteria to below concerns:
+
+1. Whether the specific CPU/GPU/FPGA/Accelerator model is well supported in Kepler?
+
+2. Whether the data/metrics of specific power source are correctly exported?
```
jichenjc, line 31: very curious on how this can be done ... e.g calculate the pod energy then aggregate them and compare to RAPL or other measurement tool?

enhancements/platform-validation.md
jiere: I am open to your suggestions:)
Could you identify the specific gap why kind does not support specific hardware platform?

enhancements/platform-validation.md
jiere: Just as I mentioned in this PR, currently the first test case is to validate if the `cpu_architecture` label is expected in Prometheus. 
Due to whether we need to verify all the three deployment scenario, this is TBD, reply on community's discussion result.

enhancements/platform-validation.md
jiere: Something like what you said, Ji.
RAPL is of coz a way to do comparison, and we could also compare the specific delta, both for Dyn and Abs power.
And if other power metering tools data is available, also could be the check base.
So I suppose test plan is general, test case steps could be hardware specific.
Anyway, such test cases' goal is to let customer believe what we Kepler exported metrics are make sense and acceptable for following operations, i.e. model training, carbon footprint, etc.

enhancements/platform-validation.md
jichenjc: at least I knew `kind` has some issue in getting some data accurately :) 
that's the reason why our test/demo rely on baremetal machines and I was told to avoid use such for real calculation
maybe I am wrong after a couple of enhancement since I was first told couple of month ago

enhancements/platform-validation.md
jichenjc: >>>Anyway, such test cases' goal is to let customer believe what we Kepler exported metrics are make sense and acceptable for following operations, i.e. model training, carbon footprint, etc.

right, don't get me wrong :) 
I am in favor of doing this but just curious on how we can do the test and have a stable (or maybe some smarter way)
way to do the calculation

enhancements/platform-validation.md
jiere: Thanks for your info, Ji.
Let me do some further investigation and get you tuned.

marceloamaral: We can expose what we need from the host to Kind. 
For example, when generating a Kind cluster using scripts, the host's `/proc` folder is mounted to facilitate Kepler in visualizing all processes.

The only downside of Kind is its lack of internet connectivity within the cluster. Consequently, we need to upload all the required images to the cluster in order to use them.

jichenjc: ok, but seems kind and VM can't get Hardware counter etc info IIRC? 
with that , the test will lack of important part ?

marceloamaral: My Kind cluster deployment can collect Hardware counters.

Not sure what is the problem that you are mentioning....Can you provide more details?
How your Kind cluster was created?

jichenjc: um... ok, I think I remember wrong thing, the kind is not functional on VM ,but it may running on baremetal
https://github.com/sustainable-computing-io/kepler/issues/373#issuecomment-1308512108

so kind is not the problem , it's the VM that has some issue ..

jiere: guys, sometimes hw counter is not the must, even for VM, we can see that Red Hat guys are trying to submit [patch](https://www.mail-archive.com/qemu-devel@nongnu.org/msg962288.html) to kernel:)

rootfs: hw counters won't be available if hypervisors don't pass through CPU features to VMs.

hack/platform-validation.sh
```diff
@@ -0,0 +1,48 @@
...
+    make push-image
+
+    echo "waiting for cluster-clean to finish"
+    if ! wait $CLEAN_PID; then
+        echo "cluster-clean failed"
+    fi
+
+    ./hack/cluster-deploy.sh
+
+    #TODO: run validation cases and generate report.
```
SamYuan1990, line 45: where are we plan to store the report?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Motivation
+
+Currently, Kepler's test cases focus on code logic verification, lack of actual platform validation.
+
+Kepler's [data](https://github.com/sustainable-computing-io/kepler/tree/main/data) files majorly contain static information, or lack of data sync mechanism.
+
+Test cases and test report are not available, which could be the evidence or check criteria to below concerns:
+
+1. Whether the specific CPU/GPU/FPGA/Accelerator model is well supported in Kepler?
```
SamYuan1990, line 29: not only for CPU, but also for memory and IO usage as disk also in kepler's goal, right? @marceloamaral , @rootfs ?

enhancements/platform-validation.md
SamYuan1990: Curious on how this can be done. +1, but I am good here as motivation.

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+2. Whether the data/metrics of specific power source are correctly exported?
+
+3. Whether the power measurement accuracy is accpetable?
+
+4. More...
```
SamYuan1990, line 35: what's more as motivation? any example?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Kepler's [data](https://github.com/sustainable-computing-io/kepler/tree/main/data) files majorly contain static information, or lack of data sync mechanism.
+
+Test cases and test report are not available, which could be the evidence or check criteria to below concerns:
+
+1. Whether the specific CPU/GPU/FPGA/Accelerator model is well supported in Kepler?
+
+2. Whether the data/metrics of specific power source are correctly exported?
+
+3. Whether the power measurement accuracy is accpetable?
```
SamYuan1990, line 33: what's the difference with the 2nd item?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+4. More...
+
+### Goals
+
+1. Define workflow to validate specific hardward platform support in Kepler.
+
+2. Design test cases for such validation. (Platform agnostic and specific cases)
```
SamYuan1990, line 41: I am curious on the different between workflow and test case here?
is that means a general workflow as pipeline and running test case for platform specific?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+3. Define test report format and release process. (General test items result and hardware specific items)
+
+### Non-Goals
+
+1. Platforms which are not supported by Kepler yet should be out of this enhancement's scope.
+
+2. Platforms which do not support power measurement could be low priority for this enhancement, since they depends on model train and the power consumption check criteria is TBD.
```
SamYuan1990, line 49: external data source as HMC or BMC?

enhancements/platform-validation.md
SamYuan1990: in my point of view, either build from source, helm, operator are containerize. we also have rpm package and binary file. + @rootfs , my point of view is that we should cover them as well.
if kepler binary able to get ebpf or cgroup data, hence there is no difference though RPM, binary build from source, container or pod of k8s. If there is some thing wrong, maybe issue with mount(of a pod). in other words, I want to avoid overlap the test of helm chart or operator if possible.

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+There are three kinds of deployment available in Kepler:
+
+Deploy using Helm Chart. Follow guide [here](https://github.com/sustainable-computing-io/kepler-helm-chart/blob/main/README.md).
+
+Deploy using Kepler Operator. Follow guide [here](https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/README.md).
+
+Deploy from Kepler Source Code. Follow guide [here](https://sustainable-computing.io/installation/kepler/).
+
+All the above deployment scenarios related validation should be automated. i.e. locate at $Kepler_ROOT/hack/platform-validation-deploy-from-xxx.sh.
```
SamYuan1990, line 80: as there may different mount path for k8s, for example kind vs minikube... I am good with the idea. But considering we need deploy helm and operator(basing on current document version).... I will leave an open question here, is `$Kepler_ROOT/hack/platform-validation-deploy-from-xxx.sh.` a good place to keep this script?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+172.19.0.2
+```
+
+4. Retrieve the metrics with `cpu_architecture` label in Prometheus.
+```
+$ curl --noproxy "*" -s http://172.19.0.2:32517/api/v1/label/cpu_architecture/values | jq -r ".data[]" | sort
+Sapphire Rapids
+```
+
+5. Check the actual `cpu_architure` info on the node:
```
SamYuan1990, line 155: is there any case that `cpu_architure` mis match? or say if we use kind. can we use uname to get CPU arch and do the compare works? as we can even deploy a kube cluster on the server... as we have ssh access right? or the steps is without SSH connection?

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+```
+$ lscpu |grep "Machine type"
+Machine type:        3906
+```
+
+
+The further test cases include not but limited to:
+
+1. Typical workloads power consumption measurement (carbon footprint) accuracy check on specific platform.
+2. Carbon footprint accuracy check in scaling scenarios (VPA/HPA) on specific platform.
```
SamYuan1990, line 177: I want to query here.
as cpu_architure is a specific metric, and the further case seems cover k8s feature as VPA? I see mismatch or different level of things.

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+2023/5/27, Jie Ren (jie.ren@intel.com), initial draft for ideas and scope.
+
+## Alternatives
+
+
+## Infrastructure Needed
+
+1. Github runner machines which cover the Kepler suppported platforms. On-prem machines, CSP's BMs/VMs, etc.
+
+2. Test cases repositories, whether rely on current ginkgo framework is TBD.
```
SamYuan1990, line 193: Why not we extend current test case?

enhancements/platform-validation.md
jiere: Sam, really agree with you, on test scenario point of view, there are BM/container(k8s)/VM scenarios. 
binary/rpm, operator/helm, these are test/deployment method. 
Perfect test cases will cover all the combination to achieve full-mesh, but due to resource limit, we may pick up the minimal set to cover the scenarios at least.
Of course Kepler and its associated tools such as operator/helm charts may not cover all the scenarios, we can deal with them case by case in execution.

rootfs:
@jiere @SamYuan1990 how about having a section for open questions and future steps? I think there will be more clarity when this proposal is being implemented.

SamYuan1990:
> @jiere @SamYuan1990 how about having a section for open questions and future steps? I think there will be more clarity when this proposal is being implemented.

I can accept with open questions for merge this RFC. As it seems a huge item or ... a suggestion as we keep an item open in our plan board for tracing? and break it down into implementation level....

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+Step 4: Generate test report and save to somewhere.
+
+Step 5: Delete the local cluster and recover the orignal environment on the test node.
+
+### Implementation Details/Notes/Constraints
+
+Our validation test is local-dev based, in another word, [Kind](https://kind.sigs.k8s.io/) based. 
+
+Since the Kind is Kubernetes cluster running in Docker container, it may be difficult for us to directly check the Kepler exposed metrics on port 9102.
```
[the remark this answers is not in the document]
marceloamaral: Why it is difficult?
We can just do `kubectl port-forward`

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+
+The easier way is to deploy Prometheus service and Kepler related Service Monitor in the cluster. Expose Prometheus service as `NodePort` type.
```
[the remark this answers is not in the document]
marceloamaral: For the integration test perspective, it might be good to test if Prometheus is collecting the metrics. Specially because there some discussion to move to OpenTelemetry in the future...

hack/platform-validation.sh
jiere: Good question, originally I am intent to add the test cases code in shell and generate report there.
But after reviewing current kepler integration test code, maybe better idea is to add a new action like .github/workflow/platform-validation.yaml, and write specific go test code under kepler/e2e/ directory?
The report is another open topic for community, store under kepler/test directory? and how about the report content format? We are thinking about this and may submit proposal for community review in the following weeks..

enhancements/platform-validation.md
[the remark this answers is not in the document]
jiere: > Why it is difficult? We can just do `kubectl port-forward`

port-forward works for containerized service, but for KIND cluster, it's something nested, after run this command, the svc's port forwarded to visible port for KIND docker container, but our test scripts run on host...

enhancements/platform-validation.md
jiere: > is there any case that `cpu_architure` mis match? or say if we use kind. can we use uname to get CPU arch and do the compare works? as we can even deploy a kube cluster on the server... as we have ssh access right? or the steps is without SSH connection?

hmm, this test case is something like white box test, because the `cpu_architecture` label is added by our Kepler metrics collector code. On Prometheus point of view(sometimes also customer point of view), such label stands for the hardware platform, so the check logic should be consistent with our code, otherwise, as Marcelo said, there is a bug...
Think about this: when customers check metrics in Prom, they see various of "unknown" `cpu_architecture` labels...

jiere:
> @jiere @SamYuan1990 how about having a section for open questions and future steps? I think there will be more clarity when this proposal is being implemented.

Yes, I am super appreciated to receive so many constructive questions/suggestions from community so far. They help us a lot to clarify something and speed up implementation.
The initiative for me to choose enhancement directory to write this document is that we can follow the current template to describe ideas of enhancement, as the template says, we can record the change history in metadata, right? So the initial idea might not be perfect and mature, but we can adjust and implement step by step, case by case, the following PRs may not stand here, but point to other repo places, i.e. doc/, e2e/, test/, etc. But above all, any changes should be recorded here as the change history means.

enhancements/platform-validation.md
jiere: `cpu_architecture` is not metric, but one of the labels of some metrics, we chose it as the first test case because it can somehow stands for the platform.
For VPA, may introduce other metrics or other labels check then.

enhancements/platform-validation.md
```diff
@@ -0,0 +1,195 @@
...
+(3) For s390x platforms, we use `lscpu` command to check the machine type:
+```
+$ lscpu |grep "Machine type"
+Machine type:        3906
+```
+
+
+The further test cases include not but limited to:
+
+1. Typical workloads power consumption measurement (carbon footprint) accuracy check on specific platform.
```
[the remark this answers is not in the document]
jiere: really new knowledge for me, thanks Marcelo.

Makefile
```diff
@@ -262,8 +262,18 @@ cluster-sync:
 cluster-up:
 	rm -rf local-dev-cluster
 	git clone https://github.com/sustainable-computing-io/local-dev-cluster.git --depth=1
-	cd local-dev-cluster && ./main.sh
+	cd local-dev-cluster && ./main.sh up
```
SamYuan1990, line 265: So far we are going to use v0.0.0 version of local-dev-cluster instead of latest.
ok... this is not only a rfc pr.
if so, I suppose the teardown support will be supported from v0.0.1 of local-dev-cluster.
Hence the suggestion here is that we have a variable for local-dev-cluster version usage.

SamYuan1990 (changes requested):
see my latest update for make file change. It should be discussed which block this PR. Sorry for just found out it's not a rfc pr.

enhancements/platform-validation.md
jichenjc: what kind of memory/IO you are talking about ? those are common metric, do they ? 
there is no different model of CPU arch , GPU vendor etc?

rootfs:
@jiere can you resolve the conflict? Let's pick this up and keep updating it with new info.

SamYuan1990:
> @jiere can you resolve the conflict? Let's pick this up and keep updating it with new info.

I really don't want to remind here.
I would like suggestion that we decouple make file change and rfc change.
as ideally we may have another repo for rfc document as recording.
and from #721 , as a new idea comes in my mind update as suggestion to @jiere ...
make be we can decouple clone local-cluster-dev from cluster up or cluster down?
```
local-cluster-dev-check:
... some shell clone local cluster dev if not existing.

local-cluster-up: local-cluster-dev-check
cd local-cluster-dev && main.sh up

local-cluster-down: local-cluster-dev-check
cd local-cluster-dev && main.sh down
```
see, this PR is already full of comments on rfc. and we adding code review.... emoji: cry out.

jiere:
> @jiere can you resolve the conflict? Let's pick this up and keep updating it with new info.

Sure, @rootfs and @SamYuan1990 , sorry for late due to browser refreshment issue, I did not realize such conflict in time... 
Let me update soon to keep this ticket as pure RFC document and also add open questions and action plans as @rootfs [suggested](https://github.com/sustainable-computing-io/kepler/pull/711#issuecomment-1570102592)

enhancements/platform-validation.md
jiere: Agree, this is out of this proposal's scope, we may discuss in #645 topic.

enhancements/platform-validation.md
jiere: Needs community's brainstorming:) So I chose to write "More..." here.

enhancements/platform-validation.md
jiere: Let's discuss related validation cases when those features ready in Kepler.

enhancements/platform-validation.md
jiere: The 2nd item focus on the physical correctness of those metrics collected from specific power source, i.e. RAPL. But most of such metrics are node level or CPU PKG level.
The 3rd item focus on the fairness, and check either ratio based power estimation or model based power estimation is reasonable, acceptable, accurate, especially on Dynamic power estimation. Which is more attractive to customers who care about their workload's carbon footprint than customers who own the whole node/the whole cluster, such as CSP. 

> what's the difference with the 2nd item?

The 2nd item focus on the physical correctness of those metrics collected from specific power source, i.e. RAPL. But most of such metrics are node level or CPU PKG level.
The 3rd item focus on the fairness, and check either ratio based power estimation or model based power estimation is reasonable, acceptable, accurate, especially on Dynamic power estimation. Which is more attractive to customers who care about their workload's carbon footprint than customers who own the whole node/the whole cluster, such as CSP.

enhancements/platform-validation.md
SamYuan1990: GHA has it's own limitation for platform support, leave this as comments and let's see what's happen in the future.

rootfs:
Good discussion!

PR #715, opened 2023-05-30
remove: remove kubelet metrics

jichenjc:


marceloamaral:
Looks good to me.
@sunya-ch wdyt?

jichenjc:
@sunya-ch can you tell me whether it's ok to merge this?

sunya-ch:
I think we should keep it but make it optional to enable. I did found some difference on kubelet metric and cgroup metric especially after API server PR merged. 
With kubelet, as long as we have a pod name we got the stat, unlike cgroup that depends on detection from BPF. 

If we cam solve these difference such as change cgroup stat to depends on pod name or have stability on BPF process detection, we may remove it.

jichenjc:
>>>If we cam solve these difference such as change cgroup stat to depends on pod name or have stability on BPF process detection,

sunya-ch:
FYI, here is the validation result of availability on each metric source I have locally (after enable apiserver with some local changes).


```
Benchmark: coremark
Benchmark query file: coremark
==== cgroup ====
1 pods: 	Zero data for ['thread=16']
39 pods: 	Valid

Valid data points:
            >0
scenarioID
thread=16   39
thread=32   60
thread=4    43
thread=8    41
==== hwc ====
15 pods: 	Zero data for ['thread=4' 'thread=8' 'thread=16' 'thread=32']
25 pods: 	Valid

Valid data points:
            >0
scenarioID
thread=16   11
thread=32   36
thread=4    22
thread=8    18
==== kubelet ====
40 pods: 	Valid

Valid data points:
            >0
scenarioID
thread=16   42
thread=32   61
thread=4    43
thread=8    40
==== bpf ====
27 pods: 	Zero data for ['thread=4' 'thread=8' 'thread=16' 'thread=32']
13 pods: 	Valid

Valid data points:
            >0
scenarioID
thread=16    7
thread=32   13
thread=4    11
thread=8    15
rapl data: 	[976]
platform data: 	[488]
======================================
```

> note: zero data mean with API server the pod is detected but there is no stat collected for it.

jichenjc:
>>>note: zero data mean with API server the pod is detected but there is no stat collected for it.

where's the code you used for this ? just want to check what's really missing caused the 0 reported ..so we can check how to fix

sunya-ch:
> > > > note: zero data mean with API server the pod is detected but there is no stat collected for it.
> 
> where's the code you used for this ? just want to check what's really missing caused the 0 reported ..so we can check how to fix

I implement it locally for the experiment but I can share this tool later. It has a dependency to the [CPE operator](https://github.com/IBM/cpe-operator) that report the created pod list corresponding to the benchmark CR (example [here](https://github.com/sustainable-computing-io/kepler-model-db/blob/main/query_response/236bac2e6940e332510d3ec9c4e76280/cr.yaml)). 
Then check the kepler query whether it reports those pods stat with number > 0.

Here is partial of the code
```python
        for rep in repetitions:
            podname = rep["pod"]
            for query in queries:
                df = query_results[query]
                filtered_df = df[df["pod_name"]==podname]
                # set validate item
                item = dict()
                item["pod"] = podname
                item["scenarioID"] = scenarioID
                item["query"] = query
                item["count"] = len(filtered_df)
                item[">0"] = len(filtered_df[filtered_df[query] > 0])
                item["total"] = filtered_df[query].max()
                items += [item]
                
def summary_validate_result(validate_df):
    for metric, query in metric_to_validate_pod.items():
        target_df = validate_df[validate_df["query"]==query]
        no_data_df = target_df[target_df["count"] == 0]
        zero_data_df = target_df[target_df[">0"] == 0]
        valid_df = target_df[target_df[">0"] > 0]
        print("==== {} ====".format(metric))
        if len(no_data_df) > 0:
            print("{} pods: \tNo data for {}".format(len(no_data_df), pd.unique(no_data_df["scenarioID"])))
        if len(zero_data_df) > 0:
            print("{} pods: \tZero data for {}".format(len(zero_data_df), pd.unique(zero_data_df["scenarioID"])))

        print("{} pods: \tValid\n".format(len(valid_df)))
        print("Valid data points:")
        print(valid_df.groupby(["scenarioID"]).sum()[[">0"]])
    for metric, query in metric_to_validate_power.items():
        target_df = validate_df[validate_df["query"]==query]
        print("{} data: \t{}".format(metric, target_df[">0"].values))
```

marceloamaral:
It seems unusual that you are able to collect kubelet metrics but not cgroup metrics.

> With kubelet, as long as we have a pod name we got the stat, unlike cgroup that depends on detection from BPF.

Regarding the apiserver, whenever a container is created and initialized, it is added to the cache. The cgroup metrics for all containers in the cache are consistently updated every 3 seconds. Therefore, the collection of cgroup metrics does not depend solely on BPF detection. BPF detection serves as a secondary method to identify the container.

@sunya-ch It is possible that Kepler does not have the necessary access permissions to collect cgroup metrics in your environment. Or, maybe the apiserver is not working in your environment...

sunya-ch:
> It seems unusual that you are able to collect kubelet metrics but not cgroup metrics.
> 
> > With kubelet, as long as we have a pod name we got the stat, unlike cgroup that depends on detection from BPF.
> 
> Regarding the apiserver, whenever a container is created and initialized, it is added to the cache. The cgroup metrics for all containers in the cache are consistently updated every 3 seconds. Therefore, the collection of cgroup metrics does not depend solely on BPF detection. BPF detection serves as a secondary method to identify the container.
> 
> @sunya-ch It is possible that Kepler does not have the necessary access permissions to collect cgroup metrics in your environment? Or, maybe the apiserver is not working in your environment...

- There was 39/40 pods that report the cgroup. So, I think it is not the permission issue.
- apiserver is working. The container metric map is there but just get zero value on one pod.

The reason I said it depends on BPF is that. BPF detect process/cgroup ID then we read the cgroup stats. 
Unlike kubelet, it read kubelet stat no matter the BPF can detect that process or not. 

My thought is that container is identified by apiserver but stat is not read due to BPF detection.

marceloamaral:
However, if you refer to the code at https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/container_cgroup_collector.go#L30, you will see that the cgroup metrics for all containers in the cache will be updated regardless the BPF.

What I'm trying to highlight is that the error you are encountering could potentially have a different cause or origin.

@sunya-ch Can you please check the logs of line L38 or L51? Maybe there was an error collecting the cgroup metric of this specific container.

sunya-ch:
> However, if you refer to the code at https://github.com/sustainable-computing-io/kepler/blob/main/pkg/collector/container_cgroup_collector.go#L30, you will see that the cgroup metrics for all containers in the cache will be updated regardless the BPF.
> 
> What I'm trying to highlight is that the error you are encountering could potentially have a different cause or origin.
> 
> @sunya-ch Can you please check the logs of line L38 or L51? Maybe there was an error collecting the cgroup metric of this specific container.

I think the point is PID is set here by BPF detection. 

https://github.com/sustainable-computing-io/kepler/blob/d989186fa621448814802f4dbcd3b59898ae7590/pkg/collector/container_hc_collector.go#L177

So, the `cgroup.NewCGroupStatManager` cannot new handler if PID is not set. 
I may miss some code that set PID. Please let me know. 

https://github.com/sustainable-computing-io/kepler/blob/d989186fa621448814802f4dbcd3b59898ae7590/pkg/collector/container_cgroup_collector.go#L30-L54

About the log, I don't have it for now. But I don't expect the deletion because in that case the kubelet stat will be also deleted.

marceloamaral:
Ok, we have two problems then.

- Conversion of PID to ContainerID in Microshift. (seems that it is still a problem)
- Finding container PID from the apiserver

@sunya-ch do you have the cgroup path of a container in Microshift? the  /proc/$pid/cgroup

jichenjc:
>>>Then check the kepler query whether it reports those pods stat with number > 0.

@sunya-ch can you extend your code a little bit to include what's report 0 so I can compare the output?
thus we know what kind of metrics missing and in dig into code etc might give me more hint on certain situation it's missing? Thanks

sunya-ch:
For more background for the above example, I ran Coremark with multiple repetition automatically generated by CPE. So that, it has multiple k8s job running sequentially right after the previous job has finished. Note that one job has one pod created. The above example has 40 jobs/pods/processes composing of 10 repetition of threads = 4,8,16,32. 

With CPE custom resource, I can list all coremark pods that were created so that I can validate if any pod information is missing. The above result should that one coremark pod with ['thread=16'] of one repetition is missing but the rest 39 cases have a value.

jichenjc:
indeed, I saw several continuous those logs in my env ( a relative simple , no big workload) 
guess that might be unexpected? 
```
I0425 04:06:16.488132       1 container_cgroup_collector.go:30] PID 0 does not have cgroup metrics since there is no /proc/0/cgroup
I0425 04:06:16.488189       1 container_cgroup_collector.go:30] PID 0 does not have cgroup metrics since there is no /proc/0/cgroup
I0425 04:06:16.488197       1 container_cgroup_collector.go:30] PID 0 does not have cgroup metrics since there is no /proc/0/cgroup

```

I assume this is

jichenjc:
>>>It is possible because c.ContainersMetrics is created by API server watcher but the pid is never set by the BPF (default=0).

right, I added some logs in the eBPF , but this doesn't print ..
```
index 2ed6241..445006f 100644
--- a/pkg/collector/container_hc_collector.go
+++ b/pkg/collector/container_hc_collector.go
@@ -175,6 +175,9 @@ func (c *Collector) updateBPFMetrics() {

                c.createContainersMetricsIfNotExist(containerID, ct.CGroupID, ct.PID, config.EnabledEBPFCgroupID)
                c.ContainersMetrics[containerID].PID = ct.PID
+               if ct.PID == 0 {
+                       klog.V(5).Infof("-----------get %s PID = 0", containerID)
+               }

```

jichenjc:
@rootfs no, I think we are still working on some metrics not able to get ,need more test to validate whether recent change can fix those

marceloamaral:
@sunya-ch can we remove the Kubelet metrics now, or we still need to use it?

sunya-ch:
> @sunya-ch can we remove the Kubelet metrics now, or we still need to use it?

Yes, we can remove it.
May need to merge the following PR first. As in the test case, it is still using KubeletOnly based power model.
- https://github.com/sustainable-computing-io/kepler/pull/947/files#diff-a2efbc8bcd1d9fd731a7d9af4aa7c7eb2c2c81e21427c8c54ec2360af839f67d

sunya-ch:
I think now we can rebase this PR to remove the kubelet metric and merge.

jichenjc:
@sunya-ch PR rebased, thanks

sunya-ch (approved):
@jichenjc Thank you for the update.

/lgtm

PR #726, opened 2023-06-06
collector: suppress debug info

rootfs:


pkg/collector/container_accelerator_collector.go
```diff
@@ -61,7 +61,7 @@ func (c *Collector) updateAcceleratorMetrics() {
 			if _, exist := c.ContainersMetrics[containerID]; !exist {
 				continue
 			}
-			klog.Infoln("DEBUG: pod", c.ContainersMetrics[containerID].PodName, "processUtilizationSample", processUtilization)
+			klog.V(6).Infoln("DEBUG: pod", c.ContainersMetrics[containerID].PodName, "processUtilizationSample", processUtilization)
```
marceloamaral, line 64: This is left over, can you just remove the line?

rootfs: removed

Issue #644, opened 2023-04-24
Support external power source: BMC/IPMI/HMC


rootfs:
### Current Kepler Architecture

- [x] HMC
- [x] BMC
![image](https://github.com/sustainable-computing-io/kepler/assets/7062400/16dd2f62-77bf-4134-b882-52bb407f6d9f)

### Out of band external power source support
![image](https://github.com/sustainable-computing-io/kepler/assets/7062400/3a2842ec-0d7c-4915-a9f3-12060b41393e)

jichenjc:
you mean add a new model that we can obtain power data through those devices and put into current kepler model 
so like previously we only consider machine itself but now we need consider related device energy ?

marceloamaral:
The power consumption of a platform (i.e. server) can be reported by the BMC/IPMI/HMC. 

In our current implementation of Kepler, we collect the platform power consumption from the motherboard sensor, which is available in most modern servers. This sensor provides data on the power consumption of components directly attached to the motherboard, such as the CPU and memory. However, it may not include the power consumption of components like disks and GPUs.

Access to the motherboard sensor is possible via the ACPI interface within the machine or through IPMI, which reads the motherboard sensor via the BMC. We currently use the ACPI interface in Kepler, but in cases where the ACPI interface is disabled and IPMI is enabled, IPMI could be used instead.

It's important to note that the power consumption data obtained through IPMI may differ if the source is BMC or `out-of-band management systems` that can consolidate the power consumption of different components, including the platform, disk, and GPU.

jichenjc:
ok, make sense to me , appreciate the detailed info~

jiere:
Please see the joint message from IPMI promoters [here](https://www.intel.com/content/www/us/en/products/docs/servers/ipmi/ipmi-home.html).
Even IPMI v2.0 is a 10+ years-old spec, there are various of open-source projects related to IPMI metrics exporter.
Shall we directly support BMC-Redfish integration for OOB power monitoring?
Another question is about the metrics usage, since BMC data is some kinds of runtime transient power, not aggregate, how could it be used in Kepler then?

marceloamaral:
> IPMI or Redfish

I am ok with any direction

> BMC data is some kinds of runtime transient power, not aggregate, how could it be used in Kepler

We do extrapolation (current power * elapsed time) and aggregate it in Kepler.

rootfs:
Let's focus on Redfish first.

eklee15:
[...]

PR #734, opened 2023-06-13
Support redfish 

rootfs:
Get power info from BMCs that support Redfish. Kepler only needs to access chassis and system to get the power info. Currently the existing redfish golang bindings based on OpenAPI client or gofish are overkill. This PR uses simple access models. 

#644

Tasks:
- [x] Redfish access credential management
- [x] Create Redfish API model
- [x] Use Redfish API model to retrieve power metrics
- [x] Kepler collector reads Redfish power metrics
- [x] Kepler emits Redfish power metrics
- [x] Test on real servers (Dell/HPE)
- [x] Document BMC power metrics
- [ ] (stretch) Support temperature, fan, humidity metrics

### DIFF
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -306,12 +306,12 @@ func (ne *NodeMetrics) SetNodeOtherComponentsEnergy() {
 	dynCPUComponentsEnergy := ne.DynEnergyInPkg.SumAllDeltaValues() +
 		ne.DynEnergyInDRAM.SumAllDeltaValues() +
 		ne.DynEnergyInGPU.SumAllDeltaValues()
-	// other component can be either platform
 	dynPlatformEnergy := ne.DynEnergyInPlatform.SumAllDeltaValues()
 	if dynPlatformEnergy > dynCPUComponentsEnergy {
 		otherComponentEnergy := dynPlatformEnergy - dynCPUComponentsEnergy
 		ne.DynEnergyInOther.SetDeltaStat(OTHER, otherComponentEnergy)
 	}
+
 	idleCPUComponentsEnergy := ne.IdleEnergyInPkg.SumAllDeltaValues() +
 		ne.IdleEnergyInDRAM.SumAllDeltaValues() +
 		ne.IdleEnergyInGPU.SumAllDeltaValues()
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -306,12 +306,12 @@ func (ne *NodeMetrics) SetNodeOtherComponentsEnergy() {
 	dynCPUComponentsEnergy := ne.DynEnergyInPkg.SumAllDeltaValues() +
 		ne.DynEnergyInDRAM.SumAllDeltaValues() +
 		ne.DynEnergyInGPU.SumAllDeltaValues()
+	// other component can be either platform
 	dynPlatformEnergy := ne.DynEnergyInPlatform.SumAllDeltaValues()
 	if dynPlatformEnergy > dynCPUComponentsEnergy {
 		otherComponentEnergy := dynPlatformEnergy - dynCPUComponentsEnergy
 		ne.DynEnergyInOther.SetDeltaStat(OTHER, otherComponentEnergy)
 	}
-
 	idleCPUComponentsEnergy := ne.IdleEnergyInPkg.SumAllDeltaValues() +
 		ne.IdleEnergyInDRAM.SumAllDeltaValues() +
 		ne.IdleEnergyInGPU.SumAllDeltaValues()
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -41,7 +41,7 @@ const (
 )
 
 var (
-	NodeName            = getNodeName()
+	NodeName            = GetNodeName()
 	NodeCPUArchitecture = getCPUArch()
 	NodeCPUPackageMap   = getCPUPackageMap()
 
--- a/pkg/nodecred/csv_cred.go
+++ b/pkg/nodecred/csv_cred.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodecred
+
+import (
+	"encoding/csv"
+	"fmt"
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+// csvNodeCredImpl is the implementation of NodeCred using on disk file
+// the file is in the format of
+// node1,admin,password,localhost
+// node2,admin,password,localhost
+// node3,admin,password,localhost
+type csvNodeCred struct {
+}
+
+var (
+	credMap map[string]string
+)
+
+func (c csvNodeCred) GetNodeCredByNodeName(nodeName, target string) (map[string]string, error) {
+	if credMap == nil {
+		return nil, fmt.Errorf("credential is not set")
+	} else if target == "redfish" {
+		cred := make(map[string]string)
+		cred["redfish_username"] = credMap["redfish_username"]
+		cred["redfish_password"] = credMap["redfish_password"]
+		cred["redfish_host"] = credMap["redfish_host"]
+		if cred["redfish_username"] == "" || cred["redfish_password"] == "" || cred["redfish_host"] == "" {
+			return nil, fmt.Errorf("no credential found")
+		}
+		return cred, nil
+	}
+
+	return nil, fmt.Errorf("no credential found for target %s", target)
+}
+
+func (c csvNodeCred) IsSupported(info map[string]string) bool {
+	// read redfish_cred_file_path from info
+	filePath := info["redfish_cred_file_path"]
+	if filePath == "" {
+		return false
+	} else {
+		nodeName := getNodeName()
+		// read file from filePath
+		userName, password, host, err := readCSVFile(filePath, nodeName)
+		if err != nil {
+			klog.V(5).Infof("failed to read csv file: %v", err)
+			return false
+		}
+		klog.V(5).Infof("read csv file successfully")
+		credMap = make(map[string]string)
+		credMap["redfish_username"] = userName
+		credMap["redfish_password"] = password
+		credMap["redfish_host"] = host
+	}
+	return true
+}
+
+func getNodeName() string {
+	nodeName := os.Getenv("NODE_NAME")
+	if nodeName == "" {
+		nodeName = "localhost"
+	}
+	return nodeName
+}
+
+func readCSVFile(filePath, nodeName string) (userName, password, host string, err error) {
+	// Open the CSV file
+	file, err := os.Open(filePath)
+	if err != nil {
+		fmt.Println("Error opening the file:", err)
+		return
+	}
+	defer file.Close()
+
+	// Create a new CSV reader
+	reader := csv.NewReader(file)
+
+	// Read all rows from the CSV file
+	rows, err := reader.ReadAll()
+	if err != nil {
+		fmt.Println("Error reading CSV:", err)
+		return
+	}
+
+	// Iterate over each row and check if the node name matches
+	for _, row := range rows {
+		if row[0] == nodeName && len(row) >= 4 {
+			userName = row[1]
+			password = row[2]
+			host = row[3]
+			return userName, password, host, nil
+		}
+	}
+	err = fmt.Errorf("node name %s not found in file %s", nodeName, filePath)
+	return
+}
--- a/pkg/nodecred/csv_cred.go
+++ b/pkg/nodecred/csv_cred.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"os"
 
+	metric_util "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+
 	"k8s.io/klog/v2"
 )
 
@@ -59,7 +61,7 @@ func (c csvNodeCred) IsSupported(info map[string]string) bool {
 	if filePath == "" {
 		return false
 	} else {
-		nodeName := getNodeName()
+		nodeName := metric_util.GetNodeName()
 		// read file from filePath
 		userName, password, host, err := readCSVFile(filePath, nodeName)
 		if err != nil {
@@ -73,14 +75,6 @@ func (c csvNodeCred) IsSupported(info map[string]string) bool {
 		credMap["redfish_host"] = host
 	}
 	return true
-}
-
-func getNodeName() string {
-	nodeName := os.Getenv("NODE_NAME")
-	if nodeName == "" {
-		nodeName = "localhost"
-	}
-	return nodeName
 }
 
 func readCSVFile(filePath, nodeName string) (userName, password, host string, err error) {
--- a/pkg/power/platform/source/redfish.go
+++ b/pkg/power/platform/source/redfish.go
@@ -1,0 +1,286 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package source
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+	"github.com/sustainable-computing-io/kepler/pkg/nodecred"
+
+	"k8s.io/klog/v2"
+)
+
+// RedfishSystemModel is the struct for the system model
+// this is generated via the following command:
+// redfishtool Systems
+type RedfishSystemModel struct {
+	OdataContext string `json:"@odata.context"`
+	OdataID      string `json:"@odata.id"`
+	OdataType    string `json:"@odata.type"`
+	Description  string `json:"Description"`
+	Members      []struct {
+		OdataID string `json:"@odata.id"`
+	} `json:"Members"`
+	MembersOdataCount int    `json:"Members@odata.count"`
+	Name              string `json:"Name"`
+}
+
+// RedfishPowerModel is the struct for the power model
+// this is generated via the following command:
+// redfishtool raw GET /redfish/v1/Chassis/System.Embedded.1/Power#/PowerControl, where "System.Embedded.1" is the system ID from the RedfishSystemModel
+// the output is then formatted via https://mholt.github.io/json-to-go/ based on sample output from https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2023.1.pdf
+type RedfishPowerModel struct {
+	OdataType     string          `json:"@odata.type,omitempty"`
+	ID            string          `json:"Id,omitempty"`
+	Name          string          `json:"Name,omitempty"`
+	PowerControl  []PowerControl  `json:"PowerControl,omitempty"`
+	Voltages      []Voltages      `json:"Voltages,omitempty"`
+	PowerSupplies []PowerSupplies `json:"PowerSupplies,omitempty"`
+	Actions       Actions         `json:"Actions,omitempty"`
+	OdataID       string          `json:"@odata.id,omitempty"`
+}
+type PowerMetrics struct {
+	IntervalInMin        int `json:"IntervalInMin,omitempty"`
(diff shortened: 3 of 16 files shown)


pkg/collector/metric/node_metric.go
```diff
@@ -61,6 +62,7 @@ type NodeMetrics struct {
 	TotalEnergyInGPU      *types.UInt64StatCollection
 	TotalEnergyInOther    *types.UInt64StatCollection
 	TotalEnergyInPlatform *types.UInt64StatCollection
+	TotalEnergyInRedfish  *types.UInt64StatCollection
```
SamYuan1990, line 65: I was thinking, if redfish, is a kind of platform energy api, then we may don't need a new var as `TotalEnergyInRedfish` but reuse `TotalEnergyInPlatform`?
as redfish api is a type of Platform API?

pkg/nodecred/csv_cred.go
```diff
@@ -0,0 +1,116 @@
...
+		klog.V(5).Infof("read csv file successfully")
+		credMap = make(map[string]string)
+		credMap["redfish_username"] = userName
+		credMap["redfish_password"] = password
+		credMap["redfish_host"] = host
+	}
+	return true
+}
+
+func getNodeName() string {
```
SamYuan1990, line 78: do we need this function? https://github.com/sustainable-computing-io/kepler/blob/f0e64598286350b7215bcb26679fd0c0864fbdd9/pkg/collector/metric/utils.go#L101 ?

pkg/collector/metric/node_metric.go
marceloamaral: I agree.

Also, if redfish is enabled, we don't need to collect the platform power from other sources.

marceloamaral (changes requested):
Let's use redfish to update the Platform Energy.
So, we should verify in the function `updatePlatformEnergy` if redfish is enabled, otherwise we use the ACPI or LPAR power sources.

marceloamaral:
The PR #748 refactor the code and make it easier to include the redfish support.

pkg/power/redfish/util.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &system, nil
+}
+
+func getRedfishPower(access RedfishAccessInfo, system string) (*RedfishPowerModel, error) {
+	var power RedfishPowerModel
+	err := getRedfishModel(access, "/redfish/v1/Chassis/"+system+"/Power/PowerControl", &power)
```
tiwatsuka, line 78: `/Power/PowerControl` is not specified as a resource in [Redfish Resource and Schema Guide](https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2023.1.pdf) and HPE iLo Redfish API does not provide it. 
cf. https://hewlettpackard.github.io/ilo-rest-api-docs/ilo5/#resource-map
Instead, all of `/Power/#PowerControl`, `/Power#PowerControl` and `/Power#/PowerControl` are available on my machine.  You can test HPE environment at https://ilorestfulapiexplorer.ext.hpe.com/ .

rootfs: thank you @tiwatsuka for the info! I am going to update the API model to work with your suggestion.

rootfs: @tiwatsuka can you get the output of `Power#/PowerControl` on your HPE? The simulator doesn't have it, thanks

tiwatsuka: I can't access on my machine right now, but it was just same as output of `Power`. On second thoughts, it is just interpreted as a URL fragment that the part after `#`.

pkg/power/redfish/redfish.go
```diff
@@ -0,0 +1,255 @@
...
+func (rf *RedFishClient) GetPower() map[string]float64 {
+	if rf.systems != nil {
+		power := make(map[string]float64)
+		for _, system := range rf.systems {
+			rf.mutex.Lock()
+			now := time.Now()
+			// calculate the elapsed time since the last power query in seconds
+			elapsed := now.Sub(system.timestamp).Seconds()
+			klog.V(5).Infof("power info: %+v\n", system)
+			power[system.system] = float64(system.consumedWatts * 1000 * int(elapsed)) // convert to mW
```
YaSuenag, line 242: This line seems to calculate value in Joule (`consumedWatts` x `elapsed`), so is it better to name to `GetEnergyFromRedfish` rather than `GetPower` in func name?

marceloamaral: It makes sense, @rootfs can you rename this function in the Class level?

marceloamaral (changes requested):
@rootfs can you change the logic to get the redfish power as the Platform power?
You can follow the logic structure in [hmc](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/platform/source/hmc.go) and the logic in [acpi](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/platform/source/acpi.go).

Also when we are using the redfish to collect the platform power (i.e. the total node power), we don't need to use the ACPI or HMC. So the [platform package](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/platform/power.go) should verify if redfish is available and use this instead of ACPI or HMC.

This should be a similar logic as we collect the component energy that can come from different power sources.

pkg/power/redfish/redfish.go
marceloamaral: Never-mind, the Class already defines the right function name: `GetEnergyFromPlatform()`

pkg/power/redfish/util.go
tiwatsuka:
```
{
  "@odata.context": "/redfish/v1/$metadata#Power.Power",
  "@odata.etag": "W/\"2613CD55\"",
  "@odata.id": "/redfish/v1/Chassis/1/Power",
  "@odata.type": "#Power.v1_3_0.Power",
  "Id": "Power",
  "Name": "PowerMetrics",
  "Oem": {
    "Hpe": {
      "@odata.context": "/redfish/v1/$metadata#HpePowerMetricsExt.HpePowerMetricsExt",
      "@odata.type": "#HpePowerMetricsExt.v2_3_0.HpePowerMetricsExt",
      "BrownoutRecoveryEnabled": true,
      "HasCpuPowerMetering": false,
      "HasDimmPowerMetering": false,
      "HasGpuPowerMetering": false,
      "HasPowerMetering": true,
      "HighEfficiencyMode": "Balanced",
      "Links": {
        "PowerMeter": {
          "@odata.id": "/redfish/v1/Chassis/1/Power/PowerMeter"
        },
        "FastPowerMeter": {
          "@odata.id": "/redfish/v1/Chassis/1/Power/FastPowerMeter"
        },
        "SlowPowerMeter": {
          "@odata.id": "/redfish/v1/Chassis/1/Power/SlowPowerMeter"
        },
        "FederatedGroupCapping": {
          "@odata.id": "/redfish/v1/Chassis/1/Power/FederatedGroupCapping"
        }
      },
      "MinimumSafelyAchievableCap": null,
      "MinimumSafelyAchievableCapValid": false,
      "SNMPPowerThresholdAlert": {
        "DurationInMin": 0,
        "ThresholdWatts": 0,
        "Trigger": "Disabled"
      }
    }
  },
  "PowerControl": [
    {
      "@odata.id": "/redfish/v1/Chassis/1/Power#PowerControl/0",
      "MemberId": "0",
      "PowerCapacityWatts": 500,
      "PowerConsumedWatts": 44,
      "PowerMetrics": {
        "AverageConsumedWatts": 40,
        "IntervalInMin": 20,
        "MaxConsumedWatts": 78,
        "MinConsumedWatts": 38
      }
    }
  ],
  "PowerSupplies": [
    {
      "@odata.id": "/redfish/v1/Chassis/1/Power#PowerSupplies/0",
      "FirmwareVersion": "1.00",
      "LastPowerOutputWatts": 44,
      "LineInputVoltage": 105,
      "LineInputVoltageType": "ACLowLine",
      "Manufacturer": "LTEON",
      "MemberId": "0",
      "Model": "865408-B21",
      "Name": "HpeServerPowerSupply",
      "Oem": {
        "Hpe": {
          "@odata.context": "/redfish/v1/$metadata#HpeServerPowerSupply.HpeServerPowerSupply",
          "@odata.type": "#HpeServerPowerSupply.v2_0_0.HpeServerPowerSupply",
          "AveragePowerOutputWatts": 44,
          "BayNumber": 1,
          "HotplugCapable": true,
          "MaxPowerOutputWatts": 58,
          "Mismatched": false,
          "PowerSupplyStatus": {
            "State": "Ok"
          },
          "iPDUCapable": false
        }
      },
      "PowerCapacityWatts": 500,
      "PowerSupplyType": "AC",
      "SerialNumber": "5WBXK0FLLG25CW",
      "SparePartNumber": "866729-001",
      "Status": {
        "Health": "OK",
        "State": "Enabled"
      }
    },
    {
      "@odata.id": "/redfish/v1/Chassis/1/Power#PowerSupplies/1",
      "MemberId": "1",
      "Oem": {
        "Hpe": {
          "@odata.context": "/redfish/v1/$metadata#HpeServerPowerSupply.HpeServerPowerSupply",
          "@odata.type": "#HpeServerPowerSupply.v2_0_0.HpeServerPowerSupply",
          "BayNumber": 2
        }
      },
      "Status": {
        "Health": "Warning",
        "State": "Absent"
      }
    }
  ]
}
```

pkg/power/redfish/redfish.go
```diff
@@ -0,0 +1,255 @@
...
+			if power.PowerControl[0].PowerMetrics.IntervalInMin > intervalInMin {
+				intervalInMin = power.PowerControl[0].PowerMetrics.IntervalInMin
+			}
+		} else {
+			klog.V(5).Infof("failed to get power info: %v\n", err)
+		}
+	}
+
+	// set a timer to check the power info every intervalInMin minutes
+	if intervalInMin > 0 {
```
tiwatsuka, line 208: `IntervalInMin` in Redfish is irrelevant to the interval of updating the value of `PowerConsumedWatts`.
I believe that `PowerConsumedWatts` is updated every one or few seconds. So it would rather be collected more frequently than using `IntervalInMin`.

pkg/power/redfish/redfish.go
```diff
@@ -0,0 +1,255 @@
...
+
+// GetPower returns the power consumption in Watt
+func (rf *RedFishClient) GetPower() map[string]float64 {
+	if rf.systems != nil {
+		power := make(map[string]float64)
+		for _, system := range rf.systems {
+			rf.mutex.Lock()
+			now := time.Now()
+			// calculate the elapsed time since the last power query in seconds
+			elapsed := now.Sub(system.timestamp).Seconds()
```
tiwatsuka, line 240: This `timestamp` is based on the ticker defined at `IsPowerSupported()`, but `GetPower()` is called periodically based on another ticker. Is the energy value calculated correctly? I suspect that some energy is added multiple times or lost due to asynchronous two tickers.

pkg/power/redfish/util.go
rootfs: thank you @tiwatsuka this looks consistent with the new model definition [here](https://github.com/sustainable-computing-io/kepler/pull/734/files#diff-1c7ba32c2f6c4c86fc187fd007b4ce14aa33f9e481377e7a2b111ee86a520db3R74)

I'll finish it next week and hopefully get into the next release in mid July.

pkg/power/redfish/redfish.go
rootfs: sounds good

pkg/power/redfish/redfish.go
rootfs: @tiwatsuka you are right, the GetPower() happens asynchronously. Since the HTTP request could take time, it is not a good idea to put it in the critical path. So I let the redfish call happen in the background and Kepler collector just reads the existing reading without having to issue HTTP calls. The accuracy and synchronization can be further improved when using a smaller probe interval window or post processing calibration.

pkg/collector/metric/node_metric.go
rootfs: sounds good, working on the refactoring

pkg/power/redfish/redfish.go
tiwatsuka: The timestamp should be refreshed when the energy is read. If not, it counts same period multiple times.
Current implementation seems like the image shown below.  (I hope it makes sense to you. )
image

rootfs: nice catch!

rootfs: @tiwatsuka thanks for the review. I have pushed the code. Will test it and ask for more reviews once I get a new system to test it again.

rootfs:
@marceloamaral redfish and ACPI are now in the same platform pkg. If redfish is available, then platform uses redfish reading, otherwise use ACPI.

pkg/power/platform/source/redfish.go
```diff
@@ -0,0 +1,286 @@
...
+			continue
+		}
+		id := split[len(split)-1]
+		res := RedfishSystemPowerResult{}
+		power, err := getRedfishPower(rf.accessInfo, id)
+		if err == nil && len(power.PowerControl) > 0 {
+			res.system = id
+			res.consumedWatts = power.PowerControl[0].PowerConsumedWatts
+			res.timestamp = time.Now()
+			rf.systems = append(rf.systems, &res)
```
tiwatsuka, line 229: `IsSystemCollectionSupported()` is called from `updatePlatformEnergy()` at `node_energy_collector` periodically. 
 So `rf.systems` should be refreshed every time.

rootfs: cool! @tiwatsuka do you like to provide your [suggested changes](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) here? I'll merge it to this PR

tiwatsuka, line 229: I mean that it should replace `consumeWatts` only if the system had already existed in `rf.systems`.
```suggestion
	for index, member := range system.Members {
		// split the OdataID by delimiter "/" and get the system ID
		split := strings.Split(member.OdataID, "/")
		if len(split) < 2 {
			continue
		}
		id := split[len(split)-1]
		res := RedfishSystemPowerResult{}
		power, err := getRedfishPower(rf.accessInfo, id)
		if err == nil && len(power.PowerControl) > 0 {
			if index < len(rf.systems) {
				rf.systems[index].consumedWatts = power.PowerControl[0].PowerConsumedWatts
			} else {
				res.system = id
				res.consumedWatts = power.PowerControl[0].PowerConsumedWatts
				res.timestamp = time.Now()
				rf.systems = append(rf.systems, &res)
			}
```

tiwatsuka, line 229: I mean that it should replace `consumeWatts` only if the system had already existed in `rf.systems`.
```suggestion
	for index, member := range system.Members {
		// split the OdataID by delimiter "/" and get the system ID
		split := strings.Split(member.OdataID, "/")
		if len(split) < 2 {
			continue
		}
		id := split[len(split)-1]
		res := RedfishSystemPowerResult{}
		power, err := getRedfishPower(rf.accessInfo, id)
		if err == nil && len(power.PowerControl) > 0 {
			if index < len(rf.systems) {
				rf.systems[index].consumedWatts = power.PowerControl[0].PowerConsumedWatts
			} else {
				res.system = id
				res.consumedWatts = power.PowerControl[0].PowerConsumedWatts
				res.timestamp = time.Now()
				rf.systems = append(rf.systems, &res)
			}
```

pkg/power/platform/source/redfish.go
```diff
@@ -0,0 +1,286 @@
...
+			if power.PowerControl[0].PowerMetrics.IntervalInMin > intervalInMin {
+				intervalInMin = power.PowerControl[0].PowerMetrics.IntervalInMin
+			}
+		} else {
+			klog.V(5).Infof("failed to get power info: %v\n", err)
+		}
+	}
+
+	// set a timer to check the power info every probeInterval seconds
+	rf.ticker = time.NewTicker(rf.probeInterval)
```
tiwatsuka, line 240:
```suggestion
	if rf.ticker == nil {
		rf.ticker = time.NewTicker(rf.probeInterval)
	}
```

pkg/power/platform/source/redfish.go
```diff
@@ -0,0 +1,286 @@
...
+		for {
+			<-rf.ticker.C
+			for _, system := range rf.systems {
+				power, err := getRedfishPower(rf.accessInfo, system.system)
+				if err == nil && len(power.PowerControl) > 0 {
+					// mutex
+					rf.mutex.Lock()
+					klog.V(5).Infof("power info: %+v\n", power)
+					system.consumedWatts = power.PowerControl[0].PowerConsumedWatts
+					system.timestamp = time.Now()
```
tiwatsuka, line 251: Timestamp refreshing is required only when the energy is calculated. It's not here.
```suggestion
```

pkg/power/platform/source/redfish.go
```diff
@@ -0,0 +1,286 @@
...
+	if rf.systems != nil {
+		power := make(map[string]float64)
+		for _, system := range rf.systems {
+			rf.mutex.Lock()
+			now := time.Now()
+			// calculate the elapsed time since the last power query in seconds
+			elapsed := now.Sub(system.timestamp).Seconds()
+			system.timestamp = time.Now()
+			klog.V(5).Infof("power info: %+v\n", system)
+			power[system.system] = float64(system.consumedWatts * 1000 * int(elapsed)) // convert to mW
```
tiwatsuka, line 273:
```suggestion
			power[system.system] = float64(system.consumedWatts*1000) * elapsed // convert to mW
```

tiwatsuka (commented):
I made some suggestions to get energy correctly.

pkg/power/platform/source/redfish_util.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &system, nil
+}
+
+func getRedfishPower(access RedfishAccessInfo, system string) (*RedfishPowerModel, error) {
+	var power RedfishPowerModel
+	err := getRedfishModel(access, "/redfish/v1/Chassis/"+system+"/Power%23/PowerControl", &power)
```
tiwatsuka, line 78: Is URL encoding required? It's not work in my environment and using '#' is working fine.

pkg/power/platform/source/redfish_util.go
```diff
@@ -0,0 +1,84 @@
...
+	"time"
+)
+
+func getRedfishModel(access RedfishAccessInfo, endpoint string, model interface{}) error {
+	username := access.Username
+	password := access.Password
+	host := access.Host
+
+	// Create a HTTP client and set up the basic authentication header
+	client := &http.Client{}
```
tiwatsuka, line 33: Doesn't it required to set `TLSClientConfig` to `&tls.Config{InsecureSkipVerify: true}`?

pkg/power/platform/source/redfish.go
rootfs: oh, i see. This function is get energy (mJ), instead of power. The comment needs to be fixed but the formula is right. See the ACPI function [here](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/power/platform/source/acpi.go#L166)

tiwatsuka: I know the formula is right. I mean cast the elapsed time from float to int might lose energy of around one second if it was 2.999... .

pkg/power/platform/source/redfish_util.go
YaSuenag: It is better to set `InsecureSkipVerify` as an option (e.g. environment variable) because it is important to secure access.

rootfs: @tiwatsuka @YaSuenag sounds good. can you suggest a change and I'll merge it here. thanks

rootfs:
@tiwatsuka @YaSuenag thank you for reviewing. My setup will be ready later this week and run the tests on it. Will keep you updated.

pkg/power/platform/source/redfish_util.go
tiwatsuka, line 33:
```suggestion
	transport := &http.Transport{}
	if config.GetRedfishSkipSSLVerify() {
		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
	}
	client := &http.Client{
		Transport: transport,
	}
```

tiwatsuka (commented):
Add config to skip SSL verification. 
@YaSuenag Does this satisfy your request?

YaSuenag:
Thanks @tiwatsuka to create the change, and @rootfs for merging! I completely agree with your change.

BTW new parameters / environment variables will be documented? Will it happen on https://github.com/sustainable-computing-io/kepler-doc ?

rootfs:
@YaSuenag yes, kepler-doc will have the redfish document. If you or your team can come up with more tests and instructions, please share with the community over there. Thanks!

pkg/power/platform/source/redfish_util.go
```diff
@@ -63,19 +74,33 @@ func getRedfishModel(access RedfishAccessInfo, endpoint string, model interface{
...
-	err = json.NewDecoder(resp.Body).Decode(model)
-	if err != nil {
-		return err
+	decErr := dec.Decode(model)
+
+	// process the error and only return significant ones
+	if decErr != nil {
+		if strings.HasPrefix(decErr.Error(), "json: unknown field ") {
+			// ignore unknown field error
+			fieldName := strings.TrimPrefix(decErr.Error(), "json: unknown field ")
```
rootfs, line 88: Dell returns some fields that are not defined in the power model, ignore them here

rootfs:
@marceloamaral redfish is now platform energy. PTAL

Here is the output:

image

rootfs:
@YaSuenag @tiwatsuka the kepler-doc update will be at this PR https://github.com/sustainable-computing-io/kepler-doc/pull/74

husky-parul (approved):
/lgtm

PR #736, opened 2023-06-14
fix: podman failing to push to local registry

sthaha:
When using podman to bring up a local cluster, using the command below,

```
export CLUSTER_PROVIDER=kind
make cluster-sync IMAGE_REPO="localhost:5001" IMAGE_TAG="devel"
```
it fails with the following error when pushing kepler image to local
registry.

```
Error: trying to reuse blob sha256:119391 at destination:
  pinging container registry localhost:5001:
  Get "https://localhost:5001/v2/": http: server gave HTTP response to HTTPS client
make[1]: *** [Makefile:128: push-image] Error 125
```

This patch fixes it by allowing push options to be specified. E.g. when
using podman, you can turn off tls-verify by ...

```
export CLUSTER_PROVIDER=kind
make cluster-sync IMAGE_REPO="localhost:5001" IMAGE_TAG="devel" CTR_CMD_PUSH_OPTIONS=--tls-verify=false
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,13 @@ endif
 
 ifdef CTR_CMD
 	CTR_CMD := $(CTR_CMD)
-else 
+else
 	CTR_CMD :=$(or $(shell which podman 2>/dev/null), $(shell which docker 2>/dev/null))
 endif
+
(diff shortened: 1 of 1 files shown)


marceloamaral (approved):
Makes sense to me.
/lgtm

PR #759, opened 2023-07-02
Add support for Microshift in CI

vprashar2929:
Since we have added support for having microshift as a cluster in the [local-dev-cluster](https://github.com/sustainable-computing-io/local-dev-cluster/pull/18) repo and updated the [kepler-action](https://github.com/sustainable-computing-io/kepler-action/pull/48) as well This PR is in continuation to add support for running Kepler integration tests on microshift cluster.

**Dependency:**
There is an open [PR](https://github.com/sustainable-computing-io/kepler-action/pull/51) in Kepler-action to update some recent [changes](https://github.com/sustainable-computing-io/local-dev-cluster/pull/20) done to local-dev-cluster repo.

CC: @SamYuan1990 @rootfs

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -8,41 +8,31 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        kube_provider: [kind]
+        cluster_provider: [kind,microshift]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
           go-version: 1.18
 
-      - name: use kepler action for kind cluster build
-        uses: sustainable-computing-io/kepler-action@v0.0.0
+      - name: use kepler action for cluster build
+        uses: sustainable-computing-io/kepler-action@main
         with:
-          runningBranch: kind
+          runningBranch: ${{matrix.cluster_provider}}
+          local_dev_cluster_version: main
+          cluster_provider: ${{matrix.cluster_provider}}
 
       - name: simple test - deploy kepler
         run: make cluster-sync
         env:
-            CLUSTER_PROVIDER: ${{matrix.kube_provider}}
-            IMAGE_REPO: "localhost:5001"
-            IMAGE_TAG: "devel"
-            CTR_CMD: docker
-            CI_ONLY: "true"
+          CLUSTER_PROVIDER: ${{matrix.cluster_provider}}
+          IMAGE_REPO: "localhost:5001"
+          IMAGE_TAG: "devel"
+          CTR_CMD: docker
+          CI_ONLY: "true"
 
-      - name: test if kepler is still alive
-        run: |
-          sleep 60
-          kubectl logs $(kubectl -n kepler get pods -o name) -n kepler
-          kubectl get all -n kepler
-      
       - name: run integration_test
-        run: |
-          docker ps -a
-          mkdir -p /tmp/.kube
-          kind get kubeconfig --name=kind > /tmp/.kube/config
-          while true; do kubectl port-forward --address localhost -n kepler service/kepler-exporter 9102:9102; done &
-          kubectl logs -n kepler daemonset/kepler-exporter
-          kubectl get pods -n kepler -o yaml
-          go test ./e2e/... --tags bcc -v --race --bench=. -cover --count=1 --vet=all
+        run: make integration-test
         env:
-            kepler_address: localhost:9102
+          CLUSTER_PROVIDER: ${{matrix.cluster_provider}}
+          kepler_address: localhost:9102
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -3,36 +3,94 @@ name: Integration test
 on:
   pull_request:
 
+env:
+  GO_VERSION: "1.18"
+  OUTPUT_DIR: "_output/"
+  FILE_NAME: "kepler.tar.gz"
+
 jobs:
+  build-kepler:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: checkout source
+        uses: actions/checkout@v3
+
+      - name: install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{env.GO_VERSION}}
+
+      - name: build manifest
+        run: make build-manifest OPTS="CI_DEPLOY"
+
+      - name: build and export Kepler image
+        run: |
+          make build_containerized
+          make save-image
+        env:
+          IMAGE_REPO: "localhost:5001"
+          IMAGE_TAG: "devel"
+          CTR_CMD: docker
+          IMAGE_OUTPUT_PATH: ${{env.OUTPUT_DIR}}${{env.FILE_NAME}}
+
+      - name: save Kepler image as artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: kepler
+          path: ${{env.OUTPUT_DIR}}${{env.FILE_NAME}}
+
   integration_test:
+    needs: [build-kepler]
     runs-on: ubuntu-20.04
     strategy:
       matrix:
         cluster_provider: [kind,microshift]
     steps:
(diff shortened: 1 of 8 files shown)


rootfs:
@vprashar2929 new microshift integration test failed:

https://github.com/sustainable-computing-io/kepler/actions/runs/5437485041/jobs/9888072090?pr=759
```
+ for opt in ${DEPLOY_OPTIONS}
+ unset CI_DEPLOY
+ for opt in ${DEPLOY_OPTIONS}
+ unset OPENSHIFT_DEPLOY
+ echo 'Done ./hack/build-manifest.sh'
make[1]: Leaving directory '/home/runner/work/kepler/kepler'
Error response from daemon: No such container: microshift
make: *** [Makefile:263: cluster-sync] Error 1
Error: Process completed with exit code 2.
```

vprashar2929:
Yes, currently it will fail as there is an open [PR](https://github.com/sustainable-computing-io/kepler-action/pull/51)  in Kepler-action for updating the logic. Once that is merged then CI will work fine

.github/workflows/integration_test.yml
```diff
@@ -8,41 +8,31 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
           go-version: 1.18
 
-      - name: use kepler action for kind cluster build
-        uses: sustainable-computing-io/kepler-action@v0.0.0
+      - name: use kepler action for cluster build
+        uses: sustainable-computing-io/kepler-action@main
```
SamYuan1990, line 19: @rootfs, @vprashar2929  for dev branch, we should keep this to main, for release we should use a release branch. Hence before we bump up version for kepler-action and local-dev-cluster, we should keep it as v0.0.0 and keep this PR open for review.

.github/workflows/integration_test.yml
```diff
@@ -8,41 +8,31 @@ jobs:
...
-            IMAGE_TAG: "devel"
-            CTR_CMD: docker
-            CI_ONLY: "true"
+          CLUSTER_PROVIDER: ${{matrix.cluster_provider}}
+          IMAGE_REPO: "localhost:5001"
+          IMAGE_TAG: "devel"
+          CTR_CMD: docker
+          CI_ONLY: "true"
 
-      - name: test if kepler is still alive
```
SamYuan1990, line 32: I hope we can keep this part of check. As for historical reason,  kepler is not started correctly, hence this part is to check if kepler pod started.

hack/cluster-deploy.sh
```diff
@@ -25,38 +25,27 @@ CLUSTER_PROVIDER=${CLUSTER_PROVIDER:-kubernetes}
...
+    then
+        $CTR_CMD image prune -a -f || true && $CTR_CMD restart microshift
+        wait_microshift_up
+    fi
+
     [ ! -d "${MANIFESTS_OUT_DIR}" ] && echo "Directory ${MANIFESTS_OUT_DIR} DOES NOT exists. Run make generate first."
     
+    if [ "$CLUSTER_PROVIDER" == "microshift" ]
+    then
+        kubectl label node --all sustainable-computing.io/kepler=''
```
SamYuan1990, line 38: we should considering change https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/exporter/patch/patch-ci.yaml

.github/workflows/integration_test.yml
SamYuan1990: hence I will suggestion as bump up version in another PR.

hack/cluster-up.sh
```diff
@@ -20,7 +20,7 @@ if [ -d local-dev-cluster ]
 then
 	echo "use local local-dev-cluster"
 else
-    echo "download local-dev-cluster with version v0.0.0"
-	git clone -b v0.0.0 https://github.com/sustainable-computing-io/local-dev-cluster.git --depth=1
+    echo "download local-dev-cluster with latest version"
+	git clone https://github.com/sustainable-computing-io/local-dev-cluster.git --depth=1
```
SamYuan1990, line 24: related with version bump up. Same comments here for version control.

SamYuan1990 (changes requested):
I will suggestion as bump up version in another PR.
As split this PR into 2 parts.
1. as adjust for logic, as add new verify.sh shell
2. bump up version to add microshift support.

SamYuan1990:
to go with a new release or not for https://github.com/sustainable-computing-io/local-dev-cluster as k3s support
to go with a new release or not for https://github.com/sustainable-computing-io/kepler-action/issues/50 as Prow support
@rootfs , @vprashar2929 , @jichenjc , @marceloamaral 
I suppose we should have a vote as an agreement before we merge this PR.
and version bump up should also apply for helm and operator.
cc: @jiangphcn , @husky-parul

.github/workflows/integration_test.yml
vprashar2929: Sure bumping up the version in another PR sounds good. Till that time I will keep it as v0.0.0

.github/workflows/integration_test.yml
vprashar2929: Yes, @SamYuan1990 I have added another [script](https://github.com/sustainable-computing-io/kepler/blob/92722eb34559bd48f75edc19ac32018fc1fafa55/hack/verify.sh) inside the hack folder which will test if kepler is still alive as well as launch e2e tests. It is called from [here](https://github.com/sustainable-computing-io/kepler/blob/92722eb34559bd48f75edc19ac32018fc1fafa55/hack/cluster-deploy.sh#L48)

SamYuan1990: ok

vprashar2929:
One more thing I would like to add is when I was working on adding micro shift to CI I realized that building Kepler image takes more than 10GB of disk space(Dependent Images) on GH runner and we are using the same runner to create cluster, build image and deploy Kepler. Since the default [disk size](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) for GH runner is around 14GB. This make's difficult for the running cluster to run properly as I have seen the case where the cluster had to evict the running pods in order to reclaim the disk space because of `NodeHasDiskPressure`. To resolve this issue as of now I have added a [step](https://github.com/sustainable-computing-io/kepler/blob/92722eb34559bd48f75edc19ac32018fc1fafa55/hack/cluster-sync.sh#L43) to stop the cluster build Kepler image push it to the local registry and then [clean the images](https://github.com/sustainable-computing-io/kepler/blob/92722eb34559bd48f75edc19ac32018fc1fafa55/hack/cluster-deploy.sh#L30) from the runner so that there is some disk space available and then restart the cluster. I am not a fan of this approach but I was thinking can we build the Kepler image on a separate runner upload it as an artifact and then download it on a different runner where we will launch our cluster and push it to the local registry. After that follow the normal process to deploy Kepler and trigger e2e tests?

@rootfs @SamYuan1990 @marceloamaral Any thoughts on this?

rootfs:
@vprashar2929 that sounds a good idea. Would you like to have multiple jobs, kind/microshift/and potentially k3s in different jobs? How about we divide the flow into image build and image test? The image build job builds and uploads kepler image artifacts and the image test steps download the image and run kind/microshift/k3s in their runners.

vprashar2929:
Yes, @rootfs  I am also thinking of the same flow to have a job that builds and uploads Kepler image artifacts and another which uses our Kepler-action to bring up kind/microshift/k3s clusters, deploy Kepler, and run e2e test's on it.

rootfs (approved):

rootfs:
@SamYuan1990 is the request done?

vprashar2929:
The current integration test will fail as I have reverted the bump up's in version for sustainable-computing-io/kepler-action. The v0.0.0 doesn't contain the functionality to launch microshift cluster.

rootfs:
@vprashar2929 got a CI failure

image

PR #762, opened 2023-07-04
CI: add path ignore to ensure rfc will not trigger CI.

SamYuan1990:
ref https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore
to add path ignore settings in CI to avoid rfc trigger an CI build.


SamYuan1990:
I will tested with my own fork to see if it works.

rootfs (approved):

Issue #767, opened 2023-07-06
Unable to access Kepler metrics from prometheus


edoblette:
### What happened?

Unable to export metrics to prometheus. Kepler appears to be `"Dropped" `from prometheus's `/service-discovery` page.
I'm reporting this problem from previous issue @ [How to select the tag of kepler-helm-chart to install Kepler?](url) from Kepler Helm deployement, following the initiative of @rootfs and @LAI-chuchi

However, metrics are available on port 9102 of the kepler pod at http://<cluster-url>:9102/metrics.

I'm using Kind for my K8s cluster, with Prometheus and Grafana already deployed. I also use Cilium without any problems.

### What did you expect to happen?

Get Kepler metrics enable on my prometheus queries dashbord.

### How can we reproduce it (as minimally and precisely as possible)?

Linux 22.04
Kernel: 5.5.0-050500-generic
kind version 0.18.0*

### Anything else we need to know?

Log from kepler pod:
<details>

```
I0706 13:50:55.221651       1 gpu.go:46] Failed to init nvml, err: could not init nvml: error opening libnvidia-ml.so.1: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
I0706 13:50:55.416231       1 acpi.go:67] Could not find any ACPI power meter path. Is it a VM?
I0706 13:50:55.714345       1 exporter.go:151] Kepler running on version: 364c44f
I0706 13:50:55.714388       1 config.go:212] using gCgroup ID in the BPF program: true
I0706 13:50:55.714445       1 config.go:214] kernel version: 5.5
I0706 13:50:55.714484       1 exporter.go:171] EnabledBPFBatchDelete: true
I0706 13:50:55.714602       1 power.go:53] use sysfs to obtain power
I0706 13:51:03.936389       1 exporter.go:184] Initializing the GPU collector
I0706 13:51:03.937016       1 watcher.go:67] Using in cluster k8s config
I0706 13:51:30.137397       1 bcc_attacher.go:171] Successfully load eBPF module with option: [-DMAP_SIZE=10240 -DNUM_CPUS=16]
I0706 13:51:30.717792       1 exporter.go:228] Started Kepler in 35.003475726s
I0706 13:51:33.720063       1 container_hc_collector.go:130] failed to get bpf table elemets, err: failed to batch get: invalid argument
I0706 13:51:33.728652       1 container_hc_collector.go:211] resetting EnabledBPFBatchDelete to false
[...]

PR #768, opened 2023-07-06
bpf: ignore IsNotExist error when deleting from ebpf map

rootfs:
this the error message from #767


rootfs:
@jichenjc test passed

jichenjc (approved):

PR #752, opened 2023-06-29
add an option to use libbpf attacher

sunya-ch:
According to issue https://github.com/sustainable-computing-io/kepler/issues/716, this PR add an option to attach BPF program using pre-compiled libbpf object instead of bcc. This PR removes dependency of kernel header library as well as bcc. 

- To generate a precompiled object file, run `make genlibbpf`. It will compile from `bpfassets/kepler.bpf.c` and its header. The precompiled object file will be in `bpfassets/$(name -m)_kepler.bpf.o`. This PR included the common one, `x86_64_kepler.bpf.o`. 
- To run go command, `LIBBPF_HEADERS` and `libbpf.a` needed to be set for cgo flag as seen in `GOENV` in `Makefile`
- This PR also refactor the attacher module as following.
    - define `config.UseLibBPFAttacher` to be configured by config map or environment variable whether to use libbpf or not (default: false)
    - define 4 core functions to be called by other module: Attach, Detach, CollectProcesses, CollectCPUFreq and use `config.UseLibBPFAttacher` to select between `bcc_attacher` or `libbpf_attacher`
         - here, this PR moves code from `node_energy_collector` (for CPUFreq) and from `container_hc_collector` (for CollectProcesses) to attacher module.
    - `libbpf_attacher` is loaded by build tag `libbpf` in the same manner as `bcc` tag.
- go version 1.18.10 is required to fix cgo issue in libbpf function. Need update in `Dockerfile.builder`.
- libbpf dependencies in Dockerfile should be included in base image 
```
RUN yum install -y clang llvm llvm-devel zlib-devel elfutils-libelf-devel make libbpf
RUN curl -LO https://github.com/libbpf/libbpf/archive/refs/tags/v0.5.0.tar.gz; mkdir -p /usr/local; tar -C /usr/local -xvzf v0.5.0.tar.gz; rm -f v0.5.0.tar.gz; cd /usr/locallibbpf-0.5.0/src && OBJDIR=/usr/lib/$(uname -m)-linux-gnu make install
```

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

PR #776, opened 2023-07-10
explicitly set default make target

sunya-ch:
Thanks to report from @jichenjc  in https://github.com/sustainable-computing-io/kepler/pull/752#issuecomment-1624791332, this PR fixes the logic change after merging the PR https://github.com/sustainable-computing-io/kepler/pull/761 by explicitly setting the default target to `kepler` before include `Makefile` in bpfassets. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+all: kepler
+
 include bpfassets/libbpf/Makefile
 
 ### env define ###


jichenjc (approved):

SamYuan1990 (approved):
approve ref https://github.com/sustainable-computing-io/kepler/blob/main/Makefile#L153

Issue #784, opened 2023-07-11
integration test: build manifest doesn't build with correct image name


sunya-ch:
### What happened?

On integration test according to this commit https://github.com/sustainable-computing-io/kepler/commit/a94cee02af53da0134da52c92b59e985b9378aeb, the manifest deployed in the cluster to test is using official image not the locally-built image. 

### What did you expect to happen?

Need to set env for building manifest using image from local registry with dev tag.

### How can we reproduce it (as minimally and precisely as possible)?

Run integration_test CI

### Kepler image tag

<details>

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

rootfs:
fixed in #785

PR #785, opened 2023-07-11
set env for build-manifest on e2e test

sunya-ch:
This PR adds env to build manifest to fix the issue https://github.com/sustainable-computing-io/kepler/issues/784.

However, I'm not sure whether which image should be used in manifest in `kepler.tar.gz` for `build-kepler` job. 
Is it expected to be official image or locally-built image? @vprashar2929  Could you please confirm this?

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -26,6 +26,10 @@ jobs:
 
       - name: build manifest
         run: make build-manifest OPTS="CI_DEPLOY"
+        env:
+          IMAGE_REPO: "localhost:5001"
+          IMAGE_TAG: "devel"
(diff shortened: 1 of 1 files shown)


vprashar2929:
Yes @sunya-ch my bad the env was not set properly due to which it takes the latest Kepler image when running integration-tests. 

Nice find.👍

SamYuan1990 (approved):
LGTM, but are those default settings?

vprashar2929:
Yes, these are the default values of env variables that are used. Here is the [link](https://github.com/vprashar2929/kepler/actions/runs/5520469895/jobs/10067372222#step:9:115) tested workflow

PR #788, opened 2023-07-13
watcher: wait pod status to be container ready to parse container id

### LINKED ISSUE #787: API server pod watcher missing event, causing missed pod metrics
### What happened?

The kube apiserver watcher missed pod status change. As a result, pods are not updated and mistakenly classified as system pods

I added some tracing along [this code block](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/kubernetes/watcher.go#L153-L157)

```go
for _, condition := range pod.Status.Conditions {
			klog.V(5).Infof("Pod %s %s status %s %s", pod.Name, pod.Namespace, condition.Type, condition.Status)
			if condition.Type != k8sv1.ContainersReady && condition.Status != k8sv1.ConditionTrue {
				continue
			}
			klog.V(5).Infof("Pod %s %s is ready with %d container statuses, %d init container status, %d ephemeral statues",
				pod.Name, pod.Namespace, len(pod.Status.ContainerStatuses), len(pod.Status.InitContainerStatuses), len(pod.Status.EphemeralContainerStatuses))
```

This is what is in the log:
```console
I0713 00:03:30.736937       1 watcher.go:154] Pod random-576f489b67-qq4hk cpu-load status PodScheduled True
I0713 00:03:30.736957       1 watcher.go:158] Pod random-576f489b67-qq4hk cpu-load is ready with 0 container statuses, 0 init container status, 0 ephemeral statues
```

The `ContainerReady` status is never received, so parsing the pod status doesn't return pod name and container ID mapping. As a result, the pod is never added to metrics cache and gets resolved.

### How can we reproduce it (as minimally and precisely as possible)?

Running Kepler on ROSA seems to produce this consistently.

### Anything else we need to know?

_No response_

### Kepler image tag

<details>
latest
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>
AWS, ROSA
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

rootfs:
fix #787


jichenjc:
lgtm, @marceloamaral may help check the watcher logic?

pkg/kubernetes/watcher.go
```diff
@@ -151,17 +151,20 @@ func (w *ObjListWatcher) handleUpdate(obj interface{}) {
...
-			w.Mx.Lock()
-			err1 := w.fillInfo(pod, pod.Status.ContainerStatuses)
-			err2 := w.fillInfo(pod, pod.Status.InitContainerStatuses)
-			err3 := w.fillInfo(pod, pod.Status.EphemeralContainerStatuses)
-			w.Mx.Unlock()
-			// only add pod to cache if all containers successfully added to map
-			if err1 == nil && err2 == nil && err3 == nil {
-				managedPods[podID] = true
+			klog.V(5).Infof("Pod %s %s status %s %s", pod.Name, pod.Namespace, condition.Type, condition.Status)
+			if condition.Type == k8sv1.ContainersReady {
```
marceloamaral, line 155: Is there any reason why you removed the `condition.Status != k8sv1.ConditionTrue`?

rootfs: when `ContainerReady` status is set, if the `Status` is `True`, then the Pods already initialized with ContainerIDs and `ContaienrMetrics` saves them. If the `Status` is `False`, then the `ContainerMetrics` is reset. 

But now I have another question, should we trust `managedPods` cache? What if the pod is updated with a new sidecar? The sidecar will not be checked since the podUID is already in the `managedPds` cache.

marceloamaral: Good point! Maybe we should not skip the events based on the managedPods....

rootfs: sounds good, i'll update

rootfs: @marceloamaral `managedPods` cache removed, all pods are to be evaluated.

husky-parul (approved):
lgtm

Issue #797, opened 2023-07-19
libbpf module attached but cannot get counter metrics


sunya-ch:
### What happened?

Successfully attach the libbpf module but cannot get counter metrics.

```
I0719 03:14:47.245575       1 bpf_perf.go:115] LibbpfBuilt: true, BccBuilt: false
I0719 03:14:47.325568       1 libbpf_attacher.go:146] Successfully load eBPF module from libbpf object
I0719 03:14:47.333567       1 node_energy_collector.go:55] System energy collection is supported
```
However, it cannot get value because the Counters variable is not properly set.
```
I0719 03:14:47.596752       1 container_metric.go:190] cannot extract: cpu_cycles
I0719 03:14:47.596758       1 container_metric.go:190] cannot extract: cpu_instr
```

### What did you expect to happen?

```
	Idle ePkg (mJ): 17914 (17914) (eCore: 0 (0) eDram: 19144 (19144) eUncore: 0 (0)) eGPU (mJ): 0 (0) eOther (mJ): 47441 (47441)
	CPUTime:  272739617 (272739617)
	NetTX IRQ: 1 (1)
	NetRX IRQ: 297 (297)
	Block IRQ: 2 (2)
	counters: map[cache_miss:819986 (819986) cpu_cycles:1329073038 (1329073038) cpu_instr:640556706 (640556706) cpu_ref_cycles:0 (0)]
```

### How can we reproduce it (as minimally and precisely as possible)?

deploy kepler with latest-libbpf tag

### Kepler image tag

<details>

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

[...]

PR #798, opened 2023-07-19
set counters before attach

sunya-ch:
This PR is mainly to solve the issue https://github.com/sustainable-computing-io/kepler/issues/797.

In addition to that, this PR also
-  apply `config.SetKernelSourceDir` only when bcc is built to avoid confusion when using libbpf from the log `failed to set kernel source dir`.
-  add more information on cgroup failure log on stat handler start error
-  update bpf.o file to latest libbpf program
-  fix kind provider support by adding patch-kind.yaml to fix https://github.com/sustainable-computing-io/kepler/issues/799


rootfs (approved):

rootfs:
CI is flaky, let's pick up the fix first.

manifests/config/exporter/patch/patch-kind.yaml
```diff
@@ -0,0 +1,15 @@
...
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: kepler-exporter
+      volumes:
+      - name: proc
+        hostPath:
+          path: /proc-host
```
sthaha, line 14: @sunya-ch could you please help me understand why we are patching this for kind cluster?

Issue #814, opened 2023-07-21
Should have cluster cleanup job in e2e test. 


jiere:
### What would you like to be added?

Current Kepler e2e integration test has no cleanup job, so after test, both the cloud-provider cluster and the kepler daemonset remain on the runner.

When new CI jobs are scheduled to the same runner, the cluster setup job will fail, still use the earlier cluster to run test, this may bring side effect.

See job error logs [example](https://github.com/jiere/kepler/actions/runs/5583266592/job/15120269564#step:6:567).

### Why is this needed?

Add cluster cleanup job in GHA e2e test workflow, since both Kind and microshift have supported the cluster-down ops in local-dev-cluster.

marceloamaral:
Good point, the scripts is actually missing the `make cluster down`
We had it before and should have it back again.

Issue #809, opened 2023-07-21
Kepler CI on BM runner always fails at integration_test job.


jiere:
### What happened?

When run current GHA integration_test job on self-hosted BM runner, the `integration_test` job always fails due to `kubectl portforward` failure like this:

```
E0703 10:28:13.885211 3604726 portforward.go:407] an error occurred forwarding 9102 -> 9102: error forwarding port 9102 to pod 8bb7390de9a310c3e1f6ae708462d18e2a5534a8f9864f4a5d942e467b1164fd, uid : failed to execute portforward in network namespace "/var/run/netns/cni-d766e3ad-a513-cbf7-244a-aaa3348c29b6": failed to connect to localhost:9102 inside namespace "8bb7390de9a310c3e1f6ae708462d18e2a5534a8f9864f4a5d942e467b1164fd", IPv4: dial tcp4 127.0.0.1:9102: connect: connection refused IPv6 dial tcp6 [::1]:9102: connect: connection refused

E0703 10:28:13.885374 3604726 portforward.go:233] lost connection to pod
```

After debugging, found the direct RCA for such issue is: we cannot run `kubectl portforward` command before Kepler completely boot success.
When we could see such words in Kepler log: `Started Kepler in`, then the portforwad command works well.

Due to #808 , we should check if the issues there are expected, then we should fix here to wait until Kepler boot success.

### What did you expect to happen?

Run CI on BM runner successfully.

### How can we reproduce it (as minimally and precisely as possible)?

Need to fork kepler repo currently and then self-hosted a Ubuntu based BM runner to run the integration test.

### Kepler image tag

<details>
main latest code.
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>
KIND
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

[...]

PR #816, opened 2023-07-21
CI enhancement: Add cleanup job, add necessary wait for Kepler ready

jiere:
PR for #809 and #814 .

### DIFF
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -56,6 +56,7 @@ function intergration_test() {
         tags="bcc"
     fi
     echo TAGS=$tags
+    kepler_ready=false
     $CTR_CMD ps -a
     mkdir -p /tmp/.kube
     if [ "$CLUSTER_PROVIDER" == "microshift" ]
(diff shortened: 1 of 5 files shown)


hack/verify.sh
```diff
@@ -64,6 +65,16 @@ function intergration_test() {
     else
         kind get kubeconfig --name=kind > /tmp/.kube/config
     fi
+    until ${kepler_ready} ; do
+        kubectl logs $(kubectl -n kepler get pods -o name) -n kepler > kepler.log
+	if [ `grep -c "Started Kepler in" kepler.log` -ne '0' ]; then
```
rootfs, line 70: can you also check if `Successfully load eBPF module` is there too? ebpf could fail to load yet Kepler can still start

jiere: One question, if ebpf really fail, how could we deal with it here? echo errmsg and exit 1?
Per current code logic, will be endless loop.

rootfs (approved):

rootfs (approved):

PR #823, opened 2023-07-24
ebpf: reducing ebpf call overhead by using sampling instead of tracing every calls

### LINKED ISSUE #668: eBPF scalability improvement
- [ ] #665 
- [x] #671
- [x] tracking ebpf performance stats, thanks to [this ](https://ebpf.io/summit-2020-slides/eBPF_Summit_2020-Lightning-Bryce_Kahle-How_and_When_You_Should_Measure_CPU_Overhead_of_eBPF_Programs.pdf)

rootfs:
fix #668 

this reduces the bpf call overhead from 1352ns to 99ns

# without this fix
```console
# sysctl -w kernel.bpf_stats_enabled=1
# bpftool prog show |grep kepler_trace |awk '{print $(NF-2)/$NF}'
1352.07
```
# with this fix
```console
# sysctl -w kernel.bpf_stats_enabled=1
# bpftool prog show |grep kepler_trace |awk '{print $(NF-2)/$NF}'
99.0167
```

- [x] make bcc work 
- [x] make libbpf able to set sampling rate by calling `InitilizeGlobalVar`
- [ ] evaluate trade-off between overhead and prediction accuracy


rootfs:
depend on #824

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -44,6 +44,9 @@ BPF_ARRAY(cache_miss, u64, NUM_CPUS);
 // cpu freq counters
 BPF_ARRAY(cpu_freq_array, u32, NUM_CPUS);
 
+int sample_rate = 1000;
+int counter = 1000;
```
marceloamaral, line 48: Shouldn't it be:
```
int sample_rate = SAMPLE_RATE;
int counter = SAMPLE_RATE;
```

bpfassets/perf_event/perf_event.c
```diff
@@ -71,6 +71,11 @@ BPF_ARRAY(cache_miss, u64, NUM_CPUS);
 // cpu freq counters
 BPF_ARRAY(cpu_freq_array, u32, NUM_CPUS);
 
+#ifndef SAMPLE_RATE
+#define SAMPLE_RATE 1000
```
marceloamaral, line 75: Isn't skipping 1000 sample to extreme?
Did you try 10 and 100?

pkg/config/config.go
```diff
@@ -80,6 +80,7 @@ var (
 	BindAddressKey               = "BIND_ADDRESS"
 	CPUArchOverride              = getConfig("CPU_ARCH_OVERRIDE", "")
 	MaxLookupRetry               = getIntConfig("MAX_LOOKUP_RETRY", defaultMaxLookupRetry)
+	BPFSampleRate                = getIntConfig("BPF_SAMPLE_RATE", 1000)
```
marceloamaral, line 83: Shouldn't we use smaller values?

marceloamaral (changes requested):
Looks great, I only concerned about the sample rate.

bpfassets/perf_event/perf_event.c
rootfs: yes, 10 or 100 does have some reduction from 1000ns to 300ns.

bpfassets/libbpf/src/kepler.bpf.c
rootfs: libbpf is pre-compiled, we cannot use compilation flags. In this case, we have to global variable to set it. I opened #824 so I can use `InitGlobalVar` function

pkg/config/config.go
rootfs: sure, let's have some stats first so users know what to choose and start from small values as default.

bpfassets/libbpf/src/kepler.bpf.c
marceloamaral: Got it!
If we need to hard code now, let's use a smaller value.

rootfs:
# kepler command
```bash
BPF_SAMPLE_RATE=1000 _output/bin/linux_amd64/kepler
```
# ebpf calculation
```
bpftool prog show |grep kepler_trace |awk '{print $(NF-2)/$NF}'
```

# result
| sample frequency  | per call time (ns)   |
|---|---|
|  0 | 1239  |
|  1 |  797 |
|  10 | 300  |
|  50 | 152  |
|  100 | 129  |
|  1000 | 93  |
```

**Comment by eklee15** [comment]
Thanks for sharing it. This is great!

**Comment by marceloamaral** [comment]
@rootfs, for now, let's set the default value to 10 since we're uncertain about the consequences of skipping samples. Moreover, using 10 seems to bring significant improvements. Once we conduct further analysis, we can increasing this value.

**Comment by rootfs** [comment]
@marceloamaral sure, let default to 10. One more question, when we sample the ebpf calls, should we also extrapolate the metrics (cpu time, cpu instructions, etc) as well?

**Comment by eklee15** [comment]
Just created a discussion 
https://github.com/sustainable-computing-io/kepler/discussions/836

**Comment by rootfs** [comment]
@eklee15 @marceloamaral @sunya-ch Let's disable sampling for now until we have a resolution on #836

File: bpfassets/perf_event/perf_event.c [hunk 4]
```diff
@@ -192,6 +197,19 @@ static inline u64 get_on_cpu_avg_freq(u32 *cpu_id, u64 on_cpu_cycles_delta, u64
...
+    u32 initial = SAMPLE_RATE, *sample_counter_value, sample_counter_key = 1234;
+    // only do sampling if sample rate is set
+    if (initial != 0) {
+        sample_counter_value = sample_rate.lookup_or_try_init(&sample_counter_key, &initial);
+        if (sample_counter_value > 0) {
+            if (*sample_counter_value > 0) {
+                (*sample_counter_value)--;
+                return 0;
+            }
+        }
```
sthaha, line 209: New to this so please excuse if this suggestion looks stupid . Would this work? 


```suggestion
        if (sample_counter_value  && *sample_counter_value > 0) {
                (*sample_counter_value)--;
                return 0;
        }
```

pkg/collector/metric/container_metric.go
```diff
@@ -274,6 +274,9 @@ func (c *ContainerMetrics) UpdateCgroupMetrics() error {
 }
 
 func (c *ContainerMetrics) GetDynEnergyStat(component string) (energyStat *types.UInt64Stat) {
+	if c == nil {
+		return
+	}
```
sthaha, line 279: This shouldn't be done we must expect the receiver to be initialised at all time. If it is not the case, then it is a programming error and must panic. By returning we are only shadowing a logical error.

pkg/model/estimator/local/ratio_process.go
```diff
@@ -74,6 +74,10 @@ func getProcessResUsage(process *collector_metric.ProcessMetrics, usageMetric st
 // UpdateProcessComponentEnergyByRatioPowerModel calculates the process energy consumption based on the energy consumption of the container that contains all the processes
 func UpdateProcessComponentEnergyByRatioPowerModel(processMetrics map[uint64]*collector_metric.ProcessMetrics, containerMetrics *collector_metric.ContainerMetrics, component, usageMetric string, wg *sync.WaitGroup) {
 	defer wg.Done()
+	if containerMetrics == nil || processMetrics == nil {
+		klog.V(5).Infoln("containerMetrics or processMetrics is nil")
```
sthaha, line 78: How about we return an error if the arguments aren't initialised properly?

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -209,7 +209,10 @@ static inline u64 get_on_cpu_avg_freq(u32 *cpu_id, u64 on_cpu_cycles_delta, u64
 SEC("tracepoint/sched/sched_switch")
 int kepler_trace(struct sched_switch_args *ctx)
 {
+<<<<<<< HEAD
     u32 next_pid = ctx->next_pid; // the new pid that is to be scheduled
+=======
+>>>>>>> 2c38bc2dc7b9aca85374c16c96db555f16784169
```
marceloamaral, line 215: merge left over

marceloamaral (changes requested):
@rootfs, could you also modify the approach for dropping samples? Instead of calculating the percentage, could we implement a method that aggregates a counter and drops samples accordingly?

For instance, after collecting 99 samples, we could skip the next 1 sample, effectively skipping 1% of the total. Then, if we decide to skip 10 samples, it would translate to a 10% reduction. To achieve this, we would need a mechanism to skip 'y' samples after gathering 'x' samples. This would provide us with the flexibility to adjust the dropout rate as needed.

rootfs:
the rebase was not successful, will reopen another PR

PR #829, opened 2023-07-26
Ignore gpu error of process not found

marceloamaral:
Ignore the error when the driver does not find GPU process metrics.


sunya-ch (approved):
/lgtm

Issue #657, opened 2023-04-26
ebpf: batch get value


rootfs:
**Is your feature request related to a problem? Please describe.**
As seen in this cpu profile (after PR #656), bpf lookup in table iterator is not quite efficiency. 

```
(pprof) top20 -cum (all)
Active filters:
   focus=(all)
Showing nodes accounting for 1090ms, 64.50% of 1690ms total
Showing top 20 nodes out of 146
      flat  flat%   sum%        cum   cum%
    1000ms 59.17% 59.17%     1020ms 60.36%  runtime.cgocall
         0     0% 59.17%      810ms 47.93%  github.com/iovisor/gobpf/bcc._Cfunc_bpf_module_create_c_from_string
         0     0% 59.17%      810ms 47.93%  github.com/iovisor/gobpf/bcc.compile
         0     0% 59.17%      810ms 47.93%  github.com/iovisor/gobpf/bcc.newModule
         0     0% 59.17%      810ms 47.93%  github.com/iovisor/gobpf/bcc.newModule.func3
         0     0% 59.17%      310ms 18.34%  github.com/sustainable-computing-io/kepler/pkg/collector.(*Collector).Update
         0     0% 59.17%      310ms 18.34%  github.com/sustainable-computing-io/kepler/pkg/manager.(*CollectorManager).Start.func1
         0     0% 59.17%      280ms 16.57%  github.com/sustainable-computing-io/kepler/pkg/collector.(*Collector).updateBPFMetrics
         0     0% 59.17%      200ms 11.83%  github.com/iovisor/gobpf/bcc.(*TableIterator).Next
         0     0% 59.17%      120ms  7.10%  github.com/iovisor/gobpf/bcc.(*TableIterator).Next.func6
         0     0% 59.17%      120ms  7.10%  github.com/iovisor/gobpf/bcc._C2func_bpf_lookup_elem
         0     0% 59.17%       80ms  4.73%  github.com/iovisor/gobpf/bcc.(*TableIterator).Next.func5
         0     0% 59.17%       80ms  4.73%  github.com/iovisor/gobpf/bcc._C2func_bpf_get_next_key
      30ms  1.78% 60.95%       80ms  4.73%  runtime.mallocgc
      60ms  3.55% 64.50%       60ms  3.55%  syscall.Syscall
         0     0% 64.50%       40ms  2.37%  github.com/sustainable-computing-io/kepler/pkg/bpfassets/attacher.TableDeleteBatch
[...]

PR #831, opened 2023-07-26
libbpf: support batch get and delete values

### LINKED ISSUE #825: Use libbpf batch operation

Current libbpf based bpf code doesn't use batch operation yet, we need to bring it on par with that using bcc

### Why is this needed?

use libbpf batch operation

rootfs:
fix #825 #657

A test image is built [quay.io/sustainable_computing_io/kepler:pr_831](quay.io/sustainable_computing_io/kepler:pr_831)

### DIFF
--- a/pkg/bpfassets/attacher/libbpf_attacher.go
+++ b/pkg/bpfassets/attacher/libbpf_attacher.go
@@ -173,44 +173,14 @@ func libbpfCollectProcess() (processesData []ProcessBPFMetrics, err error) {
 	if err != nil {
 		return
 	}
-	//processKeySize := int(unsafe.Sizeof(uint64Key))
-	iterator := processes.Iterator()
-	var ct ProcessBPFMetrics
-	//valueSize := int(unsafe.Sizeof(ProcessBPFMetrics{}))
-	keys := []uint32{}
-	retry := 0
-	next := iterator.Next()
-	for next {
-		keyBytes := iterator.Key()
-		key := ByteOrder.Uint32(keyBytes)
-		data, getErr := processes.GetValue(unsafe.Pointer(&key))
-		if getErr != nil {
-			retry += 1
-			if retry > maxRetry {
-				klog.V(5).Infof("failed to get data: %v with max retry: %d \n", getErr, maxRetry)
-				next = iterator.Next()
-				retry = 0
-			}
-			continue
-		}
-		getErr = binary.Read(bytes.NewBuffer(data), ByteOrder, &ct)
-		if getErr != nil {
-			klog.V(5).Infof("failed to decode received data: %v\n", getErr)
-			next = iterator.Next()
-			retry = 0
-			continue
-		}
-		if retry > 0 {
-			klog.V(5).Infof("successfully get data with retry=%d \n", retry)
-		}
-		processesData = append(processesData, ct)
-		keys = append(keys, key)
-		next = iterator.Next()
-		retry = 0
+	if ebpfBatchGetAndDelete {
+		processesData, err = libbpfCollectProcessBatch(processes)
 	}
-	for _, key := range keys {
-		// TODO delete keys in batch
-		processes.DeleteKey(unsafe.Pointer(&key))
+	if err == nil {
+		return
+	} else {
+		ebpfBatchGetAndDelete = false
+		processesData, err = libbpfCollectProcessSingle(processes)
 	}
 	return
 }
@@ -302,32 +272,4 @@ func unixClosePerfEvent() {
 		}
 	}
 	PerfEvents = map[string][]int{}
-}
-
-func getCPUCores() int {
-	cores := runtime.NumCPU()
-	if cpu, err := ghw.CPU(); err == nil {
-		// we need to get the number of all CPUs,
-		// so if /proc/cpuinfo is available, we can get the number of all CPUs
-		cores = int(cpu.TotalThreads)
-	}
-	return cores
-
-}
-
-func resizeArrayEntries(name string, size int) error {
-	m, err := libbpfModule.GetMap(name)
-	if err != nil {
-		return err
-	}
-
-	if err = m.Resize(uint32(size)); err != nil {
-		return err
-	}
-
-	if current := m.GetMaxEntries(); current != uint32(size) {
-		return fmt.Errorf("failed to resize map %s, expected %d, returned %d", name, size, current)
-	}
-
-	return nil
 }
--- a/pkg/bpfassets/attacher/libbpf_attacher.go
+++ b/pkg/bpfassets/attacher/libbpf_attacher.go
@@ -273,3 +273,111 @@ func unixClosePerfEvent() {
 	}
 	PerfEvents = map[string][]int{}
 }
+
+func getCPUCores() int {
+	cores := runtime.NumCPU()
+	if cpu, err := ghw.CPU(); err == nil {
+		// we need to get the number of all CPUs,
+		// so if /proc/cpuinfo is available, we can get the number of all CPUs
+		cores = int(cpu.TotalThreads)
+	}
+	return cores
+
+}
+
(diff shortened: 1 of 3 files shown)


rootfs:
The Kepler CPU usage drops from 3% to 0.3%
image

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -162,44 +162,14 @@ func libbpfCollectProcess() (processesData []ProcessBPFMetrics, err error) {
...
+	if ebpfBatchGetAndDelete {
+		processesData, err = libbpfCollectProcessBatch(processes)
 	}
-	for _, key := range keys {
-		// TODO delete keys in batch
-		processes.DeleteKey(unsafe.Pointer(&key))
+	if err == nil {
+		return
+	} else {
+		ebpfBatchGetAndDelete = false
```
jichenjc, line 171: I think we need log here to indicate it's not able to (so something might be wrong) ?

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -299,3 +269,83 @@ func unixClosePerfEvent() {
...
+			klog.V(5).Infof("successfully get data with retry=%d \n", retry)
+		}
+		processesData = append(processesData, ct)
+		keys = append(keys, key)
+		next = iterator.Next()
+		retry = 0
+	}
+	for _, key := range keys {
+		// TODO delete keys in batch
+		processes.DeleteKey(unsafe.Pointer(&key))
```
jichenjc, line 348: sorry if didn't get this correctly, but Delete keys in get process confuse me ..

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -173,44 +173,14 @@ func libbpfCollectProcess() (processesData []ProcessBPFMetrics, err error) {
...
+		processesData, err = libbpfCollectProcessBatch(processes)
 	}
-	for _, key := range keys {
-		// TODO delete keys in batch
-		processes.DeleteKey(unsafe.Pointer(&key))
+	if err == nil {
+		return
+	} else {
+		ebpfBatchGetAndDelete = false
+		processesData, err = libbpfCollectProcessSingle(processes)
```
sthaha, line 183:
```suggestion
	if err == nil {
		// log the error before ignoring it
		return
	} 
	ebpfBatchGetAndDelete = false
	processesData, err = libbpfCollectProcessSingle(processes)
```

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -331,3 +301,83 @@ func resizeArrayEntries(name string, size int) error {
...
 	return nil
 }
+
+func libbpfCollectProcessBatch(processes *bpf.BPFMap) ([]ProcessBPFMetrics, error) {
+	processesData := []ProcessBPFMetrics{}
+	var err error
+	valueSize := int(unsafe.Sizeof(ProcessBPFMetrics{}))
+	entries := MapSize / valueSize
+	isEnd := false
+	for !isEnd {
```
sthaha, line 311:
```suggestion
	for isEnd := false;  !isEnd {
```

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -331,3 +301,83 @@ func resizeArrayEntries(name string, size int) error {
...
+
+		keys := make([]uint32, entries)
+		nextKey := uint32(0)
```
sthaha, line 314: Should we create new keys for every iteration? Can this move outside the loop?

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -331,3 +301,83 @@ func resizeArrayEntries(name string, size int) error {
...
+
+		val, err := processes.GetValueAndDeleteBatch(unsafe.Pointer(&keys[0]), nil, unsafe.Pointer(&nextKey), uint32(entries))
+		// os.IsNotExist means we reached the end of the table
+		if err != nil {
+			if os.IsNotExist(err) {
+				isEnd = true
```
sthaha, line 320: should we break here?

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -331,3 +301,83 @@ func resizeArrayEntries(name string, size int) error {
...
+			var ct ProcessBPFMetrics
+			getErr := binary.Read(bytes.NewBuffer(value), ByteOrder, &ct)
+			if getErr != nil {
+				klog.V(1).Infof("failed to decode received data: %v\n", getErr)
+				continue
+			}
+			processesData = append(processesData, ct)
+		}
+	}
+	klog.V(5).Infof("successfully get data with batch get and delete with %d entries", len(processesData))
```
sthaha, line 341:
```suggestion
	klog.V(5).Infof("successfully got data using batch get and delete with %d entries", len(processesData))
```

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -331,3 +301,83 @@ func resizeArrayEntries(name string, size int) error {
...
+	return processesData, err
+}
+func libbpfCollectProcessSingle(processes *bpf.BPFMap) (processesData []ProcessBPFMetrics, err error) {
+	iterator := processes.Iterator()
+	var ct ProcessBPFMetrics
+	keys := []uint32{}
+	retry := 0
+	next := iterator.Next()
+	for next {
```
sthaha, line 350:
```suggestion
	for it := processes.Iterator(); it.Next() {
```

pkg/bpfassets/attacher/libbpf_attacher.go
sunya-ch: > sorry if didn't get this correctly, but Delete keys in get process confuse me ..

We might separate to it to function cleanup table after we call get.

pkg/bpfassets/attacher/libbpf_attacher.go
sunya-ch: > I think we need log here to indicate it's not able to (so something might be wrong) ?

Agree. We might log a message of an error and say that we are now changing to use single collect.

pkg/bpfassets/attacher/libbpf_attacher.go
sunya-ch: We should not log there because `err==nil` is always expecting true.

Instead we need log 

```suggestion
	var processes *bpf.BPFMap
	processes, err = libbpfModule.GetMap(TableProcessName)
	if err != nil {
	        // log the error here
		return
	}
	if ebpfBatchGetAndDelete {
		processesData, err = libbpfCollectProcessBatch(processes)
	        if err == nil {
		          return
	        } 
	        // log that we failed to use batch and change to false for not coming here again next time
	       ebpfBatchGetAndDelete = false
	}
	processesData, err = libbpfCollectProcessSingle(processes)
```

Also, we may consider renaming function to avoid confusion as above comment from `Collect` to `GetAndDelete` if needed.

pkg/bpfassets/attacher/libbpf_attacher.go
sunya-ch: Need another `;` for for loop syntax.

```suggestion
	for isEnd := false;  !isEnd; {
```

pkg/bpfassets/attacher/libbpf_attacher.go
sunya-ch: Similarly to above case need another `;` for correct syntax of `for`.

```suggestion
	for it := processes.Iterator(); it.Next(); {
```

sunya-ch (approved):
Logic looks good to me. 
Suggestions are good to apply.

PR #838, opened 2023-07-27
libbpf: resize maps to the actual cpu cores

rootfs:
fix #832 

before the change, per CPU core metrics uses default `NUM_CPUS`. After the change, the number of metrics is adjusted to the actual number of cores on the host:
```console
# curl localhost:8888/metrics|grep kepler_node_cpu_scaling_frequency_hertz |sort -k 2 -g
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  239k    0  239k    0     0  58.5M      0 --:--:-- --:--:-- --:--:-- 58.5M
# HELP kepler_node_cpu_scaling_frequency_hertz Current average cpu frequency in hertz
# TYPE kepler_node_cpu_scaling_frequency_hertz gauge
kepler_node_cpu_scaling_frequency_hertz{cpu="0",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="10",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="11",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="1",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="3",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="4",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="5",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="6",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="7",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="8",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="9",instance="rhel"} 0
kepler_node_cpu_scaling_frequency_hertz{cpu="2",instance="rhel"} 4.544e+06
```

PR #840, opened 2023-07-27
cgroup: use runc for cgroup v2 stat collection

mcalman:
This PR replaces the use of the containerd package for collecting cgroup stats with the analogous package from runc when using v2 cgroups. This reduces overhead as runc’s package has lower overhead for collecting v2 cgroups (see plots below). First discussed [here](https://github.com/sustainable-computing-io/kepler/issues/668#issuecomment-1541162436).
cpu-utilization
runtime
memory

Before PR:
Screenshot 2023-07-27 at 10 34 40 AM
After PR:
Screenshot 2023-07-27 at 10 47 51 AM


rootfs:
the bcc integration test failed, the fix is piggybacked in #838

pkg/cgroup/cgroup_stats_linux.go
```diff
@@ -110,32 +112,28 @@ func (c CCgroupV1StatManager) SetCGroupStat(containerID string, cgroupStatMap ma
...
-	if stat.Memory == nil {
-		return fmt.Errorf("cgroup metrics does not exist, the cgroup might be deleted")
-	}
 	// memory
-	if stat.Memory != nil {
-		cgroupStatMap[config.CgroupfsMemory].SetAggrStat(containerID, stat.Memory.Usage)
-		cgroupStatMap[config.CgroupfsKernelMemory].SetAggrStat(containerID, stat.Memory.KernelStack)
-		cgroupStatMap[config.CgroupfsTCPMemory].SetAggrStat(containerID, stat.Memory.Sock)
-	}
+	cgroupStatMap[config.CgroupfsMemory].SetAggrStat(containerID, stat.MemoryStats.Usage.Usage)
```
rootfs, line 120: do you need to check if `stat.Memory` is `nil`?

mcalman: For runc, `stat.MemoryStat` is a struct and not a pointer to a struct so it will never be `nil`. However, I have tested `c.manager.GetStats()` with a nonexistent cgroup and an error such as the following will be returned: `error while statting cgroup v2: [openat2 /sys/fs/cgroup/nonexistentcgroup/cgroup.procs: no such file or directory]`

rootfs (approved):

Issue #847, opened 2023-07-31
kepler rpm starts with 'cannot attach kprobe, probe entry may not exist'


aspanner:
### What happened?

Installed latest rpm 0.5.3 on Fedora 38.

Tried 2 different ways with systemctl start kepler and sudo kepler.
Both create the same error message:

❯ sudo dnf localinstall kepler-0.5.3-0.5.3.x86_64.rpm

Last metadata expiration check: 2:55:10 ago on Mon Jul 31 12:14:50 2023.
Dependencies resolved.
===================================================================================================================================================================================================================
 Package                                         Architecture                                    Version                                               Repository                                             Size
===================================================================================================================================================================================================================
Installing:
 kepler                                          x86_64                                          0.5.3-0.5.3                                           @commandline                                          7.5 M

Transaction Summary
===================================================================================================================================================================================================================
Install  1 Package

Total size: 7.5 M
Installed size: 32 M
Is this ok [y/N]: y
Downloading Packages:
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                           1/1 
  Installing       : kepler-0.5.3-0.5.3.x86_64                                                                                                                                                                 1/1 
  Running scriptlet: kepler-0.5.3-0.5.3.x86_64                                                                                                                                                                 1/1 
  Verifying        : kepler-0.5.3-0.5.3.x86_64                                                                                                                                                                 1/1 

Installed:
  kepler-0.5.3-0.5.3.x86_64                                                                                                                                                                                        

Complete!
❯ sudo kepler
I0731 15:10:16.761050  323170 gpu.go:46] Failed to init nvml, err: could not init nvml: error opening libnvidia-ml.so.1: libnvidia-ml.so.1: cannot open shared object file: No such file or directory
E0731 15:10:16.762861  323170 utils.go:117] getCPUArch failure: open /var/lib/kepler/data/normalized_cpu_arch.csv: no such file or directory
I0731 15:10:16.769672  323170 exporter.go:155] Kepler running on version: a7a6cb1
I0731 15:10:16.769687  323170 config.go:258] using gCgroup ID in the BPF program: true
I0731 15:10:16.769705  323170 config.go:260] kernel version: 6.3
I0731 15:10:16.769785  323170 exporter.go:179] EnabledBPFBatchDelete: true
I0731 15:10:16.769840  323170 redfish.go:169] failed to get redfish credential file path
[...]

PR #848, opened 2023-07-31
add log for attach probe failed

jichenjc:
Related to #847

### DIFF
--- a/pkg/bpfassets/attacher/bcc_attacher.go
+++ b/pkg/bpfassets/attacher/bcc_attacher.go
@@ -98,6 +98,7 @@ func loadBccModule(objProg []byte, options []string) (m *bpf.Module, err error) 
 	}
 	err = m.AttachKprobe("finish_task_switch", ftswitch, -1)
 	if err != nil {
+		klog.Infof("attaching kprobe to finish_task_switch failed, trying finish_task_switch.isra.0 instead")
 		err = m.AttachKprobe("finish_task_switch.isra.0", ftswitch, -1)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


aspanner (approved):
lgtm

aspanner (approved):
can you add the variable output as well?

vimalk78:
just adding some context to the function `finish_task_switch.isra.0` 

`isra.NNN` is the suffix added to the function name when gcc option `-fipa-sra` compiler optimization being carried out.

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html
```
-fipa-sra

    Perform interprocedural scalar replacement of aggregates, removal of unused parameters and replacement of parameters passed by reference by parameters passed by value.

    Enabled at levels -O2, -O3 and -Os.
```
So kepler looks for a function with `isra.0` suffix to attach kprobe. But why only `isra.0` ? should kepler also look for other suffixes, `isra.1`, `isra.2` ? 
Cc: @rootfs @marceloamaral

pkg/bpfassets/attacher/bcc_attacher.go
```diff
@@ -98,6 +98,7 @@ func loadBccModule(objProg []byte, options []string) (m *bpf.Module, err error)
 	}
 	err = m.AttachKprobe("finish_task_switch", ftswitch, -1)
 	if err != nil {
+		klog.Infof("finish_task_switch failed, try finish_task_switch.isra.0 instead")
```
vimalk78, line 101:
```suggestion
		klog.Infof("attaching kprobe to finish_task_switch failed, trying finish_task_switch.isra.0 instead")
```

rootfs (approved):

PR #849, opened 2023-07-31
bpf: detect correct `finish_task_switch` symbol name to attach kprobe

vimalk78:
The file `/proc/kallsyms` contains exact symbols names in kernel.

from `man proc`
```
/proc/kallsyms (since Linux 2.5.71)
      This  holds  the  kernel  exported  symbol  definitions used by the modules(X) tools to dynamically link and bind loadable modules.  In Linux 2.5.47 and earlier, a similar file with slightly different syntax was named ksyms.
```

This change detects correct symbol name at the time of building kepler, and uses the same symbol name to attach kprobe.


Makefile
```diff
@@ -15,6 +15,7 @@ GIT_VERSION     := $(shell git describe --dirty --tags --always --match='v*')
 VERSION         ?= $(GIT_VERSION)
 LDFLAGS         := "-w -s -X 'github.com/sustainable-computing-io/kepler/pkg/version.Version=$(VERSION)'"
 ROOTLESS	?= false
+KPROBE_SYM_NAME = $(shell grep finish_task_switch /proc/kallsyms | awk -F' ' '{print $$3}')
```
rootfs, line 18: the build system and runtime system are not always the same, building from a static name won't be portable when the target system uses different name

rootfs:
this is an issue with different kernel builds, bcc solves it by using dynamic [probing](https://github.com/iovisor/bcc/issues/3293), kepler does it the same [here](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/bpfassets/attacher/bcc_attacher.go#L95-L106)

Makefile
vimalk78: i was thinking since we are building it from the multistage Dockerfile, we could control the images to have the same kernel.

the current method of checking one function name after another is also at build time, and at run time the symbol could be different. we should care about the symbols in kernel running on the host machine. 
how about we find the exact symbol name at runtime, and then attach kprobe at that symbol?

rootfs: Personally I think this level of dependency belongs to ebpf library. Requiring /proc/kallsyms brings more dependencies on the kernel.

jichenjc: +1 , I think we should avoid this kind of dependency if we can use code to workaround multple options

vimalk78: the multiple options we look for are static names, and may be different from target system also.

PR #851, opened 2023-08-01
CI: Single image for integration test

SamYuan1990:
We should use single image for different k8s cluster right?
in this pr, removed strategy in image build phase of integration test to save time.

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -16,9 +16,6 @@ env:
 jobs:
   build-kepler:
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        cluster_provider: [kind,microshift]
     steps:
       - name: checkout source
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_test.yml
```diff
@@ -28,14 +25,6 @@ jobs:
         with:
           go-version: ${{env.GO_VERSION}}
 
-      - name: build manifest
-        run: make build-manifest OPTS="CI_DEPLOY"
-        env:
-          CLUSTER_PROVIDER: ${{matrix.cluster_provider}}
-          IMAGE_REPO: "localhost:5001"
-          IMAGE_TAG: "devel"
-          CTR_CMD: docker
```
sthaha, line 37: image

I see there a difference in the generation based on cluster-provider. Is that okay?

sthaha: There is also a difference in the `exporter` kustomization.yaml 

See: 
image

sthaha (approved):
Looks good to me. However, please consider separate commits for different steps. 
1. deletion of unused step
2. use a single image for all cluster-providers.

.github/workflows/integration_test.yml
sunya-ch: This affects only when building a manifest so I think the `build-kepler` should be fine.

sunya-ch (approved):
/lgtm

PR #858, opened 2023-08-08
Refactor power model class

marceloamaral:
This update brings significant improvements to the power model packages. We've organized the code and design to follow the same flow with different power models like Ratio, Linear Regression, and Sidecar Estimator.

One major change is how we estimate power. We're now using estimation-based models (LR and SE) to calculate the absolute power and idle of the node when real time power metrics are not available. But for containers, our default approach is to always use the ratio power estimation method.

Additionally, we've addressed the way idle power works for virtual machines (VMs). When we estimate absolute power, we consider both the power when the node is idle and dynamic. However, VMs add complexity. Since we can't be sure how many VMs are running, we've decided not to expose idle power when using the power model estimation. Users do have the option to enable idle power, but it's important to understand that if multiple VMs are running, the idle power value might end up duplicated.

Key Changes:

- [x] Replaced the "total power" term with "absolute power" to align with the concept of both idle and dynamic power, consistent with the Model Server approach.
- [x] Introduced the concept of normalized metrics to cater to the 3 second metric accumulation interval. The trained power model returns power instead of energy. Therefore we use normalized values to get the power and convert it later to energy by multiplying by the accumulation interval.
- [x] Resolved the issue of model output types by segregating the iota functions within distinct const declarations to prevent unintended incrementation.
- [x] Corrected the Estimator's GetComponentsPower to generate a ComponentsPower instance for all containers. It was wrongly determining the number of predicted containers.
- [x] Addressed inconsistencies in the model test, model test was return wrong power consumption of 9512J, even for 0 resource utilization.
- [x] Clarified the distinction between power and energy metrics.
- [x] Rectified metric name difference in tests through the use of constants from the config package.
- [x] Established the default use of the Ratio approach for container power estimation.
- [x] Aligned node platform power to be estimated with absolute power
- [x] Disabled estimated idle power by default when using estimated power.
- [x] Enhanced the initEstimateFunction to initialize the ratio power model.
- [x] When updating the node energy, also update the idle power when using the LR or SE power models
- [x]  Use more meaningful data during the tests, such as cpu milliseconds
- [x] fix typos
- [x] removed duplicated benchmarking test
- [x] fix dummy model server in unit tests


pkg/model/types/types.go
```diff
@@ -16,15 +16,28 @@ limitations under the License.
 
 package types
 
+type ModelType int
 type ModelOutputType int
 
 var (
 	ModelOutputTypeConverter = []string{
```
sunya-ch, line 23: Now obsoleted. 
We are now having only `AbsPower` and `DynPower`.
Weight is no more used via estimator sidecar.

pkg/config/config.go
```diff
@@ -105,20 +107,22 @@ var (
...
-	ProcessComponentsKey   = "PROCESS_COMPONENTS"
-
-	//  attribute
-	EstimatorEnabledKey = "ESTIMATOR"
-	InitModelURLKey     = "INIT_URL"
-	FixedModelNameKey   = "MODEL"
-	ModelFiltersKey     = "FILTERS"
+	ModelConfigValues map[string]string
+	// model_parameter_prefix
+	NodePlatformPowerKey        = "NODE_TOTAL"
```
sunya-ch, line 112: We may have to change later but for now "TOTAL" = "acpi" source while "COMPONENTS" = "rapl" source.

pkg/model/model.go
```diff
@@ -17,72 +17,227 @@ limitations under the License.
...
-func InitModelConfig(modelItem string) types.ModelConfig {
-	useEstimatorSidecar, selectedModel, selectFilter, initModelURL := config.GetModelConfig(modelItem)
-	modelConfig := types.ModelConfig{UseEstimatorSidecar: useEstimatorSidecar, SelectedModel: selectedModel, SelectFilter: selectFilter, InitModelURL: initModelURL}
-	klog.V(3).Infof("Model Config %s: %+v", modelItem, modelConfig)
-	return modelConfig
+// getPowerModelOutputType return the model output type for a given power source, such as platform, components, container or node power sources.
+// getPowerModelOutputType only affects LinearRegressor or EstimatorSidecar model. The Ratio model does not download data from the Model Server.
+func getPowerModelOutputType(powerSourceTarget string, modelType types.ModelType) types.ModelOutputType {
+	// Linear regression power model downloads the model weights
+	if modelType == types.LinearRegressor {
```
sunya-ch, line 218: Need to update to new model type (AbsPower and DynPower).

pkg/model/node_platform_energy.go
```diff
@@ -0,0 +1,105 @@
...
+	// TODO: Estimate the power per socket. Right now we send the aggregated values for all sockets
+	for socketID, values := range powers {
+		platformEnergy[estimatorACPISensorID+fmt.Sprint(socketID)] = values
+	}
+	return
+}
+
+// UpdateNodePlatformEnergy resets the power model samples, add new samples to the power models, then estimates the idle and dynamic energy
+func UpdateNodePlatformEnergy(nodeMetrics *collector_metric.NodeMetrics) {
+	platformPower := GetNodePlatformPower(nodeMetrics, absPower)
```
sunya-ch, line 93: I see you define absPower and idlePower in model.go. I think we need a description of this boolean variable. 
After seeing a code, it seems absPower = false because second argument GetNodePlatformPower is isIdlePower bool.  It took me minutes to go back and forward to see the function and variable. 

Can we have `GetNodePlatformPower` return two values something like below?

```
platformAbsPower, platformIdlePower =  GetNodePlatformPower(nodeMetrics)
```

sunya-ch (approved):
/lgtm 
The refactored code looks nice. Only some optional changes suggested.
Also, need some work to fix model type to the new pipeline connecting to model server.
However, it can be in a separated PR based on this refactored structure.

pkg/bpfassets/attacher/attacher_suite_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build bcc
```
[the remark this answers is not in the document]
marceloamaral: I'm not sure because I haven't made any changes to this. It's likely that libbpf requires its own set of tests.

pkg/model/types/types.go
marceloamaral: Not sure if I fully understood this, what about the linear regression model type?
Can you fix this in another PR?

pkg/model/model.go
marceloamaral: Can you fix this in another PR?

pkg/model/node_platform_energy.go
marceloamaral: Included comments in the const variables. Btw, the name of the variables to help to understand what the function will do instead of just use `true` or `false`.

PR #859, opened 2023-08-08
adding github actions to implement gosec, govet and ossf scorecard checks in kepler

### LINKED ISSUE #724: Provide go vet and gosec in Github action
Use go vet and gosec to ensure the code is following the best practice with security check

sbdtu5498:
Fixes #724
cc: @rootfs @SamYuan1990

### DIFF
--- a/.github/workflows/govet.yml
+++ b/.github/workflows/govet.yml
@@ -1,0 +1,19 @@
+name: govet
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+      - 'enhancements/**'
+      - '*.md'
+
+jobs:
+  govet_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Run go vet
+        run: go vet ./...
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -1,0 +1,29 @@
+name: Security Scan
+
+# Run workflow each time code is pushed to your repository and on a schedule.
+# The scheduled workflow runs every at 00:00 on Sunday UTC time.
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v2
+
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@master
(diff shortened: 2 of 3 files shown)


.github/workflows/govet.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  govet_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Run go vet
+        run: go vet ./...
```
marceloamaral, line 19: We have go vet in our test execution `--vet=all`.
Will this check something different?

SamYuan1990 (changes requested):
https://github.com/sustainable-computing-io/kepler/issues/724#issuecomment-1669353594
mark as RC needed, as I am confused a little. 
if go vet is covered by go lint, why we need a go vet running as https://github.com/sustainable-computing-io/kepler/pull/859/files#r1286948942?

.github/workflows/govet.yml
sbdtu5498: Nope it's just for visibility reason, we have dedicated gofmt action as well. So I think it made sense to have a separate action for govet to avoid cluttered output. govet is also implemented in golangci lint, but it also has the same troubles.

marceloamaral: We can have it separated, but we should have it in only one place.

sbdtu5498: Yes that makes more sense, I guess. I will remove them from golangci lint and test args for vet. Also, I will do the same for fmt as well as it would be redundant to run the test twice.

rootfs (approved):

.github/workflows/govet.yml
rootfs: @sbdtu5498 it is ready to merge, can you address this comment? thanks

sbdtu5498: @rootfs I will do it by today positively.

rootfs: @sbdtu5498 sounds good

SamYuan1990:
Shall we merge this PR? any comments left?

.github/workflows/govet.yml
marceloamaral: @SamYuan1990 @sbdtu5498 ^^

sbdtu5498: Just a sec. I will do it right now so that you can merge this

sbdtu5498: Okay sure. Sound's good to me. I am adding new PR on top of this one. Kindly merge so that I can fetch it.

Issue #862, opened 2023-08-14
Collect accelerator utilization through telemetry feature


Yanbo0101:
### What would you like to be added?

ref to #830, add accelerator Intel-QAT collector that can collect relevant metrics through the telemetry feature.

### Why is this needed?

collect metrics and export，prepare for the power model

jiere:
This can be the start of #645 :-)

PR #863, opened 2023-08-14
collector: add accelerator Intel-QAT collector

Yanbo0101:
This PR to add accelerator Intel-QAT collector that can collect relevant metrics through the telemetry feature. Ref to #862


pkg/power/accelerator/qat/source/qat_utils.go
```diff
@@ -0,0 +1,201 @@
...
+
+// getDevices obtain available qat devices and search for ID
+func getDevices() (map[string]interface{}, error) {
+	//Available devices
+	availableDev := make(map[string]interface{})
+
+	//use adf_ctl get qat devices status
+	commandText := "adf_ctl status"
+	cmd := exec.Command("bash", "-c", commandText)
+	out, err := cmd.CombinedOutput()
```
rootfs, line 49: can you add unit test here to validate the parsing? You can do it in a follow up PR.

pkg/power/accelerator/qat/source/qat_utils.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// getUtilization calculate utilization from each qat device
+func getUtilization(bsf string) (DeviceUtilizationSample, error) {
+
+	utilizationSample := DeviceUtilizationSample{}
+	//dataPath that can read data from telemetry
+	dataPath := fmt.Sprintf(deviceDataPath, bsf, bsf)
+
+	f, err := os.OpenFile(dataPath, os.O_RDONLY, 0666)
```
rootfs, line 107: can the data file keep open in `Init()` and close in `Shutdown()`? Just to save some file open overhead.

manifests/config/exporter/exporter.yaml
```diff
@@ -94,6 +95,8 @@ spec:
         - name: redfish
           mountPath: /etc/redfish
           readOnly: true
+        - name: qat
+          mountPath: /usr/local/bin/adf_ctl
```
rootfs, line 99: is `adf_ctl` available in rpm? We typically package utility binary such as [cpuid](https://github.com/sustainable-computing-io/kepler/blob/main/build/Dockerfile.bcc.base#L7) in the base image

Yanbo0101: adf_ctl is not available in rpm, the binary only exists on servers with QAT installed..

Yanbo0101:
> @Yanbo0101 this looks cool! I have a few comments. Please also check out the lint errors. In the mean time, can you also add the new metrics doc [here](https://github.com/sustainable-computing-io/kepler-doc/blob/main/docs/design/metrics.md)?
> 
> @jiere please help review this PR. Thanks

Thanks for your review and comments,  I have fixed all. And I will add new metrics doc later. Thanks!

jiere:
> @jiere please help review this PR. Thanks

@rootfs , actually we have an internal review before this PR, current diff LGTM, let me try approve to kick out CI workflows check first.

jiere (approved):

manifests/config/exporter/exporter.yaml
rootfs: In this case, can you make a kustomize option instead of changing the base `exporter.yaml`? You can check the CI patch [here](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/exporter/patch/patch-ci.yaml)

cc @jichenjc for info

rootfs:
The CI failure is due to a change in model server. cc @sunya-ch

manifests/config/exporter/exporter.yaml
Yanbo0101: Sure, I will take this approach. Thanks!

rootfs:
@Yanbo0101 can you rebase so to pick up the #868?

Yanbo0101:
> @Yanbo0101 can you rebase so to pick up the #868?

@rootfs Thanks. I have rebased and made a kustomize option.

rootfs (approved):

rootfs:
thanks!

Issue #870, opened 2023-08-17
metric label submitted to estimator and to prometheus are mismatch


sunya-ch:
### What happened?

According to the issue in Kepler Model Server https://github.com/sustainable-computing-io/kepler-model-server/issues/116, it  has found that the some metrics (irq and cpu time) submitted to the estimator are mismatch from what exported to the prometheus. 

### What did you expect to happen?

It should be matched to what we trained the model.

### How can we reproduce it (as minimally and precisely as possible)?

enable estimator sidecar

### Kepler image tag

<details>
targeting v0.6
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

vimalk78:
@sunya-ch is this issue closed?

PR #869, opened 2023-08-16
modify model pkg for v0.6 kepler-model-server

sunya-ch:
Corresponding to the issue https://github.com/sustainable-computing-io/kepler/issues/867,
This PR updates Kepler code syncing with v0.6 Kepler Model Server which will have only AbsPower and DynPower. 
Instead, it adds the attribute weight to differentiate whether it is applicable with LR ("Weight") or sidecar estimator ("Power") and adds the attribute energy source to differentiate between "TOTAL" (refer to acpi by default)  and "COMPONENTS" (refer to rapl by default).

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

PR #871, opened 2023-08-17
fix mismatch exported and requested metric name

sunya-ch:
As mentioned in https://github.com/sustainable-computing-io/kepler/issues/870, currently, there are many mismatch between metrics exported to prometheus and the metrics sent to estimator to estimate the power because of hard-coded metric name in collector module. 

The main goal of this PR is to use the label defined config module for the usage metric and at the same time use the energy component label defined in metric type instead of hard code. To avoid retraining the model based on currently-exporting prometheus metrics, I changed the mismatch label to what we hard coded in the collector module. 

The mismatches include
- IRQ metrics
- Kubelet metrics
- BPF (CPU time) metrics
 
This PR also fixes
- incorrect file location of defaultAbsCompURL according to previous merged PR https://github.com/sustainable-computing-io/kepler/pull/869
- remove deprecated metrics including pod_energy_stat, node_energy_stat, package_energy_millijoule


marceloamaral (approved):
Looks good
/lgtm

Issue #844, opened 2023-07-28
kepler (libbpf) misses multiple processes


sunya-ch:
### What happened?

Many processes especially the targeting workload is missing from kepler BPF collection.

```
kubectl logs -n kepler kepler-exporter-l8bn7|grep coremark|grep active
I0728 12:06:27.288384       1 metric_collector.go:126] energy from pod/container (1 active processes): name: coremark-cpeh-2340320356-nbjcg/coremark namespace: cpe-operator-system
I0728 12:06:30.054514       1 metric_collector.go:126] energy from pod/container (9 active processes): name: coremark-cpeh-2340320356-nbjcg/coremark namespace: cpe-operator-system
I0728 12:06:33.135661       1 metric_collector.go:126] energy from pod/container (8 active processes): name: coremark-cpeh-2340320356-nbjcg/coremark namespace: cpe-operator-system
I0728 12:06:36.058358       1 metric_collector.go:126] energy from pod/container (8 active processes): name: coremark-cpeh-2340320356-nbjcg/coremark namespace: cpe-operator-system
I0728 12:06:39.063118       1 metric_collector.go:126] energy from pod/container (8 active processes): name: coremark-cpeh-2340320356-nbjcg/coremark namespace: cpe-operator-system
I0728 12:06:42.061351       1 metric_collector.go:126] energy from pod/container (9 active processes): name: coremark-cpeh-2340320356-nbjcg/coremark namespace: cpe-operator-system
I0728 12:06:48.055989       1 metric_collector.go:126] energy from pod/container (0 active processes): name: coremark-cpeh-3118383817-ggvs2/coremark namespace: cpe-operator-system
I0728 12:06:51.064529       1 metric_collector.go:126] energy from pod/container (0 active processes): name: coremark-cpeh-3118383817-ggvs2/coremark namespace: cpe-operator-system
```

### What did you expect to happen?

At least one active process has been detected for each pod.

### How can we reproduce it (as minimally and precisely as possible)?

Run some workload process with high verbosity. 

### Kepler image tag

<details>
https://quay.io/repository/sustainable_computing_io/kepler/manifest/sha256:91f851348b4053ca097f14ad7dac732ae03cb0e9e72bbf821f004a01e2612f3f
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
[...]

PR #872, opened 2023-08-17
Fix missing processes issue

marceloamaral:
This PR fixes the issue #844

The changes include:

- [x] fix libbpf get batch premature return bug to collect all keys in the map
- [x] fix libbpf code to use current task to set the right cgroup id
- [x] increase the bpf map size to be able to trace more threads
- [x] fix get containerID from cgroup path bug of return containerID unknown
- [x] fix get containerID intermittently returning system_processes instead of the real containerID
- [x] fix containerID bug to return a valid container ID, some system processes were not having the system_processes ID, so that the aggregations were not properly done
- [x] reset PIDs list since it changes in each metric collection
- [x] update to use a valid PID to start a cgroup manager


marceloamaral:
example of previous invalid container ID:
```
I0817 10:39:05.777237       1 resolve_container.go:264] ##### DEBUG: extractPodContainerIDfromPath 2 containerID: /sys/fs/cgroup/system.slice/containerd.service cgroup path: /sys/fs/cgroup/system.slice/containerd.service
```

marceloamaral:
@sunya-ch @SamYuan1990 the libbpf building is failing in the integration test.
Do you know how to fix it?

rootfs:
@marceloamaral do you have a updated https://github.com/sustainable-computing-io/kepler/tree/main/bpfassets/libbpf/bpf.o

marceloamaral:
@rootfs updated!

rootfs (approved):

PR #876, opened 2023-08-22
collector: skip cgroup and kubelet metrics if they are not enabled

rootfs:


pkg/collector/metric_collector.go
```diff
@@ -102,8 +102,12 @@ func (c *Collector) Update() {
 	c.updateBPFMetrics() // collect new hardware counter metrics if possible
 
 	// TODO: collect cgroup metrics only from cgroup to avoid unnecessary overhead to kubelet
-	c.updateCgroupMetrics()  // collect new cgroup metrics from cgroup
-	c.updateKubeletMetrics() // collect new cgroup metrics from kubelet
+	if config.ExposeCgroupMetrics {
```
marceloamaral, line 105: @rootfs I understand `ExposeCgroupMetrics` as exporting prometheus metrics.

If we don't want to collect cgroup metrics, we should change to `IsCgroupMetricsEnabled()`. Also we might need to have two flags, one for exposing cgroup metrics to prometheus and another to disable cgroup metric collection.

rootfs: @marceloamaral the use case of having two flags is not clear: if the collecting flag is enabled, but the exporting flag is disabled, then the collected metrics are useless; if the collecting flag is disabled but the exporting flag is enabled, then there is an error.

rootfs: changed to `IsExposeCgroupMetricsEnabled()`

marceloamaral: Can we remove `Expose` then? Removing this will make the case clear that we can enable or disable the collection, not only exposing it to prometheus.

rootfs: @marceloamaral updated

marceloamaral (approved):
/lgtm

Issue #883, opened 2023-08-24
GPU energy collection on VM crash 


sunya-ch:
### What happened?

Got crashloopback from the following error.

```
goroutine 1 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
main.finalizing()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/cmd/exporter.go:84 +0x1d
panic({0x157c420, 0x2407c60})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/sustainable-computing-io/kepler/pkg/collector/metric.(*NodeMetrics).CalcDynEnergy(0xc00059dc90?, {0x1745b68, 0x3}, {0x176fac4, 0x1})
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/collector/metric/node_metric.go:402 +0x141
github.com/sustainable-computing-io/kepler/pkg/collector/metric.(*NodeMetrics).UpdateDynEnergy(0xc00078a200)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/collector/metric/node_metric.go:394 +0x238
github.com/sustainable-computing-io/kepler/pkg/collector.(*Collector).updateNodeEnergyMetrics(0xc00078a200)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/collector/node_energy_collector.go:124 +0x165
github.com/sustainable-computing-io/kepler/pkg/collector.(*Collector).Initialize(0xc0001917c0?)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/collector/metric_collector.go:80 +0x168
github.com/sustainable-computing-io/kepler/pkg/manager.(*CollectorManager).Start(0xc0001aeb88)
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/pkg/manager/manager.go:60 +0x25
main.main()
	/opt/app-root/src/github.com/sustainable-computing-io/kepler/cmd/exporter.go:242 +0x6b5
I0823 06:15:00.220233       1 exporter.go:87] Exiting..
```

### What did you expect to happen?

GPU energy should be collected.

### How can we reproduce it (as minimally and precisely as possible)?

Run Kepler on VM that has GPU enabled

### Anything else we need to know?

[...]

PR #884, opened 2023-08-24
handle unset idle collection

sunya-ch:
This PR is to fix and handle issue https://github.com/sustainable-computing-io/kepler/issues/883.
As mentioned in the issue, the function `components.IsSystemCollectionSupported` is to check only RAPL support.
In some environment, RAPL is not there but the GPU power is bypassed. 
So, I separate a case in function `UpdateIdleEnergyWithMinValue`. 

Not sure that should we also separate the case for platform.IsSystemCollectionSupported or not in the future. 
Currently, I keep the assumption that if components is not supported, platform is also not supported. 
If we found the case that this assumption is not valid, we may have to also separate the condition. 

Also, to prevent the failure when the idle collection is not set, I add conditional check whether the id is set in the idle stat too. 

### DIFF
--- a/pkg/collector/metric/node_metric.go
+++ b/pkg/collector/metric/node_metric.go
@@ -344,16 +344,19 @@ func (ne *NodeMetrics) SetNodeGPUEnergy(gpuEnergy []uint32, isIdleEnergy bool) {
 	}
 }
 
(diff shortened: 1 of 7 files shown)


rootfs:
> Currently, I keep the assumption that if components is not supported, platform is also not supported.

I agree

rootfs (approved):

PR #886, opened 2023-08-24
correct hardware counter settings

tiwatsuka:
The settings of hardware counters between bcc and libbpf are swapped. This PR correct it.


sunya-ch (approved):
/lgtm

sunya-ch (changes requested):
Sorry but there are more changes to make due to this fix after confirming why the code worked before.

sunya-ch:
Because the PR has merged, I will make a follow-up fix on another PR.

PR #889, opened 2023-08-25
set counter in GetEnabledHWCounters

sunya-ch:
There is another point to fix the issue in accordance to two previous merged PRs: https://github.com/sustainable-computing-io/kepler/pull/887, https://github.com/sustainable-computing-io/kepler/pull/886. 

The Counter variable is still called at the init point which is earlier than what we set value in https://github.com/sustainable-computing-io/kepler/pull/887. So, I moved the Counter variable set to right before we check the available hardware counter that is function `GetEnabledHWCounters `. 

### DIFF
--- a/pkg/bpfassets/attacher/bpf_perf.go
+++ b/pkg/bpfassets/attacher/bpf_perf.go
@@ -44,7 +44,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


marceloamaral (approved):
/lgtm

Issue #897, opened 2023-08-29
Kepler reports huge energy for some pods in K8s cluster on cloud VMs


jwjwjw3:
### What happened?

Kepler has very huge energy estimations for some pods that runs in less than an hour (at Peta Joules scale, or 10^15 J) inside a 1-master several-worker K8s cluster on IBM Cloud VMs without GPU.

Here are the output metric results of "kepler_container_joules_total" from a kepler pod:
```
# HELP kepler_container_joules_total Aggregated RAPL Package + Uncore + DRAM + GPU + other host components (platform - package - dram) in joules
# TYPE kepler_container_joules_total counter
kepler_container_joules_total{command="",container_id="325652d9b4cf1a02c37334423932361aa5b9974b2670b5f1e3eb8651d2febe0a",container_name="install-cni",container_namespace="kube-flannel",mode="dynamic",pod_name="kube-flannel-ds-j94ms"} 9.223372037599701e+14
kepler_container_joules_total{command="",container_id="5ad81f5e6bb2283d1274daf53f6f0a21ab8e7e9438d9454ff2eb2316c69e0318",container_name="install-cni-plugin",container_namespace="kube-flannel",mode="dynamic",pod_name="kube-flannel-ds-j94ms"} 9.223372037599701e+14
kepler_container_joules_total{command="",container_id="ebab60e1126208931dfe80d40bcc8396c83e93db3b55c7d8510b6b37844cd315",container_name="init-config-reloader",container_namespace="monitoring",mode="dynamic",pod_name="prometheus-k8s-0"} 9.223372037597599e+14
kepler_container_joules_total{command="",container_id="system_processes",container_name="system_processes",container_namespace="system",mode="dynamic",pod_name="system_processes"} 9.223372037599701e+14
kepler_container_joules_total{command="adapter",container_id="04f21f73f80980e6a1b88c81c19bfe9d0857e35ab8702a1faa189269c0a48147",container_name="prometheus-adapter",container_namespace="monitoring",mode="dynamic",pod_name="prometheus-adapter-757f9b4cf9-c9b44"} 9.223372037599701e+14
kepler_container_joules_total{command="alertmanag",container_id="166f9252acb14a6642bc9b0ccd527a687ef271edc69ac2e53ac3928a51578809",container_name="alertmanager",container_namespace="monitoring",mode="dynamic",pod_name="alertmanager-main-0"} 9.223372037599701e+14
[...]

PR #896, opened 2023-08-28
libbpf: fix wrong map names

rootfs:
this is to fix the error

```
 libbpf_attacher.go:143] failed to get perf event cpu_instructions_hc_reader: failed to find BPF map cpu_instructions_hc_reader: no such file or directory
```

PR #899, opened 2023-08-29
bcc: update map names

rootfs:
this is to fix an error message in #897 
```
bcc_attacher.go:120] failed to attach perf event cpu_instructions_hc_reader: passed table has wrong size: key_size=0, leaf_size=0
```

@vimalk78 previously suggested it in #896


marceloamaral (approved):
/lgtm

Issue #707, opened 2023-05-23
Find EC2 instance(s) to deploy Kepler on EKS 


nikimanoledaki:
**Is your feature request related to a problem? Please describe.**

Currently there is no documented way to deploy Kepler on Amazon EKS.

This can be challenging given that there are specific [requirements](https://sustainable-computing.io/installation/kepler/#requirements) for Kepler to work with eBPF and access RAPL.

**Describe the solution you'd like**

Find and list EC2 instances that:
- satisfy Kepler's requirements for eBPF and access to RAPL
- can be used in an EKS cluster as the AMI for a [managed node group](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html)

Then, it would be great to open a PR to the website to document these findings.

**Additional context**

Most likely the instances will have to be bare metal instances to have direct access to hardware. The tradeoff is that these are usually more expensive, but we will work with this to unblock EKS support.

[This blog post](https://medium.com/teads-engineering/building-an-aws-ec2-carbon-emissions-dataset-3f0fd76c98ac) documents how to access RAPL with the following baremetal instances: c5, m5, r5, m5zn, z1d, i3, c5n

Cross-check that these instance types are supported by EKS / https://docs.aws.amazon.com/eks/latest/userguide/choosing-instance-type.html

AntonioDiTuri:
Yes, correct! I would be happy to experiment in this direction! I wish to let you guys know more soon

SamYuan1990:
I am not sure, as in the feature, if we can have some EC2 instance added into our CI? Hence it can be a check point before we make a release tag?
May I know for EKS and EC2 instance?  if we have en EKS instance(k8s cluster), will we have the access the the EC2 instance(host VM which k8s cluster running on)?

AntonioDiTuri:
I am not sure if this asks all your questions but With EKS you have access on the ec2 instance: [more on the doc](https://aws.amazon.com/eks/features/#:~:text=Amazon%20EKS%20lets%20you%20run,Amazon%20EC2)%20and%20AWS%20Fargate.)

SamYuan1990:
> I am not sure if this asks all your questions but With EKS you have access on the ec2 instance: [more on the doc](https://aws.amazon.com/eks/features/#:~:text=Amazon%20EKS%20lets%20you%20run,Amazon%20EC2)

+@rootfs 
[...]

PR #900, opened 2023-08-30
model config: defer model config to after attacher finishes so to use the right component usage metric

rootfs:
this should fix the issue mentioned https://github.com/sustainable-computing-io/kepler/issues/707#issuecomment-1686090366 and potentially #897


sunya-ch (approved):
Nice. I think this should solve the issue. 
/lgtm

Issue #712, opened 2023-05-28
[RFC] Platform validation in Kepler.


jiere:
**Is your feature request related to a problem? Please describe.**
Currently, Kepler's test cases focus on code logic verification, lack of actual platform validation.
Kepler's data files contain static information and lack of test cases and test report as the evidence that whether the specific CPU model is well supported in Kepler, whether the data/metrics of specific power source is correctly exported, whether the power measurement accuracy is acceptable, etc. 

**Describe the solution you'd like**
We would suggest to add platform validation test cases and define content template of the test report. The specific tasks could be broken down as follows:

- [ ] Platform validation guide document. #711 
- [ ] Platform validation test cases design and implementation.
- [ ] Platform validation test report format definition, report generation and release process definition.

**Describe alternatives you've considered**
N/A.

**Additional context**
The current Kepler test framework is based on ginkgo, we will do more investigation to decide whether use Golang(align with ginkgo) or bash shell to implement the automation test logic. 

We are  proposing to contribute in this area in community.

There is an enhancement request PR [here](https://github.com/sustainable-computing-io/kepler/pull/711).

kenplusplus:
I also hope we can create a standard process or tool for early-silicon processor, so it can make Kepler can better work with the latest processor or platform quickly.

rootfs:
For baremetal machines Kepler uses ratio power attribution, so it is a lesser issue than running Kepler on VMs where ML models based estimate is used.

This platform validation will cover both baremetal and VM environments.

rootfs:
https://github.com/sustainable-computing-io/kepler-model-server/issues/100

jiere:
@rootfs , please help reopen this issue, since it has been chosen as part of 0.6.0 release features.
We will submit PR and update this feature status soon.

PR #902, opened 2023-09-01
Platform validation framework feature commit

jiere:
platform validation framework feature commit, includes:
- platform validation automation workflow definition.
- platform validation test cases design and implementation.
- new independent RAPL energy collector and power calculator for Intel X86 platforms.
- enhancement directory document update.

See issue #712


enhancements/platform-validation.md
```diff
@@ -160,102 +160,81 @@ It supports Kepler-Prometheus association by binding Kepler's specific service m
...
 ```
-$ curl --noproxy "*" -s http://172.19.0.2:32517/api/v1/label/cpu_architecture/values | jq -r ".data[]" | sort
-Sapphire Rapids
+kubectl port-forward --address localhost -n kepler service/kepler-exporter 9102:9102 &
+
+kubectl port-forward --address localhost -n monitoring service/prometheus-k8s 9090:9090 &
 ```
 
-* Deploy from kepler-helm-chart.
+To support validation cases comparison base, we are introcuding an indepentent RAPL-based energy collection and power consumption calculation tool called `validator` on Intel X86 BareMetal platforms.
```
marceloamaral, line 182: What is the problem with the current RAPL collector code? Can't you reuse it so that we avoid duplication and make sure that there is no mismatch between codes?

enhancements/platform-validation.md
```diff
@@ -275,15 +254,15 @@ Three solutions mentioned in previous comments: Kind, k3d/k3s, microshift.
 
 * Kind is current available solution in Kelper. It works well on BM, but has known issues in VM.
```
marceloamaral, line 255: Which issues?

hack/common.sh
```diff
@@ -14,25 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-# Copyright 2022 The Kepler Contributors
+# Copyright 2023 The Kepler Contributors
```
marceloamaral, line 17: We don't need to update the copyright date. Actually should should be the date that the file was created.

enhancements/platform-validation.md
jiere: There is no problem on current Kepler's RAPL code, but this PR is for validation cases design, so need to use independent way to check related data, otherwise it will be code test itself:)

enhancements/platform-validation.md
jiere: Here is not so accurate description, actually the issue is for VM's power detection, let me remove such words later.

hack/common.sh
jiere: Oh, just know about this, let me fix later.

cmd/validator/validator.go
```diff
@@ -0,0 +1,317 @@
+package main
```
rootfs, line 1: add license boiler

build/Dockerfile.platform-validation
```diff
@@ -0,0 +1,23 @@
+FROM quay.io/sustainable_computing_io/kepler_builder:ubi-9-libbpf-1.2.0-go1.18 as builder
```
rootfs, line 1: this will build `validator` in a new image, can you add these to the builder and kepler docker files?

cmd/validator/validator.go
```diff
@@ -0,0 +1,317 @@
...
+	cpu, err := ghw.CPU()
+	if err != nil {
+		fmt.Printf("get cpu info error: %s\n", err)
+		return
+	}
+	pkgNum = len(cpu.Processors)
+	samplePowers = make([]NodeComponentsPower, *sampleCount)
+	components.InitPowerImpl()
+	powerImpl := &source.PowerSysfs{}
+	pkgMax, _ = powerImpl.GetMaxEnergyRangeFromPackage()
```
rootfs, line 191: exit on error (i.e. missing RAPL domain)

cmd/validator/validator.go
```diff
@@ -0,0 +1,317 @@
...
+	coreMax, _ = powerImpl.GetMaxEnergyRangeFromCore()
+	uncoreMax, _ = powerImpl.GetMaxEnergyRangeFromUncore()
+	dramMax, _ = powerImpl.GetMaxEnergyRangeFromDram()
+
+	platform.InitPowerImpl()
+
+	csvFilePath := filepath.Join(resultDirPath, "power.csv")
+	if !isFileExists(csvFilePath) {
+		columnHeaders := []string{"Pkg", "Core", "Uncore", "Dram"}
```
rootfs, line 200: only print available RAPL domains

jiere:
> @jiere looks cool! some minor requests, it is ready to go

Thanks @rootfs , due to validator output format and whether print unsupported RAPL domain or not, please check my local workflow run log [here](https://github.com/jiere/kepler/actions/runs/6045399977/job/16405646280), are they acceptable for you?
IMO, the code should support various of X86 platforms, i.e. includes both Core and Xeon, so better to keep the four components values printed together where unsupported ones are 0.

Another question is about how to register our local runner to community's self-hosted runner? I will ttyl in Slack.

build/Dockerfile.platform-validation
rootfs, line 1: The validator image repo is created, please update the push image process to push it to this image:
podman pull quay.io/sustainable_computing_io/kepler-validator

jiere: validator is designed to be only used in platform-validation workflow which is executed ad-hoc or per release, do we need to build it regularly and maintain in official release?
And for debug use, developers also could build locally based on current Dockerfile to generate the standalone image to generate local raw node power data as comparison.

jiere: > The validator image repo is created, please update the push image process to push it to this image: `quay.io/sustainable_computing_io/kepler-validator`

How could I test this before updating the PR? Since such push operation needs dependencies like [here](https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/image.yml#L23-L26)

PR #924, opened 2023-09-11
refactor CI structure and support daily build

SamYuan1990:
in this PR.
1. we have many pipeline files specific for golang and testing....
hence, I want to change it to below settings as main entry for as CI trigger.
```
|- daily.yml for daily build
|- pr.yml
|- push.yml
|- release.yml
```
1.a IMO, in the feature, we should considering daily build for both base image and image for latest, but which will be in another PR.

2. support daily build for container.
3. btw, why should we have so many pipeline with single command as ... `go vet`?
4. nit fix on checkout action and OS to running CI.

the pr level CI will upgrade to https://github.com/sustainable-computing-io/kepler/actions/runs/6148194995?pr=924
the push level CI will upgrade to https://github.com/SamYuan1990/kepler/actions/runs/6148434127
daily build level build logs: https://github.com/SamYuan1990/kepler/actions/runs/6157468840


SamYuan1990:
k.... invalid value workflow reference: workflows must be defined at the top level of the .github/workflows/ directory

SamYuan1990:
waiting for daily build result and meanwhile for code review.

.github/workflows/developer_local.yml
```diff
@@ -1,17 +1,13 @@
-name: Unit test Mac
-
+name: developer at local
```
sthaha, line 1: nit(feel free to ignore): 
```suggestion
name: local dev env validation
```

.github/workflows/golang.yml
```diff
@@ -0,0 +1,64 @@
...
+  contents: read
+
+jobs:
+  gofmt_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{env.GO_VERSION}}
```
sthaha, line 19: This can make use of the go-version from go.mod file see:
* https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file 
* https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/.github/workflows/pr-checks.yaml#L13
*

sthaha:
Very nice @SamYuan1990 . Thanks a lot!

> btw, why should we have so many pipeline with single command as ... go vet?

Ideally we should point all CI issues on  PR. So as long as the result points out all CI issue and not only by one, I think it fine to have all checks in a single job. in fact it would be great if we had a `make lint` which devs can run locally that validates all checks locally. 
`lint` target can call other PHONY targets such as vet, gofmt, etc ..

SamYuan1990:
> Very nice @SamYuan1990 . Thanks a lot!
> 
> > btw, why should we have so many pipeline with single command as ... go vet?
> 
> Ideally we should point all CI issues on PR. So as long as the result points out all CI issue and not only by one, I think it fine to have all checks in a single job. in fact it would be great if we had a `make lint` which devs can run locally that validates all checks locally. `lint` target can call other PHONY targets such as vet, gofmt, etc ..

@sthaha ,  we have `make golint`, but ... https://github.com/sustainable-computing-io/kepler/pull/859 and https://github.com/sustainable-computing-io/kepler/issues/724 we keep the duplication for now.

.github/workflows/integration_test.yml
```diff
@@ -1,29 +1,24 @@
...
     steps:
       - name: checkout source
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: install Go
         uses: actions/setup-go@v3
         with:
-          go-version: ${{env.GO_VERSION}}
+          go-version-file: go.mod
```
jiere, line 21: Since we have aligned the go version to go.mod, shall we bump the go version in go.mod to latest then?
i.e 1.18->1.21

.github/workflows/push.yml
```diff
@@ -0,0 +1,23 @@
...
+  branch_protection_rule:
+  push:
+    branches: [ main ]
+    paths-ignore:
+      - 'doc/**'
+      - 'enhancements/**'
+      - '*.md'
+
+jobs:
+  # for each PR merge, trigger image build
```
jiere, line 12: Since we have pull_request.yml to check candidate merge PRs, and we also have daily.yaml to build image and check daily, do we still need push.yml here?

.github/workflows/image.yml
```diff
@@ -1,19 +1,14 @@
 name: image
 
 on:
-  push:
-    branches: [ main ]
-    paths-ignore:
-      - 'doc/**'
-      - 'enhancements/**'
-      - '*.md'
+  workflow_call:
```
jiere, line 5: IIRC, only image.yml here is active workflow, image_base.yml and image_pr.yml are inactive ones.
Please confirm with @rootfs, since your current PR is a workflow cleanup and refactoring.

.github/workflows/daily.yml
```diff
@@ -0,0 +1,16 @@
+name: Scheduled build
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: "12 9 * * *"
```
jiere, line 5: the timestamp here is a little bit strange, you mean 9:12 UTC daily?

.github/workflows/integration_test.yml
sthaha: How about in a different PR so that the changes introduced by this PR is minimal and can be merged?

rootfs: +1 have a min PR

sthaha:
@SamYuan1990 

> @sthaha , we have make golint, but ... https://github.com/sustainable-computing-io/kepler/pull/859 and https://github.com/sustainable-computing-io/kepler/issues/724 we keep the duplication for now.

Since I saw we have `make check` is it sufficient to have just one job that just calls `make check` ?

SamYuan1990:
> @SamYuan1990
> 
> > @sthaha , we have make golint, but ... #859 and #724 we keep the duplication for now.
> 
> Since I saw we have `make check` is it sufficient to have just one job that just calls `make check` ?

@marceloamaral , @sthaha, just IMO I suppose golangci-lint as which integrates lots of tools in one and run by single command as container is good enough for most for case. For my understanding, the reason for duplication we have today is just to provide detail informations for people who would like to use go command line tool to checking and validate the code as linting. Hence a `make check` as alias of `make golint` ... seems not the point.
IMO the point is:
as a sample of code blocks `go:=the_code`
before it been merged, every lint checking should be passed.
hence all in one as golangci-lint save a lots of works from dev ops point of view, if we treat pipeline as code, but it may stopped in the middle or other inconvenience. The `make check` as alias of `make golint`, no matter impls by golint container or go command line tools, are the same here, as in pipeline which running in a sequence, and breaks in the middle. (just for example: if go vet fails, then we know nothing about go lint result)
having go vet, go lint etc in different job as current impl, has the convenience to get the failure of lint(once it happen) and running in parallel model, but inconvenience for dev ops as pipeline as code for lots of duplicate.
...
I suppose we need another PR or issue to balance it.

.github/workflows/daily.yml
SamYuan1990: When we run it daily, in fact any random time with in a day works, hence I use Sep 12th as the day I submit this PR. the Sep to hours and 12 to min.
I don't suppose it's harmful to run go sec and scorecard daily.
From security point of view, as new CVE been reported and announce randomly at any time. A daily check is better than a weekly with a higher frequency.

.github/workflows/integration_test.yml
SamYuan1990: Resolve as this should be in another PR.

.github/workflows/push.yml
SamYuan1990: Daily build and PR level build are two different things, why not?

.github/workflows/image.yml
SamYuan1990: + @jichenjc here in loop.
- for image_pr.yml, which is used to build PR level image, as @rootfs request in the history. For our feature branch test usage, hence I suppose we need to keep it, in case of we have a big PR for example libebpf or others which need test locally. as far as I remember, we have this feature to build PR branch's image and upload to container reg, then cross validation.
- for image_base.yml, @jichenjc , @rootfs , as github agent has it's own linux header makes linux-header fails install in a ubi base image during image build.( there is no rpm package for github agent's linux header as they are ubuntu with deb package management system) Hence, any idea for how we build and daily build base image to reduce any CVE at OS level? or some yum package as linux-header?

.github/workflows/push.yml
jiere: I mean push_request.yml and push.yml.

SamYuan1990: pr, push, daily are different things.
For each PR, 
- pr means before the PR been merged. For example, we run lint before each PR and after the PR been merged the code passed the lint hence we don't need to run another turn for lint after the merge at push level or daily level.
- push means after the PR been merged. At push level, as we refresh test coverage, and build the image base on latest code.
- daily(or any schedule) is a daily build on main/default/specific branch. Take build image as an example, as we don't have PR merge per day, hence we have the image daily build to update dependences to latest as daily build. Same for go sec and score card.... SBOM as it relays on some place to store the result, so currently we make it pre releases.(store with release page on github) further if we find a place, we can make it into daily build as well.

rootfs (approved):

rootfs:
Cleaning up and consolidating the workflows is a good start. Let's take this one first.

PR #926, opened 2023-09-12
Feat: 771 remove unnecessary security concerns

### LINKED ISSUE #771: Security implications of enabling kernel-devel and pods mounting /sys/kernel/debug
We want to install kepler (https://github.com/sustainable-computing-io/kepler/) in our OpenShift 4.10 cluster.

As a prerequisite for working on OpenShift, we need to enable kernel-devel on master and worker nodes:
https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/cluster-prereqs/51-master-kernel-devel.yaml
https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/cluster-prereqs/51-worker-kernel-devel.yaml

Also, we need to mount the debugFS (/sys/kernel/debug) into the kepler Pods:
https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/exporter/patch/patch-openshift.yaml

For usage on OpenShift, kepler also requires a very powerful SecurityContextConstraints giving itself extensive privileges:
https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/exporter/openshift_scc.yaml

After seeing all this, we grew a little concerned and were wondering if enabling all this could pose a security risk or impact cluster performance.

Could you provide some insight into these aspects and if it would be secure for us to make these changes and use kepler?

BGrasnick:
fixes #771 

- feat: reduce privileges granted through SCC for OpenShift
- feat: set mounted volumes as read only as much as possible
- feat: set pod hostPID to true to avoid collisions/unwanted behavior in host PID namespace
- feat: remove unnecessary mount of /sys/kernel/debug

Same improvements were already merged in the kepler-operator: https://github.com/sustainable-computing-io/kepler-operator/pull/185
And the removal of the /sys/kernel/debug was discussed here: https://github.com/sustainable-computing-io/kepler-model-server/pull/158#discussion_r1323177090

### DIFF
--- a/manifests/config/exporter/exporter.yaml
+++ b/manifests/config/exporter/exporter.yaml
@@ -55,6 +55,7 @@ spec:
         key: node-role.kubernetes.io/master
       dnsPolicy: ClusterFirstWithHostNet
       serviceAccountName: kepler-sa
+      hostPID: true
       containers:
       - name: kepler-exporter
         image: kepler:latest
--- a/manifests/config/exporter/exporter.yaml
+++ b/manifests/config/exporter/exporter.yaml
@@ -86,8 +86,10 @@ spec:
         volumeMounts:
         - mountPath: /lib/modules
           name: lib-modules
+          readOnly: true
         - mountPath: /sys
           name: tracing
+          readOnly: true
         - mountPath: /proc
           name: proc
         - name: cfm
--- a/manifests/config/exporter/patch/patch-openshift.yaml
+++ b/manifests/config/exporter/patch/patch-openshift.yaml
@@ -19,6 +19,7 @@ spec:
         volumeMounts:
         - name: kernel-src
           mountPath: /usr/src/kernels
+          readOnly: true
         - name: kernel-debug
           mountPath: /sys/kernel/debug
         securityContext:
--- a/manifests/config/exporter/patch/patch-openshift.yaml
+++ b/manifests/config/exporter/patch/patch-openshift.yaml
@@ -20,15 +20,9 @@ spec:
         - name: kernel-src
           mountPath: /usr/src/kernels
           readOnly: true
-        - name: kernel-debug
-          mountPath: /sys/kernel/debug
         securityContext:
           privileged: true
-      volumes:      
-      - name: kernel-debug
-        hostPath:
-          path: /sys/kernel/debug
-          type: Directory
+      volumes:
       - name: kernel-src
         hostPath:
           path: /usr/src/kernels
--- a/manifests/config/exporter/openshift_scc.yaml
+++ b/manifests/config/exporter/openshift_scc.yaml
@@ -6,9 +6,9 @@ metadata:
 # To allow running privilegedContainers
 allowPrivilegedContainer: true
 allowHostDirVolumePlugin: true
-allowHostNetwork: true
-allowHostPorts: true
-allowHostIPC: true
+allowHostNetwork: false
+allowHostPorts: false
+allowHostIPC: false
 allowHostPID: true
 readOnlyRootFilesystem: true
 defaultAddCapabilities:


sthaha (commented):
Look good to me :)

manifests/config/exporter/exporter.yaml
```diff
@@ -54,6 +54,7 @@ spec:
         key: node-role.kubernetes.io/master
       dnsPolicy: ClusterFirstWithHostNet
       serviceAccountName: kepler-sa
+      hostPID: true
```
marceloamaral, line 57: This is important, otherwise we cannot get the right container ID and then pod name.

manifests/config/exporter/exporter.yaml
```diff
@@ -85,8 +86,10 @@ spec:
         volumeMounts:
         - mountPath: /lib/modules
           name: lib-modules
+          readOnly: true
```
marceloamaral, line 89: Why remove this? Enabling only read only is more secure...

manifests/config/exporter/patch/patch-openshift.yaml
```diff
@@ -19,15 +19,10 @@ spec:
         volumeMounts:
         - name: kernel-src
           mountPath: /usr/src/kernels
-        - name: kernel-debug
-          mountPath: /sys/kernel/debug
+          readOnly: true
```
marceloamaral, line 22: @sunya-ch can we remove the kernel debug path with libbpf?
Did you test that?

marceloamaral (changes requested):
We need to double check if we can remove the debug path.

manifests/config/exporter/exporter.yaml
sthaha: Doesn't the change only enables `readOnly` ?

manifests/config/exporter/patch/patch-openshift.yaml
sthaha: https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L10319 seems to indicate that it will be useful to have this mounted.

manifests/config/exporter/exporter.yaml
BGrasnick: Exactly, I enabled this for as many mounts as possible to improve security :)

jichenjc: right, this seems add the `readonly` :)

manifests/config/exporter/patch/patch-openshift.yaml
jichenjc: if above is true, maybe we need think about add some comment so we won't remove this later

BGrasnick: I tried it before when working on https://github.com/sustainable-computing-io/kepler-operator/pull/185 with both latest and latest-libbpf and I tested by running the operator locally and installing kepler + the Grafana dashboard without the /sys/kernel/debug mount. I could still see metrics in Grafana with both latest and latest-libbpf images if I recall correctly so it should work from my understanding.
If you know how we can test this more and better please tell me!
Also when I have time I will test this again just to be sure!

sunya-ch: > @sunya-ch can we remove the kernel debug path with libbpf?
> Did you test that?

sorry for delays. It is fine to remove it here because it is mounted at the beginning in the exporter ds. No need to patch.

marceloamaral: @sunya-ch where is it mounted?
Are you saying that this file is just duplicating something in another file?

sunya-ch: `/sys` path is mounted here if we deploy exporter via make manifest: https://github.com/sustainable-computing-io/kepler/blob/5272ee2f8f6ab0afc4214884ba143b6733e53f61/manifests/config/exporter/exporter.yaml#L112-L114.

patch-openshift.yaml is supposed to include only the patch that is required specific to openshift not general.

sthaha: @marceloamaral thanks for the ping. Noted, I will compare the daemonset again to ensure that they are in sync.

sthaha:
@BGrasnick would you mind if I took a stab at addressing the comments and submit another PR?

BGrasnick:
Hey @sthaha! Sorry for the delay! I was away for a while but now I am back and ready to jump back in :)

sthaha:
Thank you @BGrasnick , the PR is quite useful :)

sthaha (commented):
@marceloamaral This PR looks good to me. Since you raised a concern which seems to be addressed, I will let you review and merge if it looks fine to you.

rootfs (approved):

Issue #923, opened 2023-09-10
GPU power utilization showing up in system instead of pod


rahul-shiv:
### What happened?

I tried out using kepler to monitor GPU power utilization of a pod using kind and microk8s deployments and the gpu power usage showed up in system instead of the pod where the workload was running.

For the kind use case, I used a custom version of kind to support GPUs : https://jacobtomlinson.dev/posts/2022/quick-hack-adding-gpu-support-to-kind/.
Kind kepler build command used: make build-manifest OPTS="CI_DEPLOY PROMETHEUS_DEPLOY"

For the microk8s use case, I used the bare-metal version of kepler.
microk8s kepler build command used: make build-manifest OPTS="BM_DEPLOY PROMETHEUS_DEPLOY"

Are either of these platforms supported and if so, is there something I am doing wrong? If not, what would be the right way to deploy kepler and kubernetes on a single node setup to try out kepler's GPU power monitoring capability?

Please let me know if I can provide any relevant logs.

### What did you expect to happen?

Expected the power utilization to show up in the respective pod where the workload was running instead of the "SYSTEM" namespace.

### How can we reproduce it (as minimally and precisely as possible)?

For kind case, run a gpu workload using custom version of kind to support GPUs : https://jacobtomlinson.dev/posts/2022/quick-hack-adding-gpu-support-to-kind/

No specific steps for microk8s case.

### Kepler image tag

<details>
Built from github main branch, pulled on Sep 8th.
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```
Client Version: v1.28.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.21.2
</details>

### Cloud provider or bare metal

<details>
Bare Metal
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
[...]

PR #927, opened 2023-09-13
Fix GPU ratio bug to use GPU metrics

marceloamaral:
The issue #923 pointed out a bug in the GPU dynamic power estimation for containers.

There was a bug in the ratio formula using the wrong metrics to calculate the ratio.


sunya-ch (approved):
good catch.
/lgtm

PR #930, opened 2023-09-14
by default disable GPU

jichenjc:
it takes 6 seconds to start my kepler which is not needed since I don't have GPU (and by default should be as most of user doesn't hvae GPU)

```
I0914 08:06:49.519880       1 exporter.go:204] Initializing the GPU collector
I0914 08:06:55.525201       1 watcher.go:66] Using in cluster k8s config
```

### DIFF
--- a/manifests/config/exporter/exporter.yaml
+++ b/manifests/config/exporter/exporter.yaml
@@ -18,7 +18,7 @@ data:
   KEPLER_LOG_LEVEL: "1"
   METRIC_PATH: "/metrics"
   BIND_ADDRESS: "0.0.0.0:9102"
-  ENABLE_GPU: "true"
+  ENABLE_GPU: "false"
   ENABLE_QAT: "false"
   ENABLE_EBPF_CGROUPID: "true"
   EXPOSE_HW_COUNTER_METRICS: "true"


rootfs:
@jichenjc can you make a patch instead?

jichenjc:
you mean a kustomize patch?

rootfs:
yes, that'll make disabling GPU an option without changing the default.

marceloamaral:
I think we should not remove GPU auto-detection as the default behavior.
But, it is possible to disable GPU detection with parameters. So, @jichenjc can you use the parameters to disable it for you?

Btw, the GPU auto-detection should not take that long, maybe the problem is something else....

jichenjc:
>>>Btw, the GPU auto-detection should not take that long, maybe the problem is something else....

we wait for 5 seconds in the code
https://github.com/sustainable-computing-io/kepler/blob/main/cmd/exporter/exporter.go#L208

@marceloamaral , maybe not sleep that long?

marceloamaral:
@jichenjc we need the sleep because of another problem....

Can you add the `ENABLE_GPU=false` option to build manifest?
So that you can create a manifest with the GPU disabled.

jichenjc:
ok, I think it's ok now to build my own manifest

Issue #663, opened 2023-05-01
Remove Code and Data duplication


marceloamaral:
Currently, the Kepler code uses two data structures: `ContainerMetrics` and `ProcessMetrics`, which contain similar information. However, `ContainerMetrics` contains an aggregation of `ProcessMetrics`, resulting not only in duplicated data but also duplicated code for energy consumption calculations and duplicated Prometheus metrics for `kepler_process...` and `kepler_container...`.

Additionally, as far as I understood before, the original idea behind exposing process metrics was to allow Kepler to run standalone without the need to report container metrics. However, the current code uses the container power to estimate the process power in the ratio power model. This estimation should instead utilize the node power instead of the container power.

To enhance the code and improve consistency, we should simplify the code to have only one map, such as `ProcessMetrics` only, where the map key will correspond to the level of aggregation configured in Kepler (`Container ID`, `Pod ID`, or `Process ID`). This approach enables better scalability by reporting lower metrics as the level of aggregation increases.

When per-process metrics are enabled, there is no need to aggregate metrics per container in the map. Instead, process metrics can be exposed, including labels for process name/command, container name and ID, and pod name. This requires only the `kepler_process...` metrics. When Kepler aggregates data per container, the process-related labels could be empty if per-process metrics is not enabled, whereas if Kepler reports metrics per process, the container labels may exist or not if if Kepler is running in Kubernetes or standalone.

By adopting this approach, we can achieve a more efficient, flexible, and consistent way of exposing metrics. Users can aggregate per-container and per-pod metrics using the process metrics in a Prometheus query. 

Additional information on this topic can be found in issue #583.

SamYuan1990:
I am good with this proposal.

rootfs:
sounds good!

jichenjc:
>>>where the map key will correspond to the level of aggregation configured in Kepler (Container ID, Pod ID, or Process ID)

agree this is the right approach and yes, it's flexible and easy to handle

jichenjc:
some design as reference and discussion:

```
type ProcessMetrics struct {
        PID          uint64
        Command      string
        CounterStats map[string]*types.UInt64Stat
        // ebpf metrics
        CPUTime             *types.UInt64Stat
        SoftIRQCount        []types.UInt64Stat      

        DynEnergyxxx    ==> a list of Dyn energy 
        IdleEnergyxxx     ==> a list of Idle energy
```

container metrics (in addition to ProcessMetrics)
```
[...]

PR #931, opened 2023-09-14
collector: add kepler_vm metrics

aharivel:
Kepler is able to track all Process Metrics and thus theoretically all KVM/QEMU Virtual Machines (VM) metrics. 
However, when multiple VM are running on the same machine, we have the following problem: 

screenshot

All "command" are the same and the user can't distinguish the different VM. This is especially a problem for OpenStack user where multiple VM are running on the node. 

In order to solve this issue, this commit adds a new category of metrics dedicated to VM. All the metrics are coming from the Process Metrics. The difference is that the "command" is replaced by the "name" of the VM given by libvirt. 
During the update phase, the processes are filter by Process ID and all PID/TID that belongs to the same VM is inheriting the same "name". The addition of all metrics belonging to the same "name" represents this particular VM metrics.

Below the result of 2 different VM (instances) metrics on OpenStack with this commit :

ZFWoTb

### DIFF
--- a/pkg/collector/container_hc_collector.go
+++ b/pkg/collector/container_hc_collector.go
@@ -17,12 +17,14 @@ limitations under the License.
 package collector
 
 import (
+	"strconv"
 	"unsafe"
 
 	"github.com/sustainable-computing-io/kepler/pkg/bpfassets/attacher"
 	"github.com/sustainable-computing-io/kepler/pkg/cgroup"
 	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
 	"github.com/sustainable-computing-io/kepler/pkg/config"
+	"github.com/sustainable-computing-io/kepler/pkg/libvirt"
 
 	"k8s.io/klog/v2"
 )
@@ -33,7 +35,7 @@ import "C"
 type ProcessBPFMetrics = attacher.ProcessBPFMetrics
 
 // updateBasicBPF
-func (c *Collector) updateBasicBPF(containerID string, ct *ProcessBPFMetrics, isSystemProcess bool) {
+func (c *Collector) updateBasicBPF(containerID string, ct *ProcessBPFMetrics, isSystemProcess, isSystemVM bool) {
 	// update ebpf metrics
 	// first update CPU time
 	err := c.ContainersMetrics[containerID].CPUTime.AddNewDelta(ct.ProcessRunTime)
@@ -56,10 +58,16 @@ func (c *Collector) updateBasicBPF(containerID string, ct *ProcessBPFMetrics, is
 			}
 		}
 	}
+	// track virtual machine metrics
+	if isSystemVM && config.EnableProcessMetrics {
+		for i := 0; i < config.MaxIRQ; i++ {
+			c.VMMetrics[ct.PID].SoftIRQCount = c.ProcessMetrics[ct.PID].SoftIRQCount
+		}
+	}
 }
 
 // updateHWCounters
-func (c *Collector) updateHWCounters(containerID string, ct *ProcessBPFMetrics, isSystemProcess bool) {
+func (c *Collector) updateHWCounters(containerID string, ct *ProcessBPFMetrics, isSystemProcess, isSystemVM bool) {
 	// update HW counters
 	for _, counterKey := range collector_metric.AvailableHWCounters {
 		var val uint64
@@ -84,6 +92,10 @@ func (c *Collector) updateHWCounters(containerID string, ct *ProcessBPFMetrics, 
 				klog.V(5).Infoln(err)
 			}
 		}
+		// track virtual machine metrics
+		if isSystemVM && config.EnableProcessMetrics {
+			c.VMMetrics[ct.PID].CounterStats = c.ProcessMetrics[ct.PID].CounterStats
+		}
 	}
 }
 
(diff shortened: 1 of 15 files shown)


rootfs:
@aharivel this is great! can you rebase so we can run the CI first?

pkg/libvirt/resolve_vm.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	for _, file := range files {
+		threadIDs = append(threadIDs, file.Name())
+	}
+
+	return threadIDs
+}
+
+func GetCurrentVMPID() (map[string]string, error) {
```
rootfs, line 45: can you write a unit test for this function?

pkg/collector/metric_collector.go
```diff
@@ -118,6 +123,7 @@ func (c *Collector) Update() {
 	// calculate the process energy consumption using its resource utilization and the node components energy consumption
 	if config.EnableProcessMetrics {
 		c.updateProcessEnergy()
+		c.updateVMEnergy()
```
rootfs, line 126: we'll evaluate the VM resolution overhead later and see if we need a `EnableVMMetrics` option

pkg/model/estimator/local/ratio_vm.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"sync"
+
+	"k8s.io/klog/v2"
+
+	collector_metric "github.com/sustainable-computing-io/kepler/pkg/collector/metric"
+)
+
+func getVMResUsage(vm *collector_metric.VMMetrics, usageMetric string) float64 {
```
rootfs, line 32: since the VM power comes from the qemu process, would it make sense to create a <pid, vm> metrics mapping and reuse the process power metrics, instead of calculating VM power from scratch?

pkg/libvirt/resolve_vm.go
aharivel: done

pkg/model/estimator/local/ratio_vm.go
aharivel: I've now simplified the VM metrics by simply grab the Process metrics values that are already processed. 
So this is not needed anymore.

pkg/collector/metric_collector.go
aharivel: I've took off all the calculations that are redundant with the Process metrics so the overhead should be very insignificant

pkg/model/estimator/local/ratio_vm.go
rootfs: this new approach is great!

rootfs:
@aharivel can you clean up the lint errors?

rootfs:
@aharivel would you please add the VM metrics to the doc? https://github.com/sustainable-computing-io/kepler-doc/blob/main/docs/design/metrics.md

aharivel:
@rootfs I'm currently on PTO but you can count on me, I'll add the VM metrics into the doc asap

marceloamaral:
@aharivel this is a great initiative, but introduce some problems...

@rootfs and @aharivel, this creates a lot of duplication in our code.

Wouldn't it be better to use VMID as we use the ContainerID? In fact, I previously raised an issue to eliminate all the duplicate code related to 'process' that mirrors the container code. #663

We should update the logic by only change the ID of the map for different context. The optimal approach would be to modify the 'containerMetrics' map to be only `metrics` to handle metrics for processes, containers, and VMs...only the map key truly matters.

Maintaining duplicated code can become quite challenging in the future.
For example, the ratio power model is the same for container, process and vm now...

Would it pose any issues if we were to revert this logic later to simplify the code?

marceloamaral:
What about update our prometheus metrics removing the `process`, `containe`r and `vm` context as only `kepler_package_joules_total` and then it might have labels as `container_id`, `vm_id`, `process_id`, `pod_name`, depending on the context.

This will make our code much more consistency and with less duplication... Wdyt?

pkg/collector/metric/vm_metric.go
```diff
@@ -0,0 +1,118 @@
+/*
```
marceloamaral, line 1: This file is a good example that I am talking about introducing duplication, and we don't really need to do that.

pkg/collector/container_hc_collector.go
```diff
@@ -84,14 +92,20 @@ func (c *Collector) updateHWCounters(containerID string, ct *ProcessBPFMetrics,
...
 	}
 }
 
 // updateBPFMetrics reads the BPF tables with process/pid/cgroupid metrics (CPU time, available HW counters)
 func (c *Collector) updateBPFMetrics() {
 	foundContainer := make(map[string]bool)
 	foundProcess := make(map[uint64]bool)
+	foundVM := make(map[uint64]bool)
 	processesData, err := attacher.CollectProcesses()
+	vmPIDList, _ := libvirt.GetCurrentVMPID()
```
marceloamaral, line 108: @aharivel @rootfs We should retrieve the VM ID in the same way we do for the container ID. The libvirt ID can be found in `/proc/PID/cgroup`. This approach will allow us to maintain consistent and efficient code, eliminating the need to iterate through the vmPIDList for all ebpf entries.

aharivel:
@marceloamaral Hi Marcelo, thank you.
All your comments make total sens and in the end, refactoring code should be done to make the code more maintainable.
This initiative is just a required first step for me and your comments will lead to the next steps to make the code even better.  I will take care of this when I'll be back from PTO.

PR #932, opened 2023-09-14
fix unset environment variable

mcalman:
Sets a default value of `kepler` for `IMAGE_NAME` in the `Makefile`.

7f42672701bd8643507e0ab9ef5dd6da3f58ac76 replaced `kepler` with the environment variable `IMAGE_NAME`. Without a default value set, `make push-image` fails unless the environment variable is manually set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,44 +7,19 @@ export BIN_TIMESTAMP ?=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 export TIMESTAMP ?=$(shell echo $(BIN_TIMESTAMP) | tr -d ':' | tr 'T' '-' | tr -d 'Z')
 
 # restrict included verify-* targets to only process project files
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -28,6 +28,12 @@ else
 	BUILDER_IMAGE := quay.io/sustainable_computing_io/kepler_builder:ubi-9-libbpf-1.2.0-go1.18
 endif
 
+ifdef IMAGE_NAME
+	IMAGE_NAME := $(IMAGE_NAME)
+else
+	IMAGE_NAME := kepler
+endif
```
sthaha, line 35: How about ... 

```suggestion
IMAGE_NAME ?= kepler
```

rootfs (approved):

Makefile
mcalman: @sthaha thanks for the suggestion. I refactored the other variable definitions as well for consistency/readability.

sthaha: nice!!

PR #948, opened 2023-09-28
grafana: update kWh query

marceloamaral:
As pointed by @sthaha in the discussion #946, the Prometheus query (defined in the PR #291) that calculates the `kWh` has a problem in the logic.

The query was only considering the first `1h` interval to extrapolate the results, but it should use the `24h` interval.
The problem is if in the last 1h, the power consumption is high, the extrapolation will show a very high value, even if the early values are low.

Fortunately, the `increase` function actually does what we need. It takes the average value from the entire range and multiply by the available sample interval.

Giving that, I update all the queries that calculate the `kWh` to be as:
```
sum by (container_name, pod_name) (
  increase(
   (kepler_container_package_joules_total{}[24h:30m])
  )
) * $watt_per_second_to_kWh
```
remembering that $watt_per_second_to_kWh is 0.0000002778 because:
1J = 1W*s,
1Wh = 1J/3600
1kWh = 1/3600000 = 0.0000002778

Note that I am using a step of `30m` in a range of 24h because the queries are very heavy. 
For example, with a step of `1m` I could not set my Grafana dashboard to show a timeseries of 12h, the queries were timing out. But I could load my dashboard with `30m` step.

Therefore, this PR includes the following:

- [x] fix the `kWh` extrapolation
- [x] decrease the samples in the Prometheus queries for performance reasons
- [x] add the GPU and OTHER power consumption instead of only using Package and DRAM
- [x] decrease the number of showing containers to top 10 and top 15

### DIFF
--- a/grafana-dashboards/Kepler-Exporter.json
+++ b/grafana-dashboards/Kepler-Exporter.json
@@ -22,8 +22,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 26,
-  "iteration": 1667798877250,
+  "id": 2,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -93,7 +92,7 @@
         "showThresholdMarkers": true,
         "text": {}
       },
-      "pluginVersion": "8.5.5",
+      "pluginVersion": "9.1.0",
       "targets": [
         {
           "datasource": {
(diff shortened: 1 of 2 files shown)


rootfs (approved):

simonpasquier:
> Note that I am using a step of 30m in a range of 24h because the queries are very heavy.
For example, with a step of 1m I could not set my Grafana dashboard to show a timeseries of 12h, the queries were timing out. But I could load my dashboard with 30m step.

The expressions in the current version of the dashboard use `[24h:1m]` (https://github.com/sustainable-computing-io/kepler/blob/a1b042d87a9f46e5403ec4eecebb17e2b3916402/grafana-dashboards/Kepler-Exporter.json). Is this expected?

I haven't looked at the dashboard in detail but another improvement to reduce the load on Prometheus at query time would be to use recording rules that pre-compute the final results. The downside is that it will increase the overall series cardinality if you do each for each pod but maybe it's worth doing for namespace-level aggregations?
You could also do time-based aggregations, e.g. record power consumption for the last hour (`namespace:kepler_container_joules:increase1h`) and sum the last 24 samples to get the daily value (`sum by(namespace) (namespace:kepler_container_joules:increase1h[24h:1h]`).

sthaha:
@simonpasquier Thanks for chiming in. 

We have started using PrometheusRule when operator deploys `kepler` - https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/pkg/components/exporter/assets/dashboards/power-monitoring-by-ns.json#L100 

> You could also do time-based aggregations, e.g. record power consumption for the last hour (namespace:kepler_container_joules:increase1h) and sum the last 24 samples to get the daily value (sum by(namespace) (namespace:kepler_container_joules:increase1h[24h:1h]).

We have some something on these line here : 
https://github.com/sustainable-computing-io/kepler-operator/blob/7da8436c7f588bbbfadeaffaaf09e6c136249aef/pkg/components/exporter/exporter.go#L533

I intent to submit the same to this repo as well.  I will tag you on the PR that adds the recording rule to get your thoughts. 

Thanks again for helping out 🙇

marceloamaral:
@sthaha please also let me know when you find a better solution for this!

flaxalf:
Hi @marceloamaral , I find the query interesting and I would like to replicate it in the Flux language, do you know how I can do it? The increase function doesn't seem to be the same

marceloamaral:
Hi @flaxalf , unfortunately I am not familiar with the flux syntax.

But basically you need to aggregate the energy for the last 24h and convert it to kWh.

PR #950, opened 2023-09-30
exporter: create model before manager starts

rootfs:
The manager needs node and container models to update the metrics. So the models should be created before the manager starts.

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -229,6 +229,14 @@ func main() {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


sunya-ch (approved):
Looks good to me. 

Note:
ContainerFeatureNames is set by 

https://github.com/sustainable-computing-io/kepler/blob/da06ff8248cdcac7e7b53c4fa651402dda39252a/cmd/exporter/exporter.go#L201

which comes before line 232.

Issue #952, opened 2023-10-02
Align Kepler local estimator with updated models 


rootfs:
### What happened?

The updated models use `scale` and `weight`, while current Kepler still uses `mean` and `variance`. 

### What did you expect to happen?

Kepler local estimator should use `scale` and `weight` to align with latest models.

### How can we reproduce it (as minimally and precisely as possible)?

Running kepler post 0.5.5 image doesn't emit the right energy estimate.

### Kepler image tag

<details>
latest
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

rootfs:
Will pick fix into 0.6 release, cc @husky-parul

marceloamaral:
@rootfs the PR #952 fix this.

PR #954, opened 2023-10-02
Update LR data normalization to divide by the max value (scale)

marceloamaral:
We fixed in the model server the data normalization to divide a given metric by the max value, because the previous normalization was resulting in negative values.

This PR fix the issue #952.


rootfs:
@marceloamaral thanks, this is great! It is ready to go once you have a update to `lr_test.go`

rootfs (approved):

sunya-ch (approved):
/lgtm
could be merged after all test finished.

Issue #956, opened 2023-10-03
Double counting GPU energy for the multiple processes in a container


rinana:
### What happened?

I ran the same python script w/ torchrun command and estimated GPU energy w/ Kepler for several times by changing the number of GPUs (1,2,4,8 GPUs). 
Prometheus seems to double-count Kepler's GPU energy inside a container when multiple commands were executed although the GPU energyper node did not increase with the same number of GPUs.

GPU energy per node result from Prometheus
<img width="1387" alt="スクリーンショット 2023-10-03 12 13 16" src="https://github.com/sustainable-computing-io/kepler/assets/17348462/3b740d64-6911-441d-8cb7-60ef4fe6aff1">

GPU energy per container result from Prometheus (sum by pod/container)
The pink one should be similar to the orange one and the last green one should be similar to the blue one, but they seems to consume more.
<img width="1387" alt="スクリーンショット 2023-10-03 11 45 54" src="https://github.com/sustainable-computing-io/kepler/assets/17348462/dc79200e-c136-47fc-bfca-f5b487ba7dd2">

GPU energy per container result from Prometheus (sum by pod/container/command)
<img width="1385" alt="スクリーンショット 2023-10-03 11 45 18" src="https://github.com/sustainable-computing-io/kepler/assets/17348462/da403c72-7d03-4022-b276-62b4964411ba">

### What did you expect to happen?

I expected the Kepler's exported GPU energy per container result in almost the same with the same number of GPUs.

### How can we reproduce it (as minimally and precisely as possible)?

Run torchrun workload in a container or multiple commands that consumes GPU in turns inside a single container?

### Kepler image tag

<details>
quay.io/sustainable_computing_io/kepler:latest-libbpf
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
[...]

PR #957, opened 2023-10-03
Fix GPU power double counting by removing the container command

marceloamaral:
As described in the issue #956, as container is group of processes, Kepler randomly associates a command of one process to the container. Consequently, the GPU power will be exposed intermittently with different commands.

This PR disable the command exported in the Prometheus container metrics.

We will optionally enable it back after fixing the code duplication from issue https://github.com/sustainable-computing-io/kepler/issues/663.

### DIFF
--- a/pkg/collector/prometheus_collector.go
+++ b/pkg/collector/prometheus_collector.go
@@ -302,92 +302,92 @@ func (p *PrometheusCollector) newContainerMetrics() {
 	containerCoreJoulesTotal := prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, "container", addSuffix(collector_metric.CORE, config.AggregatedEnergySuffix)),
 		"Aggregated RAPL value in core in joules",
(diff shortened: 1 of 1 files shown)


pkg/collector/prometheus_collector.go
```diff
@@ -618,15 +618,12 @@ func (p *PrometheusCollector) updateNodeMetrics(wg *sync.WaitGroup, ch chan<- pr
...
 	for _, container := range *p.ContainersMetrics {
 		wg.Add(1)
 		go func(container *collector_metric.ContainerMetrics) {
 			defer wg.Done()
-			containerCommand := container.Command
-			if len(containerCommand) > commandLenLimit {
-				containerCommand = container.Command[:commandLenLimit]
-			}
+			// removed command because issue #956
+			containerCommand := ""
```
rootfs, line 626: can you remove the container command entirely from the metrics?

marceloamaral: @rootfs, done!

PR #968, opened 2023-10-05
Update idle power model

marceloamaral:
The previous code refactoring incorrectly defined the idle power due to a copy-and-paste error.

According to the GHG protocol, we are supposed to divide the idle power based on the application's size, which should align with the container's resource request. However, as there are containers without requests, implementing this can be challenging.

As a quick solution, this PR reverts to the previous logic of evenly distributing the idle power across all containers.


sunya-ch (approved):
/lgtm

Issue #963, opened 2023-10-04
Change default Prometheus scraping interval from 3s to 30s


mcalman:
### What would you like to be added?

I recommend we change the [Prometheus scraping interval for Kepler](https://github.com/sustainable-computing-io/kepler/blob/02dc840382575f70b6dfd8b497f7afe44f858f15/manifests/config/exporter/prometheus_service_monitor.yaml#L10C6-L10C6) from 3s to 30s.

First discussed here: https://github.com/sustainable-computing-io/kepler/discussions/915#discussioncomment-7186137

### Why is this needed?

As seen in the plots below, this will reduce overhead for both Kepler and Prometheus.

30s is the default scraping interval for Prometheus and any users who needs finer granularity can change the value in the service monitor yaml. As pointed out by @marceloamaral, Kepler uses counter metrics so no accuracy should be lost with a slower scraping frequency.

![image](https://github.com/sustainable-computing-io/kepler/assets/68564154/64ef6509-e6d5-44dd-a4ff-fa6a72e1d5ec)
![image](https://github.com/sustainable-computing-io/kepler/assets/68564154/e17de72f-e0ec-4038-a640-ddc4966e5360)
![image](https://github.com/sustainable-computing-io/kepler/assets/68564154/5b15dddb-a043-4c13-a1e5-27e087a24ae2)
![image](https://github.com/sustainable-computing-io/kepler/assets/68564154/9a444a07-fa5d-4b58-8a63-415ec2f6a135)
Experiment setup:
- Single node Kind cluster
- 40 dummy containers on node
- Average overhead values when Kepler is not deployed are subtracted to get "Kepler attributed" overhead for Prometheus

### Requirements

- [ ] Change [scrape interval default value]((https://github.com/sustainable-computing-io/kepler/blob/02dc840382575f70b6dfd8b497f7afe44f858f15/manifests/config/exporter/prometheus_service_monitor.yaml#L10C6-L10C6)) to 30s
- [ ] Add documentation informing users how to change the scrape interval

marceloamaral:
@mcalman, could you please create a PR?

Within the PR, let's include some comments in the documentation.

Btw, the only potential drawback I can think of at the moment is that there might be a delay in the appearance of metrics with the 30-second interval.

PR #970, opened 2023-10-06
config: change default prometheus scrape interval

mcalman:
Addresses #963

This makes "30s" the new Kepler default scraping interval for Prometheus. It also adds a new manifest build option (`"OPTS="PROMETHEUS_DEPLOY HIGH_GRANULARITY" make build-manifest`) to revert to the original "3s" scrape interval.

Users can also specify any scraping interval they would like by modifying PROMETHEUS_SCRAPE_INTERVAL in the ConfigMap before building the manifests.

If accepted, I will add the new deployment option to the docs [here](https://github.com/sustainable-computing-io/kepler-doc/blob/aa3a0850db39918bba5d721b6ae27faad5a03efb/docs/installation/kepler.md?plain=1#L67).


manifests/config/base/kustomization.yaml
```diff
@@ -25,4 +25,6 @@ patchesStrategicMerge: []
 # add this line to set model-server endpoint to kepler
 # - ./patch/patch-model-server-kepler-config.yaml
 # add this line to change log level for debugging
-# - ./patch/patch-debug.yaml
\ No newline at end of file
+# - ./patch/patch-debug.yaml
+# add this line to get high metric granularity (3s interval)
+# - ./patch/patch-high-granularity.yaml
```
rootfs, line 30: is this file coming next?

mcalman: @rootfs good catch. I have added the file

rootfs (approved):

sunya-ch:
> @sunya-ch can we increase the Prometheus scrapping interval?

Yes. I think it is fine. Here are results of stressng stressing on CPU with 7 different stress load on VM using estimator sidecar with non-linear power model (with PR https://github.com/sustainable-computing-io/kepler/pull/976).

Screenshot 2023-10-10 at 19 16 45

Screenshot 2023-10-10 at 19 17 36

marceloamaral:
@sunya-ch how long the workload was running?

The impact in the accuracy looks significant.... :/

sthaha:
@mcalman, thank you for the patch. We were thinking of providing a similar feature in Kepler operator to control the `fidelity | resolution | granularity` 

@marceloamaral, @rootfs, In this regard I have had a discussion offline with @simonpasquier (who is a prometheus maintainer) about the scrape interval. It seems having too low (3s) scrape interval can lead to failed scapes when the amount of metrics produced is too large. 

IIRC, the recommendation was to keep the scrape interval between 15s-2m and have prometheusrules (if any) to be evaluated at the same or greater interval

@simonpasquier, could you please shed more light on this?

marceloamaral (changes requested):
@mcalman
I think the default Prometheus scrapping interval is 15s in Vanila Kubernetes and 30s in OpenShift.
Can you update it to 15s and maybe patch to 30s for OpenShift deployment?

mcalman:
> @mcalman I think the default Prometheus scrapping interval is 15s in Vanila Kubernetes and 30s in OpenShift. Can you update it to 15s and maybe patch to 30s for OpenShift deployment?

@marceloamaral It seems different versions of Prometheus also have different defaults. For example, the version deployed in the [kube-prometheus stack](https://github.com/prometheus-community/helm-charts/blob/29916435a6c1b157b5372a3a398808a87dca9517/charts/kube-prometheus-stack/values.yaml#L2878C6-L2878C6) scrapes at 30s intervals by default. If we'd like to always use the Prometheus default value we can also remove the `interval` field from [prometheus_service_monitor.yaml](https://github.com/sustainable-computing-io/kepler/pull/970/files#diff-f3746997fef4bd7ed174d9241e8b4b47127f55f5c0487ed3901e9831f1f548ea) which will cause Prometheus's global scrape interval to be used.

Should we only specify a custom Prometheus scrape interval for Kepler when the `HIGH_GRANULARITY` option is set and let the global scrape interval be used otherwise?

sthaha:
The default scrapeInterval defined by Prometheus CRD is 30s - https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/prometheus_types.go#L217

marceloamaral (approved):
/lgtm

Great, let's set it to 30s then!

PR #955, opened 2023-10-03
Fix IsEstimatedIdlePowerEnabled function

marceloamaral:
Fix the bug in the function IsEstimatedIdlePowerEnabled

PR #975, opened 2023-10-10
config: fix expose idle power for BMs

marceloamaral:
The PR https://github.com/sustainable-computing-io/kepler/pull/955 only fixed the `IsEstimatedIdlePowerEnabled` function, but we need to also fix the `SetEnabledEstimatedIdlePower`...

This PR fixes the `SetEnabledEstimatedIdlePower`


sunya-ch (approved):
/lgtm

PR #996, opened 2023-10-15
Add ebpfprovider to align with kepler-action

### LINKED ISSUE #962: for platform validation will self host owner install ebpf before the GHA been triggered?
https://github.com/sustainable-computing-io/kepler/blob/a1b042d87a9f46e5403ec4eecebb17e2b3916402/.github/workflows/platform-validation.yml#L136

as kepler-action GHA will install ebpf(either bcc or libbpf with 0.0.2 version) hence the question here is:
for platform validation will self host owner install ebpf before the GHA been triggered?

or we should hard code a configuration here so that  kepler-action will install ebpf during process?

jiere:
1. Fix #962 
2. Change the port forwarding destination port of kepler-exporter and prometheus-k8s-0 from well-known port to avoid conflict when the test runner also joined other Kubernetes cluster where Kepler/Prometheus also deployed.

### DIFF
--- a/.github/workflows/platform-validation.yml
+++ b/.github/workflows/platform-validation.yml
@@ -136,6 +136,7 @@ jobs:
         uses: sustainable-computing-io/kepler-action@v0.0.2
         with:
           cluster_provider: ${{matrix.cluster_provider}}
+          ebpfprovider: bcc
           local_dev_cluster_version: v0.0.3
           prometheus_enable: true
           grafana_enable: true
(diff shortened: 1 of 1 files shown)


rootfs (approved):

.github/workflows/platform-validation.yml
```diff
@@ -136,6 +136,7 @@ jobs:
         uses: sustainable-computing-io/kepler-action@v0.0.2
         with:
           cluster_provider: ${{matrix.cluster_provider}}
+          ebpfprovider: bcc
```
marceloamaral, line 139: we should use libbpf

PR #1009, opened 2023-10-17
feat(dev): allow .env to be sourced when bringing up/down a cluster

sthaha:
Previously, cluster-up.sh and cluster-down.sh both had similar code to clone the repo. This is now moved to a single script (less maintenance). The scripts also `cd` in local-dev-cluster dir which meant that a .env file at the root of kepler project will not be honored by the local-dev-cluster/main.sh.

This commit modifies the cluster-up and down targets to
   * clone local-dev-cluster to tmp dir (which is ignored by git)
   * sources .env in project root where settings that needs to passed down to local-dev-cluster can be kept. 
   * E.g.  the `.env` can contain something like 
   ```sh 
    LOCAL_DEV_CLUSTER_DIR="$PROJECT/tmp/local-dev-cluster"
   PROMETHEUS_ENABLE=true
   GRAFANA_ENABLE=true
 ```


sthaha:
Additional eyes on this would be appreciated  -  @vprashar2929, @marceloamaral

hack/cluster.sh
```diff
@@ -0,0 +1,61 @@
...
+main() {
+	local op="$1"
+	shift
+
+	cd "$PROJECT_ROOT"
+	mkdir -p "${TMP_DIR}"
+
+	clone_local_dev_cluster
+	echo "Bringing cluster - $op"
+	"$LOCAL_DEV_CLUSTER_DIR/main.sh" "$op"
```
sthaha, line 58: NOTE: this allows for creation of `on_cluster_up` and `on_cluster_down` functions that can apply additional configs to the local-dev-cluster. 
E.g. https://github.com/sustainable-computing-io/kepler-operator/blob/c669624c53e5579da453e05d7254cc7c080322c6/hack/cluster.sh#L88 install OLM for `kepler-operator`

vprashar2929: Shouldn't we add rbac work in local-dev-cluster so that It can be used everywhere?🤔

hack/cluster.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+set -eu -o pipefail
+
+# NOTE: assumes that the project root is one level up
+PROJECT_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." >/dev/null 2>&1 && pwd)
+declare -r PROJECT_ROOT
+
+# NOTE: this allows common settings to be stored as `.env` file
+# shellcheck disable=SC1091
+source "$PROJECT_ROOT/.env" || true
```
vprashar2929, line 27: script fails if `.env` is not present. I think we need to add a check if env is present and then only source it

hack/cluster.sh
sthaha: Sure, I would think so but that isn't the case :-). I was only pointing at an example

hack/cluster.sh
sthaha: @vprashar2929 Are you sure? Can you please paste the  output of the run?

marceloamaral (commented):
Looks good to me.

Please double check the `.env is not present` possible problem...

sthaha:
@marceloamaral , thank you 🙇 
@vprashar2929 and I went on a call to check this. This happens only in `bash  3.xx` (osx) but I have changed it so that the file is source only if it exists.

vprashar2929 (approved):
LGTM

hack/cluster.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+set -eu -o pipefail
+
+# NOTE: assumes that the project root is one level up
+PROJECT_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." >/dev/null 2>&1 && pwd)
+declare -r PROJECT_ROOT
+
+# NOTE: this allows common settings to be stored as `.env` file
+# shellcheck disable=SC1091
+[[ -f "$PROJECT_ROOT/.env" ]] && source "$PROJECT_ROOT/.env"
```
jiere, line 27: We should define the priority for such ENV definitions, now we have at least below considerations:
1. In github action, see example [here](https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/integration_test.yml#L101-L104)
2. add a new ".env" file like this PR mentioned.
3. In the "cluster.sh" itself, the appear sequence decides the final value.
4. The "declare" command itself, "-r" means read-only?
5. Some manual export ENVs on the test environment.

So IMO, such various of ENV definitions are not good, shall we remain the only one way, either define them in GHA, or  define them in consolidate place such as ".env"?

sthaha: This feature is intended for local development usecase only and not in CI. Hence the `.env` is ignored in git. 
Common scenario. 
`make cluster-up` bring a kind cluster but I want the cluster to be 
1. 3 nodes 
2. have grafana and prometheus installed
3. have local-dev-cluster cloned to tmp/ 

This is easily supported by this script.

sthaha: `The "declare" command itself, "-r" means read-only?` yes, it means any changes beyond this point should cause failure.

jiere (approved):
LGTM

Issue #1008, opened 2023-10-17
Bugs about VM metrics


LeiZhou-97:
### What happened?

I saw kepler support to capture vm process metrics. I tried it and met so many bugs, I don't know how it can the pass the CI successfully....

1. not mount the /var/run/libvirt, which means it is impossible for kepler to obtain vm information on the host. 
2. path join error https://github.com/sustainable-computing-io/kepler/blob/8c8dab40d82fcc952f2c00651c87470c7d868dfa/pkg/libvirt/resolve_vm.go#L35
```
I1017 02:22:25.436020       1 resolve_vm.go:41] fullPath: /proc/%s/task/proc/%s/task
```
3. not exclude libvirt `driver.pid`
https://github.com/sustainable-computing-io/kepler/blob/8c8dab40d82fcc952f2c00651c87470c7d868dfa/pkg/libvirt/resolve_vm.go#L70

.....

I can help to fix these bugs, but I found this part of the code will be refactored in the near feature. 

@aharivel @marceloamaral Any update https://github.com/sustainable-computing-io/kepler/pull/931?

### What did you expect to happen?

None

### How can we reproduce it (as minimally and precisely as possible)?

None

### Kepler image tag

<details>

</details>
 

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>
 None

### Cloud provider or bare metal

<details>

</details>
 None

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

aharivel:
Hi @LeiZhou-97 ,
Did you try it by installing Kepler via the RPM ?
If not you may encounter problem with path that are absolute to the Host compute. 

[...]

PR #1017, opened 2023-10-18
vm_metrics: fix libvirt path join error

LeiZhou-97:
Link #1008 

1. mount hostpath /var/run for manifests deployment.
2. fix path join error.

### DIFF
--- a/manifests/config/exporter/exporter.yaml
+++ b/manifests/config/exporter/exporter.yaml
@@ -90,6 +90,8 @@ spec:
           name: tracing
         - mountPath: /proc
           name: proc
+        - mountPath: /var/run
+          name: var-run
         - name: cfm
           mountPath: /etc/kepler/kepler.config
           readOnly: true
@@ -117,6 +119,10 @@ spec:
       - name: proc
         hostPath:
           path: /proc
+          type: Directory
+      - name: var-run
+        hostPath:
+          path: /var/run
           type: Directory
       - name: cfm
         configMap:
--- a/pkg/libvirt/resolve_vm.go
+++ b/pkg/libvirt/resolve_vm.go
@@ -27,17 +27,11 @@ const (
 	procPath    string = "/proc/%s/task"
 )
 
-func getThreadIDsForPID(pid, extraPath string) []string {
+func getThreadIDsForPID(pid, fullPath string) []string {
 	threadIDs := []string{}
-	fullPath := ""
-
-	if procPath != "" {
-		fullPath = filepath.Join(extraPath, procPath)
-	} else {
-		fullPath = procPath
-	}
 
 	procDir := fmt.Sprintf(fullPath, pid)
+
 	files, err := ioutil.ReadDir(procDir)
 	if err != nil {
 		return nil
--- a/pkg/libvirt/resolve_vm_test.go
+++ b/pkg/libvirt/resolve_vm_test.go
@@ -65,6 +65,7 @@ func TestGetCurrentVMPID(t *testing.T) {
 	mockProcDir := t.TempDir()
 	createMockProcDir(mockProcDir)
 
+	mockProcDir = filepath.Join(mockProcDir, procPath)
 	pidFiles, err := GetCurrentVMPID(mockLibvirtDir, mockProcDir)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)


LeiZhou-97:
@aharivel  

I do a quick fix for vm metrics. Please help to review it.

manifests/config/exporter/exporter.yaml
```diff
@@ -90,6 +90,8 @@ spec:
           name: tracing
         - mountPath: /proc
           name: proc
+        - mountPath: /var/run
+          name: var
```
sthaha, line 94: var-run

pkg/libvirt/resolve_vm.go
```diff
@@ -27,17 +27,11 @@ const (
 	procPath    string = "/proc/%s/task"
 )
 
-func getThreadIDsForPID(pid, extraPath string) []string {
+func getThreadIDsForPID(pid, fullPath string) []string {
```
sthaha, line 30: I see that the previous use of extraPath was more like a `basePath` (than an extraPath) which gets added in front of `procPath` - `/proc/%s/task` giving us `/extra/path/proc/%s/task` where `%s` is  replaced by `pid`.

This change requires caller to know the format string (which looks odd to me).  I also see that no other code change is make (e.g. the actual call to the getThreadIDsForPID is not modified) which makes me wonder how the fix is supposed to work 🤔

sthaha (changes requested):
I am not sure if the code changes look right to me.

pkg/libvirt/resolve_vm.go
LeiZhou-97: Hi, @sthaha 

Please see the call chain. In existing code, the extraPath and procPath have same value, and it will cause `fullpath` join error.
`extraPath=/proc/%s/task`
`procPath=/proc/%s/task`
`fullPath=/proc/%s/task/proc/%s/task`

For detailed information, please see the issue https://github.com/sustainable-computing-io/kepler/issues/1008

https://github.com/sustainable-computing-io/kepler/blob/63fb1c4434cb693642fb7b024900dcdfda101a6a/pkg/collector/container_hc_collector.go#L115

https://github.com/sustainable-computing-io/kepler/blob/63fb1c4434cb693642fb7b024900dcdfda101a6a/pkg/libvirt/resolve_vm.go#L57

https://github.com/sustainable-computing-io/kepler/blob/63fb1c4434cb693642fb7b024900dcdfda101a6a/pkg/libvirt/resolve_vm.go#L81

aharivel: Hi @LeiZhou-97 
Sorry for the delay. So I checked your modification and I'm astonished the mistake passed under my radar. 
I've tested this morning and with or without the modification I've got the same result. Which is weird but explain why I did not see it. 
To be honest your modification looks better. I do not remember why I wanted to Join two paths but in the end, reading /proc/PID/task/ is enough to get the list of TID attached to the PID. 
So thank you for correction. 
Anthony

aharivel (approved):

manifests/config/exporter/exporter.yaml
LeiZhou-97: Updated.

pkg/libvirt/resolve_vm.go
sthaha: Thanks @LeiZhou-97  that makes sense now

LeiZhou-97: @sthaha 
From my side, they have the same effect. But I don't know if Anthony may use this `procPath` const elsewhere in the future, so I keep it. And it seems will not cause performance issue.

LeiZhou-97:
@aharivel 

Please approve it again. Update the name of mount point based on the advice. Thanks!

pkg/libvirt/resolve_vm.go
aharivel: @sthaha @LeiZhou-97 That would be indeed the same. 
Let's wait on the re-factorization of the code and we will see if this path is used somewhere else or not.

sthaha (commented):
lgtm

sthaha:
@LeiZhou-97 , I think the merging will be blocked until all tests in CI pass.

LeiZhou-97:
Hi @sthaha,

CI fixed. Plz merge it.

sthaha (approved):
Looks good to me.

PR #1023, opened 2023-10-20
fix(build): support newer version of kubectl

sthaha:
kubectl version --short has been deprecated since 1.23. In the latest kubectl, the flag is removed and causes `build-manifest` to fail when checking for the version.

This fixes the version check by using --client -o json options and using jq to validate if the version is higher than expected - 1.21

### DIFF
--- a/hack/build-manifest.sh
+++ b/hack/build-manifest.sh
@@ -17,22 +17,33 @@
 # Copyright 2022 The Kepler Contributors
 #
 
-set -ex
+set -eu -o pipefail
 
+# constants
+PROJECT_ROOT="$(git rev-parse --show-toplevel)"
+declare -r PROJECT_ROOT
+source "$PROJECT_ROOT/hack/utils.bash"
+
+export PATH="$PROJECT_ROOT/tmp/bin:$PATH"
+
+command -v jq >/dev/null 2>&1 || {
+    die "jq is not installed; run make tools to install all necessary tools"
+}
+
+kc_version_ok=$(kubectl version --client -ojson |
+    jq '.clientVersion | (.major | tonumber) >= 1 and (.minor | tonumber) > 21')
+if [[ "$kc_version_ok" != true ]]; then
+    die "Please update the kubectl version to 1.21+ to support kustomize"
+fi
+
+set -x
 export CLUSTER_PROVIDER=${CLUSTER_PROVIDER:-kind}
-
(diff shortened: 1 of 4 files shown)


hack/build-manifest.sh
```diff
@@ -26,10 +26,11 @@ export CLUSTER_PROVIDER=${CLUSTER_PROVIDER:-kind}
 DEPLOY_OPTIONS=$1
 for opt in ${DEPLOY_OPTIONS}; do export "$opt"=true; done
 
-version=$(kubectl version --short | grep 'Client Version' | sed 's/.*v//g' | cut -b -4)
-if [ 1 -eq "$(echo "${version} < 1.21" | bc)" ]; then
-    echo "You need to update your kubectl version to 1.21+ to support kustomize"
-    exit 1
+kc_version_ok=$(kubectl version --client -ojson |
+  jq '.clientVersion | (.major | tonumber) >= 1 and (.minor | tonumber) > 21')
```
sthaha, line 30: @jiere @SamYuan1990 Do you know if this is checked in CI?

jichenjc: I guess not all env has `jq` installed ?

sthaha: Yes, but we can say the same for all tools we use (including `sed`), can't we :) ? 

In those env, the script fails at which point the user should install jq. 
The way we solve this in `kepler-operator` is to provide a `make tools` target that ensures tools used by the project are installed to `<project-root>/tmp/bin` 
see: https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/hack/tools.sh which is what `make tools` invokes. 

I am happy to copy the script from there to our repo.

vprashar2929 (approved):
LGTM

hack/build-manifest.sh
jichenjc: yes, If we can do that..

sed is almost pre-installed in most of dev linux env but jq is not .. 
though the dev might find that error but if we can detect or error out it will be better

sthaha: I will start by submitting the `make tools` first and the every script can start by `validate_prerequisites` which checks if required binaries are present.

sthaha:
@jichenjc could you please take another look at this pr?

jichenjc (approved):

PR #1028, opened 2023-10-27
fix: avoid driver.pid since it's a daemon, not a VM

jichenjc:


rootfs:
@LeiZhou-97 @aharivel

aharivel:
Hi Ji,

I've been considering the importance of tracking the "driver.pid" within the context of the libvirt daemon. As you know, this identifier is closely associated with the libvirt daemon and plays a significant role in managing the necessary tools for running virtual machines (VMs).

It's crucial to acknowledge that while the "driver.pid" primarily represents the libvirt daemon, it also encapsulates a portion of the essential tools required to facilitate the operation of VMs. Considering the comprehensive tracking of resource consumption on a host node, it's necessary to factor in not only the individual VMs' power consumption (VM1.pid, VM2.pid, etc.) but also account for the resource utilization attributed to the "driver.pid."

In essence, to accurately assess the total consumption of VMs on a host node, it's imperative to consider the cumulative power consumption of all individual VMs alongside the contributions of the "driver.pid" associated with the libvirt daemon.

I believe it's an important aspect to consider for a more comprehensive understanding of resource utilization within our virtualized environment.

Looking forward to your thoughts on this matter.
Anthony

jichenjc:
>>>In essence, to accurately assess the total consumption of VMs on a host node, it's imperative to consider the cumulative power consumption of all individual VMs alongside the contributions of the "driver.pid" associated with the libvirt daemon.

kepler has a model to monitoring the `process` , so I think libvirtd is essentially represented by process already
and VM is a special process that we split from `process` model as all KVM VM actually is a qemu process

so I think we need be clear on how we count our model and monitoring the target object (container, process , VM etc) 
instead of double count .. thoughts? @aharivel

aharivel:
> so I think we need be clear on how we count our model and monitoring the target object (container, process , VM etc) instead of double count .. thoughts? @aharivel

Well yeah OK, you are right. We should stick to the monitoring target object and it's another problem (documentation,blog,...) on how  to really monitor the VMs on host node.

aharivel (approved):

marceloamaral:
The libvirt driver runs in the background and may have increased load due to virtual machine activity, but other background processes and the operating system might also have. However, correlating these background processes with specific containers and VMs is a challenging task, since the driver resource utilization is more related to api calls than the container or vm utilization....

In any case, we should account for background power and resource utilization separately from the virtual machines. Later on, we can consider allocating these background processes to the VMs. The specific phase for this is still not clear. But If Kepler is running in standalone mode, we might handle it internally. Otherwise, it kepler is within Kubernetes, it might be best for the user to manage this in Prometheus to maintain granularity.

And sorry for the delay, I will submit the code refactoring updating this soon. I am fixing the tests before submitting a PR...

jichenjc:
>>>And sorry for the delay, I will submit the code refactoring updating this soon. I am fixing the tests before submitting a PR...

@marceloamaral is the fix submitted? want to know what's the proposed change.. thanks

jichenjc:
ok, I found this , will read the code 
```

    * libvirt: find vm id from cgroup path

```

marceloamaral:
@jichenjc, the updated code now only associates processes within the VM cgroup. 
All other background processes are treated as background processes, such as the vm hypervisor.

Issue #1021, opened 2023-10-18
Explore 'CPUID' alternative


rootfs:
### What would you like to be added?

To reduce the kepler dependency, we check how to minimize other binaries used by kepler. One of those is `cpuid`. On most platforms `lscpu` is available as default. We can try to use `lscpu` if `cpuid` is missing.

### Why is this needed?

Provide a fall back plan if `cpuid` is missing

rootfs:
cc @vimalk78

jiere:
PR raised already, lack of non-Intel platform, especially AMD/ARM platform to verify.
I would be appreciated to accept any help for "lscpu" output info on different AMD/ARM architecture platforms.

vimalk78:
currently we use `cpuid -1 | grep uarch` to identify cpu vendor and _exact cpu micro architecture_ . 
```
 $ cpuid -1 | grep uarch
   (uarch synth) = Intel Willow Cove {optim of Sunny Cove}, 10nm++
```

@rootfs  `cpuid` uses [`CPUID` instruction](https://www.felixcloutier.com/x86/cpuid) to obtain this information. Using `CPUID` instruction is a reliable source of this information.

whereas `lscpu` or `cat /proc/cpuinfo` can only give 
```
Vendor ID:               GenuineIntel
  Model name:            11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
```
It does not give micro architecture of cpu.

So it depends on what we need in our power model?, if we need exact micro architecture, we need to find alternative which can get this information.

vimalk78:
we should be able to map processor model number from `lscpu` and obtain its micro architecture from intel ark site

also, there is some information available in wikichip
list of micro-architectures https://en.wikichip.org/wiki/Category:all_microarchitectures including intel architectures. 
e.g. this page https://en.wikichip.org/wiki/intel/core_i7 has links for all intel processors, and each processor's page has info about its micro-architecture. but not sure how reliable is this info.

jiere:
> So it depends on what we need in our power model?, if we need exact micro architecture, we need to find alternative which can get this information.

IIUC, we use the micro architecture as "index" for power modeling, however, not all the vendors have such microarchitecture definition list, and even for Intel, there is product codename and microarchitecture mapping, sometimes it's tricky to understand the logic of synthesize in CPUID output, see example 1:
```
$ cpuid --version
cpuid version 20230505

$ lscpu |grep Model
[...]

PR #1029, opened 2023-10-29
CPUID alternative solution

jiere:
PR for issue #1021 
    CPUID alternative solution, major changes are as follows:
    1. Move original Intel specific script and output file to `data/intel_cpu` directory.
    2. Move legacy power model csv files to `data/legacy` directory.
    3. Obsolete `normalized_cpu_arch.csv`, introduce `cpus.yaml` to involve more data mapping.
    4. Introduce `github.com/klauspost/cpuid/v2` library to replace the raw `lscpu` command call.
    5. Add new function(getCPUMircoarchitecture) to fetch uarch from `cpus.yaml`.
    6. For CPU models not matched in `cpus.yaml`, read `/sys/devices/cpu/pmu_name` as uarch.
    7. Align with latest Linux kernel for Intel CPU family models.
    8. Update README document and Dockerfile accordingly.
    9. Update kepler validator side code accordingly.


jiere:
@sunya-ch , please also help check if this PR change has any impact on kepler-model-server side? Thanks.

LeiZhou-97:
@jiere 
If you use cpuid as an alternative, I think it's better to use cpuid go pkg instead of linux raw cmd. https://github.com/klauspost/cpuid

jiere:
> @jiere If you use cpu model as an alternative solution, I think it's better to use cpuid go pkg instead of linux raw cmd. https://github.com/klauspost/cpuid

Thanks @LeiZhou-97 for pointing out this! @vimalk78  also mentioned another pkg [here](https://github.com/sustainable-computing-io/kepler/issues/1021#issuecomment-1784513126). Let me have a try and pick the better one.

jiere:
> @jiere If you use cpu model as an alternative solution, I think it's better to use cpuid go pkg instead of linux raw cmd. https://github.com/klauspost/cpuid

Just have a try, this tool is pretty cool! I think we could forget `lscpu` now:)
What a pitty is that it cannot detect those microarchitecture information in CPUID, even if the pkg name is called "cpuid" :(
otherwise we even no need to build cpuid in containerized Kepler.
Maybe we can raise an issue there~

There is a discussion [thread](https://github.com/sustainable-computing-io/kepler/discussions/1031), please feel free to share your comments there, @LeiZhou-97:)

LeiZhou-97:
> > @jiere If you use cpu model as an alternative solution, I think it's better to use cpuid go pkg instead of linux raw cmd. https://github.com/klauspost/cpuid
> 
> Just have a try, this tool is pretty cool! I think we could forget `lscpu` now:) What a pitty is that it cannot detect those microarchitecture information in CPUID, even if the pkg name is called "cpuid" :( otherwise we even no need to build cpuid in containerized Kepler. Maybe we can raise an issue there~
> 
> There is a discussion [thread](https://github.com/sustainable-computing-io/kepler/discussions/1031), please feel free to share your comments there, @LeiZhou-97:)

Before I submit the PR that uses `CPUID` tool to recognize the CPU, I've already the explore this go pkg. If we want it to support microarch or some other info, a CPU mapping list has to be created in advance, which needs huge effort.

source code of cpuid: https://salsa.debian.org/debian/cpuid/-/blob/master/cpuid.c#L2105

marceloamaral (changes requested):
I think was have discussed that a long time ago.
Why we don't use a golang library?
- https://github.com/shirou/gopsutil (VendorID, Family, Model, ModelName)
- https://github.com/klauspost/cpuid
- https://github.com/aregm/cpuid
- https://pkg.go.dev/internal/cpu (name)

vimalk78:
there is also https://github.com/intel/svr-info/tree/main/src/pkg/cpu


It can get the microarchitecture name based on `cpu-family`, `cpu-model`, and `steppings` (optional)
https://github.com/intel/svr-info/blob/main/src/pkg/cpu/cpu.go#L83

the cpu family, model, steppings can be found from `lscpu` or from other go packages.

it has cached info on intel (core,xeon), amd and arm cpus https://github.com/intel/svr-info/blob/main/src/pkg/cpu/resources/cpus.yaml

@marceloamaral ptal

jiere:
Thanks all for your input above, really helpful and full of insights:)

As @LeiZhou-97 mentioned the cpuid.c [code](https://salsa.debian.org/debian/cpuid/-/blob/master/cpuid.c#L2105) and @vimalk78 pointed svr-info library's cpu part [usage](https://github.com/intel/svr-info/blob/main/src/pkg/cpu/cpu.go#L83) shown,
both need a CPU mapping list, this is dependency, we need to find a place where the list is well maintained and updated up-to-date, either we maintain it in Kepler or we depends on other Golang pkg/library.

It looks like better [here](https://github.com/intel/svr-info/blob/main/src/pkg/cpu/resources/cpus.yaml), even though many CPU models N/A there.

So please help to confirm in discussion [thread](https://github.com/sustainable-computing-io/kepler/discussions/1031) that we are using microarchitecture for power modeling and will continue use it later.

marceloamaral:
Please, see my comment here: https://github.com/sustainable-computing-io/kepler/issues/1021#issuecomment-1788570916

pkg/collector/metric/utils.go
```diff
@@ -184,6 +187,54 @@ func getS390xArchitecture() (string, error) {
...
+	pipe, err := output.StdoutPipe()
+	if err != nil {
+		return ""
+	}
+	defer pipe.Close()
+
+	grep.Stdin = pipe
+	err = output.Start()
+	if err != nil {
+		klog.Errorf("lscpu command start failure: %s", err)
```
rootfs, line 214: `lscpu` -> `cpuid`

pkg/collector/metric/utils.go
```diff
@@ -184,6 +187,54 @@ func getS390xArchitecture() (string, error) {
...
+		return ""
+	}
+	res, err := grep.Output()
+	if err != nil {
+		klog.Errorf("grep lscpu command output failure: %s", err)
```
rootfs, line 219: same

pkg/collector/metric/utils.go
```diff
@@ -184,6 +187,54 @@ func getS390xArchitecture() (string, error) {
...
+// ...
+// Intel,SAPPHIRERAPIDS_X,143,Golden Cove
+// ...
+//
+// We can use the 'lscpu' command, which exists in most of the Linux Distros, to fetch the CPU microarchitecture information like this:
+// $ lscpu |grep Model:
+// Model:                           143
+// In this example, the Model number '143' matches Processor codename: 'SAPPHIRERAPIDS_X' and microarchitecture: 'Golden Cove'.
+func getCPUModel() string {
+	// use lscpu to get CPU Model number
```
rootfs, line 202: this function seems to use cpuid not lscpu

jiere: > this function seems to use cpuid not lscpu

sorry,  here is typo in line 204, should be "lscpu", but as you know, in this PR comments thread, folks gave lots of constructive suggestions to use golang pkg/lib instead of "lscpu", I will figure out a new diff soon.

build/Dockerfile.libbpf.kepler
```diff
@@ -29,7 +29,7 @@ COPY --from=builder /opt/app-root/src/github.com/sustainable-computing-io/kepler
 
 RUN mkdir -p /var/lib/kepler/data
 RUN mkdir -p /var/lib/kepler/bpfassets
-COPY --from=builder /opt/app-root/src/github.com/sustainable-computing-io/kepler/data/normalized_cpu_arch.csv /var/lib/kepler/data/normalized_cpu_arch.csv
+COPY --from=builder /opt/app-root/src/github.com/sustainable-computing-io/kepler/data/cpus.yaml /var/lib/kepler/data/cpus.yaml
```
marceloamaral, line 32: Can we download it from a repository that is actively updated?
I am wondering who will maintain this file updated.

Also, again, can we just use the `/proc/cpuinfo`?

jiere: > Can we download it from a repository that is actively updated?
> I am wondering who will maintain this file updated.

Yes, this is an issue, however, up till now, I have not found any golang pkg/lib which can do such maintenance.
I once submitted an [issue](https://github.com/klauspost/cpuid/issues/135) in  `github.com/klauspost/cpuid` which is imported in current PR, but the maintainer also could not commit for maintaining the list.

So I preferred to upload the current version of cpus.yaml, and for new coming CPU models, we need to add them into this yaml separately, since nobody could do this, we Kepler maintain them by ourselves...

Due to the public info download feasibility, I had some offline sync with @vimalk78 actually, we found that both the [Intel ark](https://ark.intel.com/content/www/us/en/ark.html#@Processors) and the [WikiChip](https://en.wikichip.org/wiki/WikiChip) were not adequate for the job. 
The former one is Intel only and from HTML, we could only get the mapping of `Brand name` and `uarch`, the `uarch` list itself also comes from our maintained [srcript](https://github.com/sustainable-computing-io/kepler/blob/main/data/download_intel_cpu_models.sh)
The latter one holds more CPU models from different vendors, however the webpage is not friendly for scripts or web crawlers, still huge efforts for such mapping.

So this is why the best choice is leveraging tools such as `cpuid` which is well maintained by [Todd Allen](<todd.allen@etallen.com>) and widely used by Linux Distros. 
Since we are introducing an alternative or fallback plan for `cpuid` here, we can afford such maintenance and I have finished almost the  existed not EOL ones(except some AMD/ARM/RISC-V information which are not easy to find or even cannot support in short-term)

jiere: > Also, again, can we just use the `/proc/cpuinfo`

Again, the issue for `/proc/cpuinfo` is that there are only five items related to CPU models:
i.e.
vendor_id       : GenuineIntel
cpu family      : 6
model           : 140
model name      : 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
stepping        : 1

Which one could we pick for the `id` of CPU model and without any maintenance efforts? 
There could be various of `model name` for the same microarchitecture, can we afford the efforts for training models for each?

vimalk78: @jiere , i think @marceloamaral 's suggestion is to replace `microarchitecture` with a tuple of of `{vendor_id, cpu_family, cpu_model,_no, cpu_model_name, stepping}`. If this tuple maps to a unique `microarchitecture`,  it should be able to serve same purpose as `microarchitecture` in our power model.

jiere: @vimalk78 , got your point, but actually the information of CPU model finally becomes the metrics labels, if we use tuple, means more labels will be associated to the related metrics, this will bring more confuse to customer and increase later PromQL and kepler-model-server complexity, this is what we should take consideration...

vimalk78: hmm, hadn't considered this.

rootfs: @jiere i can get cpu architecture on my laptop via
```console
# cat /sys/devices/cpu/caps/pmu_name
skylake
```

can you confirm if this also works on your end?

rootfs: the kernel code for `pmu_name` is at https://github.com/torvalds/linux/blob/master/arch/x86/events/intel/core.c

marceloamaral: @rootfs great finding! It works for me. I tested two different systems.
```
$ cat /proc/cpuinfo | grep "model name" -m 1
model name	: Intel(R) Xeon(R) Platinum 8260 CPU @ 2.40GHz
$ cat /sys/devices/cpu/caps/pmu_name
skylake

$ cat /proc/cpuinfo | grep "model name" -m 1
model name	: Intel(R) Xeon(R) CPU E5-2667 v3 @ 3.20GHz
$ cat /sys/devices/cpu/caps/pmu_name
haswell
```


Moreover, I was thinking about using the model name `Platinum_8260`,  architecture `skylake`, and the number of chips `2` (used sockets). As in the [SpecPower Data](https://spec.org/power_ssj2008/results/res2023q1/). We might utilize this data later to construct power models, so sticking to this pattern could be beneficial.

Keep in mind that different model names may have different power consumption. But in cases where a specific power model is unavailable, we can choose a model randomly for a given architecture. While this approach may not precisely reflect real power, it can provide to the users some data for reference.

rootfs: That's great! Yes, adding sockets to power models should improve the accuracy. If we do a study of running VMs with different socket configurations (single or multiple), we can get more ideas about the benefits.

vimalk78: just observation, there is a difference between value reported by `cpuid` and `pmu_name`

```
vimalkum ~ $ cat /sys/devices/cpu/caps/pmu_name 
icelake
vimalkum ~ $ cpuid -1 | grep uarch
   (uarch synth) = Intel Willow Cove {optim of Sunny Cove}, 10nm++
```

also found this repo https://github.com/archspec/archspec  will try using it and report.
docs: https://archspec.readthedocs.io/en/latest/microarchitectures.html#host-detection

jiere: Thanks guys for your great findings, I also checked on my various of local machines, several issues found yet:
1. wording issues, just as @vimalk78  found above, the string output is different with `cpuid -1 |grep uarch`
2. For some pretty new CPU models, even though the `cpu model number` has been updated in my current PR, and updated also in kernel [model definition](https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/intel-family.h#L111), but not updated accordingly on `pmu_name` side:
```
# lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         45 bits physical, 57 bits virtual
  Byte Order:            Little Endian
CPU(s):                  224
  On-line CPU(s) list:   0-223
Vendor ID:               GenuineIntel
  Model name:            GENUINE INTEL(R) XEON(R)
    CPU family:          6
    Model:               173
    Thread(s) per core:  2
    Core(s) per socket:  56
    Socket(s):           2
    Stepping:            0
```
So the result of `pmu_name` is wrong on that machine:
```
# cat /sys/devices/cpu/caps/pmu_name
sapphire_rapids
```

In a word, all the alternative solutions which leverage the current Linux SYSFS output depends on the related kernel code upstream progress, different part has different implementation, maybe done by different maintainers, the consistence  is a big issue...

jiere:
BTW, folks, we should first achieve agreement that this feature ticket is for alternative/fallback plan for `cpuid` :)
If so, the behavior between the `cpuid` solution and the alternative solution should be consistent.
No matter which one is adopted in production environment, the result of the CPU model info is used for specific power model identification, so both model train and model inference should match the same one.

marceloamaral:
Exactly, we should first try the `model name` and then, if not found, we select one model for a given `arch name`.

@jiere, let's use you PR then to find the `arch name`, as you spent a lot of time updating the cpu list. And then, we use the `pmu_name` if the CPU is not in the list.

jiere:
Thanks @rootfs and @marceloamaral for your comments, let me kick out another diff later.

jiere:
@all, I have pushed the latest diff, which supports fixing #1066 also.
Please take a look and see if any further comments? Thanks.

rootfs (approved):

PR #1030, opened 2023-10-30
Support hmc client to get energy consumption from IBM System Z

jiangphcn:
The purpose of this PR Support is to support external power source: HMC referred in https://github.com/sustainable-computing-io/kepler/issues/644.

Starting IBM Z 16, LPAR level power metrics is exposed and available. The golang zhmclient package is enhanced and open-sourced in https://github.com/zhmcclient/golang-zhmcclient/commit/6bb10a7dd31e36d973ede6b0dad56b2e467cf5b5.

Current PR is using such API to provide power consumption calculation for applications running in IBM System Z.


pkg/power/platform/source/hmc.go
```diff
@@ -16,17 +13,129 @@ limitations under the License.
...
+	lparURI := "api/logical-partitions/" + os.Getenv("LPAR_ID")
+	props := &zhmcclient.EnergyRequestPayload{
+		Range:      "last-day",
+		Resolution: "fifteen-minutes",
+	}
+	energy, _, err := hmcManager.GetEnergyDetailsforLPAR(lparURI, props)
+	if err != nil {
+		klog.V(1).Infof("Error getting energy data: %v", err.Message)
+	}
+	klog.V(1).Infof("Get energy data successfully")
```
jichenjc, line 69: this seems will lead to many logs?

pkg/power/platform/source/hmc.go
```diff
@@ -16,17 +13,129 @@ limitations under the License.
...
+	return 0, nil
+}
+
+func (r *PowerHMC) GetEnergyFromPackage() (uint64, error) {
+	return 0, nil
+}
+
+func (r *PowerHMC) GetNodeComponentsEnergy() map[int]source.NodeComponentsEnergy {
+	pkgEnergy, _ := r.GetLiveEnergyFromLpar()
+	pkgEnergy = pkgEnergy * 3
```
rootfs, line 110: what is '*3' for?

pkg/power/platform/source/hmc.go
```diff
@@ -16,17 +13,129 @@ limitations under the License.
...
+		Pkg:    pkgEnergy,
+	}
+	return componentsEnergies
+}
 
-func (a *PowerHMC) StopPower() {
+func (r *PowerHMC) GetPlatformEnergy() map[string]float64 {
+	pkgEnergy, _ := r.GetLiveEnergyFromLpar()
+	platformEnergies := make(map[string]float64)
+	platformEnergies[hmcSensorID] = float64(pkgEnergy) * 3
```
rootfs, line 127: is lpar energy at the platform level or pkg level? if it is a platform level energy source, we can probably skip getting pkg energy

jichenjc: I guess there is no pkg concept in z platform

@jiangphcn can we have a doc describe the relationship between z and x86 concept difference so we know what energy mapping to existing defined variables to give a clear picture?

pkg/power/platform/source/hmc.go
jiangphcn: It is aimed to align with `defaultSamplePeriodSec = 3` used in the kepler. Still try to finalize it and then leave explanation in the comment.

PR #1032, opened 2023-10-30
fix(Makefile): enable `libbpf` based local build

vimalk78:
This commit enables local build of kepler with libbpf.

Additional packages required to build (for fedora): `libbpf-devel` `libbpf-static`

usage: `ATTACHER_TAG=libbpf make build`


vimalk78:
Cc: @rootfs

sthaha:
Thanks for the patch. I have verified that this works.

sthaha (approved):

PR #1035, opened 2023-11-03
feat(dashboard): improve responsiveness using  prometheus rules

sthaha:
This commit adds Prometheus Rules for both (low and high resolution) to pre-compute the queries used in grafana dashboards. This reduces the time and load on Prometheus when running dashboards improving the responsiveness.


sthaha:
Please note that the initial computation takes some time (30s for low resolution) and 15s for high resolution

rootfs (approved):

rootfs:
thanks for this idea. Can you post screen shots and usage instructions in kepler doc too?

Issue #913, opened 2023-09-05
single build too long


jichenjc:
### What would you like to be added?

try to run make , every time it takes around 4+ min which is way more slower than before (I remember it's only about 1-2 min) before, it's ok to run this once slow but later on should be faster 

full log below
```
[root@jitest48 kepler]# date;make;date
Tue Sep  5 10:08:26 UTC 2023
GO111MODULE=off go get -u github.com/go-bindata/go-bindata/...
./hack/bindata.sh
go mod tidy -v
go mod vendor
./automation/presubmit-tests/gofmt.sh
Checking go format
echo BIN_TIMESTAMP==2023-09-05T10:08:29Z
BIN_TIMESTAMP==2023-09-05T10:08:29Z
/bin/podman build -t quay.io/sustainable_computing_io/kepler:v0.5.4-47-g02c16c1-linux-amd64-bcc \
        -f /root/kepler/build/Dockerfile.bcc.kepler \
        --network host \
        --build-arg SOURCE_GIT_TAG=v0.5.4-47-g02c16c1 \
        --build-arg BIN_TIMESTAMP=2023-09-05T10:08:29Z \
        --platform="linux/amd64" \
        .
[1/2] STEP 1/11: FROM quay.io/sustainable_computing_io/kepler_builder:ubi-9-bcc-0.26-go1.18 AS builder
[1/2] STEP 2/11: ARG BIN_TIMESTAMP
--> Using cache 646807b3cf5425d33d785a82b01215fae199b001cdd868c03a1defb0398e452f
--> 646807b3cf54
[1/2] STEP 3/11: ARG SOURCE_GIT_TAG
--> Using cache 136dcf519b6a8b957df5d0d9b5e7b42d74b96e7c15f057f058f570ddf365e373
--> 136dcf519b6a
[1/2] STEP 4/11: LABEL name=kepler-builder
--> Using cache 97607d165afdf8716b3f1d69a6e3087dacee10a7d7470e5e360ba6856d4640e4
--> 97607d165afd
[1/2] STEP 5/11: ENV GOPATH=/opt/app-root GO111MODULE=on GOROOT=/usr/local/go
--> Using cache a688e80219f3095633ec0717ab56cfb772f85e9a94dc538b2415ab5d7b68115b
--> a688e80219f3
[1/2] STEP 6/11: ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
[...]

PR #1054, opened 2023-11-15
CI: try to set GO111MODULE to off and use vendor folder for build(#913)

SamYuan1990:
As we have dep bot looks after our dependence , hence we don't need go mod tidy to update our dependence before our build. We can make the go build relays on our vendor to save time for image build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,7 +185,7 @@ clean-cross-build:
 build: clean_build_local _build_local copy_build_local
 .PHONY: build
 
-_build_local: tidy-vendor format
+_build_local:
 	@echo TAGS=$(GO_BUILD_TAGS)
 	@mkdir -p "$(CROSS_BUILD_BINDIR)/$(GOOS)_$(GOARCH)"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -185,7 +185,7 @@ clean-cross-build:
 build: clean_build_local _build_local copy_build_local
 .PHONY: build
 
-_build_local: tidy-vendor format
+_build_local:
```
sthaha, line 188: I think CI must check if vendor is up to date and that we don't accept patches that have difference in `vendor` directory if it isn't up to date.  With vendoring there is a potential for someone to accidentally add code they may have used for testing by editing the vendor/ files.

build/Dockerfile
```diff
@@ -8,7 +8,7 @@ ARG SOURCE_GIT_TAG
 
 LABEL name=kepler-builder
 
-ENV GOPATH=/opt/app-root GO111MODULE=on GOROOT=/usr/local/go 
+ENV GOPATH=/opt/app-root GO111MODULE=off GOROOT=/usr/local/go
```
sthaha, line 11: can we not rely on caches and use go modules?

sthaha:
> hence we don't need `go mod tidy` to update our dependence before our build. 

I think in CI we must assert that the go.mod and go.sum do not change after `go mod tidy` to ensure that these files are up to date. If we want to rely on `vendor` we should run `go mod vendor`  (and fail validation if git diff --exit) to ensure that the changes are reflected in vendor as well.

Issue #1047, opened 2023-11-12
Keep Kepler container image up to date


rootfs:
### What would you like to be added?

Some rpm packages used in the base images are subject to change. We should run `yum update` in the kepler image each time so to pick up the latest rpm packages.

For certain reason, adding `yum update -y` in the Kepler dockerfile runs into issues. Creating this issue to track the requirement

### Why is this needed?

Update rpms that are installed by the base image in Kepler image

SamYuan1990:
break down:
1. make yum packages installed by yum repo so that yum update works.
2. make CI enable to build base image.
3. bump up base image with golang 1.21 with new CI.
4. bump up golang version to golang 1.21 in go mod.
5. bump up navida base image and try to merge navida's package version bump up(reported by dep bot.)

PR #1045, opened 2023-11-11
images: add script to build and push images

rootfs:

PR #939, opened 2023-09-18
Bump up to Go 1.21

SamYuan1990:
as fix for #934

Issue #934, opened 2023-09-17
Bump up to golang 1.21


SamYuan1990:
Bump up to golang 1.21

_Originally posted by @SamYuan1990 in https://github.com/sustainable-computing-io/kepler/pull/924#discussion_r1327891431_

kmova:
can I help with this issue?

rootfs:
contributions are always welcome

SamYuan1990:
currently our base image is registry.access.redhat.com/ubi9/go-toolset:1.20 @rootfs , may I know can we have registry.access.redhat.com/ubi9/go-toolset:1.21?

SamYuan1990:
Current base image as ubi just have 1.20 golang version supported.

dave-tucker:
Closed via https://github.com/sustainable-computing-io/kepler/pull/1404

PR #1068, opened 2023-11-20
WIP chore(Dockerfile): Simplify Dockerfile

vimalk78:
- Adds Dockerfile.base which installs everything required to build kepler code base
 - Update golang to go1.20
 - Adds Dockerfile which builds kepler with libbpf attacher
 - Adds bpftool to the kepler image
 - Size of kepler image reduced from 2.3GB to 78.5 MB

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,0 +1,33 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+
+USER 0
+
+RUN yum -y install yum-utils
+RUN yum-config-manager --enable ubi-9-baseos-source 
+
+WORKDIR /elfutils-source
+RUN yumdownloader --source elfutils
+RUN yum -y install cpio
+RUN rpm2cpio elfutils-0.189-3.el9.src.rpm | cpio -iv
+RUN tar xjvf elfutils-0.189.tar.bz2
+WORKDIR /elfutils-source/elfutils-0.189
+RUN ./configure --disable-debuginfod
+RUN make install
+
+
+WORKDIR /libbpf-source
+RUN yumdownloader --source libbpf
+RUN rpm2cpio libbpf-1.2.0-1.el9.src.rpm | cpio -iv
+RUN tar xf ./linux-*el9.tar.xz
+WORKDIR /libbpf-source/linux-5.14.0-333.el9/tools/lib/bpf
+RUN make install_headers
+RUN prefix=/usr BUILD_STATIC_ONLY=y make install
+WORKDIR /libbpf-source/linux-5.14.0-333.el9/tools/bpf
+RUN make bpftool
+
+RUN yum -y install clang
+
+# enable EPEL and install cpuid
+RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
+RUN yum install -y cpuid
+RUN yum clean all
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM quay.io/sustainable_computing_io/kepler-base:latest as builder
+
+WORKDIR /workspace
+
+COPY . .
+
+RUN ATTACHER_TAG=libbpf make build
+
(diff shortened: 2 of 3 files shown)


vimalk78:
Cc: @rootfs @marceloamaral

SamYuan1990:
I think the major question is that are we able to change base image from GPU based image to others? if so maybe we can just use golang as builder and base image with libbpf from Alpine?

SamYuan1990:
btw, @vimalk78 , I hope you can take a look at https://github.com/sustainable-computing-io/kepler/pull/939 or https://github.com/sustainable-computing-io/kepler/issues/934 and https://github.com/sustainable-computing-io/kepler/issues/1047 ... to see current discussion....
@rootfs , @marceloamaral , a non GPU based image current been discussed among those topics. also https://github.com/sustainable-computing-io/kepler/pull/1045 one here.

vimalk78:
> If we are going to support a non-GPU image,

why are we calling it a non-gpu image? is it because it doesnt have `nvidia/cuda` as the base image? 

As i understood kepler looks for `libnvidia-ml.so` at run time using `"github.com/NVIDIA/go-nvml/pkg/nvml"` go package. If this shared lib is present, it is used else gpu-dummy is used. This shared library is installed by nvidia-gpu-operator, which is a pre-requisite for getting gpu power.

vimalk78:
@rootfs  i don't have a gpu based machine so i cannot test the gpu feature in the kepler image built from this Dockerfile, but  if someone has machine gpu  can use `quay.io/vimalkum/kepler:simplify-dockerfile` to test.

SamYuan1990:
> @rootfs i don't have a gpu based machine so i cannot test the gpu feature in the kepler image built from this Dockerfile, but if someone has machine gpu can use `quay.io/vimalkum/kepler:simplify-dockerfile` to test.

for an example https://github.com/sustainable-computing-io/kepler/blob/main/Makefile#L282 the mac without GPU has different tag with gpu supported. Historically, if a environment without GPU package will lead error as GPU packages fails with compile.

Dockerfile.base
```diff
@@ -0,0 +1,33 @@
...
+
+USER 0
+
+RUN yum -y install yum-utils
+RUN yum-config-manager --enable ubi-9-baseos-source 
+
+WORKDIR /elfutils-source
+RUN yumdownloader --source elfutils
+RUN yum -y install cpio
+RUN rpm2cpio elfutils-0.189-3.el9.src.rpm | cpio -iv
```
rootfs, line 11: why build it from source, can `elfutils` rpm be installed directly?

rootfs:
@vimalk78 can you rename these files, e.g add `no-gpu` suffix? We can add CI to build these no-gpu images without having to replacing existing ones

vimalk78:
> @vimalk78 can you rename these files, e.g add `no-gpu` suffix? We can add CI to build these no-gpu images without having to replacing existing ones

we can use CI to build image out of these Dockerfile, but i don't thing it needs a `no-gpu` tag, as this image should work for gpu power as well. 
Since i dont have any gpu resource, i was waiting for @marceloamaral and team to help test image built with these Dockerfiles on gpu.

rootfs:
@vimalk78 can you split into two PRs, one for the base and one for the kepler dockerfile? The base and builder container images are manually built at the moment. We have to change one at a time. Or if you can add a CI step, that'll be even better.

vimalk78:
> I verified the base image has nvidia-ml library
> 
> ```
> # podman run -ti --device nvidia.com/gpu=all registry.access.redhat.com/ubi9/go-toolset:1.20 ldconfig -p  |grep nvidia-ml
>         libnvidia-ml.so.1 (libc6,x86-64) => /lib64/libnvidia-ml.so.1
> ```

but here you are adding a device from host to the container. without `--device` option the `ldconfig` fails to find nvidia-ml library

rootfs:
@vimalk78 right, my bad. Without `--device`, the nvml library is not there. In this case, I don't think kepler container can access the gpu.

PR #1071, opened 2023-11-20
add xgboost to ld cache

rootfs:
The UBI image doesn't have `/usr/local/lib64` in ld conf. Need to add it explicitly to pick up xgboost library.


sthaha (approved):
Looks good to me

PR #1087, opened 2023-11-27
build: update rpm build spec to build libbpf based rpm

rootfs:
replacing bcc based rpms

here is the demo
[asciicast](https://asciinema.org/a/623618)

### DIFF
--- a/packaging/rpm/kepler.spec
+++ b/packaging/rpm/kepler.spec
@@ -11,14 +11,10 @@ URL:            https://github.com/sustainable-computing-io/kepler/
 Source0:        kepler.tar.gz
 
 BuildRequires: systemd
-BuildRequires: make
+BuildRequires: clang llvm llvm-devel zlib-devel make libbpf golang
 
-Requires:       cpuid
-Requires:       kmod
-Requires:       xz
-Requires:       python3  
-Requires:       bcc-devel
-Requires:       bcc
+Requires:       elfutils-libelf
+Requires:       elfutils-libelf-devel
 
 %{?systemd_requires}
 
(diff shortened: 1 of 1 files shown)


sthaha (approved):
lgtm

packaging/rpm/kepler.spec
```diff
@@ -44,9 +41,13 @@ install -d %{buildroot}%{_bindir}
 install -d %{buildroot}%{_sysconfdir}/kepler/
 
 install -d %{buildroot}/var/lib/kepler/data
+install -d %{buildroot}/var/lib/kepler/bpfassets
+install -d %{buildroot}/etc/kepler/kepler.config
 
 install -p -m755 ./_output/kepler  %{buildroot}%{_bindir}/kepler
 install -p -m644 ./packaging/rpm/kepler.service %{buildroot}%{_unitdir}/kepler.service
+install -p -m644 ./bpfassets/libbpf/bpf.o/amd64_kepler.bpf.o %{buildroot}/var/lib/kepler/bpfassets/amd64_kepler.bpf.o
+install -p -m644 ./_output/ENABLE_PROCESS_METRICS %{buildroot}/etc/kepler/kepler.config/ENABLE_PROCESS_METRICS
```
vimalk78, line 50: how does user of this rpm add more config params, as we add from operator?

vimalk78, line 50: we should make the config path configurable. it is hard coded as of now.

jichenjc: maybe we need general process for folder  `/etc/kepler/kepler.config` or we just doc this 
otherwise maintain all those config item one by one seems overkill

PR #1093, opened 2023-11-28
build: make daily rpm build pass

rootfs:
the daily rpm build failed, this is to remove the `golang` dependency from rpm build spec. Also change the version scheme.

### DIFF
--- a/packaging/rpm/kepler.spec
+++ b/packaging/rpm/kepler.spec
@@ -11,7 +11,7 @@ URL:            https://github.com/sustainable-computing-io/kepler/
 Source0:        kepler.tar.gz
 
 BuildRequires: systemd
-BuildRequires: clang llvm llvm-devel zlib-devel make libbpf golang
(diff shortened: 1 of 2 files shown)


packaging/rpm/kepler.spec
```diff
@@ -11,7 +11,7 @@ URL:            https://github.com/sustainable-computing-io/kepler/
 Source0:        kepler.tar.gz
 
 BuildRequires: systemd
-BuildRequires: clang llvm llvm-devel zlib-devel make libbpf golang
+BuildRequires: clang llvm llvm-devel zlib-devel make libbpf
```
sthaha, line 14: isn't `golang` not a  build requirement? 🤔

rootfs: yes, in a plain rpm build we need golang. However our builder image doesn't install golang rpm, we get golang directly from download, and that causes the build process break.

https://github.com/sustainable-computing-io/kepler/actions/runs/7016613096/job/19088182329

sthaha (approved):

PR #1095, opened 2023-11-29
rpm build: fix version

rootfs:
the build failed with bad version format https://github.com/sustainable-computing-io/kepler/actions/runs/7030429659/job/19129950611

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - name: Get current date
         id: date
-        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+        run: echo "::set-output name=date::$(date +'%Y%m%d')"
 
       - name: Checkout code
         uses: actions/checkout@v4.1.1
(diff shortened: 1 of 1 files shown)


rootfs:
@vimalk78

.github/workflows/rpm.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         env:
           _VERSION_ : ${{ steps.date.outputs.date }}
```
vimalk78, line 20: does it make more sense to use git commit for version?
The Makefile already sets the correct git sha as version https://github.com/sustainable-computing-io/kepler/blob/main/Makefile#L15
This `VERSION` is used to build kepler. 


```
Kepler running on version: v0.6.1-100-g0fd8a5d-dirty
```
so we already have correct version info, just need to use it.

rootfs: good point. since no '-' is allowed in `VERSION`, I am using `.` now

sthaha (approved):

vimalk78:
perhaps @rootfs wanted to do some more changes?

PR #1103, opened 2023-12-05
chore(build): Add base image Dockerfile for building kepler with libbpf

vimalk78:
- Adds Dockerfile.base which installs everything required to build kepler code base
  - Updated github workflow to build and push the base image

### DIFF
--- a/.github/workflows/image_base.yml
+++ b/.github/workflows/image_base.yml
@@ -53,5 +53,13 @@ jobs:
           platforms: linux/amd64
           push: true
           tags: quay.io/sustainable_computing_io/kepler_base:latest-libbpf-amd64
+      - name: Build and push a base image for building Kepler with libbpf
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./Dockerfile.base
+          platforms: linux/amd64
+          push: true
+          tags: quay.io/sustainable_computing_io/kepler_base:latest-libbpf-amd64-simplified
       - name: Create and upload multi-arch image
         run: make multi-arch-image-base
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,0 +1,33 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+
+USER 0
+
+RUN yum -y install yum-utils
+RUN yum-config-manager --enable ubi-9-baseos-source
+
+WORKDIR /elfutils-source
+RUN yumdownloader --source elfutils
+RUN yum -y install cpio
+RUN rpm2cpio elfutils-0.189-3.el9.src.rpm | cpio -iv
+RUN tar xjvf elfutils-0.189.tar.bz2
+WORKDIR /elfutils-source/elfutils-0.189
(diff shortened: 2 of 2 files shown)


vimalk78:
cc: @rootfs

vimalk78:
@rootfs `base` job seems to run when push happens to `base` branch. so should this PR be targeted for `base` branch?

Dockerfile.base
```diff
@@ -0,0 +1,33 @@
...
+RUN yumdownloader --source libbpf
+RUN rpm2cpio libbpf-1.2.0-1.el9.src.rpm | cpio -iv
+RUN tar xf ./linux-*el9.tar.xz
+WORKDIR /libbpf-source/linux-5.14.0-333.el9/tools/lib/bpf
+RUN make install_headers
+RUN prefix=/usr BUILD_STATIC_ONLY=y make install
+WORKDIR /libbpf-source/linux-5.14.0-333.el9/tools/bpf
+RUN make bpftool
+
+RUN yum -y install clang
```
rootfs, line 28: can you move build tools to the builder image? the base image only has the runtime dependency for kepler.

vimalk78: @rootfs the `Dockerfile.base` is the builder image. may be it should be renamed as such?

rootfs: we have 3 dockerfiles: base, builder, and kepler

The build tools are in the builder image, e.g. this one https://github.com/sustainable-computing-io/kepler/blob/main/build/Dockerfile.libbpf.builder

vimalk78: but in my refactoring there are only two Dockerfiles. `Dockerfile.base` which has all the things needed to build kepler. and `Dockerfile` which has a stage to build kapler and next stage copies everything needed to run kepler.
https://github.com/sustainable-computing-io/kepler/pull/1068

.github/workflows/image_base.yml
```diff
@@ -53,5 +53,13 @@ jobs:
...
           push: true
           tags: quay.io/sustainable_computing_io/kepler_base:latest-libbpf-amd64
+      - name: Build and push a base image for building Kepler with libbpf
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./Dockerfile.base
+          platforms: linux/amd64
+          push: true
+          tags: quay.io/sustainable_computing_io/kepler_base:latest-libbpf-amd64-simplified
```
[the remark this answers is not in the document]
vimalk78: i can add the same tag, but kept it separate to not overwrite old images.

Dockerfile.base
rootfs: Got it, so the new base is just ubi9 mimal.

rootfs (approved):

.github/workflows/image_base.yml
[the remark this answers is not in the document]
rootfs: let's add a new image first and test build the kepler image. If everything works, we can set the new tag and remove the existing build dockerfiles.

vimalk78:
will this merge trigger the action to build base image?

PR #1118, opened 2023-12-11
New build process

rootfs:
- [x] (from Vimal) Use the new base and kepler images
- [x] (from Vimal) Upgrade to golang 1.20
- [x] Remove all bcc build from Makefile and GHA workflow


.github/workflows/image_base.yml
```diff
@@ -57,9 +33,9 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          file: ./Dockerfile.base
+          file: ./build/Dockerfile.builder
```
vimalk78, line 36: @rootfs I had added Dockerfiles to t he root of repo. If we want to keep Dockerfiles under `build` folder, pls move the `Dockerfile.builder` to `build` folder.
https://github.com/sustainable-computing-io/kepler/pull/1118/files#diff-e7cd9380abeae311f2b7f9ed2e780a7ea795597555411cc965191ca1e6cedf16

rootfs: The file is copied there, I'll remove it

.github/workflows/image_base.yml
```diff
@@ -1,9 +1,7 @@
 name: base
 
 on:
-  push:
-    branches: 
-      - 'base'
+  workflow_dispatch:
```
vimalk78, line 4: base image build needs to be triggered manually now?

.github/workflows/image_base.yml
vimalk78: can you point which GH workflow step copies the `Dockerfile.builder` ?

vimalk78 (commented):
if we are removing `bcc` altogether, we can (perhaps in another PR) remove files which build with `bcc` flag
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/bpfassets/attacher/bcc_attacher.go#L1-L2
https://github.com/sustainable-computing-io/kepler/blob/main/pkg/bpfassets/attacher/attacher_suite_test.go#L1-L2
https://github.com/sustainable-computing-io/kepler/blob/main/e2e/platform-validation/platform_validation_metric_test.go#L1-L2
https://github.com/sustainable-computing-io/kepler/blob/main/e2e/integration-test/e2e_metric_test.go#L1-L2

rootfs:
@vimalk78 For now we just remove the bcc image from build and release so we don't have to release and maintain them. The entire bcc stack will removed later.

.github/workflows/image_base.yml
rootfs: the multi arch image build is still not working. Let's create a separate PR for that. For now, I just made this a manual workflow.

vimalk78: > the multi arch image build is still not working

can you point me to the pipeline?

rootfs: https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/image_base.yml

vprashar2929:
Tested the image built from the new `dockerfile` on OpenShift and it works fine.

rootfs:
thank you @vprashar2929

husky-parul (approved):

PR #1139, opened 2023-12-20
fix(test): set benchmark test timeout to 30m

sthaha:


sunya-ch (approved):
/lgtm

SamYuan1990:
may I know why we set the timeout for benchmark to 30m?
https://github.com/sustainable-computing-io/kepler/actions/runs/7320792466/job/19940365261?pr=1150
which causes lots of time wait for running this.

Issue #1128, opened 2023-12-15
Cannot run kepler test in container anymore


marceloamaral:
### What happened?

With the docker image updates I cannot run the Kepler tests locally anymore.

There is an error regarding the golang path.

```
go: cannot find GOROOT directory: /usr/local/go
go: cannot find GOROOT directory: /usr/local/go
mkdir -p /opt/app-root/src/go/bin
mkdir -p /kepler/test-bin
go mod tidy -v
go: cannot find GOROOT directory: /usr/local/go
make: *** [Makefile:189: tidy-vendor] Error 2
make: *** [Makefile:200: container_test] Error 2
```

### What did you expect to happen?

Run the test.

### How can we reproduce it (as minimally and precisely as possible)?

Just try `make container_test`

### Kepler image tag

<details>

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

PR #1150, opened 2023-12-25
[CI] ref to #1128 add local container test in CI to avoid it break.

SamYuan1990:
ref to #1128 add local container test in CI to avoid it break.


SamYuan1990:
@rootfs , @marceloamaral , @vimalk78  the new build image changed some golang related env and path settings, hence by this PR, I try to fix #1128

SamYuan1990:
but running branch mark in side container takes over 30mins to make it happen, hence I will make container test without benchmark.

rootfs (approved):

PR #1160, opened 2024-01-04
[tbd][CI] security update, daily job fix, try to recover arm64 support from base.

SamYuan1990:
in recently our daily build fails, and I further checked our commit history found out that ... after #1118 been merged
1. As the 1118 depends on some branch as `base`(for historical reason we keep the base image build steps at branch base but not main.). One of the make script/make rule which only available on branch base is lost, thus cause we fails with daily build.
2. Meanwhile, 1118 removed arch support as cross platforms build for arm64 linux/arm64
3. A security update by depends bot suggestion.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -29,7 +29,7 @@ jobs:
       uses: docker/build-push-action@v5
       with:
           context: .
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: quay.io/sustainable_computing_io/kepler:latest
           labels: latest
--- a/.github/workflows/image_base.yml
+++ b/.github/workflows/image_base.yml
@@ -29,8 +29,8 @@ jobs:
         with:
           context: .
           file: ./build/Dockerfile.builder
-          platforms: linux/amd64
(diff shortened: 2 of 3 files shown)


SamYuan1990:
btw, I know it seems not a good approach as this PR contains 3 different feature. But, as I found this issue just before go to sleep. Hence I just raise it up today as a hot fix, I would like to split this PR into 3 different PRs during weekend. Also give more time to allow others to cross check to avoid any mistake.

SamYuan1990:
if we don't merge it as hot fix, we can have use this pr to discuss the founding for daily job failure and arm64 image missing.

SamYuan1990:
Update to draft to avoid mistake as I am too tired.

rootfs (approved):

rootfs:
https://github.com/sustainable-computing-io/kepler/actions/runs/7411674436/job/20166742182
```
#2 [linux/arm64 internal] load metadata for quay.io/sustainable_computing_io/kepler_builder:ubi-9-libbpf-1.2.0
#2 ERROR: no match for platform in manifest: not found
```

rootfs:
need to get an arm64 builder image before merging this PR.

PR #1169, opened 2024-01-05
[CI][don't merge] base image multiple arch support

SamYuan1990:
tested with arm64 https://github.com/sustainable-computing-io/kepler/actions/runs/7422356767/job/20197478578?pr=1169

### DIFF
--- a/.github/workflows/image_base.yml
+++ b/.github/workflows/image_base.yml
@@ -18,17 +18,17 @@ jobs:
         uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-      - name: Login to Quay
-        uses: docker/login-action@v3
-        with:
-            registry: quay.io/sustainable_computing_io
-            username: ${{ secrets.username }}
-            password: ${{ secrets.password }}
+      #- name: Login to Quay
+      #  uses: docker/login-action@v3
+      #  with:
+      #      registry: quay.io/sustainable_computing_io
+      #      username: ${{ secrets.username }}
+      #      password: ${{ secrets.password }}
       - name: Build and push a base image for building Kepler with libbpf
         uses: docker/build-push-action@v5
         with:
           context: .
           file: ./build/Dockerfile.builder
-          platforms: linux/amd64
-          push: true
+          platforms: linux/amd64,linux/arm64,linux/s390x
+          push: false
           tags: quay.io/sustainable_computing_io/kepler_builder:ubi-9-libbpf-1.2.0
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -6,18 +6,18 @@ COPY . .
 
 RUN ATTACHER_TAG=libbpf make build
 
-FROM registry.access.redhat.com/ubi9-minimal:9.2
-RUN microdnf -y update
-RUN microdnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
-RUN if [ $(uname -i) == "x86_64" ]; then microdnf install -y --enablerepo='epel' cpuid; fi
+FROM registry.access.redhat.com/ubi9
+RUN yum -y update
+RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
+RUN if [ $(uname -i) == "x86_64" ]; then yum install -y cpuid; fi
 
 ENV NVIDIA_VISIBLE_DEVICES=all
 
 RUN INSTALL_PKGS=" \
     libbpf \
     " && \
-	microdnf install -y $INSTALL_PKGS && \
-	microdnf clean all
+	yum install -y $INSTALL_PKGS && \
+	yum clean all
 
 COPY --from=builder /workspace/_output/bin/kepler /usr/bin/kepler
 COPY --from=builder /libbpf-source/linux-5.14.0-333.el9/tools/bpf/bpftool/bpftool /usr/bin/bpftool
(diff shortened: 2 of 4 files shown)


SamYuan1990:
@rootfs , it will take lots of time to build base image on buildx... any suggestion?

SamYuan1990:
@rootfs , could you please help with a s390 image belowing?
```
#2 [linux/s390 internal] load metadata for registry.access.redhat.com/ubi9/go-toolset:1.20
#2 ERROR: no match for platform in manifest: not found
```
cc: @jiangphcn , @jichenjc

SamYuan1990:
https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?architecture=s390x&image=6571494d263ad9efd38089c4&container-tabs=gti
@rootfs , could you please help check the reason for s390x image fail with docker buildx? https://github.com/sustainable-computing-io/kepler/actions/runs/7423083117/job/20199711813?pr=1169

build/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN ATTACHER_TAG=libbpf make build
 
 FROM registry.access.redhat.com/ubi9-minimal:9.2
 RUN microdnf -y update
+RUN yum install -y cpuid
```
rootfs, line 11: yum -> microdnf

rootfs: how about limiting to x86_64?
```
RUN if [ $(uname -i) == "x86_64" ]; then microdnf install -y cpuid; fi
```

SamYuan1990: I will try

SamYuan1990:
```
Found 0 entitlement certificates
```
either for epel or cpuid there is no package with certificates.

SamYuan1990: as microdnf require a certificates which seems epel don't have. switch the image to ubi and use yum for package install.

.github/workflows/image_base.yml
```diff
@@ -18,17 +18,17 @@ jobs:
         uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-      - name: Login to Quay
```
SamYuan1990, line 21: open this PR for code review, if the code is good enough, will change back on CI settings as secret usage and remove hook in PR.

rootfs:
@vimalk78 can you take a look?

build/Dockerfile
```diff
@@ -6,20 +6,22 @@ COPY . .
 
 RUN ATTACHER_TAG=libbpf make build
 
-FROM registry.access.redhat.com/ubi9-minimal:9.2
-RUN microdnf -y update
+FROM registry.access.redhat.com/ubi9
```
vimalk78, line 9: change to `registry.access.redhat.com/ubi9:9.2` ?

build/Dockerfile
```diff
@@ -6,20 +6,22 @@ COPY . .
...
+RUN yum -y update
+RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
+RUN if [ $(uname -i) == "x86_64" ]; then yum install -y cpuid; fi
```
vimalk78, line 12: installing `epel-release` and `cpuid` in the main image causes image size to bloat thats why i kept the installation in the base image, and just copied the binary.

SamYuan1990: as cpuid just for x86, we don't have too much option here.  As cpuid is x86 specific, we can't copy the cpuid binary from base to main image, as for no x86 image, there is no cpuid rpm package support.
hence, one of the option is that we make some script specifically base on cpu arch. I am free and open to any workable option considering with multiple arch.

build/Dockerfile
SamYuan1990: should we keep to use latest tag instead of any specific version for example 9.2 ?
as the image will build daily, use a latest tag for nightly build is better?

SamYuan1990: btw, honestly, .... I hope some one can make it easier....
```
registry.access.redhat.com/ubi9
registry.access.redhat.com/ubi:9
registry.access.redhat.com/ubi9-minimal:9.2
registry.access.redhat.com/ubi9:9.2
```
what's the naming rule and difference?

vimalk78: i think it is good practice not to use `latest` tag. as it is hard to rollback in case of issues.

build/Dockerfile
vimalk78: since `cpuid` is not available on other archs, there  is no other option than what you suggest.

build/Dockerfile.builder
```diff
@@ -29,5 +29,4 @@ RUN yum -y install clang rpm-build llvm-devel
 
 # enable EPEL and install cpuid
 RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
-RUN yum install -y cpuid
```
vimalk78, line 32: since previous line, which instals epel, is also moved to builder, can we delete it also?

build/Dockerfile
SamYuan1990: k, let's try.

build/Dockerfile.builder
SamYuan1990: removed as your suggestion.

SamYuan1990:
@rootfs , @vimalk78 
I just submit changes according to pr review suggestion. It usually takes 1 hour for the job to be completed. I will take some personal issue and may absent from github in days(maybe...) hence
as following items for this PR.
1. if there is no further review comments, I will set CI hooks back and make the PR ready for merge.
2. Once merged, we need keep eye on it to see if it works.
3. If base image works well, another PR for multiple arch should be submitted and merged.

rootfs (approved):

PR #1173, opened 2024-01-06
[CI] update set output to github output

SamYuan1990:
ref https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


jichenjc (approved):

vprashar2929 (approved):
LGTM

PR #1184, opened 2024-01-15
Fix directory path type

armantorkzaban:
Fixes the following issue on Kubernetes with the type of Directory:
`MountVolume.SetUp failed for volume "proc" : hostPath type check failed: /proc-host is not a directory`

### DIFF
--- a/manifests/config/exporter/exporter.yaml
+++ b/manifests/config/exporter/exporter.yaml
@@ -113,19 +113,19 @@ spec:
       - name: lib-modules
         hostPath:
           path: /lib/modules
-          type: Directory
+          Type: DirectoryOrCreate
       - name: tracing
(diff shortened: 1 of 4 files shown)


marceloamaral:
@armantorkzaban, we exclusively utilize `/proc-host` for the [Kind](https://kind.sigs.k8s.io/) cluster due to the limitation of directly mounting the host `/proc` folder. This serves as a workaround.

If you have an error related to `/proc-host`, it indicates that the Kind cluster may not have been created correctly, or the Kepler deployment might not be configured for a Kind cluster.

sthaha:
closing this based on https://github.com/sustainable-computing-io/kepler/pull/1184#issuecomment-1893596434

PR #1186, opened 2024-01-17
[CI] enable arm64 support and fix bug for epel

SamYuan1990:
after we have base image support with arm64, we can add arm64 in image build process.
and fix bug as base image removed epel support.


rootfs:
@vimalk78 can you take a look?

rootfs (approved):

rootfs:
merging it to fix the CI error due to missing `cpuid` in container test

PR #1188, opened 2024-01-19
fix: update manifest script

vprashar2929:
This PR: 
* exporting path first to avoid duplicate installing tools
* fixed Prometheus role deploy bug


vprashar2929:
cc: @rootfs

PR #1196, opened 2024-01-23
separate metric checking into non zero and can-be-zero metrics

rootfs:
Ensure some (i.e. uncore) metrics can be ignored while useful metrics should always be valid.

### DIFF
--- a/e2e/integration-test/e2e_metric_test.go
+++ b/e2e/integration-test/e2e_metric_test.go
@@ -204,54 +204,65 @@ func (kmc *TestKeplerMetric) constructMetricName(res labels.Labels) string {
 }
 
 // checkMetricValues verifies that specified metric exists and has nonzero values.
-func checkMetricValues(keplerMetric *TestKeplerMetric, metricName string) {
+func checkMetricValues(keplerMetric *TestKeplerMetric, metricName string, zeroAllowed bool) {
 	v, ok := keplerMetric.Metric[metricName]
 	Expect(ok).To(BeTrue(), "Metric %s should exist", metricName)
 
-	nonzeroFound := false
 	for _, val := range v {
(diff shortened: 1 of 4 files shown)


rootfs:
the energy metrics is still zero, i think we need to run some stress-ng workload during the test.

e2e/integration-test/e2e_metric_test.go
```diff
@@ -260,27 +271,39 @@ var _ = Describe("Metrics check should pass", Ordered, func() {
...
 	})
 
-	_ = DescribeTable("Check node level metrics for details",
+	_ = DescribeTable("Check node level metrics for details, non zero value metric should be found",
 		func(metricName string) {
-			checkMetricValues(keplerMetric, metricName)
+			noneZeroMetricValues(keplerMetric, metricName)
+		},
+		Entry(nil, "kepler_exporter_build_info"),       // only one
+		Entry(nil, "kepler_node_core_joules_total"),    // node level
```
marceloamaral, line 279: The test is failing for `kepler_node_core_joules_total`. This metric is expected to be 0 right?

rootfs: only node platform and uncore are expected to be 0

jichenjc: my env shows non zero, is it because something in CI env?

rootfs: not sure what happened to the CI, it is always 0

rootfs: fixed

rootfs:
the stress test option will be enabled in a separate PR

vprashar2929 (approved):
LGTM

PR #1199, opened 2024-01-25
bpf: fix target arch for arm

rootfs:
fix build error 
https://github.com/sustainable-computing-io/kepler/actions/runs/7642380579/job/20821864213

### DIFF
--- a/bpfassets/libbpf/Makefile
+++ b/bpfassets/libbpf/Makefile
@@ -4,12 +4,8 @@ GOARCH=$(shell go env GOARCH)
 TARGET := kepler
 TARGET_BPF := $(TARGET).bpf.o
 BPF_SRC := bpfassets/libbpf/src/*.bpf.c
-# in libbpf if $(ARCH) is x86_64, then set TARGET_ARCH to x86
-ifeq ($(ARCH),x86_64)
-    TARGET_ARCH := x86
-else
-    TARGET_ARCH := $(ARCH)
-endif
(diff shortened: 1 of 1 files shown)


bpfassets/libbpf/Makefile
```diff
@@ -4,12 +4,8 @@ GOARCH=$(shell go env GOARCH)
...
 TARGET_BPF := $(TARGET).bpf.o
 BPF_SRC := bpfassets/libbpf/src/*.bpf.c
-# in libbpf if $(ARCH) is x86_64, then set TARGET_ARCH to x86
-ifeq ($(ARCH),x86_64)
-    TARGET_ARCH := x86
-else
-    TARGET_ARCH := $(ARCH)
-endif
+# in libbpf if $(ARCH) is x86_64, then set TARGET_ARCH to x86. if arch is aarch64, then set TARGET_ARCH to arm
+TARGET_ARCH := $(shell echo $(ARCH) | sed -e 's/x86_64/x86/' -e 's/aarch64/arm/')
```
SamYuan1990, line 8: @rootfs  may I know more details here?
+ @jiangphcn in loop, some times s390x build well and sometimes not. I am interested with libbpf binary build process...
if possible, should we build the binary away from our image build process?
Please correct me for any mistake....
for each cpu arch, if the libbpf support compile once and running anywhere as a binary... which means we can build the binary from source code and save the files on an artifactory or somewhere else? then remove it from kepler binary build by download from artifactory instead?

cc: @sunya-ch , @marceloamaral

Issue #1156, opened 2024-01-03
Missing support for linux/arm64/v8


beneiltis:
### What happened?

First of all: Great work. The operator is working great for me 👍. Thx a lot for your great work.

When i deploy the operator to my kubernetes cluster (k3s) i receive the error: 
```
Failed to pull image "quay.io/sustainable_computing_io/kepler:release-0.6.1": no matching manifest for linux/arm64/v8 in the manifest list entries
```

As far as I can see, there is only a amd64 image in the container registry. Do you pepole have any plans on multiarch support?

### What did you expect to happen?

A working image pull for arm64 systems. In this case: a M2 Mac.

### How can we reproduce it (as minimally and precisely as possible)?

Run the helm chart on a m1 or m2 mac in docker-desktop-kubernetes.

### Kepler image tag

<details>
quay.io/sustainable_computing_io/kepler:release-0.6.1
</details>

### Kubernetes version

<details>

```console
v1.28.2
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
macos 13.4.1 
$ uname -a
Darwin MBP 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:20 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T6000 arm64

```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

rootfs:
@beneiltis thanks for testing kepler on ARM. We are still working on multi arch image build at the moment. The current ARM platform kepler support is Ampere, since Ampere CPU has a hwmon that reports power consumption. If we know how to get power readings from e.g. apple silicon, we would love to support it too.

cc @vimalk78

beneiltis:
[...]

PR #1202, opened 2024-01-25
image: don't build for arm64 for now due to an issue in libbpf build

rootfs:
temporarily disable arm64 build. Will enable it after resolving libbpf module build issue in https://github.com/sustainable-computing-io/kepler/actions/runs/7659192488/job/20873789773


rootfs:
cc #1156

PR #1203, opened 2024-01-25
collector: use kernel's view of thread group (userspace view of process id) to aggregate all threads by a process, in order to properly account for resource utilization by processes

rootfs:
Idle power is attributed by the number of processes. If a process has many threads, this process is attributed with more idle power. Aggregating at user space PID level reduces such confusion.

### DIFF
--- a/pkg/bpfassets/attacher/bpf_perf.go
+++ b/pkg/bpfassets/attacher/bpf_perf.go
@@ -57,8 +57,8 @@ var (
 // must be in sync with bpf program
 type ProcessBPFMetrics struct {
 	CGroupID       uint64
-	PID            uint64
-	TGID           uint64
+	ThreadPID      uint64 /* thread id */
+	PID            uint64 /* TGID of the threads, i.e. user space pid */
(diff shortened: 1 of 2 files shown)


pkg/bpfassets/attacher/bpf_perf.go
```diff
@@ -57,8 +57,8 @@ var (
 // must be in sync with bpf program
 type ProcessBPFMetrics struct {
 	CGroupID       uint64
-	PID            uint64
-	TGID           uint64
+	ThreadPID      uint64 /* thread id */
+	PID            uint64 /* TGID of the threads, i.e. user space pid */
```
jichenjc, line 61: looks like it's just naming convention change ? e.g TGID => PID (since it's user space)

jichenjc (approved):

pkg/bpfassets/attacher/bpf_perf.go
rootfs: most of the work is done in ebpf, the userspace code is to use the tgid to track processes instead of threads.

PR #1210, opened 2024-01-30
gpu: expose both idle and dyn gpu power

rootfs:
Currently the dynamic GPU power is not shown in the metrics, this is because the metric is not yet.

In `UpdateDynEnergy()`, the initial GPU dyn stats is empty, so we need to use `AbsEnergyInGPU` stat to enumerate all the GPU IDs.


marceloamaral (approved):
/lgtm

PR #1227, opened 2024-02-12
feat: add class diagrams

vimalk78:
Generated class diagrams from source for better understanding of kepler


vimalk78:
Cc: @marceloamaral @rootfs

rootfs:
thanks @vimalk78 this is a good visual. Would you please add to the CI so that the diagrams can be updated automatically?

vimalk78:
as suggested by @sthaha in similar kepler-model-server PR, i have moved the umls to aleady existing `doc/dev` folder

rootfs (approved):

PR #1229, opened 2024-02-13
bpf: fix wrong task clock map name

rootfs:
the map name in bpf and attacher are used differently, that caused the attacher fail to load the map


rootfs:
before the fix:
```
libbpf_attacher.go:108] failed to resize array task_clock_event_reader: failed to find BPF map task_clock_event_reader: no such file or directory
```

vprashar2929 (approved):
LGTM

Issue #1233, opened 2024-02-14
Support multiple GPUs


rootfs:
### What would you like to be added?

Resource usage `ResourceUsage` can only capture one GPU usage. The NVML API also reports usage per GPU. If there are multiple GPUs in the system, then it is unclear how to translate usage of single GPU usage to multiple GPUs, especially when these GPUs are heterogenous. 

### Why is this needed?

Make `ResourceUsage` a 2D slice so it can capture usage and power on all of them

@marceloamaral @sunya-ch

marceloamaral:
@rootfs,

The `ResourceUsage` struct is defined as a map of string keys to` UInt64StatCollection` values.

The `UInt64StatCollection` struct, in turn, is a collection of` UInt64Stat` values, with each UInt64Stat representing a specific metric.

Here's the definition of UInt64StatCollection:
```go
// UInt64StatCollection keeps a collection of UInt64Stat
type UInt64StatCollection struct {
    Stat map[string]*UInt64Stat
}
```
This structure allows for the storage of metrics for multiple GPUs, with each GPU identified by its ID in the map key.

However, seeing of the code, it appears that there is a mistake. The current implementation does not store the metrics per GPU as intended, it is using a generic key instead...

To address this issue, I have submitted the PR #1236 that fixes the bug and ensures that the ResourceUsage stores the metrics per GPU. As the processes metrics are aggregated later to create the node metrics, the node metric will also have the metrics per GPU.

rootfs:
this is fixed

PR #1236, opened 2024-02-16
fix process resource usage map to use GPU ID as the map key

marceloamaral:
fixing #1233

This PR is:
- refactoring the code fixing the function name that is updating process metrics not node metrics
- fixing process resource usage map to use GPU ID as the map key

### DIFF
--- a/pkg/collector/metric_collector.go
+++ b/pkg/collector/metric_collector.go
@@ -180,14 +180,9 @@ func (c *Collector) updateNodeAvgCPUFrequencyFromEBPF() {
 // update the node metrics that are not related to aggregated resource utilization of processes
 func (c *Collector) updateNodeResourceUtilizationMetrics(wg *sync.WaitGroup) {
 	defer wg.Done()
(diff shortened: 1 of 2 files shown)


rootfs:
@marceloamaral GPU ID doesn't always help, especially when there are a MIG device. We need to pick up the device name that is created by the driver. I have added in #1238 [here](https://github.com/sustainable-computing-io/kepler/pull/1238/files#diff-48f392a5b02bbc07cc348ee2dd50feb46857153da9f9b448ad91c90981a5bcde)

marceloamaral:
@rootfs the https://github.com/sustainable-computing-io/kepler/pull/1238 [line](https://github.com/sustainable-computing-io/kepler/pull/1238/files#diff-48f392a5b02bbc07cc348ee2dd50feb46857153da9f9b448ad91c90981a5bcde) is using the same GPU id as proposed in this PR.

Also, note that this PR here is also updating other important things. So, do you mind merge this PR first?

rootfs (approved):

Issue #1244, opened 2024-02-21
Integrate curvefit function into local estimator


sunya-ch:
### What would you like to be added?

Update local estimator to support other curvefit function introduced in kepler-model-server

### Why is this needed?

To expand choice of modeling on local estimator

### TODO
- [ ] update kepler-model-server to new logarithm fitting function 
- [ ] update trained SPECPower to kepler-model-db 
- [ ] update kepler code

PR #1247, opened 2024-02-22
add regressor package for curvefit extension

sunya-ch:
This PR works on the issue https://github.com/sustainable-computing-io/kepler/issues/1244 for curvefit extension. 
I introduce a new package called regressor to replace LinearRegressor. 
`LinearRegressor` is moved into this package as well as other new functions from curvefit trainer (check [kepler-model-server issue](https://github.com/sustainable-computing-io/kepler-model-server/issues/214)). 

The notice point:
- trainer name needed to specify to use other curvefit function (using already-exists `TRAINER` key defined under [modelConfig](https://github.com/sustainable-computing-io/kepler/blob/87228e624b8643db471ee2cba7bf9fb6b23bd48f/pkg/config/config.go#L133)). 
- linear regressor has different weight definition since it can support multiple metrics while curvefit won't. 
- Curvefit will use only a single feature value. This is defined in the weight scale. 
- For log function, need https://github.com/sustainable-computing-io/kepler-model-server/pull/232 to merged first.


sunya-ch:
Any idea to fix this error?

Screenshot 2024-02-22 at 23 34 49

rootfs (approved):

rootfs:
@sunya-ch @marceloamaral there are some lint warning (duplicate code). What about we merge the PR right now for 0.7.4 and remove the duplicate code in a follow up PR in 0.7.5?

PR #1254, opened 2024-02-23
revert rpm source

rootfs:


rootfs:
reverting to avoid release failure.

PR #1262, opened 2024-02-26
base image build refactoring

rootfs:
some refactoring, so we can use manual github action and bot credentials to build builder images.

PR #1267, opened 2024-02-27
[CI] revert #1262

SamYuan1990:
Reverts https://github.com/sustainable-computing-io/kepler/pull/1262, as it makes a wrong implementation.
Just replaced workflow call by workflow_dispatch, which breaks everything in automation, either PR check or daily build.
To support workflow_dispatch, which this PR contains a temporary solution as an example.

Pre previously CI consideration/supporting for base image as https://github.com/sustainable-computing-io/kepler/issues/1047 which base image should up to date and maintained by CI.

Hence, considering the PR may breaks nightly build. Our CI for base image supports feature for:
- Nightly to fix CVEs for dependency.
- Pre PR, to validate if changes breaks CI.

### DIFF
--- a/.github/workflows/baseimage_dispatch.yml
+++ b/.github/workflows/baseimage_dispatch.yml
@@ -1,0 +1,13 @@
+name: baseimage_dispatch
+
+on:
+  workflow_dispatch:
+
+jobs:
+  baseimagebuild:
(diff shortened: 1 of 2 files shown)


SamYuan1990:
add an example as temporary solution....
the temporary will overwrite the base image tag which I don't suppose it's a good implementation.

PR #1292, opened 2024-03-06
dcgm: when processes are running in  MIGs, ratio their energy based how much their normalized utilizations are relative to the aggregate normalized utilization of only active MIGs, not the entire GPU

### LINKED ISSUE #1293: In `GetGPUPower()`, Kepler should use the per GPU Power consumption to calculate the process's how much power consumed by the process

Here is the [current implementation](https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/estimator/local/ratio.go#L168). It takes into account of all the GPU power from every devices and attributes regardless of their differences. 

```go
// GetComponentsPower returns GPU Power in Watts associated to each each process/process/pod
func (r *RatioPowerModel) GetGPUPower(isIdlePower bool) ([]float64, error) {
	nodeComponentsPowerOfAllProcesses := []float64{}

	// the number of processes is used to evernly divide the power consumption for OTHER and UNCORE
	// we do not use CPU utilization for OTHER and UNCORE because they are not necessarily directly
	numProcesses := float64(r.xidx)

	// estimate the power for each process
	for processIdx := 0; processIdx < r.xidx; processIdx++ {
		var processPower float64

		// TODO: idle power should be divided accordinly to the process requested resource
		if isIdlePower {
			processPower = r.nodeFeatureValues[GpuIdlePower] / numProcesses
		} else {
			processPower = r.getPowerByRatio(processIdx, int(GpuUsageMetric), int(GpuDynPower), numProcesses)
		}
		nodeComponentsPowerOfAllProcesses = append(nodeComponentsPowerOfAllProcesses, processPower)
	}
	return nodeComponentsPowerOfAllProcesses, nil
}
```

### Why is this needed?

We should get per GPU power and attribute only to those processes that use this GPU.

rootfs:
Fix #1293


pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
```diff
@@ -239,77 +239,71 @@ func (d *GPUDcgm) GetProcessResourceUtilizationPerDevice(device interface{}, gpu
...
+				if deviceUtilization <= 0 {
+					deviceUtilization = 0.5 // avoid division by zero
 				}
+				if deviceUtilization > 100 {
+					deviceUtilization = 100
+				}
+				// ratio of active multiprocessors to total multiprocessors
+				// the MIG metrics represent the utilization of the MIG device. We need to normalize the metric to represent the overall GPU utilization
+				// FIXME: the MIG device could have multiple processes, how to split the MIG utilization between the processes?
+				normalizedComputeUtil := deviceUtilization * multiprocessorCountRatio
```
yuezhu1, line 291: What metrics are we targeting over here? If the metrics are related to SM related usage rate (e.g., 0-1), this normalization should be fine. If we also need to query memory related metrics, the normalization rate should be consider as the proportion of MIG memory size to total gpu memory size.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
```diff
@@ -239,77 +239,71 @@ func (d *GPUDcgm) GetProcessResourceUtilizationPerDevice(device interface{}, gpu
...
-				ComputeUtil: uint32(gpuUtilization),
-			}
 		}
 	}
+	// fourth, get the process utilization
+	for _, p := range processInfo {
+		processAcceleratorMetrics[p.Pid] = ProcessUtilizationSample{
+			Pid:         p.Pid,
+			TimeStamp:   uint64(time.Now().UnixNano()),
+			ComputeUtil: uint32(processNormalizedUtil[p.Pid]/totalNormalizedUtil*100 + 0.5),
```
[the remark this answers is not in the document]
rootfs: yes. when the GPU is idle (even with a process there), the label could be zero.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
rootfs: At the moment, only SM usage.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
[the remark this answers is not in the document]
sunya-ch: I think we can add a condition logic to check if total utilization = 0, compute util = 0.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
```diff
@@ -239,77 +239,71 @@ func (d *GPUDcgm) GetProcessResourceUtilizationPerDevice(device interface{}, gpu
...
-					}
+		} else { // this is a GPU, get the GPU utilization
+			multiprocessorCountRatio = 1.0
+			vals, err = dcgm.GetLatestValuesForFields(uint(gpuID), deviceFields)
+		}
+		for _, val := range vals {
+			if val.FieldId == ratioField {
+				deviceUtilization := ToFloat64(val, 100)
+				if deviceUtilization <= 0 {
+					deviceUtilization = 0.5 // avoid division by zero
```
marceloamaral, line 283: Not sure if will make sense to divide by 0.5. It is better to change the logic to do not perform the division.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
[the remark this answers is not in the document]
marceloamaral: Right, we should not change the values. Let's just set to 0 with the division is by 0.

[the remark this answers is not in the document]
marceloamaral: Also, this metric should only be the process resource utilization, without being normalized by the total resource utilization of all processes.

This ratio will actually be calculated by the ratio power model later.

marceloamaral (changes requested):
@rootfs, it seems like you're suggesting duplicating the functionality of the ratio power model component.

Here's the ratio code: https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/estimator/local/ratio.go#L84

That is, each process's GPU utilization will be aggregated to represent the node utilization. Subsequently, the Ratio power model will calculate the process power consumption based on the ratio of the process's GPU utilization and the node GPU utilization (the node GPU utilization is the `totalNormalizedUtil` that you are proposing here).

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
```diff
@@ -239,77 +239,71 @@ func (d *GPUDcgm) GetProcessResourceUtilizationPerDevice(device interface{}, gpu
...
-					// ratio of active multiprocessors to total multiprocessors
-					// the MIG metrics represent the utilization of the MIG device. We need to normalize the metric to represent the overall GPU utilization
-					// FIXME: the MIG device could have multiple processes, how to split the MIG utilization between the processes?
-					normalizedComputeUtil := migUtilization * multiprocessorCountRatio
-					klog.V(debugLevel).Infof("pid %d computeUtil %f multiprocessor count ratio %v\n", p.Pid, normalizedComputeUtil, multiprocessorCountRatio)
-					processAcceleratorMetrics[p.Pid] = ProcessUtilizationSample{
-						Pid:         p.Pid,
-						TimeStamp:   uint64(time.Now().UnixNano()),
-						ComputeUtil: uint32(normalizedComputeUtil),
+		isMIG := p.GpuInstanceId != 0 && p.GpuInstanceId < maxMIGProfiles // if this is not a MIG, GpuInstanceId is set to e.g. 4294967295
```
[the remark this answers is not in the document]
rootfs: This function only gets called with GPU, not MIG. Upon entry, only GPU device is passed here.

When the running processes are found on this GPU, we then check if these processes use MIG, and that is the logic is about.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
rootfs: yes, this is a hack. Will use @sunya-ch suggestion.

rootfs:
> @rootfs, it seems like you're suggesting duplicating the functionality of the ratio power model component.
> 
> Here's the ratio code: https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/estimator/local/ratio.go#L84
> 
> That is, each process's GPU utilization will be aggregated to represent the node utilization. Subsequently, the Ratio power model will calculate the process power consumption based on the ratio of the process's GPU utilization and the node GPU utilization (the node GPU utilization is the `totalNormalizedUtil` that you are proposing here).

@marceloamaral the current ratio code doesn't support multiple (GPU) devices yet, so this is the gap stopper before we get there.

rootfs:
> @rootfs, it seems like you're suggesting duplicating the functionality of the ratio power model component.
> 
> Here's the ratio code: https://github.com/sustainable-computing-io/kepler/blob/main/pkg/model/estimator/local/ratio.go#L84
> 
> That is, each process's GPU utilization will be aggregated to represent the node utilization. Subsequently, the Ratio power model will calculate the process power consumption based on the ratio of the process's GPU utilization and the node GPU utilization (the node GPU utilization is the `totalNormalizedUtil` that you are proposing here).

@marceloamaral right now, the `Ratio` model only has the process index, no process pid or utilization. It will take a big refactor to get there. For the GPU case, the ratio model code is small.

bpfassets/libbpf/bpf.o/amd64_kepler.bpf.o
sthaha, line 1: Is an update to bpf expected as part of this patch?

rootfs: the module gets recompiled each time, we should git ignore it. The Makefile rebuilds the module anyway.

marceloamaral:
@rootfs that is right.

> Aggregating GPU power and distributing a process's utilization on one GPU across all GPUs is not a fair ratio model.

It works for idle power because it's a part of the node and should be divided among all processes, regardless of GPU usage.

However, for dynamic power, we need to separate it because processes' GPU utilization is specific to each GPU, different from the CPU case, where utilization is relative to the entire system.

The point here is that we should not bypass the Ratio Power model, but fix it.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
```diff
@@ -239,77 +241,72 @@ func (d *GPUDcgm) GetProcessResourceUtilizationPerDevice(device interface{}, gpu
...
-				// FIXME: As in the original NVML code, we should use here the pinfo.SmUtil from GetProcessUtilization()
-				ComputeUtil: uint32(gpuUtilization),
-			}
 		}
 	}
+	// fourth, get the process utilization
+	for _, p := range processInfo {
+		computeUtilization := 0.0
+		if totalNormalizedUtil > 0 {
+			computeUtilization = processNormalizedUtil[p.Pid]/totalNormalizedUtil*100 + 0.5
```
marceloamaral, line 301: We should not calculate the Ratio here, but in the power model.
Also, we should not sum the `+ 0.5`

marceloamaral: Also, this is not really the Ratio.
The Ratio is the sum of the utilization of all MIG slices, but here we have the utilization of only one slice.

pkg/model/process_energy.go
```diff
@@ -304,3 +283,29 @@ func addEstimatedEnergy(processIDList []uint64, processesMetrics map[uint64]*sta
...
+	for gpuID := range gpu.GetGpus() {
+		gpuName := fmt.Sprintf("%s%v", utils.GenericGPUID, gpuID)
+		numberOfProcesses := len(processIDList)
+		// calculate the gpu's processes energy consumption for each gpu
+		processIdlePower := nodeMetrics.EnergyUsage[config.IdleEnergyInGPU].GetDeltaValuesByKey(gpuID) / uint64(numberOfProcesses)
+		dynPower := nodeMetrics.EnergyUsage[config.DynEnergyInGPU].GetDeltaValuesByKey(gpuID)
+		klog.V(6).Infof("GPU %s: idle power: %v, dynamic power: %v\n", gpuName, processIdlePower, dynPower)
+		for pid, v := range processesMetrics {
+			v.EnergyUsage[config.IdleEnergyInGPU].SetDeltaStat(gpuName, processIdlePower)
+			usage := v.ResourceUsage[config.GPUComputeUtilization].GetDeltaValuesByKey(gpuName)
```
marceloamaral, line 302: The usage is not really the usage here right?
This is the GPU ratio....So, we might need to change the metric name `GPUComputeUtilization` to `GPUComputeRatio`.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
marceloamaral: This function is listing all process running in the MIG slice, and then assuming that there is only one process running in the MIG slice, it associate the MIG utilization to the process.
The MIG utilization is normalized by its size to represent a fraction of the total GPU utilization.


So, for example, if there is only one process using in a MIG slices.
The `totalNormalizedUtil` = `processNormalizedUtil`  and the ratio will be `100.5`

It means that each MIG ratio will 1, and when multiplied by the GPU power it will have the total GPU power, not the fraction of the GPU power.

pkg/model/process_energy.go
```diff
@@ -304,3 +283,29 @@ func addEstimatedEnergy(processIDList []uint64, processesMetrics map[uint64]*sta
...
+func addGPUEnergy(processIDList []uint64, processesMetrics map[uint64]*stats.ProcessStats, nodeMetrics *stats.NodeStats) {
+	// estimate the associated power comsumption of GPU for each process
+	if !gpu.IsGPUCollectionSupported() {
+		return
+	}
+	for gpuID := range gpu.GetGpus() {
+		gpuName := fmt.Sprintf("%s%v", utils.GenericGPUID, gpuID)
+		numberOfProcesses := len(processIDList)
+		// calculate the gpu's processes energy consumption for each gpu
+		processIdlePower := nodeMetrics.EnergyUsage[config.IdleEnergyInGPU].GetDeltaValuesByKey(gpuID) / uint64(numberOfProcesses)
```
marceloamaral, line 297: The idle power is a part of the node and should be divided among all processes, regardless of GPU usage.

So the process idle power should not be relative to only one GPU.

marceloamaral: So, we just need to get here the Sum of the Idle power of all GPUs.
`EnergyUsage[config.IdleEnergyInPkg].SumAllDeltaValues()`

But note that the list of GPU in `gpu.GetGpus()`, should be only the Physical GPUs. If it also has the MIG slices, the `SumAllDeltaValues` will double count the idle power of all MIG slices, as it has the GPU power.
The PR #1299 should help with this since it keeps only the  Physical GPUs in the `gpu.GetGpus()`.

pkg/sensors/accelerator/gpu/source/gpu_dcgm.go
```diff
@@ -239,77 +241,72 @@ func (d *GPUDcgm) GetProcessResourceUtilizationPerDevice(device interface{}, gpu
...
+		}
+		processAcceleratorMetrics[p.Pid] = ProcessUtilizationSample{
+			Pid:         p.Pid,
+			TimeStamp:   uint64(time.Now().UnixNano()),
+			ComputeUtil: uint32(computeUtilization),
```
marceloamaral, line 306: Same here, change `ComputeUtil` to `RatioUtil`...

rootfs:
@marceloamaral I am closing this PR for now. Please take priority in fixing the multi GPU ratio issue. Thanks

PR #1299, opened 2024-03-13
Fix listing and using the MIG device handler

### LINKED ISSUE #1293: In `GetGPUPower()`, Kepler should use the per GPU Power consumption to calculate the process's how much power consumed by the process

marceloamaral:
The current code is not listing and getting the process in the MIG device handler.

This code get the MIG device handler and store only the necessary data.

### DIFF
--- a/pkg/collector/resourceutilization/accelerator/process_gpu_collector.go
+++ b/pkg/collector/resourceutilization/accelerator/process_gpu_collector.go
@@ -19,7 +19,6 @@ package accelerator
 import (
 	"fmt"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/sustainable-computing-io/kepler/pkg/cgroup"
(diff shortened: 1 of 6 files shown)


rootfs:
> To fix the problem in https://github.com/sustainable-computing-io/kepler/issues/1293, we need to index the process GPU utilization with the parent GPU ID (not with the MIG ID). This way, the Ratio power model can correctly distribute the parent GPU power consumption among the processes in the MIG slices based on their GPU utilization.

I don't think this is sufficient. The problem in #1293 is about multiple GPUs. Each GPU may report different energy readings. The current ratio model uses aggregate utilization and energy among multiple GPU devices, this leads to problems when the GPUs are configured differently in terms of cooling, frequency, and SM spec.

marceloamaral:
@rootfs, you're correct. 

Even if we calculate the MIG power before applying the Ratio Power Model, we'll still face this issue that the Ratio Power model uses the aggregated system data....

Can we merge this PR and fix the problem to calculate the Ratio Per GPU later?

marceloamaral:
@rootfs, you're correct.

This PR does not fix the problem in #1293

This PR fix some other problems.

rootfs (approved):

rootfs:
@marceloamaral let's merge this one first. When you have time, can you also fix the multi GPU ratio in #1292?

PR #1308, opened 2024-03-19
fix(config): trim spaces and new lines in config

sthaha:
Currently a config file containing `true` (without \n at the end) is different to
```
true
```
This commit fixes the issue by trimming spaces (including newlines) when reading a config


jichenjc (approved):

sthaha:
@jichenjc thank you for the review.  Could you please take another look? I have cleaned up the function a bit more to make it easy to read/understand.

jichenjc (approved):

Issue #1315, opened 2024-03-25
[Unit test] unit test errors


SamYuan1990:
`I0320 09:35:04.845741    4157 qat.go:35] Failed to init qat-telemtry err: could not get qat status exit status 127`
ref https://github.com/sustainable-computing-io/kepler/actions/runs/8356655955/job/22874171181#step:7:94
@marceloamaral 

`I0320 09:38:10.703505    4372 watcher.go:73] failed to get config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined`
ref https://github.com/sustainable-computing-io/kepler/actions/runs/8356655955/job/22874171181#step:7:248

PR #1316, opened 2024-03-25
[CI] Nit fix for QAT

SamYuan1990:
1. It seems that QAT will always init without the config file which causes error report in CI as #1315 
2. Nit fix for code logic


SamYuan1990:
btw, @marceloamaral , @rootfs why not we use golang once to init the driver?

SamYuan1990:
it seems QAT error been removed from log https://github.com/sustainable-computing-io/kepler/actions/runs/8418037293/job/23047689366?pr=1316

rootfs (approved):

Issue #1326, opened 2024-04-04
daily.yml failed on forked repo


YaSuenag:
### What happened?

daily.yml would be invoked everyday even if it is forked repo, but it fails as following:

![image](https://github.com/sustainable-computing-io/kepler/assets/7421132/06dcdab4-0a11-46e2-81c2-ee6f6bc48a98)

### What did you expect to happen?

No workflow error on forked repo.

### How can we reproduce it (as minimally and precisely as possible)?

Fork on GitHub, and wait 1 day :)

### Anything else we need to know?

I think daily.yml should be fired on sustainable-computing.io/kepler only, otherwise image.yml should not be invoked excepts sustainable-computing.io at least.

IMHO all of jobs in daily.yml should be disabled excepts sustainable-computing.io/kepler with `if` statement. What do you think?

### Kepler image tag

<none>

### Kubernetes version

<none>

### Cloud provider or bare metal

<none>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

SamYuan1990:
hi @YaSuenag ,
is there any suggestion for the if else conditions to either check if the gha job is running with personal fork?

SamYuan1990:
hi @YaSuenag ,
could you please try `GITHUB_ACTION_REPOSITORY`  and `startwith` and submit a PR back?
ref
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
https://docs.github.com/en/actions/learn-github-actions/expressions

SamYuan1990:
[...]

PR #1327, opened 2024-04-04
Restrict repository to run daily.yml

YaSuenag:
This PR is for #1326.

I added `github.repository == 'sustainable-computing-io/kepler'` to prevent to run daily.yml on forked repo. Similar code works fine on [Carbon Aware SDK by Green Software Foundation](https://github.com/Green-Software-Foundation/carbon-aware-sdk).

https://github.com/Green-Software-Foundation/carbon-aware-sdk/commit/06c5577397815f2e867a349da7c74be06a9423d1

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,3 +1,4 @@
+# This workflow would be run on sustainable-computing-io/kepler only.
 name: Scheduled build
 on:
   branch_protection_rule:
@@ -7,9 +8,11 @@ on:
 jobs:
   # daily go security
   gosec:
+    if: github.repository == 'sustainable-computing-io/kepler'
     uses: ./.github/workflows/gosec.yml
   # daily base image build and image build
   image_base:
+    if: github.repository == 'sustainable-computing-io/kepler'
     uses: ./.github/workflows/image_base.yml
     secrets:
         username: ${{ secrets.BOT_NAME }}
@@ -22,7 +25,9 @@ jobs:
         password: ${{ secrets.BOT_TOKEN }}
   # daily openSSF scan
   scorecard:
+    if: github.repository == 'sustainable-computing-io/kepler'
     uses: ./.github/workflows/scorecard.yml
   # daily rpm
   rpm:
+    if: github.repository == 'sustainable-computing-io/kepler'
     uses: ./.github/workflows/rpm.yml


YaSuenag:
I saw similar error in push.yml, so I also fixed it.

.github/workflows/push.yml
```diff
@@ -11,6 +11,7 @@ on:
 jobs:
   # for each PR merge, trigger image build
   image:
+    if: github.repository == 'sustainable-computing-io/kepler'
```
SamYuan1990, line 14: if so.... 
https://github.com/sustainable-computing-io/kepler/pull/1327/files#diff-f3fc934cf0d89bdf07de358896ff90f1202585df812cf606206d1830a9949811R20
https://github.com/sustainable-computing-io/kepler/pull/1327/files#diff-f3fc934cf0d89bdf07de358896ff90f1202585df812cf606206d1830a9949811R25
https://github.com/sustainable-computing-io/kepler/pull/1327/files#diff-f3fc934cf0d89bdf07de358896ff90f1202585df812cf606206d1830a9949811R25
also need for this check.

YaSuenag: I can add same condition to them, but they succeeded in this PR at least. That may be because it does not include code changes, but I'm not sure.

So should I add?

SamYuan1990 (commented):
btw, I am interesting with the job if which not failure.
Does it mean the job will run success as an empty for each fork repo?
for my own fork, I disabled manually to avoid meaningless fork daily build.

.github/workflows/push.yml
SamYuan1990: from the condition, which means when you merge a PR on main branch and meet the specific path then the job been run as https://github.com/sustainable-computing-io/kepler/pull/1327/files#diff-f3fc934cf0d89bdf07de358896ff90f1202585df812cf606206d1830a9949811R4-R9
so, in fact, the changes on this file .... considering with running at a fork repo. only been triggered when the fork owner merge a PR at the fork side.

SamYuan1990: for example, as https://github.com/SamYuan1990/kepler/pull/8 this is a PR at fork side, and once the pr merged. this pipeline been triggered..... if we only focus on daily build, we should not change on this file.

SamYuan1990 (changes requested):
some discussion on going and tested may needed.
mark as request changes and don't merge before we clean up with discussion.

YaSuenag:
> Does it mean the job will run success as an empty for each fork repo?

It would be skipped.

.github/workflows/push.yml
YaSuenag: > for example, as [SamYuan1990#8](https://github.com/SamYuan1990/kepler/pull/8) this is a PR at fork side, and once the pr merged. this pipeline been triggered..... if we only focus on daily build, we should not change on this file.

You are right. I added this change because I just found it. So I can remove if it needs.

SamYuan1990:
> > Does it mean the job will run success as an empty for each fork repo?
> 
> It would be skipped.

@YaSuenag , it would be skipped as the job will run but as each pipeline doesn't meet the condition, hence we will see a job with everything skipped? or we should better disable the job at meanwhile?

as we have 153 fork repos today, this pr will safe the 153 fork from running workloads everyday.
and we should also suggestion 153 forks also disable the job to avoid an empty running?

SamYuan1990:
would you mind to add a reminder at https://github.com/sustainable-computing-io/kepler/blob/main/CONTRIBUTING.md to disable the job to avoid empty skip? or I can open another PR for that.

.github/workflows/push.yml
SamYuan1990: IMO, I recommend PR back to upstream, but not a PR just on a specific fork.
do we have any specific case to merge a PR at local only?

SamYuan1990: ok, so let's just have daily.yml in this PR. :-)

YaSuenag: I reverted change in push.yml.

YaSuenag:
I can squash, but I have to do force push. Is it ok?
IMHO it is better to do squash merge when this PR can go.

YaSuenag:
@SamYuan1990 

> could you please try git squash?

Done!

SamYuan1990 (approved):
LGTM

PR #1328, opened 2024-04-04
update QAT logic process, revert logic change at #1316

SamYuan1990:
Recently as review our logic at qat error and the package.
I found some logic change in #1316, hence this PR is to revert those logic part back.


SamYuan1990:
the GPU related questions opened at https://github.com/sustainable-computing-io/kepler/pull/1329

rootfs (approved):

PR #1333, opened 2024-04-05
update container-kepler spec; don't build kepler rpm during release

rootfs:
The kepler rpm is to be replaced by the podman rpm. Don't build it during release.


SamYuan1990:
do we need clean up https://github.com/sustainable-computing-io/kepler/blob/main/Makefile#L162?

SamYuan1990 (approved):
LGTM

PR #1337, opened 2024-04-10
Makefile: add make run-on-docker

maryamtahhan:
Add a single command (`make run-on-kind`) to spin-up a kind cluster with prometheus and grafana also deployed in the cluster. This commit also moves the go source formatting and mod updates inside the pod context, when images are built.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ include bpfassets/libbpf/Makefile
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-
 ### env define ###
 export BIN_TIMESTAMP ?=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 export TIMESTAMP ?=$(shell echo $(BIN_TIMESTAMP) | tr -d ':' | tr 'T' '-' | tr -d 'Z')
@@ -42,7 +41,6 @@ CTR_CMD            ?= $(or $(shell podman info > /dev/null 2>&1 && which podman)
 # use CTR_CMD_PUSH_OPTIONS to add options to <container-runtime> push command.
 # E.g. --tls-verify=false for local develop when using podman
 CTR_CMD_PUSH_OPTIONS ?=
-
 
 ifeq ($(DEBUG),true)
 	# throw all the debug info in!
@@ -127,7 +125,7 @@ build_image_dcgm:  image_builder_check ## Build image with DCGM.
 	$(CTR_CMD) tag $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_BUILD_TAG)-dcgm $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)-dcgm
 .PHONY: build_image_dcgm
 
-build_containerized: tidy-vendor format build_image build_image_dcgm  ## Build ALL container images.
+build_containerized: build_image build_image_dcgm  ## Build ALL container images.
 .PHONY: build_containerized
 
 save-image: ## Save container image.
@@ -312,7 +310,7 @@ CLUSTER_PROVIDER ?= kind
 LOCAL_DEV_CLUSTER_VERSION ?= main
 
 KIND_WORKER_NODES ?=2
-
+BUILD_CONTAINERIZED ?= build_image
 
 COMPOSE_DIR="$(PROJECT_DIR)/manifests/compose"
 DEV_TARGET ?= dev
@@ -327,7 +325,7 @@ compose: ## Setup kepler (latest) using docker compose
 	@echo "  * Grafana    : http://localhost:3000"
 	@echo "  * Prometheus : http://localhost:9090"
 	@echo -e "\nKepler Deployments"
-	@echo "  * latest / upstream  : http://localhost:9288/metrics" 
+	@echo "  * latest / upstream  : http://localhost:9288/metrics"
 
 .PHONY: compose-clean
 compose-clean: ## Cleanup kepler (latest) deployed using docker compose
@@ -348,7 +346,7 @@ dev: ## Setup development env using compose with 2 kepler (latest & current) dep
 	@echo "  * development version : http://localhost:9188/metrics"
 	@echo "  * latest / upstream   : http://localhost:9288/metrics"
 
-dev-clean: ## Setup kepler (current and latest) along with 
+dev-clean: ## Setup kepler (current and latest) along with
 	docker compose \
 			-f $(COMPOSE_DIR)/$(DEV_TARGET)/compose.yaml \
 		down --remove-orphans --volumes --rmi all
@@ -362,13 +360,17 @@ cluster-clean: build-manifest ## Undeploy Kepler in the cluster.
 .PHONY: cluster-clean
 
 cluster-deploy: ## Deploy Kepler in the cluster.
+	BUILD_CONTAINERIZED=$(BUILD_CONTAINERIZED) \
 	./hack/cluster-deploy.sh
 .PHONY: cluster-deploy
 
-cluster-up:  ## Create the Kind cluster
+cluster-up:  ## Create the Kind cluster, with Prometheus, Grafana and Kepler
 	CLUSTER_PROVIDER=$(CLUSTER_PROVIDER) \
 	LOCAL_DEV_CLUSTER_VERSION=$(LOCAL_DEV_CLUSTER_VERSION) \
 	KIND_WORKER_NODES=$(KIND_WORKER_NODES) \
+	BUILD_CONTAINERIZED=$(BUILD_CONTAINERIZED) \
+	PROMETHEUS_ENABLE=true \
+	GRAFANA_ENABLE=true \
 	./hack/cluster.sh up
 .PHONY: cluster-up
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,6 +4,8 @@ WORKDIR /workspace
 
 COPY . .
 
+RUN make tidy-vendor
+RUN make format
 RUN make build VERSION=${VERSION}
 
 FROM registry.access.redhat.com/ubi9:9.2
--- a/hack/cluster-clean.sh
+++ b/hack/cluster-clean.sh
@@ -36,6 +36,11 @@ function main() {
 
     sleep 2
 
+    unset PROMETHEUS_ENABLE
+    unset GRAFANA_ENABLE
+    unset CLUSTER_PROVIDER
+    unset KIND_WORKER_NODES
+
     echo "Done $0"
 }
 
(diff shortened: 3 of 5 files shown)


rootfs:
this looks good. @vprashar2929 @SamYuan1990 can you take a look? thanks

SamYuan1990 (changes requested):
we have local-dev-cluster to set up kind prometheus and grafana there. I hope we can put all those scripts together to avoid we have prometheus, grafana everywhere. now I am working on decouple those from kepler model server https://github.com/sustainable-computing-io/kepler-model-server/pull/249....

Makefile
```diff
@@ -347,6 +348,19 @@ e2e: # Run E2E integration tests.
 	./hack/verify.sh integration
 .PHONY: e2e
 
+deploy-prometheus: ## Deploy Prometheus
```
SamYuan1990, line 351: please move kube-prometheus set up to https://github.com/sustainable-computing-io/local-dev-cluster as where we set up prometheus operator. hence, I am confusing if we set up two prometheus on the kind cluster?

hack/prometheus-deploy.sh
```diff
@@ -0,0 +1,79 @@
...
+
+set -eu -o pipefail
+
+PROJECT_ROOT="$(git rev-parse --show-toplevel)"
+declare -r PROJECT_ROOT
+
+source "$PROJECT_ROOT/hack/utils.bash"
+
+declare PROM_DIR="kube-prometheus"
+declare KEPLER_EXPORTER_GRAFANA_DASHBOARD_JSON=$( curl -fsSL https://raw.githubusercontent.com/sustainable-computing-io/kepler/main/grafana-dashboards/Kepler-Exporter.json | sed '1 ! s/^/         /' )
```
SamYuan1990, line 28: I suppose this should be changed, as it always to the github version but for local development,  we should considering read this file from local.

hack/prometheus-deploy.sh
```diff
@@ -0,0 +1,79 @@
...
+
+clone_prometheus(){
+    if [ -d "$PROJECT_ROOT/$PROM_DIR" ]; then
+        return 0
+    fi
+    header "Clone kube-prometheus"
+    run git clone --depth 1 https://github.com/prometheus-operator/kube-prometheus
```
SamYuan1990, line 35: https://github.com/sustainable-computing-io/local-dev-cluster/blob/7dafcdaeec51e781bb771ad75140ef864029b758/lib/prometheus.sh#L36
please update and enhance on current scripts.

SamYuan1990:
where we clone local dev cluster. https://github.com/sustainable-computing-io/kepler/blob/main/hack/cluster.sh#L37

SamYuan1990:
> this looks good. @vprashar2929 @SamYuan1990 can you take a look? thanks

from the feature point of view maybe good catch for integration with dashboard. but from code management points of view, please enhance and add the changes into local-dev-cluster.... to avoid duplicate.

Makefile
maryamtahhan: OK will do

maryamtahhan: I didn't think/see cluster-up deploying prometheus unless I explicitly set env vars (which I probably need to remember to unset later)...

With this change I was just trying to make sure I can spin up a quick kind cluster for validation and remove the test env without having to make any changes on my host.

I will try to combine the two concepts and move/update the prometheus script in the cluster local side

vprashar2929: We can add something like this in `hack/cluster.sh` to make sure Prometheus and grafana both deployed as [local-dev-cluster](https://github.com/sustainable-computing-io/local-dev-cluster) supports deployment of both
```
declare -r GRAFANA_ENABLE=${GRAFANA_ENABLE:-false}
....
....
export GRAFANA_ENABLE
```
In `cluster-up` target:
```
...
GRAFANA_ENABLE=$(GRAFANA_ENABLE) \
KIND_WORKER_NODES=$(KIND_WORKER_NODES) \
./hack/cluster.sh up
```

maryamtahhan:
depends on: https://github.com/sustainable-computing-io/local-dev-cluster/pull/77

I've moved the Grafana dashboard automation out of this PR and into: https://github.com/sustainable-computing-io/local-dev-cluster/pull/77

I will update this PR shortly

Makefile
```diff
@@ -347,6 +348,22 @@ e2e: # Run E2E integration tests.
...
+	CLUSTER_PROVIDER=$(CLUSTER_PROVIDER) \
+	LOCAL_DEV_CLUSTER_VERSION=$(LOCAL_DEV_CLUSTER_VERSION) \
+	KIND_WORKER_NODES=$(KIND_WORKER_NODES) \
+	BUILD_CONTAINERIZED=$(BUILD_CONTAINERIZED) \
+	PROMETHEUS_ENABLE=true \
+	GRAFANA_ENABLE=true \
+	./hack/cluster.sh up
+	OPTS=$(DEV_OPTS) \
+	BUILD_CONTAINERIZED=$(BUILD_CONTAINERIZED) \
+	./hack/cluster-deploy.sh
```
vprashar2929, line 361: make [cluster-deploy](https://github.com/sustainable-computing-io/kepler/blob/main/hack/cluster-deploy.sh) extactly does the same thing:  builds and deploy kepler on kind 

`make cluster-up cluster-deploy` will do the job

maryamtahhan: `make cluster-up cluster-deploy` Doesn't automatically deploy Prometheus + Grafana + Kind into the same cluster... it also builds 2 container images not just a basic developer one without DCGM.

 I'm just trying to create an easier point of access for people to get started with the project -  if this is not desired I'm happy to drop this change.

vprashar2929: > make cluster-up cluster-deploy Doesn't automatically deploy Prometheus + Grafana + Kepler into the kind cluster... it also builds 2 container images not just a basic developer one without DCGM.

A quick way to address this can be:

Fix in cluster.sh: 
```
declare -r GRAFANA_ENABLE=${GRAFANA_ENABLE:-false}
....
....
export GRAFANA_ENABLE
```
In cluster-up target set GRAFANA_ENABLE as `true` ensuring both Prometheus and Grafana get installed:
```
GRAFANA_ENABLE=true
...
GRAFANA_ENABLE=$(GRAFANA_ENABLE) \
KIND_WORKER_NODES=$(KIND_WORKER_NODES) \
./hack/cluster.sh up
```
For only building Kepler(without dcgm) small update in cluster-deploy.sh
```
declare KEPLER_BUILD_DCGM="${KEPLER_BUILD:-false}"
...
local kepler_build_target="build_image"
$KEPLER_BUILD_DCGM && {
	kepler_build_target="build_image_dcgm"
}
run make "$kepler_build_target" \
	IMAGE_REPO="$IMAGE_REPO" \
	IMAGE_TAG="$IMAGE_TAG" \
	VERSION="devel"
```
by default, it will build Kepler without dcgm and if someone wants to play with dcgm they can pass the value during runtime `make cluster_deploy KEPLER_BUILD_DCGM=true`
@maryamtahhan Wdyt?

maryamtahhan: I did think of that option but had opted against it as targets in the Makefile are already in use with a specific intention. If we think we are OK with changing the existing intention then I've no issue making the changes.

There is an option to use [Target-specific Variable Values](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html) in the Makefile. 

So we would re-use the cluster-up target under run-on-kind

```
run-on-kind: export PROMETHEUS_ENABLE=true
run-on-kind: export GRAFANA_ENABLE=true
run-on-kind: cluster-up ## Run Prometheus, Grafana and Kepler on Kind, quick dev environment
	OPTS=$(DEV_OPTS) \
	BUILD_CONTAINERIZED=$(BUILD_CONTAINERIZED) \
	./hack/cluster-deploy.sh
.PHONY: run-on-kind
```
we wouldn't use cluster-deploy directly as this can deploy into a physical cluster as well as kind which is why I invoke it directly under the run-on-kind target

vprashar2929: The [above](https://github.com/sustainable-computing-io/kepler/pull/1337#discussion_r1575176743) changes don't impact any existing functionality. It only allows Grafana to be deployed in Kind cluster and makes the image build step variable in `cluster-deploy.sh` Instead of creating a separate make target I think it's best to enhance the existing ones whenever possible. 🙂

maryamtahhan: > The [above](https://github.com/sustainable-computing-io/kepler/pull/1337#discussion_r1575176743) changes don't impact any existing functionality. It only allows Grafana to be deployed in Kind cluster and makes the image build step variable in `cluster-deploy.sh` Instead of creating a separate make target I think it's best to enhance the existing ones whenever possible

I disagree the cluster up target is already in use with a specific intent, with the suggested change the intent is different and has a usage impact, with the script changes:

| old intent | new intent | impact |
| -------- | ------- |------- |
| Create a kind cluster | Create a kind cluster and deploy grafana + prometheus | different (more) resource usage wherever this is already used including CI pipelines |

Re the cluster-deploy.sh - making the image build step variable is fine but this alone doesn't solve the other aspect which is needing the OPTS="PROMETHEUS_DEPLOY" to do a Prometheus/grafana manifest build under the cluster-deploy target. From earlier on in the Makefile in this PR we are setting the additional `DEV_OPTS ?= "CI_DEPLOY PROMETHEUS_DEPLOY"`

the idea with the `run-on-kind` target is to create a quick dev env/validation env where everything is deployed inside kind... with the suggestion using the target specific variables we are re-using the older targets. 

I agree with @SamYuan1990 that a discussion around the targets and the scripts should be revisited as these seem to be peppered in scripts and the Makefiles. 

My feedback here is there is not a simple getting started point for onboarding new developers. The first thing any new developer is going to do is try to deploy the whole stack (Kepler + Prometheus + Grafana) on kind to get a feel for the project. Right now they need to read 2 different parts of the documentation and understand what env vars and script vars to set to get that stack up and running (that's fine - but it's slow and would probably involve some trial and error to get the exact commands right). I think we are all in agreement on this. But I would argue that whatever changes are made to scripts/deployment intent need backwards compatibility as they can impact existing users and their pipelines.

For now I will try to update the documentation with a one liner for new developers...

Makefile
```diff
@@ -313,6 +311,8 @@ build-manifest: kustomize
 CLUSTER_PROVIDER ?= kind
 LOCAL_DEV_CLUSTER_VERSION ?= main
 KIND_WORKER_NODES ?=2
+BUILD_CONTAINERIZED ?= build_image
+DEV_OPTS ?= "CI_DEPLOY PROMETHEUS_DEPLOY"
```
vprashar2929, line 315: AFAIU `CI_DEPLOY` is not required for local development. https://github.com/sustainable-computing-io/kepler/tree/main/manifests/k8s

Makefile
vprashar2929: @maryamtahhan Highly Appreciate the feedback 🙂  .  I completely agree with the documentation part. It's very confusing, outdated and needs updating. 

> I disagree the cluster up target is already in use with a specific intent, with the suggested change the intent is different and has a usage impact, with the script changes:

AFAIK we only use `make cluster-up` while setting up things locally or in `.github/workflows/developer_local.yml`. I don't see enabling Grafana/Prometheus would impact resources, especially on CI. I completely agree with the notion of having a make target that deploys a complete infra stack (Kind + Prometheus + Grafana). The intent of `make cluster-up` is to do so. That's why I suggested to amend the existing target so that it creates less confusion. Having a target that only deploys the Kind cluster and another that deploys `Kind + monitoring stack` creates more friction. Ideally monitoring stack should be enabled from the beginning.

rootfs (approved):

PR #1342, opened 2024-04-11
dev: add compose for quicker local development

sthaha:
This PR adds the ability to run kepler using docker compose for fast edit-build-test cycle. Changes include ... 
1. Prints the port kepler will be listening on
2. Add compose to bring - kepler (upstream and development) version, prometheus, grafana 
3. Makefile targets - `dev, dev-clean, dev-restart` for most common use cases 

```
Deployment Overview (compose file: hack/compose.yaml)

Services
  * Grafana    : http://localhost:3000
  * Prometheus : http://localhost:9090

Kepler Deployments
  * develpment version : http://localhost:9188/metrics
  * latest / upstream  : http://localhost:9288/metrics
```

Adds a Grafana dashboard as a starting point for comparing `upstream` vs `local` changes. 

image


sthaha:
@SamYuan1990 @vprashar2929 @marceloamaral It would be great to get your feedback on this setup. 
If this looks good, I will create one which [adds / merges](https://docs.docker.com/compose/multiple-compose-files/merge/)  estimator and model-server.

vprashar2929 (approved):

SamYuan1990:
I am confused with this PR.... may because I am still in sick.
- for unit test purposes, we need to increase test coverage, the approach as this PR.... 
  as kepler latest is the nightly build, hence which this comparing is between local code change and kepler nightly build. I suppose this not been used for any comparing between versions.
  as kepler code version and kepler latest version share same volume mount, which means, logically, if some one changed kepler's logic and makes the code version different with kepler nightly version, as unit test coverage, the unit test should be modified and we can see it from unit test.

- for integration test, we need to run on k8s cluster, or as RPM package? Which means, even if test locally, we should run on a k8s cluster right?
I am open to discuss if we should make kepler support container run like node_exporter supports binary, container deploy with docker compose and k8s.
But, as far as I know even if we running with binary, we still running to monitor a k8s cluster. Hence the docker compose without a k8s cluster for?

- we don't need a new local build for grafana, we just mount it should work.... and https://github.com/Einsteinish/Docker-Compose-Prometheus-and-Grafana there already some example for us to reference.

- for dashboard, please consider reuse our current dashboard and don't recreate a new one.

CC: @rootfs , @sthaha , @marceloamaral

sthaha:
@SamYuan1990 , 
The intended use of  the compose files is for local development and meant to complement k8s deployment than replace it. Use of compose in CI isn't intended at all at this stage. 

The advantage of the compose file is that  
1. the edit, build, deploy, test cycle is way faster than using k8s (less fiddling around with yamls)
2. you don't need to maintain and deploy monitoring like we do in k8s and can use the latest grafana and prometheus without having to deal with an Operator and ServiceMonitors etc for local develoment

> Can kepler still monitor pods deployed to kind ? 

Yes, it should be able to provided, KUBECONFIG is respected by kepler. If this isn't supported at the moment, we should definitely add that feature.

> we don't need a new local build for grafana, we just mount it should work.... and https://github.com/Einsteinish/Docker-Compose-Prometheus-and-Grafana there already some example for us to reference.

There is nothing wrong in adding a default prometheus config or a default dashboard for grafana. I would just leave it as a matter of preference than a strict rule. 

> please consider reuse our current dashboard and don't recreate a new one.

We can definitely add the existing dashboard to grafana. however, I think we still need the new one which shows you the difference between 2 version of kepler (unmodified / latest, vs development version)

PR #1357, opened 2024-04-17
fix: image workflow to pass build args correctly

vprashar2929:
This PR updates GH workflow to respect build args while building Kepler images and also updates the Makefile accordingly.
Fixes the issue where `latest-dcgm` images were not built correctly


sthaha (commented):
lgtm

sthaha (approved):

PR #1358, opened 2024-04-17
reorg: move compose files to manifest dir

sthaha:
This PR reorganises manifests into `k8s` (kubernetes) and `compose`, updates `dev` targets to use `manifest/compose` dir. All other manifest related target should work as if no changes have been made.


sthaha:
cc: @marceloamaral @rootfs

rootfs:
@sthaha can you also update the instructions in kepler-doc? thanks.

PR #1361, opened 2024-04-17
[CI] update ubi image for CVE fix

SamYuan1990:
I had checked image tags and software inside the image and https://quay.io/repository/sustainable_computing_io/kepler/manifest/sha256:b1a7c4d0faf65221667ec67b5da5c15fc75eeda21e018270873a5c0562e180a2?tab=vulnerabilities

The latest image is registry.access.redhat.com/ubi9:9.3 but personally I would like registry.access.redhat.com/ubi9:latest tag, as I don't know how to trace a new tag for registry.access.redhat.com/ubi9 hence keep it with latest tag. Once they have registry.access.redhat.com/ubi9:9.4, we can benefit without a PR.

further compare via
https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?architecture=amd64&image=65e093e60a21b531a96f93ca
https://quay.io/repository/sustainable_computing_io/kepler/manifest/sha256:b1a7c4d0faf65221667ec67b5da5c15fc75eeda21e018270873a5c0562e180a2?tab=vulnerabilities
are welcome.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -6,7 +6,7 @@ COPY . .
 
 RUN make build VERSION=${VERSION}
 
-FROM registry.access.redhat.com/ubi9
+FROM registry.access.redhat.com/ubi9:latest
 # NOTE: use bash instead of sh
 SHELL ["/bin/bash", "-c"]
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -18,6 +18,7 @@ ENV NVIDIA_DRIVER_CAPABILITIES=utility
 ENV NVIDIA_MIG_MONITOR_DEVICES=all
 ENV NVIDIA_MIG_CONFIG_DEVICES=all
 
+RUN yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical && yum clean all
 # NOTE: set -e ensure all fails results in a non-zero exit code (i.e. fail)
 RUN set -e -x ;\
 		INSTALL_PKGS=" \
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -6,7 +6,7 @@ COPY . .
 
 RUN make build VERSION=${VERSION}
 
-FROM registry.access.redhat.com/ubi9:9.2
+FROM registry.access.redhat.com/ubi9
 # NOTE: use bash instead of sh
 SHELL ["/bin/bash", "-c"]
 
--- a/build/Dockerfile.builder
+++ b/build/Dockerfile.builder
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
 
 USER 0
 
--- a/build/Dockerfile.builder
+++ b/build/Dockerfile.builder
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset as builder
 
 USER 0
 


rootfs:
@vimalk78 can you review this? thanks

build/Dockerfile.builder
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset as builder
```
sthaha, line 1: Does this always promise go 1.20 ?

SamYuan1990: for the image always go 1.20, I suppose not.
as golang 1.21 is able to compile 1.20 with go.mod I suppose it's not a big deal?

sthaha: IMHO, we must keep the `go` version the same as on `go.mod` Try this on your local development env with both versions of `go` - `1.20` and `1.21` 

```
❯ go version
go version go1.21.7 linux/amd64

❯ go version
go version go1.21.7 linux/amd64

❯ git status
       modified:   go.mod
       modified:   go.sum

```

```
❯ go version
go version go1.20 linux/amd64

❯ go mod tidy
❯ git status
# .. clean - no changes
```

In `go 1.21`  .. there is diff to `go.mod and go.sum` which now means someone has to investigate if this change 
breaks / introduces any difference in kepler which IMHO is a wasted effort.

SamYuan1990: ok, for this PR, I am going to make the builder with 1.20 tag. And apply our comments at https://github.com/sustainable-computing-io/kepler/discussions/1362#discussioncomment-9154786 
as the builder seems not due with kepler's CVE for now.
@rootfs , @sthaha , @vimalk78, @vprashar2929 , any other comments before I move to update this PR?

sthaha: Sounds good to me

build/Dockerfile
```diff
@@ -6,7 +6,7 @@ COPY . .
 
 RUN make build VERSION=${VERSION}
 
-FROM registry.access.redhat.com/ubi9:9.2
+FROM registry.access.redhat.com/ubi9
```
vimalk78, line 9: if we want we can keep it at `9.3`, but better to have version number. `Y` release of RHEL is not that frequent. We change it when a new Y version is released (e.g. `9.4`) but lets not have it without tag.

SamYuan1990: I don't want to have it as 9.3, as for each 9.x release, we have to bump up the version manually or if depend bot can help us bump up the version, we have the 9.3. @vimalk78  do you know if dependabot can support the version bump up for registry.access.redhat.com? otherwise, keep it as latest.

SamYuan1990: I prefer to use latest as https://docs.github.com/en/code-security/security-advisories/working-with-global-security-advisories-from-the-github-advisory-database/browsing-security-advisories-in-the-github-advisory-database the CVE for registry.access.redhat.com/ubi9 is not listed on the github database, hence we can't use dependency bot to trigger a CVE fix patch for us.

SamYuan1990: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#supported-repositories-and-ecosystems redhat container is not on the default support list of dependabot either. @vimalk78 , I suppose you can make anther PR for dependabot support, otherwise, I don't want to bump up version manually each time.

vimalk78: there is a `latest` which points to [9.3-1610](https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?tag=9.3-1610) as of today. but can it change to some build of 9.2 if there is a new build? So prefer to not use `latest`

vimalk78: The `Y` version of RHEL does not change that frequently. perhaps 2 Y versions in a year.

SamYuan1990: > there is a `latest` which points to [9.3-1610](https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?tag=9.3-1610) as of today. but can it change to some build of 9.2 if there is a new build? So prefer to not use `latest`

I suppose @sthaha give us lots of reference and @vimalk78 , could you please check with other colleague and confirm if the latest tag may switch between 9.2 and 9.3?
if we choose 9.2 image, which is not a recommend one. Hence, I suppose for redhat stream, seems the tag will not downgrade to 9.2 from 9.3 anymore.

> Red Hat strongly recommends updating to the newest image version [9.3-1610](https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?image=65e093e60a21b531a96f93ca&architecture=amd64) unless otherwise defined by the support policy of [Red Hat](https://access.redhat.com/support).

ref
https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?tag=9.3-1610&architecture=amd64&image=652fc5bc9252cb8029f46161

and other ref from https://access.redhat.com/support/policy/updates/errata also as @sthaha provided, 
> For example, a Red Hat Enterprise Linux 8 RHSA will be applied cumulatively to the latest Red Hat Enterprise Linux 8 version and patch set.

if the major version...well for this image we don't have `registry.access.redhat.com/ubi9:9` but just `registry.access.redhat.com/ubi9:latest` I have to guess(as I don't have redhat account). If the major version and sub version share the same logic, the latest tag will always points to the largest version?

SamYuan1990: > The `Y` version of RHEL does not change that frequently. perhaps 2 Y versions in a year.

ref here https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b/history
I don't see overlap between 9.0 and 9.1 or 9.1 and 9.2 or 9.2 and 9.3
even if it's not changed frequently, how should we get the notice? as dependency bot is not able to get the update from https://catalog.redhat.com/ by default.
@vimalk78 if we use 9.3 we need to know when it changes to 9.4 and bump up the version, please provide an alternative for dependency bot.

SamYuan1990: @vimalk78 , I have to ask you to provide more evidences and I am open to change to 9.3 with your new evidence.
For your current worries, learn from the history https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b/history, once the version bump up, the previous version Y is not updated. You can further check as there 9.1, the 9.0 is not updated. same for 9.2, and 9.3. hence I don't worry about the latest tag change back to previous.

even if the Y version may just have 2 within a year. I suppose we don't want to have a monthly check for version bump up if we can benefits from latest tag, or forget to bump up version and relay on yum update to make our images build costs a lot of time for yum update.

why not enjoy with latest?

vimalk78: not sure if this is evidence, but It is always advised to not use latest tag. 
https://vsupalov.com/docker-latest-tag/

SamYuan1990: > not sure if this is evidence, but It is always advised to not use latest tag. https://vsupalov.com/docker-latest-tag/

I don't suppose it's a valid evidence, as we use ubi but not fedora from docker registry.
@vimalk78 I don't mind to take a long time to review this PR.

as this document conflicts with redhat's document for how to use ubi image at link here https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/building_running_and_managing_containers/index#con_characteristics-of-ubi-images_assembly_types-of-container-images
and I quote it again.
> Red Hat only supports the latest version of UBI and does not support parking on a dot release. If you need to park on a specific dot release, please take a look at [Extended Update Support](https://access.redhat.com/articles/rhel-eus).

SamYuan1990: for an example, https://github.com/ossf/scorecard/discussions/3145 @vimalk78 , a strict rule is to use a specific hash value instead of the X.Y tag. which is from openSSF point of view same for any kind of dependency as either container file or GHA. But for OpenSSF there still some people do agree that we need to case by case.
in this case, as your colleague from Redhat announced in the public document that they only supports latest version by default, as our community image usage case, I suppose is a case following default.
I am confused with the reason your insist on X.Y version here? either not a strict as hash value or following user document from your colleague.

vimalk78: @SamYuan1990 i checked the ubi9 repository. it is a single stream repository, so perhaps `latest` will always be the highest `Y` version.
I would still request to use `9.3` tag, but if it helps us we can go for `latest` tag

vimalk78, line 9: LGTM if use `:latest` tag

build/Dockerfile
```diff
@@ -18,6 +18,7 @@ ENV NVIDIA_DRIVER_CAPABILITIES=utility
 ENV NVIDIA_MIG_MONITOR_DEVICES=all
 ENV NVIDIA_MIG_CONFIG_DEVICES=all
 
+RUN yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical && yum clean all
```
vimalk78, line 21: with `:latest` image, do we need the yum update?

SamYuan1990: A good question, and for safety, I would like to say yes.
ref: https://catalog.redhat.com/software/containers/ubi9/618326f8c0d15aff4912fe0b?architecture=amd64&image=65e093e60a21b531a96f93ca&container-tabs=security
taking today's latest image as an example, even if the tag is latest, the new CVE comes after the image been created. So there is a time window, which yum can fix some CVE for us.
On the other hand, as the CVE is related with any package, impart us or not.

Therefore.... we hope the yum install can help as table below during the time window according to severity level.

|             | Related to us | no related |
|-------------|---------------|------------|
| Critical    | Fixed         | Fixed      |
| Important   | Fixed         | Fixed      |
| High or low | n/A           | n/A        |

vimalk78 (approved):
/lgtm

sthaha (approved):

PR #1364, opened 2024-04-18
reorg: split compose into 2 targets - latest & dev

sthaha:
The monolith compose is now split into 2 compose
1. compose.yaml - deploys kepler:latest image + grafana + prometheus
2. compose.dev.yaml - builds and deploys kepler, make the default dashboard - latest vs dev

Make targets:
* A new make target `compose` is added to deploy kepler:latest image
* `dev` target now accepts DEV_TARGET param to pick the file that overrides default compose.yaml (defaults to compose.dev.yaml)


sthaha:
cc: @rootfs @marceloamaral @vimalk78

vprashar2929 (approved):

rootfs (approved):

PR #1371, opened 2024-04-19
support Habana GPU

rootfs:
Tested on a setup with Habana
```
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="0",source="habana"} 206.895
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="1",source="habana"} 345.609
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="2",source="habana"} 128.661
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="3",source="habana"} 362.46
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="4",source="habana"} 157.596
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="5",source="habana"} 327.219
kepler_node_gpu_joules_total{instance="",mode="dynamic",package="6",source="habana"} 299.94
kepler_node_gpu_joules_total{instance="",mode="idle",package="0",source="habana"} 1314.276
kepler_node_gpu_joules_total{instance="",mode="idle",package="1",source="habana"} 921.663
kepler_node_gpu_joules_total{instance="",mode="idle",package="2",source="habana"} 1363.764
kepler_node_gpu_joules_total{instance="",mode="idle",package="3",source="habana"} 988.407
kepler_node_gpu_joules_total{instance="",mode="idle",package="4",source="habana"} 1260.441
kepler_node_gpu_joules_total{instance="",mode="idle",package="5",source="habana"} 1005.987
kepler_node_gpu_joules_total{instance="",mode="idle",package="6",source="habana"} 1041.636
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,10 @@ endif
 
 GENERAL_TAGS := 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo libbpf '
 GPU_TAGS := ' gpu '
+ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
+	GPU_TAGS := $(GPU_TAGS)'habana '
+endif
+
 GO_LD_FLAGS := $(GC_FLAGS) -ldflags "-X $(LD_FLAGS)" $(CFLAGS)
 
 # set GOENV
@@ -103,6 +107,7 @@ build_image: image_builder_check ## Build image without DCGM.
 	$(CTR_CMD) build -t $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_BUILD_TAG) \
 		-f $(DOCKERFILE) \
 		--build-arg INSTALL_DCGM=false \
+		--build-arg INSTALL_HABANA=false \
 		--build-arg VERSION=$(VERSION) \
 		--platform="linux/$(GOARCH)" \
 		.
@@ -114,13 +119,25 @@ build_image_dcgm:  image_builder_check ## Build image with DCGM.
 	$(CTR_CMD) build -t $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_BUILD_TAG)-"dcgm" \
 		-f $(DOCKERFILE) \
 		--build-arg INSTALL_DCGM=true \
+		--build-arg INSTALL_HABANA=false \
 		--build-arg VERSION=$(VERSION) \
 		--platform="linux/$(GOARCH)" \
 		.
 	$(CTR_CMD) tag $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_BUILD_TAG)-dcgm $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)-dcgm
+
+build_image_habana: image_builder_check ## Build image with Habana.
+	# build kepler with habana
+	$(CTR_CMD) build -t $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_BUILD_TAG)-"habana" \
+		-f $(DOCKERFILE) \
+		--build-arg INSTALL_HABANA=true \
+		--build-arg INSTALL_DCGM=false \
+		--build-arg VERSION=$(VERSION) \
+		--platform="linux/$(GOARCH)" \
+		.
+	$(CTR_CMD) tag $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_BUILD_TAG)-habana $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)-habana
 .PHONY: build_image_dcgm
 
-build_containerized: build_image build_image_dcgm  ## Build ALL container images.
+build_containerized: build_image build_image_dcgm  build_image_habana## Build ALL container images.
 .PHONY: build_containerized
 
 save-image: ## Save container image.
--- a/manifests/k8s/config/exporter/kustomization.yaml
+++ b/manifests/k8s/config/exporter/kustomization.yaml
@@ -24,6 +24,8 @@ patchesStrategicMerge: []
 # - ./patch/patch-qat.yaml
 # add this line for dcgmi patch
 # - ./patch/patch-dcgmi.yaml
+# add this line for habana patch
+# - ./patch/patch-habana.yaml
 
 secretGenerator:
 - name: redfish
--- a/pkg/sensors/accelerator/gpu/source/gpu_habana.go
+++ b/pkg/sensors/accelerator/gpu/source/gpu_habana.go
@@ -1,0 +1,112 @@
+//go:build gpu && habana
+// +build gpu,habana
+
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package source
+
+import (
+	"time"
+
+	hlml "github.com/HabanaAI/gohlml"
+	"k8s.io/klog/v2"
+
+	"github.com/sustainable-computing-io/kepler/pkg/config"
+)
+
+type GPUHabana struct {
+	collectionSupported bool
+	devices             map[int]Device
+}
+
+func (d *GPUHabana) GetName() string {
+	return "habana"
+}
+
+func (d *GPUHabana) InitLib() error {
+	return nil
+}
+
+// todo: refactor logic at invoking side, if gpu is not set?
+func (d *GPUHabana) Init() error {
+	ret := hlml.Initialize()
+	if ret != nil {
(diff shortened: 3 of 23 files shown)


pkg/sensors/accelerator/gpu/source/gpu_habana.go
```diff
@@ -0,0 +1,111 @@
...
+	gpuEnergy := []uint32{}
+	for _, device := range d.devices {
+		power, ret := device.HabanaDeviceHandler.(hlml.Device).PowerUsage()
+		if ret != nil {
+			klog.V(2).Infof("failed to get power usage on device %v: %v\n", device, ret)
+			continue
+		} else {
+			energy := uint32(uint64(power) * config.SamplePeriodSec)
+			gpuEnergy = append(gpuEnergy, energy)
+		}
```
sthaha, line 72:
```suggestion
		if ret != nil {
			klog.V(2).Infof("failed to get power usage on device %v: %v\n", device, ret)
			continue
		} 
		energy := uint32(uint64(power) * config.SamplePeriodSec)
		gpuEnergy = append(gpuEnergy, energy)

```

pkg/sensors/accelerator/gpu/source/gpu_habana.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (d *GPUHabana) GetGpus() map[int]Device {
+	count, ret := hlml.DeviceCount()
+	if ret != nil {
+		return devices
+	}
+	d.devices = make(map[int]Device, count)
+	for i := 0; i < int(count); i++ {
+		device, ret := hlml.DeviceHandleByIndex(uint(i))
```
sthaha, line 84: Is this `ret` or `err` ?

pkg/sensors/accelerator/gpu/source/gpu_habana.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+	return gpuEnergy
+}
+
+func (d *GPUHabana) GetGpus() map[int]Device {
+	count, ret := hlml.DeviceCount()
+	if ret != nil {
+		return devices
+	}
```
sthaha, line 81: where is `devices` define 🤔  ?

sthaha: also could you please elaborate the logic here? wouldn't get `GetGpus()` compute the devices over and over?

pkg/sensors/accelerator/gpu/source/gpu_habana.go
```diff
@@ -0,0 +1,111 @@
...
+			d.devices[i] = Device{
+				HabanaDeviceHandler: device,
+				NVMLDeviceHandler:   nil,
+			}
+		}
+	}
+	return d.devices
+}
+
+func (d *GPUHabana) GetMIGInstances() map[int]map[int]Device {
```
sthaha, line 95: can we simplify the `map[int]map[int]Devices` to something easier to understand?

pkg/sensors/accelerator/gpu/source/gpu_habana.go
rootfs: it is `ret` (as in hlml go pkg style)

pkg/sensors/accelerator/gpu/source/gpu_habana.go
rootfs: will address that in a later refactoring (we used to think of just nvidia, now we have more than one :D )

rootfs:
the builder image needs to be updated first to build container image. #1374

vendor/github.com/HabanaAI/gohlml/hlml.go
```diff
@@ -0,0 +1,695 @@
+/*
```
maryamtahhan, line 1: Is there a reason why we have all these go libraries and packages in the Kepler code base?

pkg/sensors/accelerator/gpu/source/gpu_habana.go
rootfs: this is fixed

rootfs:
the new builder image is proposed in #1376

manifests/k8s/config/exporter/kustomization.yaml
```diff
@@ -24,6 +24,8 @@ patchesStrategicMerge: []
 # - ./patch/patch-qat.yaml
 # add this line for dcgmi patch
 # - ./patch/patch-dcgmi.yaml
+# add this line for habana patch
+# - ./patch/patch-habana.yaml
```
maryamtahhan, line 28: Where is the `patch-habana.yaml` file? I don't see any added files under `manifests/k8s/config/exporter/patch` or maybe I'm just missing something

Makefile
```diff
@@ -54,6 +54,10 @@ endif
 
 GENERAL_TAGS := 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo libbpf '
 GPU_TAGS := ' gpu '
+ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
+	GPU_TAGS := $(GPU_TAGS)'habana '
```
maryamtahhan, line 58: Do we expect a habana and an nvidia gpu to be installed in the same system? if not then we can probably use one flag or the other rather than bot wdyt?

pkg/sensors/accelerator/gpu/source/gpu_habana.go
```diff
@@ -0,0 +1,112 @@
+//go:build gpu && habana
```
maryamtahhan, line 1: I think we can get away with just one of the tags rather than both enabled at the same time... ?

pkg/sensors/accelerator/gpu/variables_habana.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	gpu_source "github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator/gpu/source"
+)
+
+var (
+	acceleratorOrder = []acceleratorInterface{
+		&gpu_source.GPUDcgm{},
+		&gpu_source.GPUNvml{},
+		&gpu_source.GPUHabana{},
```
maryamtahhan, line 30: I have a question here re the initialisation in gpu.go and the order in this accelerator interface, the init loop is shown below

```
// init initialize the acceleratorImpl and start it
func init() {
	var errLib error
	for i := 0; i < len(acceleratorOrder); i++ {
		klog.Infof("Trying to initialize GPU collector using %s", acceleratorOrder[i].GetName())
		acceleratorImpl = acceleratorOrder[i]
		errLib = acceleratorImpl.InitLib()
		if errLib == nil {
			klog.Infof("Using %s to obtain gpu power", acceleratorImpl.GetName())
			return
		} else {
			klog.Infof("Error initializing %s: %v", acceleratorImpl.GetName(), errLib)
		}
	}
	klog.Infof("no gpu collector available: %v", errLib)
```

based on this init loop is `acceleratorImpl` left set at the last item in the loop once the loop completes. or am I understanding this wrong?

I guess I can't see how this works when acceleratorImpl used in power.go is set to the last item (dummy) in the list rather than dcgm or nvml or habana explicitly.

Apologies if this is something obvious that I missed.

pkg/sensors/accelerator/gpu/variables_habana.go
```diff
@@ -0,0 +1,33 @@
+//go:build gpu && habana
```
maryamtahhan, line 1: I think if we use only the habana flag then we can drop nvml, dummy and dcgm sources from the `acceleratorOrder` variable file...

```
var (
	acceleratorOrder = []acceleratorInterface{
		&gpu_source.GPUHabana{},
	}
)
```

pkg/sensors/accelerator/gpu/source/gpu_habana.go
maryamtahhan: esp when the `InitLib()` function for Habana is not doing anything

pkg/sensors/accelerator/gpu/variables_habana.go
maryamtahhan: edit: the question isn't necessarily related to habana... but in general

rootfs:
@maryamtahhan would you like to take the comments into the GPU code refactoring? Thanks

pkg/sensors/accelerator/gpu/source/gpu_habana.go
rootfs: cc @maryamtahhan

maryamtahhan:
> @maryamtahhan would you like to take the comments into the GPU code refactoring? Thanks

Will do

manifests/k8s/config/exporter/kustomization.yaml
rootfs: added

Makefile
rootfs: this is a good question! Let's have a follow up on the potential use cases. If we build all the GPU support in the same image, that'll be quite big. But maybe in certain use cases, image sizes don't matter.

sthaha (approved):

pkg/sensors/accelerator/gpu/variables_habana.go
maryamtahhan: Just to close this out I missed the return in:

```
		if errLib == nil {
			klog.Infof("Using %s to obtain gpu power", acceleratorImpl.GetName())
			return
```

once it inits the accelerator based on the accelerator order it returns. which is how the nvml/dcgm accelerator is selected

PR #1384, opened 2024-04-26
chore(bpf): Fix Kepler Licensing

dave-tucker:
The eBPF code needs to have a GPL-compatible license. This is due to its use of Linux Kernel structs (like task_struct) and its use of GPL-only eBPF helpers (e.g bpf_probe_read) [1].

eBPF loaders default to GPL when loading code without a license section, which is why this might not have prevented the probes from working correctly. You can verify this by inspecting the loaded probes with bpftool.

This commit adjusts the license of the eBPF files to be:

- GPL-2.0-only OR BSD-2-Clause for Kepler code
- GPL-2.0-only for the generated vmlinux.h files given they come from the Linux Kernel.

It adjusts the README to display these changes.

[1]: https://github.com/torvalds/linux/blob/c942a0cd3603e34dd2d7237e064d9318cb7f9654/kernel/trace/bpf_trace.c#L243-L250

The following contributors will need to agree to the relicensing of this work according to the git history.

- [x] @rootfs 
- [x] @jichenjc
- [x] @marceloamaral
- [x] @sunya-ch
- [x] @jiangphcn

This licensing arrangement is established in mature eBPF projects like Cilium.
Userspace components are unaffected.
There is a [license exception](https://github.com/cncf/foundation/issues/474#issuecomment-1739796978) in the CNCF to allow it.


rootfs:
Thank you @dave-tucker for the investigation and the PR!

@marceloamaral @sunya-ch @jiangphcn @jiangphcn please take a look and let us know if you are good with the transition. Thanks!

sunya-ch (commented):
@dave-tucker Thank you for the PR. This transition looks good to me.

rootfs:
@marceloamaral @jiangphcn can you take a look?

jiangphcn:
LGTM.

PR #1385, opened 2024-04-26
Clean up C development environment

dave-tucker:
This PR contains a few commits.

1. Fixes the bpfassets/libbpf Makefile to now build bytecode
2. Removes the checked-in object files since these should not be version controlled
3. Dynamically links against libbpf - the current build is statically linked which:
  a. will cause issues with RPM packaging (i.e we'd need to rebuild Kepler to fix a vuln in libbpf, libelf or libz)
  b. won't work out of the box on fedora or other linux distros where libbpf is packaged as a shared library
4. Consistently formats the C code with clang-format, using the same style guide from the Cilium project


SamYuan1990:
if we make in dynamic link but not static link.... will it works if the container OS different with host OS? for example what if the host OS missing some library ....

dave-tucker:
> if we make in dynamic link but not static link.... will it works if the container OS different with host OS?

Yes it will work fine. The only limitation is that libbpf must be installed in the Container OS, so in other words you'd be prevented from using a `scratch` image.

> for example what if the host OS missing some library ....

For host OS installation, you'd just add a dependency on libbpf in the RPM/Debian packaging.

Another benefit is that I believe we can remove `Dockerfile.builder` and just build using the UBI image directly.

dave-tucker:
Marking as draft while I ensure that CI is green

PR #1386, opened 2024-04-29
reorg(compose): move monitoring to separate compose

sthaha:
This is needed to support targets that do not need kepler-latest. E.g. in bare-metal and VM validation compose, these targets do not need a kepler-latest to be running all the time.

This PR 
* creates a monitoring only compose file that is reusable 
* reuses monitoring compose target in default compose file


jichenjc (approved):

vprashar2929 (approved):
LGTM

PR #1388, opened 2024-04-30
cleanup startup-profile replaced by pprof http endpoint

sthaha:
The commit - f5910e28d4f0b adds pprof HTTP endpoints to kepler thus removing the need to support -cpu-profile or -memory-profile explicitly.

See also: https://github.com/sustainable-computing-io/kepler/pull/1383

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -21,12 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
-	"os"
-	"os/signal"
-	"runtime"
 	"runtime/debug"
-	"runtime/pprof"
-	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


marceloamaral:
@sthaha could you please also add a documentation in https://github.com/sustainable-computing-io/kepler/tree/main/doc/dev?

sthaha:
@marceloamaral , happy to do that. Since this is a new section to the dev docs, would you mind if this is added as a separate PR?

rootfs (approved):

sthaha:
@marceloamaral as promised - https://github.com/sustainable-computing-io/kepler/pull/1396/files?short_path=f17b7da#diff-f17b7da5234d300e8e070e8795622e174780b3ad8b0ca24e1663580009354b9e .

PR #1392, opened 2024-04-30
feat(test): initial version of validator tool

sthaha:
This commit adds a validator tool that is able to run a stress tests on a remote VM and compare the results by fetching the metrics from prometheus.


e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+@validator.command()
+@click.option(
+    "--script-path", "-s", 
+    default="scripts/stressor.sh", 
+    type=str,
+)
+@pass_config
+def stress(cfg: Config, script_path: str):
+    PROM_QUERIES = {
+        "vm_process_joules_total": {"name": "kepler_process_package_joules_total", "base_labels": {"job": "metal", "pid": "2093543"}},
```
rootfs, line 115: is the pid always 2093543?

rootfs (approved):

e2e/tools/validator/src/validator/cli/__init__.py
vprashar2929: I think it gets updated here: https://github.com/sustainable-computing-io/kepler/pull/1392/files#diff-23a6e727f292aee124b881c947aa1cfff3bcb9c8a6126d2d131e0a610bfdd8b1R140

PR #1393, opened 2024-05-01
tools: add install_yq

maryamtahhan:
Add `yq` to the tools script - it's needed for Kepler Grafana dashboard installation automation 

> Note:  if `yq` isn't available on the platform the grafana deployment script simply skips the step that adds the grafana dashboard to grafana and deploys grafana with all the preconfigured dashboards

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -293,6 +293,7 @@ TOOLS = govulncheck \
 		jq \
 		kubectl \
 		kustomize \
+		yq \
 
 tools:
 	./hack/tools.sh
--- a/hack/tools.sh
+++ b/hack/tools.sh
@@ -34,6 +34,8 @@ declare -r KUSTOMIZE_INSTALL_SCRIPT="https://raw.githubusercontent.com/kubernete
 
 declare -r JQ_VERSION=${JQ_VERSION:-1.7}
 declare -r JQ_INSTALL_URL="https://github.com/jqlang/jq/releases/download/jq-$JQ_VERSION"
+declare -r YQ_VERSION=${YQ_VERSION:-v4.34.2}
+declare -r YQ_INSTALL_URL="https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_${GOOS}_${GOARCH}"
 
 source "$PROJECT_ROOT/hack/utils.bash"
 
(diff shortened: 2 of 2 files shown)


dave-tucker (approved):
Tested. LGTM 🎉

rootfs (approved):

Makefile
```diff
@@ -293,6 +293,7 @@ TOOLS = govulncheck \
 		jq \
```
SamYuan1990, line 293: @rootfs， @maryamtahhan I have to ask a further information for yq?
ref
https://github.com/sustainable-computing-io/local-dev-cluster/blob/main/main.sh#L190
as yq and other dependencies been installed via local-dev-cluster, why should we have yq install here for?
if we want to have grafana dashboard with us in CI, as the grafana been installed through local-dev-cluster with yq....
mean while, if we want to handle json, we have jq here. so it's confusion for yq here.

maryamtahhan: The reason why I added yq here is because we are not guaranteed that someone is going to run the prerequisites install from local-cluster-dev. I don't see that option in the Makefile here or in the cluster.sh... it would be great to streamline all the prerequisites into one place so as to avoid having to make tooling changes in a bunch of different locations TBH.  The prerequisites for local-dev-cluster aren't automated under the scripts/Makefile in the Kepler repo it's only mentioned in the (dev readme)[https://github.com/sustainable-computing-io/kepler/blob/f97a4487af022639f3cde41d8f7efb822fc6c5b0/doc/dev/README.md?plain=1#L25]

The intention behind my change isn't for CI - it's for development workflow - foe e.g. my workflow for development is I would like to test the full stack when I make changes. I start by making changes to the Kepler repo and then I spin up a kind cluster to verify everything - including wanting to see my stats (esp if I have new ones in kepler, prometheus + grafana)... I know I don't have to spin up a cluster to test kepler but TBH I like to keep my development setup clean - a kind cluster is easy to bring up and tear down without having any residual side affects on my system (which I use for multiple projects). If yq isn't installed on my system then I have to manually add the grafana dashboard every time I launch test environment for my changes as the prerequisites for local-dev-cluster aren't automated and the  local-dev-cluster will skip the grafana dashboard automation step if yq isn't installed. 

jq is not used for creating the appropriate yaml configurations for the grafana dashboards - so I'm not sure why it's being referenced. 

I guess I'm just trying to make things easier for the developer. I thought the flow most folks would follow would be:
- clone the Kepler repo
- run `make tools all`
- make changes in kepler
- run `make cluster up`
- run `make deploy`
- check changes through the full stack...

PR #1394, opened 2024-05-02
cleanup(validator): simplify imports and modules

sthaha:
This commit refactors
* config to its on module - validator/config
* creates a validator/stressor.Remote class for remote execution
* Uses ScriptResult Tuple to return result of a script execution


e2e/tools/validator/src/validator/config/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    with open(config_file, 'r') as file:
+        config = yaml.safe_load(file)
+
+    remote_config = config['remote']
+    remote = Remote(
+        host=remote_config['host'],
+        port=remote_config.get('port', 22),
+        user=remote_config.get('username', 'fedora'),
+        password=remote_config.get('password', None),
+        pkey=os.path.expanduser(remote_config.get('pkey', '~/.ssh/id_rsa')),
```
KaiyiLiu1234, line 56: assuming null is passed to pkey, this will produce an error. This means for users that do not need to pass a pkey or want to use password cannot really set this to null.

KaiyiLiu1234 (approved):
LGTM. Only concern is regarding error appearing when passing pkey to be null in validator.yaml.

e2e/tools/validator/src/validator/config/__init__.py
sthaha: Good find! fixed now with a unittest

rootfs (approved):

Issue #1390, opened 2024-04-30
Remove BCC Code


dave-tucker:
### What would you like to be added?

- Remove the older bcc probe code
- Simplify the `pkg/bpfassets` load code as an interface won't be required - only one attachment method is supported.

### Why is this needed?

Given that libbpf is backwards compatible to old kernels, we lose nothing by removing the older bcc-based probe code.

PR #1397, opened 2024-05-02
chore(bpfassets): Remove BCC

dave-tucker:
This removes the BCC code from the repository and
from the bpfassets package. The bpfassets package
has been simplified given we will only support a
single "attacher".

Fixes: #1390


rootfs (approved):

rootfs (approved):

sunya-ch (approved):
Thank you for the PR. I agree that we can now focus on a single attacher. This PR looks good to me.

sthaha (approved):

PR #1399, opened 2024-05-02
feat(bpf): Build for bpfel and bpfeb

dave-tucker:
Since we now have portable probes, we can simplify building only 2 variants of our eBPF bytecode. One for big endian and one for little endian systems. The correct probe is picked at runtime.

### DIFF
--- a/pkg/bpfassets/attacher/libbpf_attacher.go
+++ b/pkg/bpfassets/attacher/libbpf_attacher.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	objectFilename       = "kepler.bpf.o"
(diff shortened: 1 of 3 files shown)


sthaha (approved):
Thank you! Good improvement!

pkg/bpfassets/attacher/libbpf_attacher.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	objectFilename       = "kepler.bpf.o"
+	objectFilename       = "kepler.%s.o"
```
sthaha, line 40: nit: 

```suggestion
	objectFilePattern       = "kepler.%s.o"
```

PR #1400, opened 2024-05-03
Revert "tools: add install_yq"

SamYuan1990:
Reverts sustainable-computing-io/kepler#1393

There seems no need for a PR to add yq.
In CI if we have grafana features, which yq installed at https://github.com/sustainable-computing-io/local-dev-cluster/blob/main/main.sh#L190 this covers case which we build up instance on a new created ec2.

and if we need to handle json, yq and jq been installed via GHA naivety.
https://github.com/actions/runner-images/blob/7bb1d84f7071bfa9c350d7552d9631d9a69bfdb0/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L72-L86

So, what's the reason we add yq install in our code?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -291,7 +291,6 @@ TOOLS = govulncheck \
 		jq \
 		kubectl \
 		kustomize \
-		yq \
 
 tools:
 	./hack/tools.sh
--- a/hack/tools.sh
+++ b/hack/tools.sh
@@ -34,8 +34,6 @@ declare -r KUSTOMIZE_INSTALL_SCRIPT="https://raw.githubusercontent.com/kubernete
 
 declare -r JQ_VERSION=${JQ_VERSION:-1.7}
 declare -r JQ_INSTALL_URL="https://github.com/jqlang/jq/releases/download/jq-$JQ_VERSION"
-declare -r YQ_VERSION=${YQ_VERSION:-v4.34.2}
-declare -r YQ_INSTALL_URL="https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_${GOOS}_${GOARCH}"
 
 source "$PROJECT_ROOT/hack/utils.bash"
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -291,7 +291,6 @@ TOOLS = govulncheck \
 		jq \
 		kubectl \
 		kustomize \
-		yq \
```
sthaha, line 294: Instead of removing it, how about we maintian a different target called `ci-tools` and have 

```
tools: ci-tools 
```
YQ is quite useful since we have a lot of yaml files in `manifests`

maryamtahhan: I'm all for a new target `ci-tools` is ok - but it doesn't encompass the development side.

could we not just have a `deps` target that installs all the depdendencies and centralise all the dependencies/tools in one repo...

From a dev experience POV - it would make the onboarding experience a lot smoother IMHO. Right now we rely on the developer to manually run a command to install the tools they need after finding the appropriate reference in the Readme rather than automating this step

maryamtahhan: My honest feedback is the developer experience for Kepler right now could be greatly improved. It's not just about CI...

maryamtahhan:
> So, what's the reason we add yq install in our code?

To improve developer experience.  If YQ isn't installed then `local-dev-cluster` will skip the grafana dashboard automation step and the developer will need to manually import the dashboard.

TBH as a developer it would be great if the first thing I could do after cloning the kepler repo is run `make deps` or some other target that installs all the tools that I need to deploy the full stack. I don't see a place where the `local-dev-cluster` pre-requisites are automated for developer experience and we rely on the (new) developer to run a bunch of steps to get up and running with kepler. 

IMHO - the kepler new developer experience could be greatly improved.

rootfs:
I tested `jq` and `yq` on my end, for the use case kepler needs (i.e. parsing kubectl output), both can do the job. Since we only need one, what about replacing existing `jq` reference with `yq`?

PR #1404, opened 2024-05-03
chore: Bump Go Version

dave-tucker:
Go 1.20 is not supported since Go 1.22 was released in February 2024.

### DIFF
--- a/build/Dockerfile.builder
+++ b/build/Dockerfile.builder
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.21 as builder
 
 USER 0
 


build/Dockerfile.builder
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.21 as builder
```
SamYuan1990, line 1: as this repo is a single repo image, https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=6633d168b5bb692a240e57b3
should we use latest tag?
@rootfs , @sthaha , @vimalk78

dave-tucker: I'd prefer being explicit. That way the Go version used in CI doesn't drift from what we expect it to be.

rootfs (approved):

PR #1406, opened 2024-05-07
Fixed Validation Script to compare metrics based on timestamps

KaiyiLiu1234:
WIP: Updated prometheus to compare metrics based on timestamps. Currently investigating inconsistencies in prometheus timestamp datapoints, between requests api and prometheus api client functions. 

Signed-off-by: Kaiyi <kaiyiliu21@gmail.com>


e2e/tools/validator/src/validator/prometheus/__init__.py
```diff
@@ -1,66 +1,127 @@
...
+    print(actual_metrics)
+    # clean data to acquire only lists
+    expected_data = merge_prom_metric_list(expected_metrics)
+    actual_data = merge_prom_metric_list(actual_metrics)
+    
+    # remove timestamps that do not match
+    return disjunct_on_timestamps(expected_data, actual_data)
+
+
+def absolute_percentage_error(expected_data, actual_data) -> List[float]:
```
sthaha, line 78: I feel these pure functions can be easily unit-tested. Could you please add to the existing test suite?

e2e/tools/validator/src/validator/prometheus/__init__.py
```diff
@@ -1,66 +1,127 @@
...
+        for list_one_datapoint in prom_data_list_one:
+            if int(list_one_datapoint[0]) == timestamp:
+                list_one_metrics.append(list_one_datapoint[1])
+        for list_two_datapoint in prom_data_list_two:
+            if int(list_two_datapoint[0]) == timestamp:
+                list_two_metrics.append(list_two_datapoint[1])
+    return list_one_metrics, list_two_metrics
 
+
+def compare_metrics(endpoint: str, disable_ssl, start_time: datetime, end_time: datetime, expected_query: str, expected_query_labels: dict, actual_query: str, actual_query_labels: dict) -> Tuple[List[float], List[float]]:
```
sthaha, line 45: we should move this function back to `MetricValidator` that holds all information about prometheus.

KaiyiLiu1234:
I think the best solution is to just include rate(). Grafana uses that and the function should handle missing datapoints and deal with the alternating timestamp issue because both metal and vm use the same scrape interval of 3s, we can use a consistent step of 12s and use the same rate interval (which follows this formula: max(step + Scrape interval, 4 * Scrape interval)). I will incorporate these changes immediately.

KaiyiLiu1234:
Changes have been completed. The last task is to create unit tests for classes and top level functions. @sthaha

e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -39,11 +38,11 @@ def validator(ctx: click.Context, config_file: str):
...
-    PROM_QUERIES = {
-        "vm_process_joules_total": {"name": "kepler_process_package_joules_total", "base_labels": {"job": "metal", "pid": "2093543"}},
-        "platform_joules_vm": {"name": "kepler_node_platform_joules_total", "base_labels": {"job": "vm"}},
-        # "platform_joules_vm_bm" : "kepler_vm_platform_joules_total{job='metal'}"
-    }
+    # PROM_QUERIES = {
+    #     "vm_process_joules_total": {"name": "rate(kepler_process_package_joules_total)", "base_labels": {"job": "metal", "pid": "2093543"}},
+    #     "platform_joules_vm": {"name": "kepler_node_platform_joules_total", "base_labels": {"job": "vm"}},
+    #     # "platform_joules_vm_bm" : "kepler_vm_platform_joules_total{job='metal'}"
+    # }
```
sthaha, line 45: delete ?

sthaha (approved):

Issue #1402, opened 2024-05-03
No value reported by Kepler latest on OpenShift


vprashar2929:
### What happened?

When Kepler using the `latest` is deployed against OpenShift 4.15(VM/BM) it does not produce any metric value. All the metric values are 0. 

Some screenshots for reference:

![Screenshot 2024-05-03 at 4 55 34 PM](https://github.com/sustainable-computing-io/kepler/assets/115542213/32b62faa-53c1-4f11-ac90-c1de3fa15ebf)

![Screenshot 2024-05-03 at 4 58 54 PM](https://github.com/sustainable-computing-io/kepler/assets/115542213/a6680c5d-f9e4-47e5-9565-ce80c880345e)

### What did you expect to happen?

Kepler should report energy values

### How can we reproduce it (as minimally and precisely as possible)?

Deploy Kepler on OpenShift (VM/BM) using the latest image and then check the metrics it exports.

### Kepler image tag

<details>
latest
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>
VM
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

vprashar2929:
Attaching Kepler logs for reference:

Kepler on BM: 
[kepler-bm.log](https://github.com/sustainable-computing-io/kepler/files/15218467/kepler-bm.log)

Kepler on VM: 
[kepler.log](https://github.com/sustainable-computing-io/kepler/files/15218474/kepler.log)

vprashar2929:
cc: @rootfs

vimalk78:
[...]

PR #1407, opened 2024-05-07
fix(bpf): Fix issue(s) introduced with bpf refactor

dave-tucker:
This fixes a few issues identified with the bpf code format and refactor.

1. Zero initialize all variables
2. Use the bpf_perf_event_read_value helper exlusively
3. Cherry-pick of PR #1410 - reported by @vimalk78 - which correctly sizes the `processes` and `pid_time` maps. Without this patch entries won't be created in the `processes` map for pids that are > `number of cpu cores` which fundamentally breaks power reporting.

In addition, add some logging around eBPF array resizing.
    
Fixes: #1402 #1411

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -6,14 +6,14 @@ struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, u32);
 	__type(value, process_metrics_t);
-	__uint(max_entries, NUM_CPUS);
+	__uint(max_entries, MAP_SIZE);
 } processes SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, u32);
 	__type(value, u64);
-	__uint(max_entries, NUM_CPUS);
+	__uint(max_entries, MAP_SIZE);
 } pid_time SEC(".maps");
 
 struct {
@@ -128,7 +128,7 @@ static inline u64 calc_delta(u64 *prev_val, u64 *val)
 static inline u64 get_on_cpu_cycles(u32 *cpu_id)
 {
 	u64 delta, val, *prev_val;
-	int error;
+	long error;
 	struct bpf_perf_event_value c = {};
 
 	error = bpf_perf_event_read_value(
@@ -147,7 +147,7 @@ static inline u64 get_on_cpu_cycles(u32 *cpu_id)
 static inline u64 get_on_cpu_instr(u32 *cpu_id)
 {
 	u64 delta, val, *prev_val;
-	int error;
+	long error;
 	struct bpf_perf_event_value c = {};
 
 	error = bpf_perf_event_read_value(
@@ -166,7 +166,7 @@ static inline u64 get_on_cpu_instr(u32 *cpu_id)
 static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
 {
 	u64 delta, val, *prev_val;
-	int error;
+	long error;
 	struct bpf_perf_event_value c = {};
 
 	error = bpf_perf_event_read_value(
@@ -205,7 +205,7 @@ int kepler_sched_switch_trace(struct sched_switch_info *ctx)
 	pid_t cur_pid;
 
 	struct process_metrics_t *cur_pid_metrics, *prev_pid_metrics;
-	struct process_metrics_t buf;
+	struct process_metrics_t buf = {};
 
 	if (SAMPLE_RATE > 0) {
 		if (counter_sched_switch > 0) {
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -95,8 +95,9 @@ int counter_sched_switch = 0;
 
 static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
 {
-	u64 cpu_time, *prev_ts;
-	pid_time_t prev_pid_key, new_pid_key;
+	u64 cpu_time = 0;
+	u64 *prev_ts;
+	pid_time_t prev_pid_key, new_pid_key = {};
 
 	prev_pid_key.pid = prev_pid;
 	prev_ts = bpf_map_lookup_elem(&pid_time, &prev_pid_key);
@@ -115,7 +116,7 @@ static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
 
 static inline u64 calc_delta(u64 *prev_val, u64 *val)
 {
-	u64 delta;
+	u64 delta = 0;
 
 	if (prev_val) {
 		if (*val > *prev_val)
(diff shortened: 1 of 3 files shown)


bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -127,15 +127,14 @@ static inline u64 calc_delta(u64 *prev_val, u64 *val)
...
-	struct bpf_perf_event_value c;
 	int error;
 
 	// TODO: Fix Verifier errors upon changing this to bpf_perf_event_read_value
 	error = bpf_perf_event_read(&cpu_cycles_event_reader, *cpu_id);
 	if (error < 0) {
 		return delta;
 	}
-	val = c.counter;
+	val = error;
```
sthaha, line 137: could you please help me understand how this fixes the issue?
I see that `get_on_cpu_instr` also makes use of `c.counter`. Should that be changed as well?

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -127,15 +127,14 @@ static inline u64 calc_delta(u64 *prev_val, u64 *val)
 static inline u64 get_on_cpu_cycles(u32 *cpu_id)
 {
 	u64 delta, val, *prev_val;
-	struct bpf_perf_event_value c;
```
sthaha, line 130: shouldn't this have been `struct bpf_perf_event_value c = {};` to 0 out all values?

bpfassets/libbpf/src/kepler.bpf.c
vimalk78: looking at the [old code](https://github.com/sustainable-computing-io/kepler/blob/4e8abc8d09cc54b57bc26d95a12c5e938862bda2/bpfassets/libbpf/src/kepler.bpf.c#L116-L141), function `bpf_perf_event_read_value` is getting used in the newer kernels, which i assume to be the case for openshift node, and is also the recommended function over `bpf_perf_event_read`

bpfassets/libbpf/src/kepler.bpf.c
vimalk78: and initialize variables `delta` etc

bpfassets/libbpf/src/kepler.bpf.c
dave-tucker: That's not entirely true. `BPF_PERF_EVENT_READ_VALUE_AVAILABLE` was always true given it's a compile time check + all CI systems (and likely all developer systems) are using > 5.12 kernels.

`bpf_perf_event_read_value` is the desired helper to use here, but see #1411 as to why that's not being used here.

This bug stems from me attempting to use ☝️ but not backing that change out completely.
The other comments are reasonable, I've forced pushed a commit that implements those.

bpfassets/libbpf/src/kepler.bpf.c
dave-tucker: I can take another pass over the bpf code and ensure everything is zero initalized. Will address that in a follow up.

dave-tucker:
@sthaha @vimalk78 thanks for your review and comments.

I took a pass over the code and ensured that everything that needed to be zero'd, was zero'd.
In doing so I was able to fix #1411 so we now use `bpf_perf_event_read_value` everywhere.
In addition, I was able to verify that this solves the verifier error encountered in #1402.

Force pushed and updated commit message - I'd appreciate you both taking another look.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -203,7 +204,8 @@ int kepler_sched_switch_trace(struct sched_switch_info *ctx)
 	u64 pid_tgid, cgroup_id, cur_ts;
 	pid_t cur_pid;
 
-	struct process_metrics_t *cur_pid_metrics, *prev_pid_metrics, buf;
+	struct process_metrics_t *cur_pid_metrics, *prev_pid_metrics;
+	struct process_metrics_t buf;
```
sthaha, line 208: should this be zeroed out as well?

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -127,14 +128,14 @@ static inline u64 calc_delta(u64 *prev_val, u64 *val)
 static inline u64 get_on_cpu_cycles(u32 *cpu_id)
 {
 	u64 delta, val, *prev_val;
-	struct bpf_perf_event_value c;
 	int error;
```
sthaha, line 131: nit: https://man7.org/linux/man-pages/man7/bpf-helpers.7.html#:~:text=long%20bpf_perf_event_read_value(struct%20bpf_map%20*map%2C%20u64%20flags%2C%0A%20%20%20%20%20%20%20struct%20bpf_perf_event_value%20*buf%2C%20u32%20buf_size) says - `long` instead of `int`

sthaha:
@dave-tucker , thanks a lot for cleaning up the code quite a bit 👼  !

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -203,7 +204,8 @@ int kepler_sched_switch_trace(struct sched_switch_info *ctx)
 	u64 pid_tgid, cgroup_id, cur_ts;
 	pid_t cur_pid;
 
-	struct process_metrics_t *cur_pid_metrics, *prev_pid_metrics, buf;
+	struct process_metrics_t *cur_pid_metrics, *prev_pid_metrics;
+	struct process_metrics_t buf;
 
 	if (SAMPLE_RATE > 0) {
```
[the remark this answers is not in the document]
dave-tucker: re: `SAMPLE_RATE` it's 5 in the code - but can be changed back to 0.
Regardless, at program load time it gets patched [here](https://github.com/sustainable-computing-io/kepler/blob/5cf570b941c1f2e4d82651af0c42bad4595069a9/pkg/bpfassets/attacher/libbpf_attacher.go#L122) to whatever the user has configured where the default of 0 remains unchanged.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -127,14 +128,14 @@ static inline u64 calc_delta(u64 *prev_val, u64 *val)
 static inline u64 get_on_cpu_cycles(u32 *cpu_id)
 {
 	u64 delta, val, *prev_val;
```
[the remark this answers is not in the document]
dave-tucker: delta is assigned in the call to `calc_delta` and it's never used before being assigned so it's safe to not zero initiatlize - same with val. at least that's my understanding. I'll happily take another pass through and zero out all u64/u32/int etc... as well to be extremely cautious if you prefer.

sthaha:
@dave-tucker , could you please rebase this PR? There are some open comments as well which you may want to address.

vimalk78 (approved):
tested, works fine. `/lgtm
```
1583: hash  name processes  flags 0x0
	key 4B  value 112B  max_entries 32768  memlock 6034304B
	btf_id 812
	pids kepler(750518)
1584: hash  name pid_time  flags 0x0
	key 4B  value 8B  max_entries 32768  memlock 2624352B
	btf_id 812
	pids kepler(750518)
```

rootfs (approved):

rootfs:
@dave-tucker for 5.4 kernel, bpf_perf_event_read_value has an issue (discussed [here](https://github.com/iovisor/gobpf/issues/317#issuecomment-1829741823))

PR #1410, opened 2024-05-07
fix(bpf): Incorrect map size for processes and pid_time

dave-tucker:
Reported by @vimalk78 - Thanks 🙏

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -6,14 +6,14 @@ struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, u32);
 	__type(value, process_metrics_t);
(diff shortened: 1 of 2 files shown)


sthaha:
@dave-tucker , could you please shed some light into what this fixes? I understand the earlier code had a larger buffer but  
I was thinking more in terms of missing tests. Some test should have flagged the regression which we are missing and probably an opportunity to add it.

dave-tucker:
@sthaha I'd gladly add tests but unfortunately doing so is non-trivial given there is no test rigging for ebpf tests in the project (yet). I'll open an issue to add them as well as highlighting any decisions we'll want to make up-front about how to handle that type of testing.
Without a larger buffer we're limited to capturing data for $number_of_cpu_core processes, which could be very low (single digits) 😱

PR #1415, opened 2024-05-09
fix(bpf): fix pid key for querying processes map

vimalk78:
`bpf_get_current_pid_tgid` gives both process id and group id, need to use mask to extract pid from the return value.
Same is used while creating the map entry in `kepler_sched_switch_trace` 

Cc: @rootfs @dave-tucker

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -266,7 +266,7 @@ int kepler_irq_trace(struct trace_event_raw_softirq *ctx)
 	struct process_metrics_t *process_metrics;
 	unsigned int vec;
 
-	cur_pid = bpf_get_current_pid_tgid();
+	cur_pid = bpf_get_current_pid_tgid() & 0xffffffff;
 	vec = ctx->vec;
 	process_metrics = bpf_map_lookup_elem(&processes, &cur_pid);
 	if (process_metrics != 0) {
@@ -286,7 +286,7 @@ int kepler_read_page_trace(void *ctx)
 	u32 cur_pid;
 	struct process_metrics_t *process_metrics;
 
-	cur_pid = bpf_get_current_pid_tgid();
+	cur_pid = bpf_get_current_pid_tgid() & 0xffffffff;
 	process_metrics = bpf_map_lookup_elem(&processes, &cur_pid);
 	if (process_metrics) {
 		process_metrics->page_cache_hit++;
@@ -301,7 +301,7 @@ int kepler_write_page_trace(void *ctx)
 	u32 cur_pid;
 	struct process_metrics_t *process_metrics;
 
-	cur_pid = bpf_get_current_pid_tgid();
+	cur_pid = bpf_get_current_pid_tgid() & 0xffffffff;
 	process_metrics = bpf_map_lookup_elem(&processes, &cur_pid);
 	if (process_metrics) {
 		process_metrics->page_cache_hit++;


bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -266,7 +266,7 @@ int kepler_irq_trace(struct trace_event_raw_softirq *ctx)
 	struct process_metrics_t *process_metrics;
 	unsigned int vec;
 
-	cur_pid = bpf_get_current_pid_tgid();
+	cur_pid = bpf_get_current_pid_tgid() & 0xffffffff;
```
dave-tucker, line 269: `cur_pid` is u32, but the value returned by `bpf_get_current_pid_tgid` is `u64` - when it's cast to u32 (IIRC) it takes the lowest bits. This is obviously an issue given that the pid in the `pid_tgid` is at the highest (left-most) value.
Bit masking won't solve that problem.

What we should be doing is something like this:

```c
u64 pid_tgid;
u32 pid;

pid_tgid = bpf_get_current_pid_tgid();
pid = pid_tgid >> 32;
```

Or you could shift at assignment time to avoid the intermediate variable:
```
cur_pid = bpf_get_current_pid_tgid() >> 32;
```

dave-tucker (changes requested):

bpfassets/libbpf/src/kepler.bpf.c
vimalk78: thanks for the review @dave-tucker 

as per `man bpf-helpers`

```
       u64 bpf_get_current_pid_tgid(void)

              Description
                     Get the current pid and tgid.

              Return A 64-bit integer containing the current tgid and pid, and created as such: current_task->tgid << 32 | current_task->pid.
```

and in `kepler_sched_switch_trace`  it is 

```
	u64 pid_tgid
	pid_t cur_pid;

	pid_tgid = bpf_get_current_pid_tgid();
	cur_pid = pid_tgid & 0xffffffff;
	tgid = pid_tgid >> 32;
```
[kepler.bpf.c#L217-L219](https://github.com/sustainable-computing-io/kepler/blob/main/bpfassets/libbpf/src/kepler.bpf.c#L217-L219)

so it seems there no issue to begin with if `cur_pid` is 32 bit?

dave-tucker: See: https://github.com/search?q=bpf_get_current_pid_tgid&type=code
You'll note that almost all examples in the wild use `pid` as `bpf_get_current_pid_tgid() >> 32`
It would seem kepler has inverted tgid/pid by accident?

dave-tucker: > I looked into kernel code, and am confused.

Ha yes me too. If pid is in the lower bits then there is no bitmask needed - the upper bits are dropped when it's cast from u64 to u32.

To be 100% certain about what's happening I'll write a test program 😆 and compare it with `bpftrace` which helpfully exposes the `pid` and `tid` as variables.

dave-tucker: Ok so here's the verdict. I have a VM running as PID 93307 on my system (as observed from userspace).

Using bpftrace's pid and tid helpers I see the following:

```
$ sudo bpftrace -e 'tracepoint:sched:sched_switch { printf("pid:%d tgid:%d\n", pid, tid)}' | grep 933907
pid:933907 tgid:933919
pid:933907 tgid:933921
pid:933907 tgid:933919
pid:933907 tgid:933919
pid:933907 tgid:933919
pid:933907 tgid:933921
```

Note: PID is correct (as observed from userspace).

Here's my test eBPF program, mimicking what kepler is doing.

```c
#include <linux/types.h>

#include <bpf/bpf_helpers.h>

SEC("tp/sched/sched_switch")
int sched_switch(void *ctx)
{
	__u64 pid_tgid;
	__u32 tgid, pid;

	pid_tgid = bpf_get_current_pid_tgid();
	pid = pid_tgid; // cast to u32 drops the high bits
	tgid = pid_tgid >> 32;
	bpf_printk("pid: %d, tgid: %d", pid, tgid);
	return 0;
}

char __license[] SEC("license") = "Dual BSD/GPL";
```

And the result of running this and attaching to the tracelog is as follows:

```console
$ clang -c -target bpf -I ./bpfassets/libbpf/include -D __TARGET_ARCH_x86_64 -O2 test.bpf.c -o test.bpf.o
$ sudo bpftool prog load test.bpf.o /sys/fs/bpf/kepler autoattach
sudo bpftool prog tracelog | grep 933907                       
       CPU 1/KVM-933919  [010] d..41 612674.668881: bpf_trace_printk: pid: 933919, tgid: 933907
  kvm-pit/933907-933923  [007] d..41 612674.689231: bpf_trace_printk: pid: 933923, tgid: 933923
    vhost-933907-933916  [004] d..41 612674.712091: bpf_trace_printk: pid: 933916, tgid: 933907
       CPU 1/KVM-933919  [004] d..41 612674.712301: bpf_trace_printk: pid: 933919, tgid: 933907
       CPU 1/KVM-933919  [004] d..41 612674.724466: bpf_trace_printk: pid: 933919, tgid: 933907
```

NOTE: The tgid and pid are **reversed** of what you'd expect to observe from userspace.

The best explanation I can find for this [is here](https://stackoverflow.com/questions/9305992/if-threads-share-the-same-pid-how-can-they-be-identified) but TL;DR is that the kernel and userspace view PID differently.

So TL;DR we should probably assign `pid = bpf_get_current_pid_tgid() >> 32;` and add some comments to the code.
Unless the 2 are being switched around anywhere else in the codebase.

sthaha: IIUC, `tgid` refers to the the `pid` (a process's id)  in the traditional sense of that term. My understanding is as follows

```
foobar(tgid: 9000, pid: 9000)
| --- foobar:thread-1( tgid: 9000, pid: 9100)
| --- foobar:thread-2( tgid: 9000, pid: 9200)
|
| --- <clone> foobar(tgid: 12000, pid: 12000 | ppid: 9000)
           | --- foobar: thread-1 (tgid: 12000, pid: 20000)
```


Hence what we are interested in is the tgid (the parent) of all threads instead of individual threads (pid), so `u32 tgid = bpf_get_current_pid_tgid()` should be what it is. 

@marceloamaral @rootfs could you please shed some light into this?

sthaha (changes requested):

bpfassets/libbpf/src/kepler.bpf.c
sthaha: see also: https://github.com/torvalds/linux/blob/448b3fe5a0eab5b625a7e15c67c7972169e47ff8/Documentation/accounting/taskstats.rst#terminology

vimalk78:
if we see the overall picture:
- In the kernel, the entity that gets scheduled on the cpu is a process (aka thread in user space), and the key in kernel space is `pid`. when we are counting instructions, time, cache misses, page reads etc , the key there should also be `pid`. 
- In user space we read the power from power sources (rapl etc) and divide that power among the entities that get scheduled (threads) as per kepler logic (ratio, lr etc). then when we sum the power usage for a user space process, we should use `tgid` which is the `pid` in userspace. 
checking further.

vimalk78:
the methods `kepler_irq_trace` , `kepler_read_page_trace`, and `kepler_write_page_trace` are getting the correct value for key in the `processes` bpf map. 
the changes in this PR are not required. Closing the PR.

vimalk78:
> Unless the 2 are being switched around anywhere else in the codebase.

@dave-tucker it is getting switched, my bad i should have read more code.

`kepler.bpf.h`:
```C
typedef struct process_metrics_t {
	u64 cgroup_id;
	u64 pid;  // pid is the kernel space view of the thread id
	u64 tgid; // tgid is the user space view of the pid
	u64 process_run_time;
	u64 task_clock_time;
	u64 cpu_cycles;
	u64 cpu_instr;
	u64 cache_miss;
	u64 page_cache_hit;
	u16 vec_nr[10];
	char comm[16];
} process_metrics_t;
```
`bpf_perf.go`:
```go
// must be in sync with bpf program
type ProcessBPFMetrics struct {
	CGroupID       uint64
	ThreadPID      uint64 /* thread id */
	PID            uint64 /* TGID of the threads, i.e. user space pid */
	ProcessRunTime uint64 /* in ms */
	TaskClockTime  uint64 /* in ms */
	CPUCycles      uint64
	CPUInstr       uint64
	CacheMisses    uint64
	PageCacheHit   uint64
	VecNR          [config.MaxIRQ]uint16 // irq counter, 10 is the max number of irq vectors
	Command        [16]byte
}
```

PR #1408, opened 2024-05-07
Dockerfile.builder: update libbpf + elfutils pkgs

maryamtahhan:
updating libbpf + elfutils package versions to align with local-dev-cluster as these were causing build failures in the local-dev-cluster unit tests (as old packages didn't exist)

https://github.com/sustainable-computing-io/local-dev-cluster/pull/80/files#r1588896032

PR #1417, opened 2024-05-09
Dockerfile: update libbpf-source linux version

maryamtahhan:
update the linux version to match the changes in https://github.com/sustainable-computing-io/kepler/commit/f1a8267b6dea1795124bb4d9827d507f5d3f6ec9 
It seems this was missed in the checks for https://github.com/sustainable-computing-io/kepler/pull/1408 - only found it today when running `make build_image`

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -39,8 +39,8 @@ RUN set -e -x ;\
 		yum clean all
 
 COPY --from=builder /workspace/_output/bin/kepler /usr/bin/kepler
-COPY --from=builder /libbpf-source/linux-5.14.0-333.el9/tools/bpf/bpftool/bpftool /usr/bin/bpftool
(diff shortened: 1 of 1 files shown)


sthaha (commented):
lgtm
I will let @SamYuan1990 and @rootfs take another look before merging

dave-tucker (approved):
LGTM

dave-tucker:
Note that CI is red on #1418 and #1419 and I believe that will be fixed by this PR.

sthaha (approved):

SamYuan1990:
so... which means we bumped up linux kernel version? to be specific, which means when kepler running on a kube node, the linux kernel version should be bump up accordingly with this PR? @rootfs / @marceloamaral

Issue #1403, opened 2024-05-03
Audit usage of build tags


dave-tucker:
### What would you like to be added?

While working on #1390 I fell into a trap.
Build tags are currently being used to gate:
- bcc
- libbpf
- gpu
- darwin

Due to the way CI is set up (lint for example) and how tooling like `go lint` and `go vet` work it's quite possible that the linters aren't being run on all files - meaning that reviewers will need to be ultra cautious on vetting changes to these files.

### Why is this needed?

To ensure everything builds and lints properly it would be a good idea to drop all unnecessary build tags.

Ideally that would include the `darwin` stubs as this seems to create a lot of work for very little gain - i.e you'd be better building inside a Linux Container on macOS (given this software is macOS specific) than having to maintain a sprawl of stub files.

PR #1423, opened 2024-05-13
chore: Enable linter for libbpf/gpu tags

dave-tucker:
This PR enables the linter for some packages that were disabled due to build tags.
It fixes the issues identified by the linter so we can pass CI.

Relates To: #1403


dave-tucker:
I'm going to close this for now. This still needs to be done, but I'd rather see #1413 land first.

PR #1435, opened 2024-05-15
Rename pkg/bpfassets/attacher.Attacher to pkg/bpf.Exporter

dave-tucker:
Follow up to #1413
It was pointed out in the review that we could rename the package to `pkg/bpf`.
This was done.
It was also pointed out that `Attacher` wasn't a helpful name.
I decided to rename this as `Exporter` since it is responsible for exporting data from bpf.


vimalk78 (approved):

vimalk78:
lgtm, request @rootfs @marceloamaral @sthaha to take a look

rootfs (approved):

PR #1437, opened 2024-05-15
update rpm build with new licenses

rootfs:


dave-tucker:
Do we need to add the license files again? Surely they can just be re-used from the repository root 🤔

dave-tucker (approved):

PR #1443, opened 2024-05-16
feat: Fixed eBPF Feature Detection

dave-tucker:
On working to migrate to cilium/ebpf I started to hit issues with
nil pointer dereferences that indicated that something odd was happening
between eBPF feature detection and metrics export.

I discovered that there were additional global variables in many of the
packages whose values are set before eBPF feature detection has
happened.

It appears that in these cases - through the various layers of
indirection - we may try and read/write metrics that do not exist
causing a nil pointer dereference.

There appear to be several places where we're accessing map keys
without checking they are nil - that will need to be fixed in a
follow up. This PR attempts to at least ensure that from the BPF
side that the keys we expect to exist, do actually exist in the
maps that matter.

eBPF feature detection how produces a bpf.SupportedMetrics
struct, which is now consumed in various places in the codebase
to ensure that the metrics we're exporting match those supported
by the eBPF exporter.


vimalk78:
pls share the NPE, and steps to reproduce

dave-tucker:
NPE:
```
goroutine 71 [running]:
github.com/sustainable-computing-io/kepler/pkg/collector/stats/types.(*UInt64StatCollection).SumAllAggrValues(0x16d0[40](https://github.com/sustainable-computing-io/kepler/actions/runs/9114066283/job/25057276989?pr=1438#step:11:41)0?)
	/workspace/pkg/collector/stats/types/types.go:144 +0x16
github.com/sustainable-computing-io/kepler/pkg/metrics/utils.CollectResUtil(0x16d3460?, {0x179f7e0?, 0xc0000e81c0?}, {0x18f5588, 0xd}, {0x1b67f28, 0xc000066368})
	/workspace/pkg/metrics/utils/utils.go:1[41](https://github.com/sustainable-computing-io/kepler/actions/runs/9114066283/job/25057276989?pr=1438#step:11:42) +0x6ed
github.com/sustainable-computing-io/kepler/pkg/metrics/utils.CollectResUtilizationMetrics(0x1?, {0x179f7e0, 0xc0000e81c0}, 0xc0029d8180?, 0x0)
	/workspace/pkg/metrics/utils/utils.go:48 +0xc5
github.com/sustainable-computing-io/kepler/pkg/metrics/container.(*collector).Collect(0xc0025f8ed0, 0xc000061f60?)
	/workspace/pkg/metrics/container/metrics.go:105 +0xd8
github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
	/workspace/vendor/github.com/prometheus/client_golang/prometheus/registry.go:[45](https://github.com/sustainable-computing-io/kepler/actions/runs/9114066283/job/25057276989?pr=1438#step:11:46)5 +0x102
created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather in goroutine 1
	/workspace/vendor/github.com/prometheus/client_golang/prometheus/registry.go:5[47](https://github.com/sustainable-computing-io/kepler/actions/runs/9114066283/job/25057276989?pr=1438#step:11:48) +0xbab
```

`utils.CollectResUtil` is the offending function, where it's assuming that a hashmap entry exists - which it may not.

As far as I can tell the NPE was caused by the following chain of events:
- `CollectResUtilizationMetrics` gets called  via the prometheus collector
- It iterates over several arrays of constants e.g `consts.SCMetricNames` and `consts.IRQMetricNames` and then calls down into `utils.CollectResUtil`
  - NOTE: `config.IsIRQCounterMetricsEnabled()` gets checked before it tries to do anything with the IRQ metrics
- `CollectResUtil` takes an `interface {}` argument (in this case operating on a `ContainerStats` object)
- It indexes directly into `ContainerStats.ResourceUsage["bpf_net_tx_irq"]` without checking it exists

In which circumstances would "bpf_net_tx_irq" not exist then?

- We can trace `NewContainerStats` -> `NewStats` and see that the `ResourceUsage` map is populated from a global variable called `AvailableBPFSWCounters`
- `InitAvailableParamAndMetrics` sets the global ☝️ **after** the bpf programs are loaded - it was empty beforehand

So the only way this can occur is when `config.IsIRQCounterMetricsEnabled() == true` **BUT**  we don't have any of the IRQ keys in the array of supported eBPF software counters. I'll continue to debug here to see why that's happening.

I'm not sure what I managed to do with the cilium/ebpf port and with this patch to make that issue worse... but there is definitely something fishy happening 🐟  Still trying to untangle this to find out what's going on.

vimalk78:
> NPE:

on which commit? does `main` + this PR reproduce it? or `main` + cillium PR ?

vimalk78:
> > NPE:
> 
> on which commit? does `main` + this PR reproduce it? or `main` + cillium PR ?

got it. `main` + cillium PR

dave-tucker:
> main + cillium PR ?

That will reproduce it (for now). I'm only able to reproduce it in CI so using this PR to help debug it.

dave-tucker:
This PR is ready to go.

As for the root cause, I ended up hooking up a debugger and I found out that the issue was that in #1413 I'd forgotten to attach one of the perf_events (`config.TaskClock`). I've fixed this and I'll push into CI tomorrow to see if it resolves the NPE.

Either way the issue is basically because we have 2 competing definitions of "what's supported by ebpf". This PR fixes that.

pkg/bpf/exporter.go
```diff
@@ -98,16 +86,11 @@ type perfCounter struct {
...
-func (e *exporter) GetEnabledBPFSWCounters() []string {
-	return e.enabledSoftwareCounters
-}
-
-func (e *exporter) HardwareCountersEnabled() bool {
-	return e.hardwareCountersEnabled
+func (e *exporter) SupportedMetrics() SupportedMetrics {
+	return SupportedMetrics{
+		HardwareCounters: e.enabledHardwareCounters,
+		SoftwareCounters: e.enabledSoftwareCounters,
```
sthaha, line 92: Could we return a copy instead of pointer to the real object to avoid and unintended side effects.

pkg/bpf/exporter.go
```diff
@@ -66,11 +56,10 @@ type exporter struct {
 	cpuCores              int
 	// due to performance reason we keep an empty struct to verify if a new read is also empty
 	emptyct                 ProcessBPFMetrics
-	hardwareCountersEnabled bool
 	byteOrder               binary.ByteOrder
 	perfEventFds            []int
-	enabledHardwareCounters []string
-	enabledSoftwareCounters []string
+	enabledHardwareCounters map[string]struct{}
+	enabledSoftwareCounters map[string]struct{}
```
sthaha, line 62: since we already depend on apimachinery, https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set fits perfectly here.

pkg/bpf/exporter.go
```diff
@@ -185,23 +168,26 @@ func (e *exporter) attach() error {
 	if _, err = prog.AttachGeneric(); err != nil {
 		klog.Infof("failed to attach tracepoint/sched/sched_switch: %v", err)
```
sthaha, line 169: should we exit/return here?

pkg/bpf/exporter.go
```diff
@@ -185,23 +168,26 @@ func (e *exporter) attach() error {
...
-		if err != nil {
-			klog.Warningf("could not get kepler_irq_trace: %v", err)
-			// disable IRQ metric
-			config.ExposeIRQCounterMetrics = false
-		} else {
+		err := func() error {
+			// attach softirq_entry tracepoint to kepler_irq_trace function
+			irq_prog, err := e.module.GetProgram("kepler_irq_trace")
+			if err != nil {
+				return fmt.Errorf("could not get kepler_irq_trace: %v", err)
```
sthaha, line 179:
```suggestion
				return fmt.Errorf("could not get kepler_irq_trace: %w", err)
```

pkg/bpf/exporter.go
```diff
@@ -243,30 +232,36 @@ func (e *exporter) attach() error {
 		config.TaskClock:      {unix.PERF_TYPE_SOFTWARE, unix.PERF_COUNT_SW_TASK_CLOCK},
 	}
 
+	cleanup := func() error {
+		unixClosePerfEvents(e.perfEventFds)
+		e.perfEventFds = []int{}
+		e.enabledHardwareCounters = map[string]struct{}{}
```
sthaha, line 238: should we clear off  the software counters as well?

pkg/bpf/exporter.go
```diff
@@ -243,30 +232,36 @@ func (e *exporter) attach() error {
...
-				klog.Warningf("could not attach perf event %s: %v. Are you using a VM?\n", bpfPerfArrayName, perfErr)
-				// if any counter is not enabled, we need disable HardwareCountersEnabled
-				e.hardwareCountersEnabled = false
-			}
-			for i, fd := range fds {
-				err = bpfMap.Update(unsafe.Pointer(&i), unsafe.Pointer(&fd))
-				if err != nil {
-					return fmt.Errorf("failed to update bpf map: %v", err)
-				}
+			return cleanup()
```
sthaha, line 247: earlier, we were skipping (continue) to the next counter, whereas here we are quitting the function. Is this intended?

pkg/bpf/exporter.go
```diff
@@ -243,30 +232,36 @@ func (e *exporter) attach() error {
...
+		}
+		fds, perfErr := unixOpenPerfEvent(counter.EvType, counter.EvConfig, e.cpuCores)
+		if perfErr != nil {
+			klog.Warningf("could not attach perf event %s: %v. Are you using a VM?\n", bpfPerfArrayName, perfErr)
+			return cleanup()
```
sthaha, line 252: same as above

pkg/bpf/exporter.go
dave-tucker: Good catch - I've been doing Rust too long 🦀 and had forgotten that Go maps are passed by reference 😮‍💨

pkg/bpf/exporter.go
dave-tucker: yes same case here as well. the only edge case is the `config.TaskClock` counter.
That one 100% needs to be attached, even if the others fail. Failure to do so should be an error - I've updated the logic to reflect this.

pkg/bpf/exporter.go
```diff
@@ -226,47 +213,75 @@ func (e *exporter) attach() error {
...
 		} else {
-			if !slices.Contains(e.enabledSoftwareCounters, config.PageCacheHit) {
-				e.enabledSoftwareCounters = append(e.enabledSoftwareCounters, config.PageCacheHit)
-			}
+			e.enabledSoftwareCounters[config.PageCacheHit] = struct{}{}
 		}
 	}
 
+	if !config.ExposeHardwareCounterMetrics {
+		klog.Infof("Hardware counter metrics are disabled")
```
sthaha, line 221: nit: How about we Warn?

sthaha (commented):
lgtm

rootfs (approved):

PR #1444, opened 2024-05-17
fix(bpf): cleanup initialising structs and nested ifs

sthaha:
This commit changes all declarae and then initialise code of form below

```
struct foobar x; x.a = a; x.b = b;
```

to
```
struct foobar x = { .a = a, .b = b, };
```
The difference in the code produced is as follows

```
;  new_pid_key.pid = cur_pid;                             | pid_time_t new_pid_key = {.pid = cur_pid };
  0: 63 6a fc ff 00 00 00 00	*(u32 *)(r10 - 0x4) = r6  |  0:	79 a1 68 ff 00 00 00 00	r1 = *(u64 *)(r10 - 0x98)
  1: bf a2 00 00 00 00 00 00	r2 = r10                  |  1:	63 1a fc ff 00 00 00 00	*(u32 *)(r10 - 0x4) = r1
  2: 07 02 00 00 fc ff ff ff	r2 += -0x4                |  2:	bf a2 00 00 00 00 00 00	r2 = r10
  3: bf a3 00 00 00 00 00 00	r3 = r10                  |  3:	07 02 00 00 f0 ff ff ff	r2 += -0x10
  4: 07 03 00 00 70 ff ff ff	r3 += -0x90

```

Modifies calc_delta to remove unnecessary pointer deferencing of second arg - `val`

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -96,11 +96,10 @@ int counter_sched_switch = 0;
 static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
 {
 	u64 cpu_time = 0;
-	u64 *prev_ts;
-	pid_time_t prev_pid_key, new_pid_key = {};
+	pid_time_t prev_pid_key = { .pid = prev_pid };
+	pid_time_t new_pid_key = { .pid = cur_pid };
 
-	prev_pid_key.pid = prev_pid;
-	prev_ts = bpf_map_lookup_elem(&pid_time, &prev_pid_key);
+	u64 *prev_ts = bpf_map_lookup_elem(&pid_time, &prev_pid_key);
 	if (prev_ts) {
 		// Probably a clock issue where the recorded on-CPU event had a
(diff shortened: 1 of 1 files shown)


bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -109,19 +108,18 @@ static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
...
 		}
 	}
-	new_pid_key.pid = cur_pid;
+
 	bpf_map_update_elem(&pid_time, &new_pid_key, &cur_ts, BPF_NOEXIST);
 	return cpu_time;
 }
 
-static inline u64 calc_delta(u64 *prev_val, u64 *val)
+static inline u64 calc_delta(u64 *prev_val, u64 val)
```
sthaha, line 116: note that the current implementation already assumes `u64 *val` is not nil.

dave-tucker: while you're here wouldn't it be better to make this `calc_delta(u64 prev_val, u64 val)`?

dave-tucker (approved):
One small nit but otherwise LGTM

rootfs (approved):

PR #1455, opened 2024-05-20
feat(cgroup): Add podman support

dave-tucker:
This commit adds "libpod" to the list of supported container runtimes. This enables stats for podman containers to be correctly reported.


dave-tucker:
I've tested this on my local system and confirm that both rootful and rootless podman containers now show up when querying kepler metrics.

pkg/cgroup/resolve_container.go
```diff
@@ -231,49 +222,23 @@ func validContainerID(id string) string {
...
-		tmp := strings.Split(path, "/")
-		containerID := tmp[len(tmp)-1]
-		return validContainerID(containerID), nil
-	}
 
-	return utils.SystemProcessName, fmt.Errorf("failed to find pod's container id")
+	path = strings.TrimSuffix(path, "/container")
+	path = strings.TrimSuffix(path, ".scope")
+
+	// get the last 64 characters of the path
```
rootfs, line 236: can you post the reference to this 64-char path length?

dave-tucker: https://docs.docker.com/engine/reference/run/#container-identification

I've tried looking for a reference in the OCI spec but I couldn't find one, but from what I can see this appears to be the de-facto standard.

We could split the path string on all `/` and use the last path component, which wouldn't cause an issue in the case of something that's not exactly 64 chars long.

dave-tucker: I could yes, but consider this - it's a 10x performance improvement over the legacy code + all of our unit tests (including the newly added ones from podman) pass. Not only that, it's called on when the K8s watcher isn't enabled.

However, when the k8s watcher isn't enabled (i.e linux only, no k8s) this function is in the hot path - i.e it's called on in `createContainerStatsIfNotExist` which is called multiple times every time we collect metrics. Adding a fallback path is expensive in terms of performance - I'd rather that users submit issues so we can expand our tests cases and have a performant and correct solution. 

```
go test -benchmem -run=^$ -bench ^BenchmarkExtractPodContainerIDfromPathWithCgroup github.com/sustainable-computing-io/ke
pler/pkg/cgroup 
goos: linux
goarch: amd64
pkg: github.com/sustainable-computing-io/kepler/pkg/cgroup
cpu: AMD Ryzen 5 5600G with Radeon Graphics         
BenchmarkExtractPodContainerIDfromPathWithCgroup-12              1222920               930.0 ns/op             0 B/op          0 allocs/op
BenchmarkExtractPodContainerIDfromPathWithCgroupLegacy-12         123741              9918 ns/op             595 B/op          9 allocs/op
PASS
ok      github.com/sustainable-computing-io/kepler/pkg/cgroup   3.463s
```

PR #1458, opened 2024-05-21
fix(model): Use correct variable in `IsNodeComponentPowerModelEnabled()`

vimalk78:
Use `NodeComponentPowerModel` instead of `NodePlatformPowerModel`

This fixes https://github.com/sustainable-computing-io/kepler/issues/1459


vimalk78:
Cc: @rootfs @marceloamaral

rootfs (approved):

PR #1460, opened 2024-05-21
fix: remove logging while collecting GPU metrics

vprashar2929:
removes debugging logging from GPU metrics

### DIFF
--- a/pkg/metrics/utils/utils.go
+++ b/pkg/metrics/utils/utils.go
@@ -57,7 +57,6 @@ func CollectResUtilizationMetrics(ch chan<- prometheus.Metric, instance interfac
 			CollectResUtil(ch, instance, collectorName, collectors[collectorName])
 		}
 	}
-	klog.Info("Collecting GPU metrics")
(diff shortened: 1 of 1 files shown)


vprashar2929:
cc: @dave-tucker

dave-tucker (approved):
LGTM. The line can also be removed entirely - it was left over from debugging and I'd forgotten to remove it.

vprashar2929:
Ah in that case I am happy to remove it as well

vprashar2929:
cc: @rootfs

pkg/metrics/utils/utils.go
```diff
@@ -57,7 +57,6 @@ func CollectResUtilizationMetrics(ch chan<- prometheus.Metric, instance interfac
 			CollectResUtil(ch, instance, collectorName, collectors[collectorName])
 		}
 	}
-	klog.Info("Collecting GPU metrics")
```
sthaha, line 60: We may want to log this somewhere at V(5) or so  probably inside the `if`

sthaha (approved):

PR #1465, opened 2024-05-22
feat(ci)add prometheus and scrape kepler on equinix runner

rootfs:
The test results are [here](https://github.com/sustainable-computing-io/ci-playground/actions/runs/9196509105/job/25294851855)


rootfs:
At the moment, the self hosted runner could mysteriously vanish if there are more than 1 jobs. So all the commands are embeded in the same job. Will move the commands to a script after the runner issue is fixed.

SamYuan1990:
is this duplicate with test case in e2e_metric_test.go?
as we tested prometheus format there....

PR #1467, opened 2024-05-23
RFC: Async

dave-tucker:
This PR contains some fixes to the way async/waitgroups were being used previously and allows for Kepler to be gracefully shutdown - this doesn't appear to have any measureable performance impact.

The last commit in this series is an experiment on asynchronously collecting the Process Metrics (from eBPF) instead of doing so synchronously (i.e from the `Update` method of the manager).

You can see the result of this on the collector update here:
https://gist.github.com/dave-tucker/57f944a586e0d4463c5e19a586a585c7

I'll want to get a pprof of this and compare to Kepler on main also before we go further with this.

TL;DR The collector update takes on average 600ms less, and the variance (StdDev) reduces by 500ms.

I opened this PR to get some feedback as I'm quite keen to push forward in adapting the Kepler architecture to be more async. I'll open an issue and provide a diagram with what I'm thinking shortly...

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -17,11 +17,16 @@ limitations under the License.
 package main
 
 import (
+	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
-	"runtime/debug"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
 	"time"
 
 	"github.com/sustainable-computing-io/kepler/pkg/bpf"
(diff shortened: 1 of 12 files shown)


sthaha:
> This PR contains some fixes to the way async/waitgroups were being used previously and allows for Kepler to be gracefully shutdown - this doesn't appear to have any measureable performance impact.

A nice library I have come across which abstracts this use-case is - https://github.com/oklog/run?tab=readme-ov-file .

See usages
* In Thanos: https://github.com/thanos-io/thanos/blob/main/cmd/thanos/main.go#L96 
* In Prometheus: https://github.com/prometheus/prometheus/blob/8894d65cd6135635c4ac9cf100464e6aacd29593/cmd/prometheus/main.go#L954

PR #1470, opened 2024-05-24
accelerator: refactor accelerator device code

maryamtahhan:
Refactor pkg/sensors/accelerator to use a more generic device abstractions that different devices can plug into.

Todo 
- [ ] Retest Habana
- [x] Deprecate QAT for reasons listed [here](https://gist.github.com/maryamtahhan/81ded3940f32c9eb53fb84c7fa5d117a)
- [x] Retest DCGM 
- [x] Pull in QAT under device/sources


pkg/sensors/accelerator/acclerator.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+var (
+	accelerators map[string]Accelerator
+)
+
+// Accelerator represents an implementation of... equivalent Accelerator device.
+type Accelerator interface {
+	// StartupAccelerator ...
+	StartupAccelerator() error
```
dave-tucker, line 43: No need to suffix these methods with Accelerator.

pkg/sensors/accelerator/acclerator.go
```diff
@@ -0,0 +1,219 @@
...
+	// IsRunning ...
+	IsRunning() bool
+	// StopAccelerator ...
+	StopAccelerator()
+}
+
+type accelerator struct {
+	sync.Mutex
+	acc           device.AcceleratorInterface // Device Accelerator Interface
+	accType       string                      // NVML|DCGM|Habana|Dummy|QAT
```
dave-tucker, line 57: This should be an enum:

```go
type AcceleratorType int

const (
  NVML = iota
  GPU
  Habana
  Dummy
  QAT
 )
 ```

File: pkg/sensors/accelerator/acclerator.go [hunk 3]
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator/device"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+
+	// Add supported devices.
+	_ "github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator/device/sources"
+)
+
+var (
+	accelerators map[string]Accelerator
```
dave-tucker, line 37: Not a fan of globals

I would rather that you create a:

```go 
type AcceleratorRegistry struct {
   registry map[AcceleratorType]Accelerator
}
```

That would allow us to easily create various mock registry configurations without having to override package globals etc...

It can also be protected under mutex if we registrations might occur from different goroutines.

pkg/sensors/accelerator/acclerator.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	} else {
+		klog.Errorf("No %s Accelerator collectors found", strings.ToUpper(atype))
+		return errors.New("unsupported accelerator")
+	}
+	return nil
+}
+
+// GetAccelerators returns a map of supported accelerators.
+func GetAccelerators() (map[string]Accelerator, error) {
```
dave-tucker, line 112: This would move to `AcceleratorRegistry`

pkg/sensors/accelerator/acclerator.go
```diff
@@ -0,0 +1,219 @@
...
+	if len(accelerators) == 0 {
+		return nil, errors.New("no accelerators found")
+	}
+	return accelerators, nil
+}
+
+// GetActiveAcceleratorsByType returns a map of supported accelerators based on the specified type gpu|qat|...
+func GetActiveAcceleratorsByType(t string) (map[string]Accelerator, error) {
```
dave-tucker, line 120: As would this.

pkg/sensors/accelerator/acclerator.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+	if len(accs) == 0 {
+		return nil, errors.New("accelerators not found")
+	}
+	return accs, nil
+}
+
+// NewAccelerator creates a new Accelerator instance [NVML|DCGM|DUMMY|HABANA|QAT] for the local node.
+func NewAccelerator(accType string) Accelerator {
```
dave-tucker, line 135: And this would become something like `AcceleratorRegistry.Register()`

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,141 @@
...
+	"k8s.io/klog/v2"
+)
+
+var (
+	gpuDevices   = map[string]deviceStartupFunc{} // Static map of supported gpuDevices.
+	dummyDevices = map[string]deviceStartupFunc{} // Static map of supported dummyDevices.
+	qatDevices   = map[string]deviceStartupFunc{} // Static map of supported qatDevices.
+)
+
+type AcceleratorInterface interface {
```
dave-tucker, line 33: Should this be `type Device interface`?

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"golang.org/x/exp/maps"
+	"k8s.io/klog/v2"
+)
+
+var (
```
dave-tucker, line 27: Per previous comment re: Accelerators, I'd prefer encapsulating this into a `typ DeviceRegistry struct`

pkg/sensors/accelerator/device/sources/dummy.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (d *Dummy) GetDevicesByID() map[int]any {
+	return nil
+}
+
+func (d *Dummy) GetDevicesByName() map[string]any {
+	return nil
+}
+
+func (d *Dummy) GetDeviceInstances() map[int]map[int]interface{} {
```
dave-tucker, line 87: Mixed use of `any` and `interface{}`.
I'd assume this could also be written as `map[int]map[int]any`

pkg/sensors/accelerator/device/device.go
dave-tucker: If it **really** needs to be global because it's being used to register plugins via `init` functions - not 100% sure I can think of an easier way to deal with this right now....

Then let's add some scaffolding like so:

```
var globalRegistry *DeviceRegistry
var once sync.Once

// DeviceRegistry gets the default device registry instance
func DeviceRegistry() *DeviceRegistry {
    once.Do(func() {
        globalRegistry = NewDeviceRegistry()
    })
    return globalRegistry
}

// SetDeviceRegistry replaces the global registry instance
// NOTE: All plugins will need to be manually registered
// after this function is called.
func SetDeviceRegistry(registry *DeviceRegistry) {
    globalRegistry = registry
}
```

That would allow us to replace the registry for testing purposes.

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,141 @@
...
+	gpuDevices   = map[string]deviceStartupFunc{} // Static map of supported gpuDevices.
+	dummyDevices = map[string]deviceStartupFunc{} // Static map of supported dummyDevices.
+	qatDevices   = map[string]deviceStartupFunc{} // Static map of supported qatDevices.
+)
+
+type AcceleratorInterface interface {
+	// GetName returns the name of the device
+	GetName() string
```
dave-tucker, line 35: Here and everywhere else, don't use `Get` it should just be `Name() string`

dave-tucker (changes requested):
Overall looking good and much cleaner than it was before. Some initial comments.
Look forward to reviewing once you've finished testing this.

pkg/sensors/accelerator/acclerator.go
maryamtahhan: Fair enough. I will change these

pkg/sensors/accelerator/acclerator.go
maryamtahhan: I guess we can do that.

pkg/sensors/accelerator/device/sources/dummy.go
maryamtahhan: yeah I will tidy this up

pkg/sensors/accelerator/device/device.go
maryamtahhan: lol, Ok :)

pkg/sensors/accelerator/device/device.go
maryamtahhan: I couldn't easily get rid of the global aspect as there are functions invoked from different packages throughout kepler to get the stats. 
So either I created a registry at startup and then passed a reference to it to every place that needed it in Kepler (passing it to all the stats collection functions at init time so they have a reference) or add the scaffolding as you suggested throughout - I opted for the scaffolding to avoid too many changes in the existing code base

Makefile
```diff
@@ -57,8 +57,16 @@ endif
 
 GENERAL_TAGS := 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo libbpf '
 GPU_TAGS := ' gpu '
```
maryamtahhan, line 59: Note: These GPU build tags mean that we are filtering out GPU implementations from the base Kepler image. 

This places a requirement that the Kepler operator pick the right image for the accelerators supported on the node where Kepler is deployed.

The reasoning behind excluding the GPU specific implementations from the base Kepler image is that DCGM grows the image size by 2GB...

 We can continue this trend of a separate image per accelerator, or perhaps we can just combine all the accelerator builds into a single image (as this is going to be large anyway)... OR
 
 we just build everything into the default Kepler image (removing that requirement from the operator to pick and deploy specific images)

pkg/sensors/accelerator/device/sources/nvml.go
```diff
@@ -28,23 +28,22 @@ import (
 
 	"github.com/sustainable-computing-io/kepler/pkg/config"
 	_dev "github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator/device"
```
dave-tucker, line 30: There's no conflict here so no need to rename the import

dave-tucker (changes requested):
One small nit, but otherwise this LGTM

maryamtahhan:
I will rebase now :(

pkg/sensors/accelerator/device/sources/nvml.go
maryamtahhan: this was mainly to avoid the clash in the 1st param of:

```
func (n *GPUNvml) ProcessResourceUtilizationPerDevice(device any, since time.Duration) (map[uint32]any, error) 
```

I can change the function signature in all the underlying devices so the first param isn't device and then put the import back to the original name - NGL I went for the option with smaller impact but I will tidy it up

cmd/exporter/exporter.go
```diff
@@ -92,6 +90,22 @@ func healthProbe(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func finalizing() {
```
maryamtahhan, line 93: I must;ve missed this in the rebase... 
I do need the finalizing to do the accelerator clean up... let me see if I can work around this

pkg/config/config.go
```diff
@@ -68,8 +68,8 @@ var (
 
 	KeplerNamespace              = getConfig("KEPLER_NAMESPACE", defaultNamespace)
 	EnabledEBPFCgroupID          = getBoolConfig("ENABLE_EBPF_CGROUPID", true)
+	EnabledDummy                 = getBoolConfig("ENABLE_DUMMY", false)
```
maryamtahhan, line 71: remove this

maryamtahhan:
Will look into the integration test failures... not sure what happened :(

PR #1475, opened 2024-05-28
make: quote ldflags

rootfs:
fix [this](https://github.com/sustainable-computing-io/kepler/actions/runs/9271102170/job/25505700049) build error


sthaha (approved):
sorry I missed this when I changed the LDFLAGS

PR #1477, opened 2024-05-29
fix: ensure all entries from bpf map are copied

sthaha:
Previously only a `MapSize` (lesser than size of processes) of entries used to be copied from processes bpf map. The right use of a batch size less than the size of map is as follows:

Ref: https://lwn.net/Articles/797808/
```
    p_key = NULL;
    p_next_key = &key;
    while (true) {
       err = bpf_map_lookup_batch(fd, p_key, &p_next_key, keys, values,
                                  &batch_size, elem_flags, flags);
       if (err) ...
       if (p_next_key) break; // done
       if (!p_key) p_key = p_next_key;
    }
```

This PR fixes the issue by creating a entries byte array with the same size as the `processes` map.

Additionally, this commit makes use of `bytes.Reader` instead of bytes.Buffer to avoid unnecessary copy of ephemeral byte array.

### DIFF
--- a/pkg/bpf/exporter.go
+++ b/pkg/bpf/exporter.go
@@ -45,7 +45,6 @@ const (
 	bpfPerfArraySuffix = "_event_reader"
 	TableProcessName   = "processes"
 	TableCPUFreqName   = "cpu_freq_array"
-	MapSize            = 10240
 	CPUNumSize         = 128
 )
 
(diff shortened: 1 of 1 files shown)


pkg/bpf/exporter.go
```diff
@@ -327,7 +326,7 @@ func (e *exporter) CollectCPUFreq() (cpuFreqData map[int32]uint64, err error) {
 	if err != nil {
 		return
 	}
-	//cpuFreqkeySize := int(unsafe.Sizeof(uint32Key))
+	// cpuFreqkeySize := int(unsafe.Sizeof(uint32Key))
```
sthaha, line 329: `go fmt` artefact

rootfs (approved):

PR #1479, opened 2024-05-30
bpf: fix the process parameter order to match the c and go code

marceloamaral:
fix the process parameter order to match the c and go code


marceloamaral:
related to PR #1468

PR #1438, opened 2024-05-15
chore(pkg/bpf): Replace libbpfgo with cilium/ebpf

dave-tucker:
cilium/ebpf is a pure Go eBPF package and is used in a number of popular
cloud-native projects. The benefits to Kepler are:

1. Bytecode is built using bpf2go and the C and Go structs are kept in
   sync automatically
2. There is no requirement for Cgo anymore and therefore no requirement
   to have libbpf or libelf installed to compile and/or to be
   dynamically linked at runtime
3. Simplified packaging as the correct bytecode is contained within the
   kepler binary

Overall I'm happy with this change, but there is only one thing that
bugs me.

We have to check in the bytecode object files (e.g kepler.bpfel.o) or
the Go tooling (go lint/go vet) complains about the missing files.
I couldn't reliably get `go generate ./...` to work to compile these
files in CI. This is something which should be relatively easy to fix
in the Makefile/CI environment before we cut a release.

Depends-On: #1435 
Fixes: #1424

PR #1481, opened 2024-05-30
Fix bug in the ebpf metric collection

marceloamaral:
As stated in PR #1468, the ebpf metric collection is not correct.

This PR is fixing multiple problems:

- Fixed incorrect padding in some structures to ensure it is 8 bytes (unsigned long long).
- Updated the schedule switch to collect only on-CPU time and ignore idle time. As detailed [here](https://litux.nl/mirror/kerneldevelopment/0672327201/ch03lev1sec1.html), during a schedule switch, the previous task might not have been in the running state. Inspired by the libbpf-tools [cpudist](https://github.com/iovisor/bcc/blob/c93a19aaf3f7eae9d6c9070309cc785c18575767/libbpf-tools/cpudist.bpf.c#L100), I modified the code to save data only if the task was previously running, thereby excluding idle or blocked times. Collecting data from a non-running task could lead to gathering hardware counters from another task that was actually running on the CPU.
- Updated metric collection to gather hardware counter metrics only when CPU time can be reliably collected. This ensures the metrics are correlated, addressing issues where the start timestamp of the task might be missing or there could be clock issues.
- Minimized Kepler overhead by aggregating thread data into process IDs within the eBPF code (kernel space).  Note that the Kernel tgid is the user-space PID, and the Kernel pid is the user-scape TID.
- Changed the eBPF code to use microseconds instead of milliseconds to improve precision and accurately identify when a process was not active.

With the fixes, both the `scaphandre` and `kepler` baremetal power consumption measurements of a VM process is similar.
As we can see in the figure below after running `stress-ng` into the VM.

image

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -3,18 +3,25 @@
 
 #include "kepler.bpf.h"
 struct {
-	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
 	__type(key, u32);
 	__type(value, process_metrics_t);
 	__uint(max_entries, MAP_SIZE);
 } processes SEC(".maps");
 
 struct {
-	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
 	__type(key, u32);
 	__type(value, u64);
 	__uint(max_entries, MAP_SIZE);
-} pid_time SEC(".maps");
+} pid_time_map SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
+	__type(key, u32);
+	__type(value, u32);
+	__uint(max_entries, MAP_SIZE);
+} pid_tgid_map SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
@@ -59,29 +66,17 @@ struct {
 } cache_miss SEC(".maps");
 
 struct {
-	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
-	__type(key, int);
-	__type(value, u32);
-	__uint(max_entries, NUM_CPUS);
-} task_clock_ms_event_reader SEC(".maps");
-
-struct {
-	__uint(type, BPF_MAP_TYPE_ARRAY);
-	__type(key, u32);
-	__type(value, u64);
-	__uint(max_entries, NUM_CPUS);
-} task_clock SEC(".maps");
-
-struct {
 	__uint(type, BPF_MAP_TYPE_ARRAY);
 	__type(key, u32);
 	__type(value, u32);
 	__uint(max_entries, NUM_CPUS);
 } cpu_freq_array SEC(".maps");
 
+// The sampling rate should be disabled by default because its impact on the
+// measurements is unknown.
 SEC(".rodata.config")
 __attribute__((
-	btf_decl_tag("Sample Rate"))) static volatile const int SAMPLE_RATE = 5;
+	btf_decl_tag("Sample Rate"))) static volatile const int SAMPLE_RATE = 0;
 
 SEC(".rodata.config")
 __attribute__((btf_decl_tag("CPU Reference Frequency"))) static volatile const int
@@ -93,34 +88,29 @@ __attribute__((btf_decl_tag("Hertz Multiplier"))) static volatile const int HZ =
 
 int counter_sched_switch = 0;
 
-static inline u64 get_on_cpu_time(u32 cur_pid, u32 prev_pid, u64 cur_ts)
-{
-	u64 cpu_time = 0;
-	pid_time_t prev_pid_key = { .pid = prev_pid };
-	pid_time_t new_pid_key = { .pid = cur_pid };
-
-	u64 *prev_ts = bpf_map_lookup_elem(&pid_time, &prev_pid_key);
-	if (prev_ts) {
-		// Probably a clock issue where the recorded on-CPU event had a
-		// timestamp later than the recorded off-CPU event, or vice versa.
-		if (cur_ts > *prev_ts) {
-			cpu_time = (cur_ts - *prev_ts) / 1000000; // convert to ms
-			bpf_map_delete_elem(&pid_time, &prev_pid_key);
-		}
-	}
-
-	bpf_map_update_elem(&pid_time, &new_pid_key, &cur_ts, BPF_NOEXIST);
-	return cpu_time;
-}
-
 static inline u64 calc_delta(u64 *prev_val, u64 val)
 {
 	u64 delta = 0;
-	if (prev_val && val > *prev_val) {
+	// Probably a clock issue where the recorded on-CPU event had a
+	// timestamp later than the recorded off-CPU event, or vice versa.
+	if (prev_val && val > *prev_val)
 		delta = val - *prev_val;
-	}
 
 	return delta;
+}
+
+static inline u64 get_on_cpu_elapsed_time_us(u32 prev_pid, u64 curr_ts)
+{
+	u64 cpu_time = 0;
+	u64 *prev_ts;
+
+	prev_ts = bpf_map_lookup_elem(&pid_time_map, &prev_pid);
+	if (prev_ts) {
+		cpu_time = calc_delta(prev_ts, curr_ts) / 1000;
+		bpf_map_delete_elem(&pid_time_map, &prev_pid);
+	}
+
+	return cpu_time;
 }
 
 static inline u64 get_on_cpu_cycles(u32 *cpu_id)
@@ -169,9 +159,8 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
 
 	error = bpf_perf_event_read_value(
 		&cache_miss_event_reader, *cpu_id, &c, sizeof(c));
-	if (error) {
+	if (error)
 		return 0;
-	}
 	val = c.counter;
 	prev_val = bpf_map_lookup_elem(&cache_miss, cpu_id);
 	delta = calc_delta(prev_val, val);
@@ -180,11 +169,51 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
 	return delta;
 }
 
+static inline void register_new_process_if_not_exist()
+{
+	u64 cgroup_id, pid_tgid;
+	u32 curr_pid, curr_tgid;
+	struct process_metrics_t *curr_tgid_metrics;
+
+	pid_tgid = bpf_get_current_pid_tgid();
+	curr_pid = (u32)pid_tgid;
+	curr_tgid = pid_tgid >> 32;
+
+	// create new process metrics
+	curr_tgid_metrics = bpf_map_lookup_elem(&processes, &curr_tgid);
+	if (!curr_tgid_metrics) {
+		cgroup_id = bpf_get_current_cgroup_id();
+		// the Kernel tgid is the user-space PID, and the Kernel pid is the
+		// user-space TID
+		process_metrics_t new_process = {
+			.pid = curr_tgid,
+			.cgroup_id = cgroup_id,
+		};
+		bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
+		bpf_map_update_elem(
+			&processes, &curr_tgid, &new_process, BPF_NOEXIST);
+
+		// add new thread id (curr_pid) to the precess id (tgid) list
+		bpf_map_update_elem(
+			&pid_tgid_map, &curr_pid, &curr_tgid, BPF_NOEXIST);
+	}
+}
+
+static inline void collect_metrics_and_reset_counters(
+	struct process_metrics_t *buf, u32 prev_pid, u64 curr_ts, u32 cpu_id)
+{
+	buf->cpu_cycles = get_on_cpu_cycles(&cpu_id);
+	buf->cpu_instr = get_on_cpu_instr(&cpu_id);
+	buf->cache_miss = get_on_cpu_cache_miss(&cpu_id);
+	// Get current time to calculate the previous task on-CPU time
+	buf->process_run_time = get_on_cpu_elapsed_time_us(prev_pid, curr_ts);
+}
+
 // This struct is defined according to the following format file:
 // /sys/kernel/tracing/events/sched/sched_switch/format
 struct sched_switch_info {
 	/* The first 8 bytes is not allowed to read */
-	unsigned long pad;
+	u64 pad;
 
 	char prev_comm[16];
 	pid_t prev_pid;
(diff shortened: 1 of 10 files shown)


bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -81,7 +89,7 @@ struct {
 
 SEC(".rodata.config")
 __attribute__((
-	btf_decl_tag("Sample Rate"))) static volatile const int SAMPLE_RATE = 5;
```
sunya-ch, line 84: Could you put some comment here where this number can be overwrote?

marceloamaral: The sampling rate should be disabled by default since we don't know the impact in the measurement.

sunya-ch: I agree to have this disabled and set to zero. Thank you for adding the comment above.

pkg/metrics/utils/utils.go
```diff
@@ -109,6 +109,14 @@ func collectEnergy(ch chan<- prometheus.Metric, instance interface{}, metricName
 	}
 }
 
+func convertUnit(metricName string, val uint64) float64 {
```
sunya-ch, line 112: It might be clearer to have function convertU64ToFloat. 
Then move the logic if microseconds to milliseconds outside after the conversion and define a clear constant like MilliMicroConversionFactor float64 = 1000. 

Or, we can see single function but changing the function name to clarify the task such as convertResUtilToPromValue

sunya-ch (approved):
This PR should fix a vital issue on BPF time in the current Kepler. I think we should have this merge asap.

rootfs:
@marceloamaral can you fix the lint? it is ready to go!

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -3,311 +3,345 @@
...
 }
 
 // This struct is defined according to the following format file:
 //  /sys/kernel/tracing/events/irq/softirq_entry/format
 struct trace_event_raw_softirq {
-	/* The first 8 bytes is not allowed to read */
-	unsigned long pad;
-	unsigned int vec;
+  /* The first 8 bytes is not allowed to read */
+  unsigned long long pad;
```
sthaha, line 298: could we use u64 to be sure we are skipping 8 bytes?

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -3,311 +3,345 @@
...
+  cur_pid = bpf_get_current_pid_tgid();
+  vec = ctx->vec;
+  process_metrics = bpf_map_lookup_elem(&processes, &cur_pid);
+  if (process_metrics != 0) {
+    if (vec < 10) {
+      u16 count = process_metrics->vec_nr[vec];
+      count++;
+      process_metrics->vec_nr[vec] = count;
+    }
+  }
```
sthaha, line 317:
```suggestion
  if (process_metrics != 0 && vec < 10) {
      u16 count = process_metrics->vec_nr[vec];
      count++;
      process_metrics->vec_nr[vec] = count;
  }
```

bpfassets/libbpf/src/kepler.bpf.c
dave-tucker: What @vimalk78 said. The number here is irrelevant as it's set from userland. There was previously an issue where this got optimized out when it was `0` but since it changed to be `static volatile const` that shouldn't happen in future.

The comment should be updated to reflect the current state.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -180,11 +182,52 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
 	return delta;
 }
 
+static inline void register_new_process_if_not_exist()
+{
+	u64 cgroup_id;
+	u64 pid_tgid = bpf_get_current_pid_tgid();
```
dave-tucker, line 188: Style nit: Split declaration and assignment:

`u64 cgroup_id, pid_tgid, curr_pid, curr_tgid;`

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -180,11 +182,52 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
...
+	u32 curr_pid = (u32)pid_tgid;
+	u32 curr_tgid = pid_tgid >> 32;
+
+	// create new process metrics
+	struct process_metrics_t *curr_tgid_metrics =
```
dave-tucker, line 193: Again, declare this before assignment.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -180,11 +182,52 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
...
+			&processes, &curr_tgid, &new_process, BPF_NOEXIST);
+
+		// add new thread id (cur_pid) to the precess id (tgid) list
+		bpf_map_update_elem(
+			&pid_tgid_map, &curr_pid, &curr_tgid, BPF_NOEXIST);
+	}
+}
+
+static inline void collect_metrics_and_reset_counters(
+	struct process_metrics_t *buf, u32 prev_pid, u64 cur_ts, u32 cpu_id)
```
dave-tucker, line 214: style nit; split the declarations by kind:
```
process_metrics_t *buf;
u32 cpu_id, prev_pid;
u64 cur_ts
```

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -198,13 +241,27 @@ struct sched_switch_info {
 SEC("tp/sched/sched_switch")
 int kepler_sched_switch_trace(struct sched_switch_info *ctx)
 {
-	u32 prev_pid, tgid, cpu_id;
-	u64 pid_tgid, cgroup_id, cur_ts;
-	pid_t cur_pid;
+	u32 prev_pid, next_pid, cpu_id;
+	u64 *prev_tgid;
+	long prev_state;
+	u64 cur_ts = bpf_ktime_get_ns();
```
dave-tucker, line 247: nit: split declaration and assignment here too.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -198,13 +241,27 @@ struct sched_switch_info {
...
 
-	struct process_metrics_t *cur_pid_metrics, *prev_pid_metrics;
+	struct process_metrics_t *prev_tgid_metrics, *curr_tgid_metrics;
```
dave-tucker, line 249: It seems unnecessary to rename these and reorder them.
Can we pick a naming convention and stick with it - either `cur` or `curr`.
Then sorting alphabetically is always preferred.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -198,13 +241,27 @@ struct sched_switch_info {
...
 
+	prev_state = ctx->prev_state;
+	prev_pid = (u32)ctx->prev_pid;
+	next_pid = (u32)ctx->next_pid;
+	cpu_id = bpf_get_smp_processor_id();
+
+	// Collect metrics
+	// Regardless of skipping the collection, we need to update the hardware
+	// counter events to keep the metrics map current.
+	collect_metrics_and_reset_counters(&buf, prev_pid, cur_ts, cpu_id);
```
dave-tucker, line 260: I'm concerned about the performance impact here. It potentially negates the benefit of sampling at all (which is still relatively expensive anyway).

Have you profiled the older version + this version so we can see how much CPU time is spent where? Map reads/writes are the most expensive operations and if we can't avoid those then I would expect sampling to be pointless.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -14,7 +14,14 @@ struct {
...
 	__uint(max_entries, MAP_SIZE);
-} pid_time SEC(".maps");
+} pid_time_map SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u32);
+	__type(value, u32);
+	__uint(max_entries, MAP_SIZE);
+} pid_tgid_map SEC(".maps");
```
dave-tucker, line 24: This naming is problematic to me. See where we access this map in the code using `prev_pid` as the key. We should take this opportunity to use sensible names here.

dave-tucker: @vimalk78 and I had a lengthy discussion about this in a PR a week or so ago and it's very easy to get confused

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -180,11 +182,52 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
...
+	u32 curr_pid = (u32)pid_tgid;
+	u32 curr_tgid = pid_tgid >> 32;
+
+	// create new process metrics
+	struct process_metrics_t *curr_tgid_metrics =
+		bpf_map_lookup_elem(&processes, &curr_tgid);
+	if (!curr_tgid_metrics) {
+		cgroup_id = bpf_get_current_cgroup_id();
+		// the Kernel tgid is the user-space PID, and the Kernel pid is the
+		// user-scape TID
```
sthaha, line 198: nit: typo 

```suggestion
		// user-space TID
```

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -213,41 +270,38 @@ int kepler_sched_switch_trace(struct sched_switch_info *ctx)
...
+		prev_tgid = bpf_map_lookup_elem(&pid_tgid_map, &prev_pid);
+		if (prev_tgid) {
+			// The process_run_time is 0 if we do not have the previous timestamp of
+			// the task or due to a clock issue. In either case, we skip collecting
+			// all metrics to avoid discrepancies between the hardware counter and CPU
+			// time.
+			if (buf.process_run_time > 0) {
+				prev_tgid_metrics = bpf_map_lookup_elem(
+					&processes, prev_tgid);
+				if (prev_tgid_metrics) {
```
vimalk78, line 284: if we are aggregating power consumption for the whole process in bpf code only, do we need [`pid` and `tgid` separately](https://github.com/sustainable-computing-io/kepler/blob/54c6e85cb6af9e4e1d2b46053545f5cc9257fda4/bpfassets/libbpf/src/kepler.bpf.h#L94-L95) in `process_metrics_t`  ?

vimalk78: was this aggregation also happening in user space kepler ?

bpfassets/libbpf/src/kepler.bpf.c
marceloamaral: That is right, but just in case it is better to set the default value here to 0.

pkg/metrics/utils/utils.go
marceloamaral: I agree, but let's fix this in another PR

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -180,11 +182,52 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
...
+
+	// create new process metrics
+	struct process_metrics_t *curr_tgid_metrics =
+		bpf_map_lookup_elem(&processes, &curr_tgid);
+	if (!curr_tgid_metrics) {
+		cgroup_id = bpf_get_current_cgroup_id();
+		// the Kernel tgid is the user-space PID, and the Kernel pid is the
+		// user-scape TID
+		process_metrics_t new_process = {
+			.pid = curr_tgid,
```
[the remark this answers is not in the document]
marceloamaral: Thread Group ID (TGID) in Kernel is the PID in user space.
On the other hand, the PID in Kernel space is the the thread ID in user space.
See: 
- https://github.com/iovisor/bcc/pull/674
- https://github.com/iovisor/bcc/blob/master/tools/offcputime.py#L69

So that, we are interested in in the PID in user space, which is the TGID in the Kernel.

bpfassets/libbpf/src/kepler.bpf.c
marceloamaral: Yes, it should not be updated!

marceloamaral: Right, let's use `cur` only.

bpfassets/libbpf/src/kepler.bpf.c
marceloamaral: While skipping the counter update may seem beneficial for performance, it would lead to inaccurate data. By neglecting to update the counter, we would misattribute skipped threads metrics to subsequent thread. This would result in a significant overestimation of resource usage for individual threads.

Additionally, our current default deployment does not take advantage of sampling, and as you mentioned, more in-depth analysis is needed before employing it.

bpfassets/libbpf/src/kepler.bpf.c
marceloamaral: What about `pid_to_tgid_map`?
We just want to get the `tgid` of a give `pid`, which in case will be the `prev_pid`

bpfassets/libbpf/src/kepler.bpf.c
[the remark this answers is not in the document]
marceloamaral: In fact, for simplicity's sake, I don't think we need to export the Kernel space PID (user TID). 
We can just have the user PID (Kernel TIGID) in the golang code.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -180,11 +182,52 @@ static inline u64 get_on_cpu_cache_miss(u32 *cpu_id)
...
+			.cgroup_id = cgroup_id,
+		};
+		bpf_get_current_comm(&new_process.comm, sizeof(new_process.comm));
+		bpf_map_update_elem(
+			&processes, &curr_tgid, &new_process, BPF_NOEXIST);
+
+		// add new thread id (cur_pid) to the precess id (tgid) list
+		bpf_map_update_elem(
+			&pid_tgid_map, &curr_pid, &curr_tgid, BPF_NOEXIST);
```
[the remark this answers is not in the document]
marceloamaral: When an eBPF hash map fills up in the kernel, eBPF employs an eviction strategy to handle the insertion of a new key. LRU (Least Recently Used) Eviction is the most common strategy. So, it will remove a old key and replace with a new one.

This should not be a big problem, but we could introduce the logic later to reset this map times to times to avoid trashing problem. However, we should not always reset the map to have a complete list of pids to present loose datapoints. Clean up should be done in a longer interval than the process map reset.

bpfassets/libbpf/src/kepler.bpf.c
marceloamaral: Good point, I have actually already mentioned that in another thread.
So, we don't need the `pid` and `tgid` in the `process_metrics_t`.
We can just have the `pid`, which will be the `tgid` from the kernel space.

bpfassets/libbpf/src/kepler.bpf.c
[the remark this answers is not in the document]
dave-tucker: You need to use `BPF_MAP_TYPE_LRU_HASH` to have LRU-based eviction.

dave-tucker:
@marceloamaral do you need any help implementing the review feedback and getting the unit tests to pass?
keen to get this merged to unblock #1438

sunya-ch:
I've just found that there is some issue left even we patch this PR.
I run the validation docker compose. Now, I can see BPF CPU time reported at BM is as expected (0 to 700 ms).
However, at the VM, there is one process ([kworker/u8:3-btrfs-endio]) that the CPU utilization is zero but the CPU time reported by Kepler covers 1 vcore (1000ms) at idle state. When stress is loaded, this process drop to 400ms while stressng is up to 100ms (the total is still 500ms). 

I think it might be something related to the process state in the BPF code. Is it possible to fix it here or better create a new issue? 

Screenshot 2024-06-06 at 10 46 25

sthaha:
@marceloamaral, like @sunya-ch reported, I tried the same on my machine with a single cpu allocated to the VM 
The view from metal seems to match schaphandre (but out of phase as seen earlier) 

### process_package{metal} vs schaphandre
image

### 15 mins
image

But VM dynamic paints a totally different picture and it could be because of models and estimator.  

image

However, the CPU BPF seems to match well 
image

marceloamaral:
@sunya-ch and @sthaha 
I think eBPF might not be the cause of the CPU time difference between bare metal and virtual machine environments. The hypervisor could be the problem. We have actually further discussed about that before and the hypervisor can introduce the following limitations:
- Timer Resolution: Hypervisors might rely on different clock sources or employ less accurate timing mechanisms in the guest OS kernel, leading to inconsistencies in CPU time measurements.
- Timekeeping Divergence: VMs might leverage paravirtualized clock sources, which will have a different behavior from the physical hardware clocks in the bare metal.
- Hypervisor Overhead: The hypervisor introduces an overhead layer that can influence the accuracy of CPU time measurements.

pkg/bpf/bpf_suite_test.go
```diff
@@ -19,14 +19,12 @@ func TestBpf(t *testing.T) {
 func checkDataCollected(processesData []ProcessBPFMetrics) {
 	// len > 0
 	Expect(len(processesData)).To(BeNumerically(">", 0))
-	Expect(processesData[0].PID).To(BeNumerically(">", 0))
+	Expect(processesData[0].PID).To(BeNumerically(">=", uint64(0)))
```
sthaha, line 22: Should we track pid 0 ?

marceloamaral: We were always tracking PID 0, not sure why the unit test was only > 0.

marceloamaral:
The ebpf code is not collecting cpu time in the CI....
I cannot reproduce this problem locally, so I am trying to debug it in CI.... I will fix the PR's commits later.

marceloamaral:
@dave-tucker @sthaha @sunya-ch the PR is ready to be reviewed again!

dave-tucker (approved):

PR #1484, opened 2024-05-31
chore(validator)support remote prometheus

rootfs:
The validator should retrieve actual metrics from metal prometheus and expected metrics from remote VM

### DIFF
--- a/e2e/tools/validator/src/validator/cli/__init__.py
+++ b/e2e/tools/validator/src/validator/cli/__init__.py
@@ -80,20 +80,15 @@ def stress(cfg: Validator, script_path: str):
     metrics_validator = MetricsValidator(cfg.prometheus, cfg.remote_prometheus)
     test_case_result = test_cases.load_test_cases()
     click.secho("Validation results during stress test:")
-    click.secho("Validation results during stress test:")
     for test_case in test_case_result.test_cases:
         expected_query = test_case.expected_query
         actual_query = test_case.actual_query
-        print(f"expected_query: {expected_query}")
-        print(f"actual_query: {actual_query}")
         print(f"start_time: {result.start_time}, end_time: {result.end_time}")
         metrics_res = metrics_validator.compare_metrics(result.start_time, 
                                                         result.end_time, 
                                                         expected_query, 
                                                         actual_query)
 
-        click.secho(f"Expected Query Name: {expected_query}", fg='bright_yellow')
-        click.secho(f"Actual Query Name: {actual_query}", fg='bright_yellow')      
         click.secho(f"Expected Query Name: {expected_query}", fg='bright_yellow')
         click.secho(f"Actual Query Name: {actual_query}", fg='bright_yellow')      
         click.secho(f"Absolute Errors during stress test: {metrics_res.ae}", fg='green')
--- a/e2e/tools/validator/src/validator/cli/__init__.py
+++ b/e2e/tools/validator/src/validator/cli/__init__.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: APACHE-2.0
 
 import click
+import subprocess
 from validator.__about__ import __version__
 from validator.stresser import ( Remote )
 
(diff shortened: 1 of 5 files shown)


KaiyiLiu1234 (commented):
Currently, I have made a change which I will push today: There will no longer be "actual" or "expected" metrics in the validator. Rather, we will just have the validator take a single promql query like: 
```
abs(1 - 
  (
    sum (rate(kepler_process_bpf_cpu_time_ms_total{pid="192759", job="metal"}[1m]))
    /  
    sum (rate(kepler_process_bpf_cpu_time_ms_total{job="vm"}[1m]))
  )
)
```
. sum (rate(kepler_process_bpf_cpu_time_ms_total{pid="192759", job="metal"}[1m])) would be from metal and  sum (rate(kepler_process_bpf_cpu_time_ms_total{job="vm"}[1m])) would be from vm. This should produce a value which shows how much the metal deviates from the vm. Also, vm jobs and metal jobs already both push to the prometheus on metal. Since this change is to try and introduce a vm/remote prometheus to track the expected, there will be conflict as the plan is to just have the validator only handle running a single query. Would this be an acceptable change or do you propose something different? @rootfs

e2e/tools/validator/src/validator/config/__init__.py
```diff
@@ -32,6 +32,7 @@ class Validator(NamedTuple):
     remote: Remote
     metal: Metal
     prometheus: Prometheus
+    remote_prometheus: Prometheus
```
sthaha, line 35: Could you please shed some light into why we need this? 
I would expect the same prometheus on the Baremetal to scrape kepler running on VM.

e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -77,23 +78,28 @@ def stress(cfg: Validator, script_path: str):
...
-    metrics_validator = MetricsValidator(cfg.prometheus)
+    metrics_validator = MetricsValidator(cfg.prometheus, cfg.remote_prometheus)
     test_case_result = test_cases.load_test_cases()
     click.secho("Validation results during stress test:")
-    click.secho("Validation results during stress test:")
+    # run git describe command and get the output as the report name
+    tag = ""
+    git_describe = subprocess.run(["git", "describe", "--tag"], stdout=subprocess.PIPE)
+    if git_describe.stdout:
+        tag = git_describe.stdout.decode().strip()
```
sthaha, line 88: Could you please submit another PR for this change since supporting remote prometheus shouldn't require `git describe`

e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -77,23 +78,28 @@ def stress(cfg: Validator, script_path: str):
...
+    # save all the print result into a markdown file as a table
+    report = open(f"/tmp/report-{tag}.md", "w")
+    # create table header
+    report.write("| Expected Query | Actual Query | MAE | MAPE | MSE | RMSE |\n")
+    report.write("|----------------|--------------|-----|------|-----|------|\n")
```
sthaha, line 93: could you please make use of `with` ?

sthaha (changes requested):
I don't think we need to deploy 2 prometheus for validation.

sthaha:
> The validator should retrieve actual metrics from metal prometheus and expected metrics from remote VM

IMHO, it shouldn't. the setup should be more like this

image

e2e/tools/validator/src/validator/config/__init__.py
rootfs: @sthaha this is to support the configuration of having two prometheus instances running in metal and vm. If you configure the same prometheus to scrape metal and vm, you can set the same url for the remote and local prometheus

rootfs:
> Currently, I have made a change which I will push today: There will no longer be "actual" or "expected" metrics in the validator. Rather, we will just have the validator take a single promql query like:
> 
> ```
> abs(1 - 
>   (
>     sum (rate(kepler_process_bpf_cpu_time_ms_total{pid="192759", job="metal"}[1m]))
>     /  
>     sum (rate(kepler_process_bpf_cpu_time_ms_total{job="vm"}[1m]))
>   )
> )
> ```
> 
> . sum (rate(kepler_process_bpf_cpu_time_ms_total{pid="192759", job="metal"}[1m])) would be from metal and sum (rate(kepler_process_bpf_cpu_time_ms_total{job="vm"}[1m])) would be from vm. This should produce a value which shows how much the metal deviates from the vm. Also, vm jobs and metal jobs already both push to the prometheus on metal. Since this change is to try and introduce a vm/remote prometheus to track the expected, there will be conflict as the plan is to just have the validator only handle running a single query. Would this be an acceptable change or do you propose something different? @rootfs

@KaiyiLiu1234 can you post your PR and let's see what is the plan there?

rootfs:
> If we are to support remote_prometheus, it must be optional, so that missing the remote-prometheus would fallback to using the metal one.

@sthaha that's good, let me make this optional

e2e/tools/validator/src/validator/cli/__init__.py
rootfs: that'll shift quite a few more lines below :D

sthaha:
@rootfs , I just realised that having a single prometheus that holds both keplers's metrics would simplify comparisons / validations quite a bit since we can make use of built in functions. E.g. 

```
abs(
    ( sum( rate(kepler_bpf_cpu_time{job="vm"}) ) -    sum( rate(kepler_bpf_cpu_time{job="metal", pid="<vm-pid>"})   )
    /
    rate(kepler_bpf_cpu_time{job="metal", pid="<vm-pid>"})
) > 0.1
```
If we were to do the same using 2 prometheus, there is quite some work that we will have to reinvent.

PR #1489, opened 2024-06-03
Replace expected and actual query with single query

KaiyiLiu1234:
Instead of using two separate queries which the validator will use to calculate the error metrics, replace with a single promql query which will calculate and return the error metric to the validator. Please give this a try @rootfs. Note if an error occurs, it's most likely because the "/" symbol in the promql query results in an empty query.

### DIFF
--- a/e2e/tools/validator/query.json
+++ b/e2e/tools/validator/query.json
@@ -1,18 +1,4 @@
 [
-    {
-        "expected_query": "rate(kepler_{level}_package_joules_total{{{query}, mode='dynamic'}}[{interval}])",
-        "actual_query": "rate(kepler_node_platform_joules_total[{interval}])"
-    },
-    {
-        "expected_query": "rate(kepler_{level}_platform_joules_total{{{query}, mode='dynamic'}}[{interval}])",
-        "actual_query": "rate(kepler_node_platform_joules_total[{interval}])"
-    },
-    {
-        "expected_query": "rate(kepler_{level}_bpf_cpu_time_ms_total{{{query}}}[{interval}])",
-        "actual_query": "sum by(__name__, job) (rate(kepler_process_bpf_cpu_time_ms_total[{interval}]))"
-    },
-    {
-        "expected_query": "rate(kepler_{level}_bpf_page_cache_hit_total{{{query}}}[{interval}])",
-        "actual_query": "sum by(__name__, job) (rate(kepler_process_bpf_page_cache_hit_total[{interval}]))"
-    }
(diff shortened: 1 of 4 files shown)


KaiyiLiu1234:
I am not entirely sure if it's a great idea to just have a single prom ql and removing expected and actual query. It does simplify the code, but after some discussion with @sthaha regarding selecting the type of error metric we will produce, it sort of just looped back around to using MAPE again. What is your opinion @sthaha @rootfs

e2e/tools/validator/query.json
```diff
@@ -1,18 +1,4 @@
...
-    },
-    {
-        "expected_query": "rate(kepler_{level}_bpf_cpu_time_ms_total{{{query}}}[{interval}])",
-        "actual_query": "sum by(__name__, job) (rate(kepler_process_bpf_cpu_time_ms_total[{interval}]))"
-    },
-    {
-        "expected_query": "rate(kepler_{level}_bpf_page_cache_hit_total{{{query}}}[{interval}])",
-        "actual_query": "sum by(__name__, job) (rate(kepler_process_bpf_page_cache_hit_total[{interval}]))"
-    }
+    "abs(1 - (rate(kepler_{level}_package_joules_total{{{query}, mode='dynamic'}}[{interval}]) / rate(kepler_node_platform_joules_total{{job='vm'}}[{interval}])))",
```
rootfs, line 2: can you make `job='vm'` a param, so i can replace the vm name?

sthaha: How about we require a label to indicate `vm` other than `job` ? 
e.g. `target="vm"` ? and `target="metal"` ?

sthaha, line 2:
```suggestion
    """
     abs(
        1 - (
           rate(kepler_{level}_package_joules_total{{{query}, mode='dynamic'}}[{interval}]) 
            / 
             rate(kepler_node_platform_joules_total{{job='vm'}}[{interval}])
         )
      )
",
```

KaiyiLiu1234: Please look at my recent change to query.json. This would not work because the labels and id are not the same. So I had to add a new label. It makes the input a lot more verbose however. We can rewrite the verbose part of introducing a new label into the code instead of having users do it themselves in query.json. Also I get an error when I try to use """.

KaiyiLiu1234: Converted job='vm' into a param. For now I put job='vm' as the placeholder

sthaha:
@KaiyiLiu1234 is this still WIP?

rootfs (approved):

PR #1503, opened 2024-06-07
Fix use of goroutines and implement graceful shutdown

dave-tucker:
Some small fixes while investigating #1467:

- In `cmd/validator` adds to the `wg` before it's passed into another function. `updateNodePower` is then run as `go updateNodePower`
- Consistently initilize wg's as `wg := &sync.WaitGroup{}`
- Call `go c.updateNodeResourceUtilizationMetrics(wg)` and `go c.updateProcessResourceUtilizationMetrics(wg)` in their own goroutines. Otherwise the use of a wg is pointless since they would be run sequentially
- Allows for Kepler to gracefully shutdown when a SIGINT or SIGTERM is received

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -17,11 +17,16 @@ limitations under the License.
 package main
 
 import (
+	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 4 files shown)


sthaha (approved):
/lgtm

PR #1505, opened 2024-06-10
chore: Group dependabot updates

dave-tucker:
This groups updates into go-dependencies and github-actions.
The schedule is changed to run weekly on mondays.


SamYuan1990 (approved):
LGTM

PR #1515, opened 2024-06-11
chore(manifests): cleanup unused CGROUP_METRICS=*

sthaha:
Cleaning this config since there is seems to be no reference to this setting in code.


sunya-ch (approved):
/lgtm 
This config really seems out of date.

dave-tucker (approved):

PR #1508, opened 2024-06-10
chore(validator): add raw queries

rootfs:
Current script only keeps compound prom query. This PR makes the raw query results persisted so we can use them to replay the validation.

PR #1522, opened 2024-06-12
Fix MSE and MAPE Single Queries

KaiyiLiu1234:
Fixed MSE Queries and added MAPE Queries and included Query outputs in cli. These changes are fixed to @rootfs [PR](https://github.com/sustainable-computing-io/kepler/pull/1508)


KaiyiLiu1234:
Note queries that do not work are anyones which are {job='vm', mode='idle'}. VM does not appear to account for idle. Only dynamic.

rootfs (approved):

rootfs:
> Note queries that do not work are anyones which are {job='vm', mode='idle'}. VM does not appear to account for idle. Only dynamic.

you need to turn on idle power on the VM (i.e. `EXPOSE_ESTIMATED_IDLE_POWER_METRICS=true`)
https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/ansible/kepler_playbook.yml#L26

PR #1523, opened 2024-06-12
Added MSE and MAPE Dashboards

KaiyiLiu1234:
Added MSE, MAPE, SE, APE Panels for Core, Dram, Package, Platform Joules. Added MSE Panels for Core, Dram,
Package, Platform, CPU Time, and Cache Hits. Included Idle Power Estimation Metrics. 

Image shows APE and SE Dynamic Power Error Metrics
metalvsvm
Image shows sample of metrics used for Core, Dram, Package, Platform MAPE Dynamic Power Error Metrics
sample_metrics
Image shows sample of metrics used for Core, Dram, Package, Platform MSE Dynamic Power Error Metrics
sample_2
Image shows thresholds added for MAPE Panels
thresholds


KaiyiLiu1234:
Followup PR to these dashboards will include Idle Power Comparisons.

manifests/compose/validation/metal/grafana/dashboards/validation/dashboard.json
```diff
@@ -29,6 +29,648 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "PDE6745920139CE56"
+          },
+          "editorMode": "code",
+          "expr": "(rate(kepler_process_bpf_cpu_time_ms_total{pid=\"${process}\", job=\"metal\"}[$__rate_interval]) - on() sum by(__name__, job) (rate(kepler_process_bpf_cpu_time_ms_total{job=\"vm\"}[$__rate_interval])))^2",
```
sthaha, line 656: could you please format the queries for bette readablity ? 
e..g it makes it harder to see that `on () sum by (__name__, job)` could just just written as `sum(..)`

sthaha (changes requested):
could you please attach a screenshot with commits describing what is being changed? 
e.g. https://github.com/sustainable-computing-io/kepler/pull/1429

manifests/compose/validation/metal/grafana/dashboards/validation/dashboard.json
KaiyiLiu1234: How would I format it? As in with multilines? JSON isn't built to do that but I think there are workarounds like string concatenation.

KaiyiLiu1234:
Added Images

manifests/compose/validation/metal/grafana/dashboards/validation/dashboard.json
sthaha: @KaiyiLiu1234 , See existing dashboards for example the dev dashboard to understand how queries can be formatted.

sthaha:
@KaiyiLiu1234 , I see that the dashboard adds 
(metal - vm)^2 which isn't MSE or MAPE. I think it may be worth exploring if we can use avg_over_time() to get the MSE of the time - range selected in the grafana.

rootfs (approved):

KaiyiLiu1234:
> @KaiyiLiu1234 , I see that the dashboard adds (metal - vm)^2 which isn't MSE or MAPE.
> 
> I think it may be worth exploring if we can use `avg_over_time()` to get the MSE (Single Value) for the time - range selected in the grafana.

@sthaha  you are right it is not MSE, it is Squared Error. I should rename it. The reason being that since the error metrics are being graphed, it doesn't make sense to graph the mean of it. Rather you want to graph what the error is at every timestamp. I think using avg_over_time is not very applicable in the case of graphing error metrics, however I can explore it and show an image.

manifests/compose/validation/metal/grafana/dashboards/validation/dashboard.json
KaiyiLiu1234: Fixed

KaiyiLiu1234: Resolved

KaiyiLiu1234:
@sthaha Added MSE and MAPE by using the Stat Visualization. Stat Visualization will look at the data according to the global time variable range and automatically includes a mean function. 

mean_values

sthaha:
@vprashar2929 , would you mind taking a look at the dashboard changes once more?

vprashar2929:
@KaiyiLiu1234 For the values of MAPE which comes to be `100000` due to dividing by zero we can show something like this on dashboard:
Screenshot 2024-06-21 at 2 23 18 PM
Wdyt?

KaiyiLiu1234:
@vprashar2929  Originally @sthaha wanted to include the clamping so an actual value is shown. But this should also be acceptable as well and I think this way is ok to show as well. I will remove the clamp_max

KaiyiLiu1234:
@vprashar2929 is this acceptable?

KaiyiLiu1234:
@sthaha Can you approve this PR?

sthaha (approved):

PR #1527, opened 2024-06-13
feat(validator): load validations from validations.yaml

sthaha:
This commit is rewrite of existing validator. 

Changes include
  * remove queries.json in favour of validations.yaml
  * move prometheus job names - metal and vm under `prometheus` field
  * Simplify parsing queries making it testable
  * Add run-validation command to rerun validations
  * Add logging that can be controlled by `log_level` in validator.yaml
  * Fix use of hard-coded ssh-key to generate VM specs

Minor
  * All validator files formatted using `ruff`

Produces the following report - https://gist.github.com/sthaha/80b30096d317bcf2e11f4f8f2ece3976


e2e/tools/validator/scripts/stressor.sh
```diff
@@ -2,29 +2,16 @@
...
-# # Stress at 25% load for 20 seconds
-# stress-ng --cpu $(nproc) --cpu-load 25 --timeout 20s
-#
-# # Stress at 10% load for 20 seconds
-# stress-ng --cpu $(nproc) --cpu-load 10 --timeout 20s
-
-# cool off for 40 seconds
-# idle
-stress-ng --cpu $(nproc) --cpu-load 0 --timeout 10s
+stress-ng --cpu "$cpu" --cpu-load 0 --timeout 10s # idle for 2 scrapes
```
rootfs, line 7: how about this one?
```bash
 for level in $(seq 10 10 100); do
              stress-ng --cpu $(nproc) --cpu-load $level --timeout 10s
```

rootfs:
can you post a copy of the report file? Want to check if we need to sync the CI with the validation report [here ](https://sustainable-computing-io.github.io/kepler-metal-ci/kepler-model-validation.html)

e2e/tools/validator/scripts/stressor.sh
sthaha: https://gist.github.com/sthaha/fc5c33ae6f92945902296fa665b41447 is what I run on my VM

The idea is to have 100% running for longer

sthaha:
@rootfs  - https://gist.github.com/sthaha/d3f669f53e9f0cf26d4052352ab93b9a

e2e/tools/validator/scripts/stressor.sh
sthaha: fixed now

e2e/tools/validator/src/validator/config/__init__.py
```diff
@@ -50,49 +60,57 @@ def load(config_file: str) -> Validator:
...
 
-    remote_config = config['remote']
+    remote_config = config["remote"]
     # NOTE: set default path to pkey if password is not set
 
-    pkey = remote_config.get('pkey', '')
-    if pkey != '':
-        pkey=os.path.expanduser(pkey)
+    pkey = remote_config.get("pkey", "")
+    if pkey != "":
```
vprashar2929, line 70: in case of 
```
remote:
  host: xxx
  username: xx
  password: xxx
  pkey:
```
this check fails and tries to connect with `pkey` Shouldn't this be `if pkey is not None`?

e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -2,153 +2,201 @@
...
+
+def new_report() -> Report:
+    # run git describe command and get the output as the report name
+    git_describe = subprocess.run(["git", "describe", "--tag"], stdout=subprocess.PIPE)
+    tag = git_describe.stdout.decode().strip()
+
+    results_dir = os.path.join("/tmp", f"validator-{tag}")
+    os.makedirs(results_dir, exist_ok=True)
+
+    path = os.path.join("/tmp", f"report-{tag}.md")
```
vprashar2929, line 62: Can we use current directory for storing the results instead of `/tmp`?

e2e/tools/validator/src/validator/config/__init__.py
sthaha: good find!

e2e/tools/validator/validations.yaml
```diff
@@ -0,0 +1,136 @@
+validations:
+  - name: platform - dynamic
+    actual: |
+      rate(
+        kepler_{level}_platform_joules_total{{
+          job="{metal_job_name}",
+          {vm_selector},
+          mode="dynamic",
```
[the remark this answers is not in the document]
sthaha: it works, and makes it easier to add an additional line. so my preference is to keep it.

e2e/tools/validator/src/validator/cli/__init__.py
sthaha: Lets add a reports param to `validator.yaml` in a different pr.

e2e/tools/validator/src/validator/config/__init__.py
sthaha: fixed now 🤞

e2e/tools/validator/validator.yaml
```diff
@@ -1,16 +1,10 @@
...
-  # password: supersecret
-  # pkey: ~/.ssh/id_rsa
+  pkey: ~/.ssh/kvm
 
 metal:
-  metal_job_name: metal
-  vm_job_name: vm
   vm:
-    pid: 2093543
+    pid: 1819485
```
sunya-ch, line 7: shall this be some fix number such as 0? In my understanding, this will be replaced automatically by the VM process id.

e2e/tools/validator/validations.yaml
```diff
@@ -0,0 +1,136 @@
+validations:
```
sunya-ch, line 1: suggest to add "vm" in the name for clarification. 
can we add a field for query description? do not have to be in this PR anyway.

sunya-ch (approved):
This PR looks good to me overall. 
Just left some suggestions. Do not necessary to be added in this PR.

e2e/tools/validator/validations.yaml
sthaha: I will think of a better way to do this (in a separate PR) since I have the idea of using the same for comparing 2 kepler running on the same machine.

e2e/tools/validator/validator.yaml
sthaha: I feel we should delete `validator.yaml` and `.gitignore` it so that everyone would make a copy of `validator.yaml.sample`

PR #1532, opened 2024-06-14
fix(bpf): Fix map lookup for IRQ/Page Cache

dave-tucker:
The processes map keys are kernel tgids.
This corrects the code in the softirq and pache cache functions to lookup using the tgid.


sthaha (approved):
LGTM

PR #1534, opened 2024-06-14
chore: refactor code for reporting

husky-parul:


e2e/tools/validator/src/validator/stresser/__init__.py
```diff
@@ -26,12 +26,12 @@ def __repr__(self):
     def connect(self):
         print(f"connecting -> {self.user}@{self.host}")
 
-        if self.pkey is not None:
-            pkey = paramiko.RSAKey.from_private_key_file(self.pkey)
-            self.ssh_client.connect(
-                hostname=self.host, port=self.port,
-                username=self.user, pkey=pkey,
-            )
+        if self.pkey is not None and self.pkey is not "" :
```
sthaha, line 29:
```suggestion
        if  not self.pkey :
```
is more pythonic

sthaha (approved):

e2e/tools/validator/src/validator/stresser/__init__.py
```diff
@@ -75,3 +75,14 @@ def run_script(self, script_path):
...
         )
+    
+    def run_command(self, command):
+        self.connect()
+        _, stdout, stderr = self.ssh_client.exec_command(command)
+        exit_status = stdout.channel.recv_exit_status()
+        result = stdout.read().decode('ascii').strip("\n")
+        
+        if exit_status != 0:
+            print(f"Something went wrong in running command: ", command)
```
sthaha, line 86: It is better to return an Object that holds exit_code, stdout and stderr and let the caller decide what to do than ignore this.

sthaha (changes requested):
could you please fix run_command to return RunResult(stdout, stderr, exit_code)

sthaha:
@husky-parul , if you don't mind,  I am making major changes to the validator script. I will incorporate this in that patch. 
See #1527

Issue #1530, opened 2024-06-13
kepler_process_bpf_cpu_time_ms_total metrics turned to 0 in latest code


rootfs:
### What happened?

We need BPF CPU time to estimate dynamic power on VM. But the metrics are recently unstable.

# In release-0.7.8, the kepler_process_bpf_cpu_time_ms_total is correct:
```console
kepler_process_bpf_cpu_time_ms_total{command="systemd",container_id="system_processes",pid="188268",source="bpf",vm_id=""} 2
kepler_process_bpf_cpu_time_ms_total{command="sshd",container_id="system_processes",pid="592344",source="bpf",vm_id=""} 5
kepler_process_bpf_cpu_time_ms_total{command="pmdalinux",container_id="system_processes",pid="1478",source="bpf",vm_id=""} 6
kepler_process_bpf_cpu_time_ms_total{command="redis-server",container_id="system_processes",pid="1281",source="bpf",vm_id=""} 7
kepler_process_bpf_cpu_time_ms_total{command="pmproxy",container_id="system_processes",pid="4020821",source="bpf",vm_id=""} 9
kepler_process_bpf_cpu_time_ms_total{command="pmdaproc",container_id="system_processes",pid="1476",source="bpf",vm_id=""} 13
kepler_process_bpf_cpu_time_ms_total{command="kworker/4:1",container_id="kernel_processes",pid="590323",source="bpf",vm_id=""} 7082
```

# In release-0.7.10, all the CPU time goes to pid 0:
```console
kepler_process_bpf_cpu_time_ms_total{command="etcd",container_id="system_processes",pid="996366",source="bpf",vm_id=""} 0
kepler_process_bpf_cpu_time_ms_total{command="in:imjournal",container_id="system_processes",pid="1299",source="bpf",vm_id=""} 0
kepler_process_bpf_cpu_time_ms_total{command="node",container_id="system_processes",pid="284272",source="bpf",vm_id=""} 0
kepler_process_bpf_cpu_time_ms_total{command="node",container_id="system_processes",pid="284297",source="bpf",vm_id=""} 0
kepler_process_bpf_cpu_time_ms_total{command="prometheus",container_id="system_processes",pid="512773",source="bpf",vm_id=""} 0
kepler_process_bpf_cpu_time_ms_total{command="swapper/0",container_id="kernel_processes",pid="0",source="bpf",vm_id=""} 1203
```

# In latest code, all cpu time is zero
```console
[...]

PR #1542, opened 2024-06-17
fix(bpf): Use BTF-Defined Raw Tracepoints

dave-tucker:
It turns out that RHEL is carrying a patch that increases the size of the unreadable tracepoint-common area from 8 bytes to 12 bytes. This means that our reads into the context area are off, causing the probes to return invalid data.

To fix this we use BTF-Defined Raw Tracepoints that were added in Kernel 5.5 (less than our 5.12 minimum version). This type of tracepoint program uses BTF and CO-RE to access arguments from the tracepoint context.

Fixes: #1530


rootfs (approved):
Great!

PR #1550, opened 2024-06-17
fix(bpf): Access __state from task_struct

dave-tucker:
In older kernels prev_state isn't passed as part of the context of the
function. Therefore it's better to access prev_task-->_state directly.

The field was renamed from state to __state in Kernel 5.14.
Given that we need to support Kernel 5.12+ we include the
bpf_core_read.h file such that we can use the bpf_core_field_exists
function that allows us to see which field is present in task struct
and return the correct value.

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -198,6 +198,22 @@ struct task_struct {
 	int pid;
 } __attribute__((preserve_access_index));
 
+struct task_struct___o {
+	unsigned int state;
+};
+
+struct task_struct___x {
+	unsigned int __state;
+};
+
+static unsigned int get_task_state(void *task)
+{
+	struct task_struct___x *t = task;
+	if (bpf_core_field_exists(t->__state))
+		return t->__state;
+	return ((struct task_struct___o *)task)->state;
+}
+
 SEC("tp_btf/sched_switch")
 int kepler_sched_switch_trace(u64 *ctx)
 {
@@ -212,7 +228,6 @@ int kepler_sched_switch_trace(u64 *ctx)
 
 	prev_task = (struct task_struct *)ctx[1];
 	next_task = (struct task_struct *)ctx[2];
-	prev_state = (unsigned int)ctx[3];
 
 	prev_pid = (u32)prev_task->pid;
 	next_pid = (u32)next_task->pid;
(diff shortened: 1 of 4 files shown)


bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -196,6 +196,7 @@ static inline void collect_metrics_and_reset_counters(
 
 struct task_struct {
 	int pid;
+	unsigned int __state;
```
sthaha, line 199: Is it fixed / defined what the size of  `int` is ?

dave-tucker: Yes. https://github.com/sustainable-computing-io/kepler/blob/main/bpfassets/libbpf/src/kepler.bpf.h#L8

dave-tucker: Also, these fields are BTF relocated - which will perform the necessary checks around int size and signedness at program load time.

sthaha: Thanks Dave. 
Related question ...  shouldn't we use `unsigned int` or `u32`  for `pid` as well?

marceloamaral: Shouldn't we use BPF_CORE_READ() to read the pointers in a safe manner?

dave-tucker: @sthaha I'm just using whatever is used in `task_struct` in the kernel - see https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L977

if you follow the links from `pid_t` down to a concrete type you get to https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/posix_types.h#L28

@marceloamaral no you don't need to use BPF_CORE_READ in tp_btf programs.
https://mozillazg.com/2022/06/ebpf-libbpf-btf-powered-enabled-raw-tracepoint-common-questions-en.html#hidthe-difference-between-btf-raw-tracepoint-and-raw-tracepoint

marceloamaral (approved):
/lgtm

bpfassets/libbpf/src/kepler.bpf.c
sthaha: Thanks a lot for clarifications, @dave-tucker 🙏

PR #1551, opened 2024-06-18
fix(bpf): Fix kepler_write_page_cache attach

dave-tucker:
We need to use a standard tracepoint for kepler_write_page_cache to
avoid BTF relocation errors on systems where the writeback_dirty_folio
tracepoint doesn't exist.

This is due to the rename of writeback_dirty_page to
writeback_dirty_folio that happened in Kernel 5.16.

This patch adjust the attach logic to fall back to the old name if the
writeback_dirty_folio tracepoint does not exist.

### DIFF
--- a/pkg/bpf/exporter.go
+++ b/pkg/bpf/exporter.go
@@ -195,9 +195,13 @@ func (e *exporter) attach() error {
 	if err != nil {
 		return fmt.Errorf("failed to get kepler_write_page_trace: %w", err)
 	} else {
-		_, err = page_write.AttachTracepoint("writeback", "writeback_dirty_folio")
-		if err != nil {
-			klog.Warningf("failed to attach tp/writeback/writeback_dirty_folio: %v. Kepler will not collect page cache write events. This will affect the DRAM power model estimation on VMs.", err)
+		category := "writeback"
+		name := "writeback_dirty_page"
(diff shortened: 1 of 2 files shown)


pkg/bpf/exporter.go
```diff
@@ -195,9 +195,13 @@ func (e *exporter) attach() error {
 	if err != nil {
 		return fmt.Errorf("failed to get kepler_write_page_trace: %w", err)
 	} else {
-		_, err = page_write.AttachTracepoint("writeback", "writeback_dirty_folio")
-		if err != nil {
-			klog.Warningf("failed to attach tp/writeback/writeback_dirty_folio: %v. Kepler will not collect page cache write events. This will affect the DRAM power model estimation on VMs.", err)
+		category := "writeback"
+		name := "writeback_dirty_page"
+		if _, err := os.Stat("/sys/kernel/debug/tracing/events/writeback/writeback_dirty_folio"); err == nil {
+			name = "writeback_dirty_folio"
```
marceloamaral, line 201: What is the main difference between `writeback_dirty_page` and `writeback_dirty_folio`?

dave-tucker: There is no difference - it was just renamed. See:
https://github.com/torvalds/linux/commit/b9b0ff61eef51469eae962b80bae094ce2c8c2a5

PR #1554, opened 2024-06-18
Remove tgid map

marceloamaral:
With the new task structure in the tracepoint in the code, we can now directly get the previous running task tgid, so that we can remove the map to track and translate the pid to tgid.

### DIFF
--- a/bpfassets/libbpf/src/kepler.bpf.c
+++ b/bpfassets/libbpf/src/kepler.bpf.c
@@ -17,13 +17,6 @@ struct {
 } pid_time_map SEC(".maps");
 
 struct {
-	__uint(type, BPF_MAP_TYPE_LRU_HASH);
-	__type(key, u32);
-	__type(value, u32);
-	__uint(max_entries, MAP_SIZE);
-} pid_tgid_map SEC(".maps");
-
-struct {
 	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
 	__type(key, int);
 	__type(value, u32);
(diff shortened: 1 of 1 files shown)


bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -198,6 +186,10 @@ struct task_struct {
 	int pid;
 } __attribute__((preserve_access_index));
 
+struct task_struct___t {
+	unsigned int tgid;
+};
+
```
dave-tucker, line 192: You don't need to create an new struct - this wouldn't work anyway as tgid isn't the first field (that's state)

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -198,6 +186,10 @@ struct task_struct {
 	int pid;
 } __attribute__((preserve_access_index));
```
dave-tucker, line 187: Add `unsigned int tgid;` into `struct task_struct`.
It will get CO:RE relocated to the right offset in task_struct at load time.

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -214,11 +206,16 @@ static unsigned int get_task_state(void *task)
 	return ((struct task_struct___o *)task)->state;
 }
 
+static unsigned int get_task_tgid(void *task)
```
dave-tucker, line 209: No need for a function here. With the change above you can just use `prev_task->tgid` directly.

dave-tucker (changes requested):

bpfassets/libbpf/src/kepler.bpf.c
```diff
@@ -231,6 +220,7 @@ int kepler_sched_switch_trace(u64 *ctx)
 
 	prev_pid = (u32)prev_task->pid;
 	next_pid = (u32)next_task->pid;
+	prev_tgid = next_task->tgid;
```
dave-tucker, line 223: Should this be prev_task->tgid?

marceloamaral: Yeap, fixed!

rootfs (approved):

dave-tucker (approved):

PR #1557, opened 2024-06-19
accelerator: refactor gpu code

maryamtahhan:
Refactor pkg/sensors/accelerator to use a more generic device abstractions that different devices can plug into.

Provide a more generic interface the accelerator devices can plug into. 

Deprecate QAT for the following [reasons](https://gist.github.com/maryamtahhan/81ded3940f32c9eb53fb84c7fa5d117a#introduction)


Makefile
```diff
@@ -43,8 +43,14 @@ CTR_CMD_PUSH_OPTIONS ?=
 
 GENERAL_TAGS := 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo libbpf '
 GPU_TAGS := ' gpu '
```
dave-tucker, line 45: Should this be `GPU_TAGS := ' '` given we're not using `gpu` anywhere I believe

Makefile
```diff
@@ -43,8 +43,14 @@ CTR_CMD_PUSH_OPTIONS ?=
...
+ifeq ($(shell ldconfig -p | grep -q libnvml_injection.so && echo exists),exists)
+	GPU_TAGS := ' nvml '
```
dave-tucker, line 47:
```suggestion
	GPU_TAGS := "$(GPU_TAGS) nvml"
```

Makefile
```diff
@@ -43,8 +43,14 @@ CTR_CMD_PUSH_OPTIONS ?=
...
+endif
+ifeq ($(shell ldconfig -p | grep -q libdcgm.so && echo exists),exists)
+	GPU_TAGS := ' dcgm '
```
dave-tucker, line 50:
```suggestion
	GPU_TAGS := "$(GPU_TAGS) dcgm"
```

Makefile
```diff
@@ -43,8 +43,14 @@ CTR_CMD_PUSH_OPTIONS ?=
...
+endif
 ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
-	GPU_TAGS := $(GPU_TAGS)'habana '
+	GPU_TAGS := ' habana '
```
dave-tucker, line 53:
```suggestion
	GPU_TAGS := "$(GPU_TAGS) habana"
```

Makefile
dave-tucker: IMHO this should append to the list - so you could build with all things supported if your system has all libraries available.

dave-tucker (commented):
Some small Makefile comments but overall LGTM

Makefile
maryamtahhan: Yeah I can change that

Makefile
maryamtahhan: I think it's unlikely to have a system that has Habana + Nvidia GPUs. And TBH if the system has both NVML  libraries and DCGM we prioritise the use of dcgm. so rather than building aggregating the tag I had just set it flat out.

maryamtahhan: From an image POV atm we have separate images also:
Screenshot 2024-06-19 at 14 54 19

For these reasons I'm less inclined to aggregate the tag (ATM). If we decide to change the image builds then absolutely

dave-tucker (approved):
LGTM 🎉

dave-tucker:
@rootfs could you take a quick look over this? IMHO it's ready to go but would appreciate a second set of eyes 👀

dave-tucker:
@sthaha would you mind taking a quick pass over these changes? I've reviewed this twice now and I'm pretty sure it's good to go. Would appreciate a second set of eyes 👀

cmd/exporter/exporter.go
```diff
@@ -144,31 +140,11 @@ func main() {
...
-		// therefore, we wait up to 1 min to allow the gpu operator initialize
-		for i := 0; i <= maxGPUInitRetry; i++ {
-			err = gpu.Init()
-			if err == nil {
-				break
-			} else {
-				time.Sleep(6 * time.Second)
+		if r := accelerator.Registry(); r != nil {
+			if err := accelerator.CreateAndRegister(accelerator.GPU, r, true); err != nil {
+				klog.Errorf("failed to init GPU accelerators: %v", err)
```
sthaha, line 145: This confuses me a bit .. here we obtain the global registry (which can't be nil) and then use  CreateAndRegister register with `r` . This feels like the Register should be a function of `registry` instead of `accelerator`

cmd/exporter/exporter.go
```diff
@@ -223,6 +199,7 @@ func main() {
 		klog.Fatalf("%s", fmt.Sprintf("failed to listen and serve: %v", err))
 	case <-signalChan:
 		klog.Infof("Received shutdown signal")
+		accelerator.Shutdown(accelerator.GPU)
```
sthaha, line 202: Could you please explain why `Shutdown` requires a type?

pkg/collector/metric_collector.go
```diff
@@ -154,19 +153,18 @@ func (c *Collector) updateResourceUtilizationMetrics() {
 
 // update the node metrics that are not related to aggregated resource utilization of processes
 func (c *Collector) updateNodeResourceUtilizationMetrics(wg *sync.WaitGroup) {
-	defer wg.Done()
-	if config.IsExposeQATMetricsEnabled() && qat.IsQATCollectionSupported() {
-		accelerator.UpdateNodeQATMetrics(stats.NewNodeStats(c.bpfSupportedMetrics))
-	}
+	wg.Done()
 }
```
sthaha, line 157: should we remove this function?

pkg/collector/energy/node_energy_collector.go
```diff
@@ -66,10 +66,14 @@ func UpdateNodeComponentsEnergy(nodeStats *stats.NodeStats, wg *sync.WaitGroup)
 // UpdateNodeGPUEnergy updates each GPU power consumption. Right now we don't support other types of accelerators
 func UpdateNodeGPUEnergy(nodeStats *stats.NodeStats, wg *sync.WaitGroup) {
 	defer wg.Done()
-	if config.EnabledGPU && gpu.IsGPUCollectionSupported() {
-		gpuEnergy := gpu.GetAbsEnergyFromGPU()
-		for gpu, energy := range gpuEnergy {
-			nodeStats.EnergyUsage[config.AbsEnergyInGPU].SetDeltaStat(fmt.Sprintf("%d", gpu), uint64(energy))
+	if config.EnabledGPU {
+		if gpus, err := acc.Registry().ActiveAcceleratorsByType(acc.GPU); err == nil {
+			for _, g := range gpus {
```
sthaha, line 71: Nearly all calls to ActiveAccelerators use `acc.GPU`. this makes me feel that perhaps dummy shouldn't an accelerator type. Is a Dummy type different to `GPU` in any the registry treats it?

pkg/collector/stats/types/types.go
```diff
@@ -132,8 +132,16 @@ func (s *UInt64StatCollection) SetDeltaStat(key string, newDelta uint64) {
...
+	}
 	for _, stat := range s.Stat {
-		sum += stat.Delta
+		if stat != nil {
+			sum += stat.Delta
+		} else {
+			klog.V(3).Info(" Error retrieving stat")
+		}
 	}
 	return sum
```
sthaha, line 146: This seems to be fixing an issue than a required patch for the abstraction. Is that right? 
Also makes me wonder when is it that `s` and `stat` can be nil?

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+	} else {
+		return errors.New("No Accelerator collectors found")
+	}
+	return nil
+}
+
+// NewAccelerator creates a new Accelerator instance with a specific device [NVML|DCGM|DUMMY|HABANA] for the local node.
+func NewAccelerator(devType device.DeviceType, atype AcceleratorType, r *AcceleratorRegistry) error {
```
sthaha, line 158: IMHO, we shouldn't mix creation of an object / accelerator with registering it in a registry.  If that is the case, then new accelerator should be a method of the registry.

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+	var d device.DeviceInterface
+	var err error
+
+	_, ok := r.Registry[devType] // e.g. accelerators[nvml|dcgm|habana|dummy]
```
sthaha, line 162: Shouldn't you mutex lock for read? I think sync.Map may be a better candidate if we feel registry will be accessed by multiple threads.

sthaha (changes requested):
@maryamtahhan , thank you for the patch. 

I feel the abstraction that is proposed needs a bit more work to make it easier to understand and maintain, especially around the concept of dummy devices,  registering and creating accelerators.

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+		return errors.New("unsupported accelerator")
+	}
+
+	if numDevs != 0 {
+		devices := getDevices()
+
+		klog.V(5).Infof("Initializing the %s Accelerator collectors in type %v", atype.String(), devices)
+
+		for _, devType := range devices {
+			for i := 0; i <= maxDeviceInitRetry; i++ {
```
sthaha, line 136:
```suggestion
			for i := 0; i < maxDeviceInitRetry; i++ {
```

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+				// Create a new accelerator and add it to the registry
+				if err := NewAccelerator(devType, atype, r); err != nil {
+					klog.Errorf("Could not init the %s Accelerator collector going to try again", devType.String())
+					if sleep {
+						// The GPU operators typically takes longer time to initialize than kepler resulting in error to start the gpu driver
+						// therefore, we wait up to 1 min to allow the gpu operator initialize
+						time.Sleep(6 * time.Second)
```
sthaha, line 143:
```suggestion
						time.Sleep(60 * time.Second)
```

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+func SetRegistry(registry *AcceleratorRegistry) {
+	globalRegistry = registry
+}
+
+// Devices returns a map of supported accelerators.
+func (r *AcceleratorRegistry) Accelerators() (map[device.DeviceType]Accelerator, error) {
+	if len(r.Registry) == 0 {
+		return nil, errors.New("no accelerators found")
+	}
+	return r.Registry, nil
```
sthaha, line 94: This exposes internal registry to the caller and now allows caller to make changes to the internal map.

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+		if a.AccType() == t && a.IsRunning() {
+			d := a.Device()
+			if d.IsDeviceCollectionSupported() {
+				acc[d.DevType()] = a
+			}
+		}
+	}
+	if len(acc) == 0 {
+		return nil, errors.New("accelerators not found")
+	}
```
sthaha, line 110: IMHO, not having any active accelerator shouldn't be an error since it is an acceptable state.

Makefile
```diff
@@ -42,9 +42,15 @@ CTR_CMD            ?= $(or $(shell podman info > /dev/null 2>&1 && which podman)
...
 GENERAL_TAGS := 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo libbpf '
-GPU_TAGS := ' gpu '
+GPU_TAGS := ' '
+ifeq ($(shell ldconfig -p | grep -q libnvml_injection.so && echo exists),exists)
+	GPU_TAGS := ' nvml '
+endif
+ifeq ($(shell ldconfig -p | grep -q libdcgm.so && echo exists),exists)
+	GPU_TAGS := ' dcgm '
+endif
 ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
```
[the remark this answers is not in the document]
maryamtahhan: We are only prioritising dcgm over NVML. 

Dcgm/nvml won't exist in the same system

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,156 @@
...
+	once           sync.Once
+)
+
+type DeviceType int
+
+// Function prototype to create a new deviceCollector.
+type deviceStartupFunc func() DeviceInterface
+type DeviceRegistry struct {
+	gpuDevices   map[DeviceType]deviceStartupFunc // Static map of supported gpuDevices.
+	dummyDevices map[DeviceType]deviceStartupFunc // Static map of supported dummyDevices.
```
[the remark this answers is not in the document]
maryamtahhan: We only use the dummy devices for testing amd is built in to have at least one thing built in to the pkg. I thought it would be more efficient to have separate registries for different device types (one being qat again in the future,  as the code that's there today for qat will not work in RHEL and the plan is to fix that in the very near future)...

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,156 @@
...
+	return [...]string{"DUMMY", "HABANA", "DCGM", "NVML"}[d]
+}
+
+// Registry gets the default device DeviceRegistry instance
+func Registry() *DeviceRegistry {
+	once.Do(func() {
+		globalRegistry = &DeviceRegistry{
+			gpuDevices:   map[DeviceType]deviceStartupFunc{},
+			dummyDevices: map[DeviceType]deviceStartupFunc{},
+		}
```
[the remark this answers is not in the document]
maryamtahhan: The dummy devices are just for testing. We had it in the old code. I thought it was a useful implementation that can be leveraged for unit testing. We also need at least 1 thing that's built into the device pkg. So I kept it

pkg/sensors/accelerator/accelerator.go
```diff
@@ -0,0 +1,231 @@
...
+// SetRegistry replaces the global registry instance
+// NOTE: All plugins will need to be manually registered
+// after this function is called.
+func SetRegistry(registry *AcceleratorRegistry) {
+	globalRegistry = registry
+}
+
+// Devices returns a map of supported accelerators.
+func (r *AcceleratorRegistry) Accelerators() (map[device.DeviceType]Accelerator, error) {
+	if len(r.Registry) == 0 {
```
sthaha, line 91: I see that we are using `mutex` to protect `r.Registry` but there places where `Registry` map is accessed and exposed to the world outside. This isn't thread safe. We have to make a decision if the `Registry` needs to be threadsafe and prevent access to internal members though mutex (preferably RWMutex) or use `sync.Map`

pkg/collector/energy/node_energy_collector.go
maryamtahhan: There was also QAT, which I plan on readding after I reimplement it as The current implementation will not work on RHEL. 

There will be other devices I'm sure in the future also DPUs and IPUs... we are just starting with fixing up what's currently supported in kepler

pkg/collector/stats/types/types.go
maryamtahhan: I can't remember what HW and test generated a null pointer dereference but this is something that manifested in one of the setups when I ran kepler

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,156 @@
...
+	gpuDevices   map[DeviceType]deviceStartupFunc // Static map of supported gpuDevices.
+	dummyDevices map[DeviceType]deviceStartupFunc // Static map of supported dummyDevices.
+}
+type DeviceInterface interface {
+	// Name returns the name of the device
+	Name() string
+	// DevType returns the type of the device (nvml, dcgm, habana ...)
+	DevType() DeviceType
+	// DevTypeName returns the type of the device (nvml, dcgm, habana ...) as a string
+	DevTypeName() string
```
sthaha, line 37: isn't this always `DevType()`.String()?

pkg/sensors/accelerator/accelerator.go
maryamtahhan: Originally I had this all separate.  But TBH I ended up invoking the functions that are currently wrapped separately and thought it was cleaner from an exporter perspective if we just wrapped things in a single function. 

The exporter is creating a registery with the intent of adding an accelerator.  So calling things in one place seemed more efficient

pkg/sensors/accelerator/device/sources/dcgm.go
```diff
@@ -447,7 +529,17 @@ func ToFloat64(value dcgm.FieldValue_v1, multiplyFactor float64) float64 {
 		}
 
 	default:
-		klog.Errorf("DCGM metric type %s not supported: %v\n", value.FieldType, value)
+		klog.Errorf("DCGM metric type %v not supported: %v\n", value.FieldType, value)
 		return defaultValue
 	}
 }
+
+func nvmlErrorString(errno nvml.Return) string {
```
sthaha, line 537: why is `nvmlErrorString` redeclared here?

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,156 @@
...
+	// InitLib the external library loading, if any.
+	InitLib() error
+	// Init initizalizes and start the metric device
+	Init() error
+	// Shutdown stops the metric device
+	Shutdown() bool
+	// DevicesByID returns a map with devices identifying then by id
+	DevicesByID() map[int]any
+	// DevicesByName returns a map with devices identifying then by name
+	DevicesByName() map[string]any
```
sthaha, line 49: Perhaps we need a better name for `DeviceInterface` since this seems to suggest that a `device` has `DevicesByName()` that a `Device` needs to implement. 

I also wonder if the implementation would change between Devices, which if you look at dcgm is 
```

func (d *GPUDcgm) DevicesByName() map[string]any {
	devices := make(map[string]any)
	return devices
}
```
For Habana is 
```
func (g *GPUHabana) DevicesByName() map[string]any {
	devices := make(map[string]interface{})
	return devices
}

```
Is this used?

pkg/sensors/accelerator/accelerator.go
sthaha: Its about coupling, An accelerator shouldn't have to know existence of a registry and if it has to and only a single instance of the accelerator should be created and registered, then the responsibility should be delegated to the registry / factory .

pkg/collector/stats/types/types.go
sthaha: IMO, calling `SumAllDeltaValues()` on a nil UInt64StatCollection   is a logical error and should be fixed than masking it.

pkg/sensors/accelerator/device/device.go
```diff
@@ -0,0 +1,156 @@
...
+
+// SetRegistry replaces the global registry instance
+// NOTE: All plugins will need to be manually registered
+// after this function is called.
+func SetRegistry(registry *DeviceRegistry) {
+	globalRegistry = registry
+}
+
+// AddDeviceInterface adds a supported device interface, prints a fatal error in case of double registration.
+func AddDeviceInterface(dtype DeviceType, accType string, deviceStartup deviceStartupFunc) {
```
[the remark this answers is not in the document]
sthaha: Lets look at the coupling here ...

All devices (currently those under `accelerator/sources`) folder have to be
registered here.

This requires calling (AddDeviceInterface aka `Register`) with the following information
* type of device which is `DUMMY` , `HABANA`, `DCGM`, `NVML`
* further more it needs to specify the type of accelerator DUMMY, GPU

So the devices registry knows about all the devices it support and also that there is an accel type

However, this function also already hardcodes information about DCGM and knows what AccelType each device type is . 
So I would argue that  AccelType need not be specified by the caller. 

Further more adding or deleting a new device requires us to change the
DeviceType enum (coupling).

Next the point of this function is to register the device to the global registry, so the function needs to be part of registry since
the package that is trying to register. 
Changing this to a convenience function is fine though.
```
func Register(dtype Type, accType string, deviceStartup deviceStartupFunc) {
	globalRegistry.Register(dtype, accType, deviceStartup)
}
```

Regarding the call to Register, A Device that is registering already knows its
type and how to start, so shouldn't this  be a method of the device itself. In short, shouldn't dgcm look as follows

```
import "pkg/sensors/accelerator/devices"

type dcgm struct {
}

func (d *dcgm) Type() device.Type {
	return device.DCGM
}

func (d *dcgm) Start() error  {
	...
}

func Init() {
  devices.Register(&dcgm{})
}
```

sthaha (changes requested):

Makefile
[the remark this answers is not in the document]
sthaha: nothing actually prevents all 3 shared libraries being on the same machine, so that's really beyond our control. However if the build machine has all 3 shared libraries, then the makefile will choose `habana` over `dcgm` over `nvml` - right?

dave-tucker:
@maryamtahhan @sthaha had a chat about the path forward for this PR.

@sthaha is going to take a pass over the proposed interfaces and make some changes:

1. To ensure we're not exposing methods in interfaces that are never used (we found a couple)
2. To ensure that the interfaces don't provide a leaky abstraction

These will appear as commits on top of the work you've already done - it's the foundation needed for us to make these changes 🙏 Thanks!!!

As soon as the interfaces look good we'll get this merged - any further nits and/or cleanup of the implemenatation details of any of the GPU/Accel device code can happen in follow up PRs.

Hold tight. I'm hoping we'll have something to look at soon

cmd/exporter/exporter.go
maryamtahhan: Yeah this should be changed. I will change it

Makefile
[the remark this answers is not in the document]
maryamtahhan: Someone could install all the libraries on the system I guess. I think it would be rather unconventional as it's unlikely that you would find an NVIDIA GPU plugged into a system with a Habana CPU. 
If we think we need to cater for all three we can make the build flag build up all the devices together

sthaha:
> @maryamtahhan - @sthaha and I had a chat about the path forward for this PR.
> 
> @sthaha is going to take a pass over the proposed interfaces and make some changes:
> 

@dave-tucker @maryamtahhan since 0.7.11 is released, lets go ahead and merge this PR. I will work on refactor next week (or week after) in a different PR.

maryamtahhan:
> > @maryamtahhan - @sthaha and I had a chat about the path forward for this PR.
> > 
> > @sthaha is going to take a pass over the proposed interfaces and make some changes:
> > 
> 
> @dave-tucker @maryamtahhan since 0.7.11 is released, lets go ahead and merge this PR. I will work on refactor next week (or week after) in a different PR. 
> 
> 

@sthaha thanks. Just doing some minor cleanup and will fix the linting issue. Will comment again when that's done .

maryamtahhan:
Closing to reduce workflow noise - will reopen shortly

maryamtahhan:
Cleanup + rebase complete

maryamtahhan:
Moved to https://github.com/sustainable-computing-io/kepler/pull/1613 as I couldn't re-open this PR

PR #1559, opened 2024-06-20
fix: Fix uncomment of YAML in hack/build-manifests.sh

dave-tucker:
When we formatted the code with yamllint it was not immediately clear that we had a shell script that was performing uncommenting of lines.

When done by a human, their IDE (or eyes) would see that the indentation didn't align and would adjust. This is not clear to machines, for example when performing `make build-manifests`

This commit adds spaces to commented YAML lines in any kustomization.yaml to ensure that once the leading '#` has been removed that they are correctly indented.

In addition the uncomment() function in hack/build-manifests.sh has been adjusted to preserve any spaces that occur after the leading `#`

### DIFF
--- a/hack/build-manifest.sh
+++ b/hack/build-manifest.sh
@@ -79,7 +79,7 @@ uncomment_path() {
 uncomment() {
 	regex="${1:?}"
 	file="${2:?}"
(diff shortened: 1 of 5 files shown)


husky-parul (approved):
LGTM

PR #1560, opened 2024-06-20
fix: Bash functions should exit on error

dave-tucker:
The run command wrapper was silently ignoring errors. This commit ensures that a failed command will exit, ensuring that we don't accidentally ignore errors anymore.

### DIFF
--- a/hack/utils.bash
+++ b/hack/utils.bash
@@ -64,7 +64,7 @@ info_run() {
 
 run() {
 	echo -e " ❯ $*\n" >&2
-	"$@"
+	"$@" || die "Failed to run: $*"
 }
 
 die() {


hack/utils.bash
```diff
@@ -64,7 +64,7 @@ info_run() {
 
 run() {
 	echo -e " ❯ $*\n" >&2
-	"$@"
+	"$@" || die "Failed to run: $*"
```
sthaha, line 67: This shouldn't be necessary as we should have `set -e` if isn't already.

sthaha (changes requested):
Run shouldn't fail rather the script including `utils` should have `set -eu -o pipefail` as default.

dave-tucker:
> Run shouldn't fail rather the script including `utils` should have `set -eu -o pipefail` as default.

That's not what was observed in #1559 
You can easily reproduce by applying this diff:
```
diff --git a/manifests/k8s/config/exporter/kustomization.yaml b/manifests/k8s/config/exporter/kustomization.yaml
index 55514604..ffcbfa12 100644
--- a/manifests/k8s/config/exporter/kustomization.yaml
+++ b/manifests/k8s/config/exporter/kustomization.yaml
@@ -3,7 +3,7 @@ resources:
   # - openshift_scc.yaml
   - exporter.yaml
 #  uncomment this line if prometheus deployed
-#  - prometheus_common_service_monitor.yaml
+# - prometheus_common_service_monitor.yaml
 #  - prometheus_common_rules.yaml
 #  - prometheus_high_granularity_rules.yaml
```

Then running `make build-manifest OPTS="CI_DEPLOY PROMETHEUS_DEPLOY"`
This will make the `run kustomize edit ...` commands exit with error code 1... but somehow that error is being ignored 🤔 

In other words, applying `set -eu -o pipefail` (which is present in hack/build-manifests.sh) doesn't appear to be operating as expected.

sthaha:
My main objection to this patch is that with `die` being called from `run` the script can exit at any point. The only purpose of run is to print the command it runs and nothing more. 

So ... as a quick test, I removed all `run` from hack/build-manifest.sh` function `build_manifest` and then ran the script and it exhibits the same erroneous behaviour.  

```
	build_manifest || {
		fail "Fail to build the manifests"
		return 1
	}
```

Doing a lot more than what is intended. Please allow me some time to look into this.

sthaha:
@dave-tucker I guess the fix really is to just not use any `if` or 'fn || fail` construct in `main` . IOW, 

```
main() {
	export PATH="$PROJECT_ROOT/tmp/bin:$PATH"

	local opts=$1
	shift
	ensure_all_tools

	for opt in ${opts}; do
		info "Setting $opt as True"
		eval "$opt=true"
	done

	info "move to untrack workspace ${MANIFESTS_OUT_DIR}"
	run rm -rf "${MANIFESTS_OUT_DIR}"
	run mkdir -p "${MANIFESTS_OUT_DIR}"
	run cp -r manifests/k8s/config/* "${MANIFESTS_OUT_DIR}"/
	build_manifest
}
```
should fix it

PR #1561, opened 2024-06-20
path fix for reading cpus.yaml file bug when building locally

gdozortsev:
Running "make build" was throwing an "unable to read cpus.yaml" error because the path to cpus.yaml relied on a directory created only when the image is built. 
Fix: added a step to check local location of cpus.yaml in addition to checking /var/lib/kepler


pkg/collector/stats/utils.go
```diff
@@ -224,8 +224,16 @@ func getS390xArchitecture() (string, error) {
 func getCPUMicroArchitecture(family, model, stepping string) (string, error) {
 	yamlBytes, err := os.ReadFile(CPUModelDataPath)
 	if err != nil {
-		klog.Errorf("failed to read cpus.yaml: %v", err)
-		return "", err
+		currentFilePath, err := os.Getwd()
+		if err != nil {
+			return "", err
+		}
+		LocalCPUModelDataPath := strings.SplitAfter(currentFilePath, "kepler")[0]
```
dave-tucker, line 231:
```suggestion
		localCPUModelDataPath := strings.SplitAfter(currentFilePath, "kepler")[0]
```
Because an uppercase variable name signals that this is part of the public API of a module

pkg/collector/stats/utils.go
```diff
@@ -224,8 +224,16 @@ func getS390xArchitecture() (string, error) {
...
+		yamlBytes, err = os.ReadFile(LocalCPUModelDataPath + "/data/cpus.yaml")
```
dave-tucker, line 232:
```suggestion
		yamlBytes, err = os.ReadFile(localCPUModelDataPath + "/data/cpus.yaml")	
```

pkg/collector/stats/utils.go
```diff
@@ -224,8 +224,16 @@ func getS390xArchitecture() (string, error) {
...
-		klog.Errorf("failed to read cpus.yaml: %v", err)
-		return "", err
+		currentFilePath, err := os.Getwd()
+		if err != nil {
+			return "", err
+		}
+		LocalCPUModelDataPath := strings.SplitAfter(currentFilePath, "kepler")[0]
+		yamlBytes, err = os.ReadFile(LocalCPUModelDataPath + "/data/cpus.yaml")	
+		if err != nil {
+			klog.Errorf("failed to read cpus.yaml: %v", err)
```
dave-tucker, line 234:
```suggestion
```
It's better to return the error to the caller without logging (this was an issue in the existing code). That way the caller can choose to log it, or not.

dave-tucker (changes requested):
Great work @gdozortsev! You'll need to sign your commits.
`git commit --amend --signoff`
[This guide](https://docs.pi-hole.net/guides/github/how-to-signoff/) explains the process.

pkg/collector/stats/utils.go
```diff
@@ -224,8 +224,15 @@ func getS390xArchitecture() (string, error) {
...
-		return "", err
+		currentFilePath, err := os.Getwd()
+		if err != nil {
+			return "", err
+		}
+		localCPUModelDataPath := strings.SplitAfter(currentFilePath, "kepler")[0]
+		yamlBytes, err = os.ReadFile(localCPUModelDataPath + "/data/cpus.yaml")	
+		if err != nil {
+			return "", err
+		}
```
sthaha, line 235: isn't this same as `osReadFile('./data/cpus.yaml') ?

gdozortsev: I tried this first, it still produced the same error as in the original

dave-tucker:
@gdozortsev almost there! Please use `git commit --amend` to edit the commit message.
It's currently showing as many commit messages combined into one.

sthaha (changes requested):
It seems the pr changes a lot more than fixing the path to cpus.yaml. Could you please try rebasing ?

.github/workflows/image.yml
```diff
@@ -51,12 +51,20 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4.1.1
+        with:
+          # NOTE:  setting fetch-depth to 0 to retrieve the entire history
+          # instead of a shallow-clone so that all tags are fetched as well.
+          # This is necessary for computing the VERSION using `git describe`
+          fetch-depth: 0
+
```
sthaha, line 59: Is this a rebase error?

pkg/collector/stats/utils.go
```diff
@@ -224,8 +224,15 @@ func getS390xArchitecture() (string, error) {
...
+		currentFilePath, err := os.Getwd()
+		if err != nil {
+			return "", err
+		}
+		localCPUModelDataPath := strings.SplitAfter(currentFilePath, "kepler")[0]
+		yamlBytes, err = os.ReadFile(localCPUModelDataPath + "/data/cpus.yaml")
+		if err != nil {
+			return "", err
+		}
 	}
```
sthaha, line 236: @gdozortsev I tested this locally and works fine 
```
func readCPUModelData() ([]byte, error) {
	data, err := os.ReadFile(CPUModelDataPath)
	if err == nil {
		return data, nil
	}
	if !os.IsNotExist(err) {
		return nil, err
	}

	klog.Infof("reading local cpus.yaml")
	return os.ReadFile("./data/cpus.yaml")
}

func getCPUMicroArchitecture(family, model, stepping string) (string, error) {
	yamlBytes, err := readCPUModelData()
```

Tested using ... 
```
make build && sudo ./_output/bin/kepler 2>&1
```

sthaha: @gdozortsev could you please try if this works - https://github.com/sustainable-computing-io/kepler/pull/1572/files

PR #1565, opened 2024-06-25
chore: Flatten Uint64StatCollection

dave-tucker:
This commit fixes a couple of small issues.
Firstly, in Go maps are reference types (like pointers) so
using *map[string]any for example isn't required.
Secondly, this removes the structure that backs
Uint64StatCollection and replaces it with a map.
Doing so make the code much easier to read (IMHO).

Finally, operations in Uint64Stat have been made
atomic to ensure correct operation if methods are
called from different threads. A test has been added
to verify this behaviour - this test fails when
run against the main branch.

There are some additional small fixes to address lint issues
that appear to have eluded our CI. Firstly the `dupl` threshold
has been moved to 150 to stop it complaining about Gingko-style
tests (which are very similar). Secondly the YAML for the `shadow`
has been fixed and issues with variable shadowing also fixed.

Below are the benchmark results vs. the implementaton in main.
``` console
$ benchstat ../main.txt ../uintflat.txt
goos: linux
goarch: amd64
pkg: github.com/sustainable-computing-io/kepler/pkg/collector/stats/types
cpu: AMD Ryzen 5 5600G with Radeon Graphics         
                                                      │ ../main.txt  │            ../uintflat.txt            │
                                                      │    sec/op    │    sec/op      vs base                │
UInt64StatCollectionSetAggrStatMissed-12                480.0n ± 20%    472.8n ± 17%        ~ (p=0.739 n=10)
UInt64StatCollectionSetAggrStatBothRandom-12            478.7n ±  2%    458.1n ±  3%   -4.28% (p=0.002 n=10)
UInt64StatCollectionSetAggrStatCachedRandomNumber-12    13.15n ±  1%    13.90n ±  0%   +5.71% (p=0.000 n=10)
UInt64StatCollectionSetAggrStatCached-12                5.301n ±  1%    5.609n ±  3%   +5.81% (p=0.000 n=10)
UInt64StatCollectionAddDeltaStatMissed-12               485.6n ±  3%    467.1n ±  8%        ~ (p=0.197 n=10)
UInt64StatCollectionAddDeltaStatBothRandom-12           478.2n ±  5%    400.8n ± 18%  -16.20% (p=0.009 n=10)
UInt64StatCollectionAddDeltaStatCachedRandomNumber-12   9.995n ±  1%   10.535n ±  1%   +5.40% (p=0.000 n=10)
UInt64StatCollectionAddDeltaStatCached-12               5.540n ±  1%    6.543n ±  1%  +18.09% (p=0.000 n=10)
UInt64StatCollectionSetDeltaStatMissed-12               487.8n ±  4%    473.0n ±  6%        ~ (p=0.165 n=10)
UInt64StatCollectionSetDeltaStatBothRandom-12           475.4n ±  3%    461.5n ±  3%   -2.91% (p=0.035 n=10)
UInt64StatCollectionSetDeltaStatCachedRandomNumber-12   10.26n ±  1%    10.79n ±  0%   +5.11% (p=0.000 n=10)
UInt64StatCollectionSetDeltaStatCached-12               5.785n ±  0%    6.792n ±  1%  +17.41% (p=0.000 n=10)
geomean                                                 61.34n          62.40n         +1.74%

                                                      │  ../main.txt  │           ../uintflat.txt            │
                                                      │     B/op      │    B/op      vs base                 │
UInt64StatCollectionSetAggrStatMissed-12                163.5 ± 33%     164.5 ± 33%       ~ (p=0.381 n=10)
UInt64StatCollectionSetAggrStatBothRandom-12            168.0 ±  1%     172.0 ±  2%  +2.38% (p=0.000 n=10)
UInt64StatCollectionSetAggrStatCachedRandomNumber-12    0.000 ±  0%     0.000 ±  0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetAggrStatCached-12                0.000 ±  0%     0.000 ±  0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionAddDeltaStatMissed-12               163.0 ±  2%     167.5 ±  4%  +2.76% (p=0.008 n=10)
UInt64StatCollectionAddDeltaStatBothRandom-12           168.0 ±  2%     110.0 ± 56%       ~ (p=0.493 n=10)
UInt64StatCollectionAddDeltaStatCachedRandomNumber-12   0.000 ±  0%     0.000 ±  0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionAddDeltaStatCached-12               0.000 ±  0%     0.000 ±  0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetDeltaStatMissed-12               164.0 ±  3%     167.0 ±  3%       ~ (p=0.098 n=10)
UInt64StatCollectionSetDeltaStatBothRandom-12           170.0 ±  1%     173.0 ±  1%  +1.76% (p=0.017 n=10)
UInt64StatCollectionSetDeltaStatCachedRandomNumber-12   0.000 ±  0%     0.000 ±  0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetDeltaStatCached-12               0.000 ±  0%     0.000 ±  0%       ~ (p=1.000 n=10) ¹
geomean                                                             ²                -2.72%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                                                      │ ../main.txt  │           ../uintflat.txt           │
                                                      │  allocs/op   │ allocs/op   vs base                 │
UInt64StatCollectionSetAggrStatMissed-12                2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetAggrStatBothRandom-12            2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetAggrStatCachedRandomNumber-12    0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetAggrStatCached-12                0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionAddDeltaStatMissed-12               2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionAddDeltaStatBothRandom-12           2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionAddDeltaStatCachedRandomNumber-12   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionAddDeltaStatCached-12               0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetDeltaStatMissed-12               2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetDeltaStatBothRandom-12           2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetDeltaStatCachedRandomNumber-12   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
UInt64StatCollectionSetDeltaStatCached-12               0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                            ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

Note there is no significant performance difference between these 2 implementations.

### DIFF
--- a/pkg/collector/stats/types/types.go
+++ b/pkg/collector/stats/types/types.go
@@ -3,22 +3,30 @@ package types
 import (
 	"fmt"
 	"math"
+	"sync/atomic"
 
 	"k8s.io/klog/v2"
 )
 
 type UInt64Stat struct {
-	Aggr  uint64
-	Delta uint64
+	aggr  atomic.Uint64
+	delta atomic.Uint64
 }
 
-func (s UInt64Stat) String() string {
-	return fmt.Sprintf("%d (%d)", s.Delta, s.Aggr)
+func NewUInt64Stat(aggr, delta uint64) *UInt64Stat {
+	stat := &UInt64Stat{}
+	stat.aggr.Store(aggr)
+	stat.delta.Store(delta)
+	return stat
+}
+
+func (s *UInt64Stat) String() string {
+	return fmt.Sprintf("%d (%d)", s.delta.Load(), s.aggr.Load())
 }
 
 // ResetDelta resets current value
 func (s *UInt64Stat) ResetDeltaValues() {
-	s.Delta = uint64(0)
+	s.delta.Swap(uint64(0))
 }
 
 // AddNewDelta add a new read (process) delta value (e.g., from bpf table that is reset, computed delta energy)
@@ -38,65 +46,59 @@ func (s *UInt64Stat) SetNewDeltaValue(newDelta uint64, sum bool) error {
 		return nil
 	}
 	// verify overflow
-	if s.Aggr >= (math.MaxUint64 - newDelta) {
-		// we must set the the value to 0 when overflow, so that prometheus will handle it
-		s.Aggr = 0
+	if s.aggr.Load() >= (math.MaxUint64 - newDelta) {
+		// we must set the value to 0 when overflow, so that prometheus will handle it
+		s.aggr.Swap(0)
 		return fmt.Errorf("the aggregated value has overflowed")
 	}
 	if sum {
 		// sum is used to accumulate metrics from different processes
-		s.Delta += newDelta
+		s.delta.Add(newDelta)
 	} else {
-		s.Delta = newDelta
+		s.delta.Swap(newDelta)
 	}
-	s.Aggr += newDelta
+	// update aggregated value
+	s.aggr.Add(newDelta)
 	return nil
 }
 
 // SetNewAggr set new read aggregated value (e.g., from cgroup, energy files)
 func (s *UInt64Stat) SetNewAggr(newAggr uint64) error {
-	if newAggr == 0 || newAggr == s.Aggr {
+	currAggr := s.aggr.Load()
+	if newAggr == 0 || newAggr == currAggr {
 		// if a counter has not changed, we skip it
 		return nil
 	}
 	// verify aggregated value overflow
 	if newAggr == math.MaxUint64 {
-		// we must set the the value to 0 when overflow, so that prometheus will handle it
-		s.Aggr = 0
+		// we must set the value to 0 when overflow, so that prometheus will handle it
+		s.aggr.Swap(0)
 		return fmt.Errorf("the aggregated value has overflowed")
 	}
-	if (s.Aggr > 0) && (newAggr > s.Aggr) {
-		s.Delta = newAggr - s.Aggr
+	if (currAggr > 0) && (newAggr > currAggr) {
+		s.delta.Swap(newAggr - currAggr)
 	}
-	s.Aggr = newAggr
+	s.aggr.Swap(newAggr)
 	return nil
 }
 
 func (s *UInt64Stat) GetDelta() uint64 {
-	return s.Delta
+	return s.delta.Load()
 }
 
 func (s *UInt64Stat) GetAggr() uint64 {
-	return s.Aggr
+	return s.aggr.Load()
 }
 
-func NewUInt64StatCollection() *UInt64StatCollection {
-	return &UInt64StatCollection{
-		Stat: make(map[string]*UInt64Stat),
-	}
+func NewUInt64StatCollection() UInt64StatCollection {
+	return make(map[string]*UInt64Stat)
 }
 
-// UInt64StatCollection keeps a collection of UInt64Stat
-type UInt64StatCollection struct {
-	Stat map[string]*UInt64Stat
-}
+type UInt64StatCollection map[string]*UInt64Stat
 
-func (s *UInt64StatCollection) SetAggrStat(key string, newAggr uint64) {
-	if instance, found := s.Stat[key]; !found {
-		s.Stat[key] = &UInt64Stat{
-			Aggr:  newAggr,
-			Delta: 0,
-		}
+func (s UInt64StatCollection) SetAggrStat(key string, newAggr uint64) {
+	if instance, found := s[key]; !found {
+		s[key] = NewUInt64Stat(newAggr, 0)
 	} else {
 		if err := instance.SetNewAggr(newAggr); err != nil {
 			klog.V(3).Infoln(err)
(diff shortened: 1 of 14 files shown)


pkg/collector/stats/types/types.go
```diff
@@ -38,90 +46,92 @@ func (s *UInt64Stat) SetNewDeltaValue(newDelta uint64, sum bool) error {
...
-	if instance, found := s.Stat[key]; !found {
-		s.Stat[key] = &UInt64Stat{
-			Aggr:  newAggr,
-			Delta: 0,
-		}
+func (s UInt64StatCollection) SetAggrStat(key string, newAggr uint64) {
+	if instance, found := s[key]; !found {
+		stat := &UInt64Stat{}
+		stat.aggr.Store(newAggr)
+		s[key] = stat
```
sthaha, line 109:
```suggestion
		s[key] = NewUInt64Stat(newAggr, 0)
```

pkg/collector/stats/types/types.go
```diff
@@ -38,90 +46,92 @@ func (s *UInt64Stat) SetNewDeltaValue(newDelta uint64, sum bool) error {
...
-		s.Stat[key] = &UInt64Stat{
-			Aggr:  newDelta,
-			Delta: newDelta,
-		}
+func (s UInt64StatCollection) AddDeltaStat(key string, newDelta uint64) {
+	if instance, found := s[key]; !found {
+		stat := &UInt64Stat{}
+		stat.aggr.Store(newDelta)
+		stat.delta.Store(newDelta)
+		s[key] = stat
```
sthaha, line 122: How about we use the constructors / NewX instead of handling the internal datatypes (now not exported) ? 

```suggestion
		s[key] = NewUnit64Stat(newDelta, newDelta)
```

pkg/collector/stats/types/types.go
```diff
@@ -38,90 +46,92 @@ func (s *UInt64Stat) SetNewDeltaValue(newDelta uint64, sum bool) error {
...
 		return fmt.Errorf("the aggregated value has overflowed")
 	}
 	if sum {
 		// sum is used to accumulate metrics from different processes
-		s.Delta += newDelta
+		s.delta.Add(newDelta)
 	} else {
-		s.Delta = newDelta
+		if swapped := s.delta.CompareAndSwap(s.delta.Load(), newDelta); !swapped {
+			return fmt.Errorf("the delta value has been changed by another process")
```
sthaha, line 59: How about using just `Swap` ? 
```suggestion
		s.delta.Swap(newDelta)
```
Or use a loop to until it is swapped ? 
E.g. see: 

https://github.com/prometheus/client_golang/blob/bccd68204bf41e9cce27e8561b70e0d1cb350bcb/prometheus/counter.go#L137-L143

pkg/collector/stats/types/types.go
```diff
@@ -38,90 +46,92 @@ func (s *UInt64Stat) SetNewDeltaValue(newDelta uint64, sum bool) error {
...
 func (s *UInt64Stat) SetNewAggr(newAggr uint64) error {
-	if newAggr == 0 || newAggr == s.Aggr {
+	currAggr := s.aggr.Load()
+	if newAggr == 0 || newAggr == currAggr {
 		// if a counter has not changed, we skip it
 		return nil
 	}
 	// verify aggregated value overflow
 	if newAggr == math.MaxUint64 {
 		// we must set the the value to 0 when overflow, so that prometheus will handle it
```
sthaha, line 76:
```suggestion
		// we must set the value to 0 when overflow, so that prometheus will handle it
```

pkg/collector/stats/types/types.go
```diff
@@ -38,90 +46,92 @@ func (s *UInt64Stat) SetNewDeltaValue(newDelta uint64, sum bool) error {
...
-		s.Aggr = 0
+		s.aggr.Store(0)
```
sthaha, line 77: shouldn't we also reset `delta`?

pkg/collector/stats/types/types.go
```diff
@@ -130,25 +140,25 @@ func (s *UInt64StatCollection) SetDeltaStat(key string, newDelta uint64) {
 }
 
 // SumAllDeltaValues aggregates the delta metrics of all sources (i.e., stat keys)
-func (s *UInt64StatCollection) SumAllDeltaValues() uint64 {
+func (s UInt64StatCollection) SumAllDeltaValues() uint64 {
 	sum := uint64(0)
```
sthaha, line 144: Unrelated, but caught my eye .. wouldn't this overflow as well? 
Shouldn't we use https://pkg.go.dev/math/big here?

sthaha (commented):
Looks good overall, some points to consider.

pkg/collector/stats/types/types.go
dave-tucker: I'm not entirely sure we should be resetting at all.
I know the expectation from prometheus's side is that counters get reset to 0 on restart.
I would think it makes sense to pick the right precision of counter in the first place 🤔

sthaha: Note necessarily reset to 0 because it can't scrape exactly when the reset to 0 happened. 
AFAIK, the expectation from Prometheus is that counters are monotonic and are ever increasing, so if the previous scrape is greater than current then, `rate | sum| etc..` would assume there was a reset in between. 
see: https://www.robustperception.io/how-does-a-prometheus-counter-work/#:~:text=A%20common%20question%20is%20what%20happens%20when%20the%20process%20restarts%20and%20the%20counter%20is%20reset%20to%200%3F%20rate()%C2%A0will%C2%A0automatically%20handle%20this.%20Any%20time%20a%20counter%20appears%20to%20decrease%20it%27ll%20be%20treated%20as%20though%20there%20was%20a%20reset%20to%200%20right%20after%20the%20first%20data%20point.

rootfs:
Per 07/02 meeting, @sunya-ch will take a final check and plan it for the next release.

sthaha:
@dave-tucker could you please rebase this pr so that we can include it in 0.7.11 ?

pkg/collector/stats/types/types.go
dave-tucker: Tracking this discussion here:
https://github.com/sustainable-computing-io/kepler/issues/1594

dave-tucker:
@sthaha done! and opened issues to investigate some of the outstanding comments on this code

sunya-ch:
@sthaha when you approve, I think we are good to have this PR merge.

sthaha (approved):

PR #1570, opened 2024-06-26
feat(validator): Add workflow for validator tests

vprashar2929:
This PR introduces a workflow to execute validator tests on CI Additionally, it includes a Makefile within `e2e/tools/validator`, enabling the execution of targets such as `make fmt` and `make test` for local testing

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -26,6 +26,9 @@ jobs:
   # for each PR run integration test
   integration_test:
     uses: ./.github/workflows/integration_test.yml
+  # for each PR run validator test
+  validator:
+    uses: ./.github/workflows/validator.yml
 
   changes_image:
     runs-on: ubuntu-latest
--- a/.github/workflows/validator.yml
+++ b/.github/workflows/validator.yml
@@ -1,0 +1,37 @@
+name: Validator tests
+
+on: # yamllint disable-line rule:truthy
+  workflow_call:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      # checkout source code
+      - name: checkout source
+        uses: actions/checkout@v4.1.1
+
+      # setup Python environment
+      - name: set up Python
+        uses: actions/setup-python@v5.1.0
+        with:
+          python-version: "3.11"
+
+      # install hatch
+      - name: install hatch
+        run: |
+          python -m pip install --upgrade pip
+          pip install hatch
+
+      # run the formatter using hatch
+      - name: run formatter with hatch
+        run: |
+          cd ./e2e/tools/validator
+          make fmt
+          git diff --exit-code
+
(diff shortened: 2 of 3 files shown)


vprashar2929:
Sample CI run: https://github.com/sustainable-computing-io/kepler/actions/runs/9683848586/job/26720262220?pr=1570

Makefile
```diff
@@ -286,6 +286,13 @@ unit-test: ginkgo-set tidy-vendor ## Run unit tests.
 		--race --count=1 \
 		$(TEST_PKGS)
 
+.PHONY: validator-test
+validator-test: ## Run validator tests.
+	@echo "Running validator tests..."
+	@cd ./e2e/tools/validator && \
+	hatch env create && \
+	hatch run test || { echo 'validator-test failed'; exit 1; }
```
sthaha, line 294:
```suggestion
	hatch run test
```

.github/workflows/validator_test.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      # checkout source code
+      - name: checkout source
+        uses: actions/checkout@v4.1.1
+      # setup Python environment
+      - name: Set up Python
+        uses: actions/setup-python@v5.1.0
+        with:
+          python-version: "3.12"
```
sthaha, line 17:
```suggestion
          python-version: "3.11"
```

Makefile
```diff
@@ -286,6 +286,13 @@ unit-test: ginkgo-set tidy-vendor ## Run unit tests.
 		--race --count=1 \
 		$(TEST_PKGS)
 
+.PHONY: validator-test
+validator-test: ## Run validator tests.
+	@echo "Running validator tests..."
+	@cd ./e2e/tools/validator && \
+	hatch env create && \
```
sthaha, line 293: I think it is better to have a `validator-env` target that `test` depends on

.github/workflows/validator_test.yml
vprashar2929: Do we have any hard requirements on `3.11` ?

sthaha: Indeed we do :( .. rhel 9

sthaha: I have pinned the python version in #1573

KaiyiLiu1234:
This is great to add to the github actions workflow. The unit tests however need some more work last time I checked. Maybe we can improve it? @vprashar2929 I can raise it as an issue and hop on that if you would like.

Makefile
sthaha: @vprashar2929 , gentle reminder for a separate makefile for Validator :) but actually think `hatch` makes it really unnecessary to have a Makefile

sthaha (changes requested):

sthaha (approved):

PR #1572, opened 2024-06-27
fix: fallback to reading cpus.yaml relative to current dir

sthaha:
This commit fixes the issue of having to copy cpus.yaml to /var/lib/kepler (the hardcoded path for the cpus.yaml) by reading the cpus.yaml file relative to the project root.

Thus the following now works fine.
```
make build
sudo ./_output/bin/kepler
```


gdozortsev:
Tried this locally and it works now

dave-tucker (approved):

PR #1589, opened 2024-07-03
Remove the libbpf and darwin build tags

dave-tucker:
This series of commits removes the `darwin` and `libbpf` build tags.

This is because:

1. Kepler doesn't work on `macOS` so building for `darwin` isn't required
1. Since we removed the `bcc` code, there is no need to gate libbpf support behind a build tag

Doing this means that `golangci-lint` actually gets run on portions of the codebase that were previously skipped due to build tags 🎉 The last 2 commits in this PR fix those lint issues for the `bpf` and `e2e` packages.

Updates: #1403


.github/workflows/c.yml
```diff
@@ -21,7 +21,16 @@ jobs:
...
-          sudo apt-get update && sudo apt-get install -y libelf-dev libbpf-dev
+          sudo apt remove libbpf-dev
+          sudo apt install libelf-dev
+          mkdir temp-libbpf
+          cd temp-libbpf
+          git clone https://github.com/libbpf/libbpf
+          cd libbpf/src
+          sudo make install_headers
+          sudo make install_uapi_headers
+          sudo prefix=/usr  BUILD_STATIC_ONLY=y make install
```
sthaha, line 34: shouldn't we checkout a particular version instead of main?

sthaha (approved):

.github/workflows/c.yml
dave-tucker: we checkout main in the unit test code and I just copied that here.
this is only temporary until #1438 gets in when it will be removed.
I just wanted to get the build tags removed first to avoid bloating the cilium-ebpf PR more.

PR #1590, opened 2024-07-03
chore(validator): use nearest timestamps to align prom queries

rootfs:
If the two prom queries are not fully synchronized, we should choose to use the nearest timestamps.


rootfs:
closing this one. The root cause of poor MAPE is fixed in #1592

PR #1592, opened 2024-07-03
chore(collector): convert cpu time in collection time instead of metrics reporting time to avoid inconsistent use of cpu time in models

rootfs:
The bad MAPE results in latest code (as in contrast to 0.7.8) is related to the inconsistent usage of cpu time. The models trained using early 0.7 code uses [microsecond cpu time ](https://github.com/sustainable-computing-io/kepler/blob/release-0.7.8/bpfassets/libbpf/src/kepler.bpf.c#L68). #1481  [changed it](https://github.com/sustainable-computing-io/kepler/pull/1481/files#diff-0916c401e16b7d26ca96720a88f24fe4fa5117113d56e05b85b515c9b23cea73R109) and compensated the delta at metrics reporting time. This causes the model to use wrong CPU time.

I tested the fix on my env:
After the change:
```
kepler_node_core_joules_total{instance="rhel",mode="dynamic",package="0",source="trained_power_model"} 87.354
kepler_node_core_joules_total{instance="rhel",mode="idle",package="0",source="trained_power_model"} 1468.257
```

Before the change (aka main branch)
```
kepler_node_core_joules_total{instance="rhel",mode="dynamic",package="0",source="trained_power_model"} 96386.637
kepler_node_core_joules_total{instance="rhel",mode="idle",package="0",source="trained_power_model"} 1678.008
```

### DIFF
--- a/pkg/collector/resourceutilization/bpf/process_bpf_collector.go
+++ b/pkg/collector/resourceutilization/bpf/process_bpf_collector.go
@@ -39,7 +39,7 @@ func updateSWCounters(key uint64, ct *ProcessBPFMetrics, processStats map[uint64
 	for counterKey := range bpfSupportedMetrics.SoftwareCounters {
 		switch counterKey {
 		case config.CPUTime:
-			processStats[key].ResourceUsage[config.CPUTime].AddDeltaStat(utils.GenericSocketID, ct.ProcessRunTime)
+			processStats[key].ResourceUsage[config.CPUTime].AddDeltaStat(utils.GenericSocketID, ct.ProcessRunTime/1000 /* convert microseconds to miliseconds */)
 		case config.PageCacheHit:
 			processStats[key].ResourceUsage[config.PageCacheHit].AddDeltaStat(utils.GenericSocketID, ct.PageCacheHit/(1000*1000))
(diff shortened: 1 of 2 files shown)


rootfs:
@sthaha can you evaluate it for 0.7.11 release? Thanks

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
```diff
@@ -32,14 +32,19 @@ import (
 
 type ProcessBPFMetrics = bpf.ProcessBPFMetrics
 
+const (
+	// convert microseconds to miliseconds
+	MicrosecondsToMiliseconds = 1000
```
dave-tucker, line 37: This shouldn't be package global as its not used anywhere else. Given its only used once I would simply divide by 1000 and put a comment in the line above explaining what you are converting and why

dave-tucker (changes requested):

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
rootfs: done

sthaha (approved):

PR #1605, opened 2024-07-09
fix: Makefile for running on macOS

dave-tucker:
A few fixes to more easily run Kepler in a KIND cluster when running make on a macOS system:

- Consistently indent using tabs
- Check for the presence of ldconfig to avoid errors
- export variables required by the cluster-* scripts
- deploy the serivcemonitors for prometheus by default

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ IMAGE_REPO         ?= quay.io/sustainable_computing_io
 BUILDER_IMAGE      ?= quay.io/sustainable_computing_io/kepler_builder:ubi-9-libbpf-1.3.0
 IMAGE_NAME         ?= kepler
 IMAGE_TAG          ?= latest
-CTR_CMD            ?= $(or $(shell podman info > /dev/null 2>&1 && which podman), $(shell docker info > /dev/null 2>&1 && which docker))
+export CTR_CMD     ?= $(or $(shell command -v podman), $(shell command -v docker))
 
 # use CTR_CMD_PUSH_OPTIONS to add options to <container-runtime> push command.
 # E.g. --tls-verify=false for local develop when using podman
@@ -43,7 +43,7 @@ CTR_CMD_PUSH_OPTIONS ?=
 
 GENERAL_TAGS := 'include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo '
 GPU_TAGS := ' gpu '
-ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
+ifeq ($(shell command -v ldconfig && ldconfig -p | grep -q libhlml.so && echo exists),exists)
 	GPU_TAGS := $(GPU_TAGS)'habana '
 endif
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -400,31 +402,18 @@ cluster-clean: build-manifest ## Undeploy Kepler in the cluster.
 .PHONY: cluster-clean
 
 cluster-deploy: ## Deploy Kepler in the cluster.
-	BUILD_CONTAINERIZED=$(BUILD_CONTAINERIZED) \
 	./hack/cluster-deploy.sh
 .PHONY: cluster-deploy
 
 cluster-up:  ## Create the Kind cluster, with Prometheus, Grafana and Kepler
-	CLUSTER_PROVIDER=$(CLUSTER_PROVIDER) \
```
maryamtahhan, line 408: Are we happy that there are reasonable defaults configured for all these params in `./hack/cluster.sh`
in my workflow I run a lot of these make targets directly, is it expected that I now set these variables when I run the make target?

maryamtahhan: nvm I see the export above. LGTM

sthaha (changes requested):
LGTM! But `requesting-change` since this requires changes to be tested manually on both Linux and Osx .

cc: @vprashar2929

Makefile
```diff
@@ -341,19 +341,21 @@ build-manifest: kustomize
 .PHONY: build-manifest
 
 ##@ Development Env
-CLUSTER_PROVIDER ?= kind
-LOCAL_DEV_CLUSTER_VERSION ?= main
-
-KIND_WORKER_NODES ?=2
-BUILD_CONTAINERIZED ?= build_image
+export CLUSTER_PROVIDER ?= kind
```
vprashar2929, line 344: I think for local dev we should stick container runtime to docker as with podman it won't work as expected. For instance while doing `make cluster_up` as of now it will fail due to the unavailability of `kind` network since podman doesn't create a network automatically if not present, unlike [docker](https://github.com/sustainable-computing-io/local-dev-cluster/blob/a3aaeb684fc94f6c4d167b18d27c5aa8df8f810d/providers/kind/kind.sh#L103).

dave-tucker: I tested this on my mac with podman desktop and it worked 🤔 maybe something has changed there as I saw the failure you described in CI.

vprashar2929: That's strange 🤔 Is the `kind` network already present if you do `podman network ls`?

dave-tucker:
```console
$ podman network ls
NETWORK ID    NAME                     DRIVER
24edfc3079df  kind                     bridge
```

I didn't create that network

sthaha: What we (Vibhu & I) observed is that the kind create fails the first time. It however creates `kind` network and fails. Re running `kind create` succeeds the second time

sthaha (approved):

Makefile
vprashar2929: @dave-tucker I tried on a fresh fedora using the latest podman and kind. `make cluster-up` failed due to the same issue `network kind was not found` 
```
 ❯ /usr/bin/podman run -d --restart=always --name kind-registry -p 127.0.0.1:5001:5000 --network kind registry:2

Resolved "registry" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/registry:2...
Getting image source signatures
Copying blob bc8f2b8a18ff done   |
Copying blob d49090716641 done   |
Copying blob ad02dd2076d6 done   |
Copying blob 9d41963883ad done   |
Copying blob 73baa7ef167e done   |
Copying config 6a3edb1d5e done   |
Writing manifest to image destination
Error: unable to find network with name or ID kind: network not found
make: *** [Makefile:411: cluster-up] Error 125

❯ podman version
Client:       Podman Engine
Version:      5.1.1
API Version:  5.1.1
Go Version:   go1.22.3
Built:        Tue Jun  4 05:30:00 2024
OS/Arch:      linux/amd64

❯ podman network ls
NETWORK ID    NAME        DRIVER
2f259bab93aa  podman      bridge

❯ kind version
kind v0.23.0 go1.21.10 linux/amd64

❯ podman network ls
NETWORK ID    NAME        DRIVER
2f259bab93aa  podman      bridge
```

PR #1608, opened 2024-07-09
feat(validator): Add Health Checks and Autoheal to Docker Compose

KaiyiLiu1234:
Added Health Checks and Autoheal for verifiable services like kepler, scaphandre, prometheus, and grafana.

Note for Scaphandre and Kepler Model Server, curl and wget were not installed in the image. For scaphandre, curl was manually installed. If necessary, I can also manually install curl for the Model Server. 

Model Server and Estimator were not included in this implementation as I cannot curl their ports.


build/Dockerfile.scaphandre
```diff
@@ -0,0 +1,8 @@
+# Use the base image
+FROM hubblo/scaphandre:latest
+
+# Install curl
+RUN apt-get update && apt-get install -y curl
+
+# Optionally clean up apt cache
+RUN rm -rf /var/lib/apt/lists/*
```
sthaha, line 8:
```suggestion
# Install curl
RUN apt-get install -y curl && \
         apt-get clean
```
Each run creates a seperate layer so removing files in created by the previous run won't yield smaller images. 

Additionally, lets not run `apt-get update` without a good reason since it affects the time taken to build the container.

sthaha (changes requested):
Requesting change for correcting the scaphandre dockerfile.

manifests/compose/autoheal/compose.yaml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  autoheal:
+    deploy:
+      replicas: 1
+    image: willfarrell/autoheal:latest
+    network_mode: none
+    restart: always
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - /var/run/docker.sock:/var/run/docker.sock
```
sthaha, line 12: @KaiyiLiu1234  , I am not convinced we need autoheal for compose files. 
Too vary of running unverified container and give it docker.sock access.

sthaha (changes requested):
LGTM, except the autoheal (which I missed y'day :( )

build/Dockerfile.scaphandre
vprashar2929, line 1: Shouldn't we keep the Dockerfile inside `manifests/compose/validation` as it is something that is specific to VM validations?

sthaha: good catch!

sthaha: @KaiyiLiu1234 could you please fix this and resubmit?

KaiyiLiu1234: Moved into manifests/compose/validation under dockerfiles.

manifests/compose/validation/dockerfiles/Dockerfile.scaphandre
```diff
@@ -0,0 +1,6 @@
+# Use the scaph base image
+FROM hubblo/scaphandre:latest
+
+# Install only curl
+RUN apt-get update && apt-get install -y curl
```
sthaha, line 5:
```suggestion
RUN apt-get install -y curl
```

sthaha (approved):
LGTM!
Will wait for @vprashar2929 to take a final look

build/Dockerfile.scaphandre
KaiyiLiu1234: I should mention that it is necessary to update. There is a scenario when building the image for the first time that without an update curl fails to install.

vprashar2929:
@KaiyiLiu1234 Shouldn't this fail while running docker compose up
```
services:
  kepler-metal:
    build:
      context: ../../../../
      dockerfile: build/Dockerfile
      args:
        INSTALL_DCGM: false
        INSTALL_HABANA: false

    ports:
      # NOTE: use 9888 to keep the host 8888 port free for any local testing
      # e.g. sudo ./bin/kepler
      - 9188:8881
    privileged: true
    pid: host

    entrypoint: [/usr/bin/bash, -c]

    command:
      - |
        set -x;
        /usr/bin/kepler \
          -address "0.0.0.0:8881" \
          -v "8" \
          -enable-gpu=false
    networks:
      - kepler-network

    healthcheck:
      test: curl -f http://localhost:8888/metrics || exit 1
      interval: ${HEALTHCHECK_INTERVAL:-50s}
      timeout: ${HEALTHCHECK_TIMEOUT:-30s}
      retries: ${HEALTHCHECK_RETRIES:-3}
      start_period: ${HEALTHCHECK_START_PERIOD:-1m}
```
as the test referring to the wrong port?

KaiyiLiu1234:
@vprashar2929 It does fail.
failureproof
Health check will label services as unhealthy if the test fails. autoheal will attempt to restart. I think autoheal might be able to terminate a service when unhealthy.

PR #1614, opened 2024-07-15
chore(compose/grafana): allow anonymous login with admin role

sthaha:
This commit allows grafana to be accessed without logging in as admin user. It also solves the nagging change password issue.


maryamtahhan (approved):

sunya-ch (approved):
/lgtm 
Thanks. Shall we note somewhere that this is for testing purpose?

PR #1620, opened 2024-07-17
resolve security concern with unsigned container images

arthurus-rex:
Follow-up to a security advisory I raised 3 weeks ago about unsigned images within Kepler workflow.  In this PR, I have introduced cosign into the image-building workflow (specifically in `image.yml`).  I tested it successfully on my own registry and then adjusted the URLs.


.github/workflows/image.yml
```diff
@@ -18,6 +18,10 @@ on: # yamllint disable-line rule:truthy
 jobs:
   image_build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
```
sthaha, line 24: mind adding a note why these are needed?

sthaha (changes requested):
Thanks for the patch @arthurus-rex . I am requesting change to rebase the required change on top of the latest main.

.github/workflows/image.yml
arthurus-rex: Yes, the id-token must be set to write in order for the OIDC to be used during the signing.  And, in GitHub Actions, when any permissions are specified, all others are set to none.  So, the other two permissions allow this action to read/write from registry.

arthurus-rex:
I had a serious issue with the rebasing, I will be making a new pull request shortly.

Issue #1625, opened 2024-07-22
Add the possibility to use metadata in libvirt to customize the vm_id of VM metrics


aharivel:
### What would you like to be added?

By default the name chosen to define the Virtual Machine ID is the one found in /proc/<pid>/cgroup/.
Kepler should support the metadata in libvirt in order to get the same name as a the IaaS that spawned the VM. 

### Why is this needed?

 Many IaaS platform like Openstack use metadata in libvirt to customize among other things the instance's ID.

PR #1626, opened 2024-07-22
Libvirt metadata vm_id usage

aharivel:
This should answer the Enhancement #1625  

By default the name chosen to define the Virtual Machine ID is the one found in /proc/\<pid\>/cgroup/.
Many IaaS platform like Openstack use metadata in libvirt to customize among other things the instance's ID.
Allow the user to use the metadata in libvirt to set the name of the vm metrics and gives a better user experience.
    
    - Set "LIBVIRT_METADATA_URI" with the uri of the XML namespace identifier.
    - Set "LIBVIRT_METADATA_TOKEN" to choose the string used for VM ID. By default, the token "name" is used.
 
If no URI is set or if any errors occurs, the default method to set the name is kept.

Before: 
```
# HELP kepler_vm_bpf_block_irq_total Aggregated value in bpf_block_irq value from bpf
# TYPE kepler_vm_bpf_block_irq_total counter
kepler_vm_bpf_block_irq_total{source="bpf",vm_id="machine-qemu-1-fedora-40"} 0
kepler_vm_bpf_block_irq_total{source="bpf",vm_id="machine-qemu-2-alpinelinux3.19"} 0
# HELP kepler_vm_bpf_cpu_time_ms_total Aggregated value in bpf_cpu_time_ms value from bpf
# TYPE kepler_vm_bpf_cpu_time_ms_total counter
kepler_vm_bpf_cpu_time_ms_total{source="bpf",vm_id="machine-qemu-1-fedora-40"} 113
kepler_vm_bpf_cpu_time_ms_total{source="bpf",vm_id="machine-qemu-2-alpinelinux3.19"} 25
[...]

```
After
```
# HELP kepler_vm_bpf_block_irq_total Aggregated value in bpf_block_irq value from bpf
# TYPE kepler_vm_bpf_block_irq_total counter
kepler_vm_bpf_block_irq_total{source="bpf",vm_id="instance-alpine"} 0
kepler_vm_bpf_block_irq_total{source="bpf",vm_id="machine-qemu-1-fedora-40"} 0
# HELP kepler_vm_bpf_cpu_time_ms_total Aggregated value in bpf_cpu_time_ms value from bpf
# TYPE kepler_vm_bpf_cpu_time_ms_total counter
kepler_vm_bpf_cpu_time_ms_total{source="bpf",vm_id="instance-alpine"} 0
kepler_vm_bpf_cpu_time_ms_total{source="bpf",vm_id="machine-qemu-1-fedora-40"} 577
[...]
```

This is obtains with the "instance-alpine" libvirt XML :
```
<domain type='kvm' id='2'>
  <name>alpinelinux3.19</name>
  <uuid>45885a2b-6e3a-4b8a-bfbc-178942gf803</uuid>
  <metadata>
    <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
      <libosinfo:os id="http://alpinelinux.org/alpinelinux/3.19"/>
    </libosinfo:libosinfo>
    <nova:instance xmlns:nova="http://openstack.org/xmlns/libvirt/nova/1.1">
      <nova:package version="27.3.1-18.0.202445915514.dd48561540.el9osttrunk"/>
      <nova:name>instance-alpine</nova:name>
      <nova:creationTime>2024-07-11 16:41:29</nova:creationTime>
      <nova:flavor name="alpine_base_flavor">
        <nova:memory>4096</nova:memory>
        <nova:disk>20</nova:disk>
        <nova:swap>0</nova:swap>
        <nova:ephemeral>0</nova:ephemeral>
        <nova:vcpus>4</nova:vcpus>
      </nova:flavor>
      <nova:owner>
        <nova:user uuid="8abfae245945675313521fa574e94b">admin</nova:user>
        <nova:project uuid="3456f07e078789546318104d29c2868f59">admin</nova:project>
      </nova:owner>
      <nova:root type="image" uuid="56aa78956h45-2856-49e7-ajdk-b748946"/>
      <nova:ports>
        <nova:port uuid="78895418-727d-4dac-86c8-382345456">
          <nova:ip type="fixed" address="10.10.10.10" ipVersion="4"/>
        </nova:port>
        <nova:port uuid="43d7897895e6-ca52-44e5-b2878-7864153">
          <nova:ip type="fixed" address="10.10.20.10" ipVersion="4"/>
        </nova:port>
      </nova:ports>
    </nova:instance>
  </metadata>
  <memory unit='KiB'>2097152</memory>
[...]
```


dave-tucker (changes requested):
The libvirt dependency and performance of this code are likely to be deal breakers for me.

I hear what you say in https://github.com/sustainable-computing-io/kepler/issues/1625, but maybe there's another way we an achieve what you're looking for?

e.g what if instead of reading the cgroup we were to instead look at:

```
$ xargs --null -n1 -a /proc/3905059/cmdline | grep uuid -n1
20-4,sockets=4,cores=1,threads=1
21:-uuid
22-8fc5d2df-80d0-4971-a198-d2fc2602f5f7
```

Which gets the dom uuid (avoiding having to list all + search).

Another alternative might be to add all visited doms to the cache (assuming we're able to get mapping to a pid/tgid).

Using https://github.com/digitalocean/go-libvirt would get around the Cgo concerns.

aharivel:
Hi @dave-tucker 
First, thanks for your feedback.

So if I resume what we have done so far to detect which processes are a VM:
- 10 month ago I proposed to read "/run/libvirt/qemu/<name>.pid" to match the pid<-> name
- 4 month later @marceloamaral changed it to read "/proc/<pid>/cgroup" to unify with the containers filtering
- now you propose to filter /proc/<pid>/cmdline to get uuid of the VM.

Some questions come to my mind:
- How do you know that that <pid> is a VM before running the command you proposed ? Are we going to run it for any processes (!!!) or are we first read the "/proc/<pid>/cgroup" then get the uuid of the VM by reading the cmdline ?
- We have a map[uint64]string that gives the pair "pid<->name" so we would change the name by uuid ? i.e "pid<->uuid"
- I can check the digitalocean libvirt library if that suits the performance better. But in the end, if we don't do anything with "libvirt" (not reading libvirt folders, nor using libvirt librairies) we should change the name of this folder, because it's really misleading.

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/HabanaAI/gohlml v1.16.0
 	github.com/NVIDIA/go-dcgm v0.0.0-20240118201113-3385e277e49f
 	github.com/NVIDIA/go-nvml v0.12.0-1
+	github.com/beevik/etree v1.4.0
```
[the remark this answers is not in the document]
aharivel: Only My habit to use this library. I'll try encoding/xml then !

go.mod
```diff
@@ -26,6 +27,7 @@ require (
 	k8s.io/apimachinery v0.29.6
 	k8s.io/client-go v0.29.6
 	k8s.io/klog/v2 v2.120.1
+	libvirt.org/go/libvirt v1.10005.0
```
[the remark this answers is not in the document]
aharivel: If we can go with https://github.com/digitalocean/go-libvirt then we might be able to achieve the same result. I'm going to check this.

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
...
+	domainName := ""
+
+	parts := strings.Split(vmid, "-")
+	if len(parts) < 3 {
+		return "", fmt.Errorf("the vmid provided is not `machine-qemu-*number*-*name*` : %v", vmid)
+	} else {
+		domainName = strings.Join(parts[3:], "-")
+	}
+
+	conn, err := libvirt.NewConnect("qemu:///system")
```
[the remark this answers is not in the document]
aharivel: Kepler is running as a privileged user to be able to use eBPF so this is no problem

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
...
+	if err != nil {
+		return "", fmt.Errorf("Failed to connect to hypervisor: %v", err)
+	}
+
+	defer conn.Close()
+
+	domains, err := conn.ListAllDomains(libvirt.CONNECT_LIST_DOMAINS_ACTIVE)
```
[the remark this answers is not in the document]
aharivel: I'll check this point with the uuid and digitalocean lib

go.mod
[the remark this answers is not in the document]
aharivel: Ok so this small example: 
```
package main

import (
	"fmt"
	"log"
	"net/url"

	"github.com/digitalocean/go-libvirt"
)

func main() {
	uuid_t := libvirt.UUID{0x01, 0x88, 0x5a, 0x2b, 0x6e, 0x3a, 0x4b, 0x8a, 0xbf, 0xbc, 0x13, 0x72, 0x05, 0xc3, 0xc8, 0x03}

	uri, _ := url.Parse(string(libvirt.QEMUSystem))
	l, err := libvirt.ConnectToURI(uri)
	if err != nil {
		log.Fatalf("failed to connect: %v", err)
	}

	d, errr := l.DomainLookupByUUID(uuid_t)
	if errr != nil {
		log.Fatalf("Domainlookup fail %v", errr)
	}

	fmt.Println("ID\tName\t\tUUID")
	fmt.Printf("--------------------------------------------------------\n")
	fmt.Printf("%d\t%s\t%x\n", d.ID, d.Name, d.UUID)
}
```
is returning me :
$ go run main.go 
ID	Name		UUID
--------------------------------------------------------
1	alpinelinux3.19	01885a2b6e3a4b8abfbc137205c3c803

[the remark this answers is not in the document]
aharivel: So we could straight after do something like this: 
```
uri := libvirt.optstring{"http://openstack.org/xmlns/libvirt/nova/1.1"}                                                              
str, _ := l.domaingetmetadata(d, 0x2, uri, 0)  
```
and we received the metadata, then we filter with the right token.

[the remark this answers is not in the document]
aharivel: So in the end, we trade a C-go calls to RPC calls. I'm not that of an expert in this kind of optimization so I'll go with what you prefer.

pkg/libvirt/resolve_vm.go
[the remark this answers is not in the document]
aharivel: PS: all my smalls tests (digitalocean libvirt) above are done without any modification. So I guess that it works without any privilege.

pkg/libvirt/resolve_vm.go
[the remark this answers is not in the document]
aharivel: See my comments above that remove this.

go.mod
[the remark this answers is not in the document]
aharivel: @dave-tucker The issue here is that we don't know in advance the struct of the XML. So no possible way to use Unmarshall because it maps XML element to a struct. 
Etree lets you create a document and then FindElement into it.
If you have a better idea, please tell me, I'm making a second version now.

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
...
 }
+
+func getMetadataVMID(vmid string) (string, error) {
+	metadataURI := config.GetLibvirtMetadataURI()
+	token := config.GetLibvirtMetadataToken()
+	domainName := ""
+
+	parts := strings.Split(vmid, "-")
+	if len(parts) < 3 {
+		return "", fmt.Errorf("the vmid provided is not `machine-qemu-*number*-*name*` : %v", vmid)
```
sthaha, line 123:
```suggestion
		return "", fmt.Errorf("the vmid does not match machine-qemu-number-name` : %v", vmid)
```

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
...
 	cacheNotExist[pid] = false
 }
+
+func getMetadataVMID(vmid string) (string, error) {
+	metadataURI := config.GetLibvirtMetadataURI()
+	token := config.GetLibvirtMetadataToken()
+	domainName := ""
+
+	parts := strings.Split(vmid, "-")
+	if len(parts) < 3 {
```
sthaha, line 122: won't there be 4 parts?

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
...
+		return "", fmt.Errorf("the vmid provided is not `machine-qemu-*number*-*name*` : %v", vmid)
+	} else {
+		domainName = strings.Join(parts[3:], "-")
+	}
```
sthaha, line 126:
```suggestion
	} 
	
	domainName = strings.Join(parts[3:], "-")

```

sthaha:
I am marking this with `do-not-merge` since this change needs to be tested manually.

pkg/libvirt/resolve_vm.go
```diff
@@ -66,7 +70,16 @@ func getVMID(pid uint64, fileName string) (string, error) {
...
-	addVMIDToCache(pid, vmID)
+	if config.GetLibvirtMetadataURI() != "" {
+		if metaID, err := getMetadataVMID(vmID); metaID != "" && err == nil {
+			addVMIDToCache(pid, metaID)
+		} else {
+			addVMIDToCache(pid, vmID)
+		}
+	} else {
+		addVMIDToCache(pid, vmID)
+	}
```
sthaha, line 81:
```suggestion
	if config.GetLibvirtMetadataURI() != "" {
		// NOTE: use VM Metadata instead 	
		if metaID, err := getMetadataVMID(vmID); metaID != "" && err == nil {
			vmID = metaID
		} 
	}
	addVMIDToCache(pid, vmID)
```

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
 	}
 	cacheNotExist[pid] = false
 }
+
+func getMetadataVMID(vmid string) (string, error) {
+	metadataURI := config.GetLibvirtMetadataURI()
+	token := config.GetLibvirtMetadataToken()
```
sthaha, line 118: can we move to close to the usage?

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +112,62 @@ func addToNotExistCache(pid uint64) {
...
+	domains, err := conn.ListAllDomains(libvirt.CONNECT_LIST_DOMAINS_ACTIVE)
+	if err != nil {
+		return "", fmt.Errorf("Failed to list domains: %v", err)
+	}
+
+	defer func() {
+		for _, dom := range domains {
+			dom.Free()
+		}
+	}()
```
sthaha, line 144: Doesn't this re-iterate all domains (say 100)  even if we the vm was found in the first iteration in the loop below? 

I think we should simplify the loop below as 

```
for _, dom := range domains {
  name, err := parseDomainForVmID(dom)
  dom.Free()

  if err == nil {
    return name
  }
}

return "", fmt.Errorf("...")
```

pkg/libvirt/resolve_vm.go
aharivel: Thanks for your review. I will include all of your feedback  in the next iteration. 
However I don't understand this question. Could you please elaborate ?

pkg/libvirt/resolve_vm.go
aharivel: On my next iteration using the libvirt library from DigitalOcean that Dave suggested, I directly get the metadata once I have the name of the VM so no more iteration !

aharivel:
Hi @dave-tucker and @sthaha ,
I've pushed a new version of my commit.
I ended up using digitalocean libvirt library and simplify the way I retrieve the metadata name.
I did not use any UUID. I think we (I?) can change that on a second phase.
My intention is to disturb the least possible the actual code while being able to evolve at the same time with the new requirement (UUID instead of name?)

Please tell me what you would like me to change on the base of that.

PS: I'll deal with the failing check as soon as they are finished.

dave-tucker:
> My intention is to disturb the least possible the actual code while being able to evolve at the same time with the new requirement (UUID instead of name?)


Just confirming that ☝️ works for me.
Name from metadata using the digital ocean go library as step 1
We can open an issue to discuss using UUID as a future enhancement

pkg/libvirt/resolve_vm.go
sthaha: Sorry, I wasn't clear :( 
What I meant is that as a good practice we should keep variables close to its usage. E.g. `token` is initialised here but used may be 20 lines below. For a reader / maintainer look at the code, they have to keep a context of this when they read its usage 20 lines later.  But if the variables are moved closer to its actual usage, it reduces the cognitive load.

aharivel: Oh ok I see. I'm a long year C developer some habits are hard to get rid of !
I'm moving the variable close to its usage.

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +111,49 @@ func addToNotExistCache(pid uint64) {
...
+	l, err := libvirt.ConnectToURI(uri)
+	if err != nil {
+		return "", fmt.Errorf("libvirt failed to connect: %v", err)
+	}
+
+	defer func() {
+		if err := l.Disconnect(); err != nil {
+			return
+		}
+	}()
```
sthaha, line 137: can't we `defer l.Disconnect()` instead?

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +111,49 @@ func addToNotExistCache(pid uint64) {
...
+
+	d, err := l.DomainLookupByName(domainName)
+	if err != nil {
+		return "", fmt.Errorf("libvirt Domainlookup fail %v", err)
```
sthaha, line 141:
```suggestion
		return "", fmt.Errorf("libvirt Domainlookup fail %w", err)
```

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +111,49 @@ func addToNotExistCache(pid uint64) {
...
+	}
+
+	metaURI := libvirt.OptString{metadataURI}
+	// Domain Metadata Element = 0x2
+	metaXML, _ := l.DomainGetMetadata(d, 0x2, metaURI, 0)
+
+	doc := etree.NewDocument()
+	if err := doc.ReadFromString(metaXML); err != nil {
+		return "", fmt.Errorf("failed to read metadata XML: %v", err)
```
sthaha, line 150:
```suggestion
		return "", fmt.Errorf("failed to read metadata XML: %w", err)
```

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +111,49 @@ func addToNotExistCache(pid uint64) {
...
+	}
+
+	nameElement := doc.FindElement(fmt.Sprintf("//%s", token))
+	if nameElement != nil {
+		return nameElement.Text(), nil
+	} else {
+		return "", fmt.Errorf("no <%s> element found", token)
+	}
```
sthaha, line 158:
```suggestion
	if nameElement == nil {
		return "", fmt.Errorf("no <%s> element found", token)
	} 
	return nameElement.Text(), nil

```

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +111,49 @@ func addToNotExistCache(pid uint64) {
...
+	if len(parts) < 3 {
+		return "", fmt.Errorf("the vmid provided doest not match machine-qemu-number-name : %v", vmid)
+	}
+
+	domainName = strings.Join(parts[3:], "-")
+
+	uri, _ := url.Parse(string(libvirt.QEMUSystem))
+	l, err := libvirt.ConnectToURI(uri)
+	if err != nil {
+		return "", fmt.Errorf("libvirt failed to connect: %v", err)
```
sthaha, line 130:
```suggestion
		return "", fmt.Errorf("libvirt failed to connect: %w", err)
```

pkg/libvirt/resolve_vm.go
```diff
@@ -99,3 +111,49 @@ func addToNotExistCache(pid uint64) {
...
+	metadataURI := config.GetLibvirtMetadataURI()
+	token := config.GetLibvirtMetadataToken()
+	domainName := ""
+
+	parts := strings.Split(vmid, "-")
+	if len(parts) < 3 {
+		return "", fmt.Errorf("the vmid provided doest not match machine-qemu-number-name : %v", vmid)
+	}
+
+	domainName = strings.Join(parts[3:], "-")
```
sthaha, line 125: wouldn't this crash if len(parts) == 3 ?

sthaha (changes requested):
lgtm overall! some minor comments to address.

pkg/libvirt/resolve_vm.go
sthaha: Ahem ahem :)

pkg/libvirt/resolve_vm.go
aharivel: I did that but the linter says I don't check the return value of "l.Disconnect()"...

pkg/libvirt/resolve_vm.go
aharivel:
```
vmids := []string{
		"machine-qemu-1234-name",
		"machine---5678-anothername",
		"machine-qemu-91011-differentname",
		"invalid-vmid",
		"machine-0000-lastname",
		"qemu-machine-",
		"qemu-machine--123456-last-",
	}

	for _, vmid := range vmids {
		parts := strings.Split(vmid, "-")
		if len(parts) < 4 {
			fmt.Printf("The vmid provided does not match machine-qemu-number-name: %v\n", vmid)
			continue
		}

		domainName := strings.Join(parts[3:], "-")
		fmt.Println(domainName)
	}
```
Output:
```
name
5678-anothername
differentname
The vmid provided does not match machine-qemu-number-name: invalid-vmid
The vmid provided does not match machine-qemu-number-name: machine-0000-lastname
The vmid provided does not match machine-qemu-number-name: qemu-machine-
123456-last-
```

Any suggestion of name that could be problematic ?

sthaha: @aharivel the snipped you pasted, has len(parts) < 4 instead of 3 (as in the code).

pkg/libvirt/resolve_vm.go
sthaha: we can silence the linter :) when we purposely ignore errors

sthaha: On second thought, how we return the error if Disconnect fails?

pkg/libvirt/resolve_vm.go
aharivel: I literally have no idea what happen there but thanks for pointing that out.

pkg/libvirt/resolve_vm.go
aharivel: I'm going to put the `defer l.Disconnect()` back into the code but the linter will failed on that part.

sthaha: @aharivel , I think we should return the error when running disconnect. A common pattern is to have something like 

```
func foo() (val, err error) {
   defer func() {
       err = l.Disconnect()
   }()
}
```

This way the caller gets to handle and report the error.

aharivel: Right, I do prefer this me too.

sthaha (commented):
LGTM! Thanks a lot @aharivel !
I will let @dave-tucker take another look before merging.

sthaha:
@aharivel in case you didn't notice, this PR requires rebasing

aharivel:
> @aharivel in case you didn't notice, this PR requires rebasing

oh sorry, I thought I had no right to do that. Let me do it now. thanks !

aharivel:
@sthaha I think the error above is not related to my patch. Am I wrong ?

sthaha (approved):
LGTM. I will let @dave-tucker do a final review before merging.

marceloamaral (approved):
Looks good to me too.
/lgtm

aharivel:
It still needs you review @dave-tucker ! 
Everything should be fairly OK by now as already 2 out-of 3 reviewers approved the changes !

dave-tucker (approved):

sthaha:
@aharivel , I was going to merge it this one but can't due to merge conflicts. 
Could you please rebase one last time (without any code changes 🙏 ) .

aharivel:
Hey @sthaha , everything is green now :)

Issue #1611, opened 2024-07-11
Reduce eBPF Probe Execution Time


dave-tucker:
### What would you like to be added?

# Summary

#1535 adds some microbenchmarks for the eBPF probes.
The sched_switch probe is the most critical and it currently measures at almost 3 microseconds on my system - we'd like to get this somewhere in the order of a few hundred nanoseconds.
This is critical since that code is executed every time a task is scheduled on or off the CPU.

I've profiled the probe as follows:
- Running kepler
- Using `perf record` to gather samples
- Analyze using `sudo perf annotate -l bpf_prog_eb5663401a302a92_kepler_sched_switch_trace` - you can get the tag from `bpftool prog`

In summary the hot parts of the probe appear to be as follows:

| Code | Percent |
|-----|---------|
| Preparing Context | 46.32% | 
| bpf_map_lookup_elem(&processes, &prev_tgid) | 4.35% |
| bpf_perf_event_read_value - cache miss | 2.9% |
| bpf_get_current_pid_tgid() | 2.79% |
| bpf_map_update_elem(&cpu_instructions, cpu_id, &val, BPF_ANY);  | 2.76% |
| bpf_map_lookup_elem(&processes, &curr_tgid) | 2.62% |
| bpf_map_delete_elem(&pid_time_map, &prev_pid) | 2.38% |
| bpf_map_update_elem(&cache_miss, cpu_id, &val, BPF_ANY); | 1.98% |
| bpf_ktime_get_ns() | 1.82% |
| bpf_perf_event_read_value - cpu_instructions | 1.63% |

The time spent preparing the context is a fixed overhead - mostly due to spectre/meltdown prevention code in kernel that we are not in control of.
When looking at what out probe code contributes you could view this as `map operations` being the key contributor to the overall probe execution time.

# Proposal

Firstly we're only going to collect the following information on a sched_switch event:

1. Timestamp - bpf_ktime_get_ns()
2. Which CPU this event is for
2. `prev_task->tgid`
3. `next_task->tgid`
4. Value of the `cpu_cycles` hardware counter

Secondly that information is going to immediately be sent back to userland using a [BPF_MAP_TYPE_RINGBUF](https://docs.kernel.org/6.6/bpf/ringbuf.html)

Userland is going to constant read from that ring buffer and will be responsible for performing the delta calculations that were previously done in the kernel.

This should get our eBPF probe execution time down really low.

### Why is this needed?

[...]

Issue #1607, opened 2024-07-09
bpf sampling no longer reduces overhead


rootfs:
### What happened?

Latest the `EXPERIMENTAL_BPF_SAMPLE_RATE` in kepler ebpf code no longer shows any difference in reducing ebpf overhead. Note, in the latest, the [map collection](https://github.com/sustainable-computing-io/kepler/blob/main/bpfassets/libbpf/src/kepler.bpf.c#L213) is still executed even sampling is enabled. This is not the case in 0.7.8 code.

I compared 0.7.8 and latest code and calculated the ebpf per call overhead in ns. In 0.7.8 code, when setting `EXPERIMENTAL_BPF_SAMPLE_RATE` to 1000, the per call overhead is reduced from 12043.3 ns to 141.279 ns. While in latest code, there is almost no difference (10187.8 ns vs 10100 ns).
![image](https://github.com/sustainable-computing-io/kepler/assets/7062400/ac2bc532-01f8-45b6-856a-5ddf289aa0e1)

### What did you expect to happen?

EXPERIMENTAL_BPF_SAMPLE_RATE should be able to reduce ebpf overhead 

### How can we reproduce it (as minimally and precisely as possible)?

Run the bpftool command on kepler latest and 0.7.8

```
sysctl kernel.bpf_stats_enabled=1

# latest without sampling
echo 0 | sudo tee  /etc/kepler/kepler.config/EXPERIMENTAL_BPF_SAMPLE_RATE
bpftool prog show name kepler_sched_switch_trace | head -n 1 | awk '{print "rt_ns:", $(NF-2), "count: ", $NF, "avg: ", $(NF-2)/$NF } '

# enable sampling 
echo 1000 | sudo tee  /etc/kepler/kepler.config/EXPERIMENTAL_BPF_SAMPLE_RATE
bpftool prog show name kepler_sched_switch_trace | head -n 1 | awk '{print "rt_ns:", $(NF-2), "count: ", $NF, "avg: ", $(NF-2)/$NF } '


```

### Kepler image tag

<details>
0.7.8 and latest
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>
KVM 
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
[...]

Issue #1354, opened 2024-04-16
Kepler not reporting correct process name in metrics


vprashar2929:
### What happened?

When Kepler using the `latest` deployed on a machine currently it reports the wrong process name in the exported metrics.

Attaching some screenshots for reference:
* Actual process name and PID running on the system: 

```
ps -ef | grep 75577
qemu       75577       1  8 Apr15 ?        01:10:07 /usr/bin/qemu-system-x86_64 -name guest=fedora39,debug-threads=on -S 
```

Output from pstree command:

```
pstree -p | grep qemu
           |-qemu-system-x86(75577)-+-{qemu-system-x86}(75605)
           |                        |-{qemu-system-x86}(75617)
           |                        |-{qemu-system-x86}(75618)
           |                        |-{qemu-system-x86}(75619)
           |                        |-{qemu-system-x86}(75620)
           |                        |-{qemu-system-x86}(75622)
           |                        |-{qemu-system-x86}(109718)
           |                        |-{qemu-system-x86}(109719)
           |                        |-{qemu-system-x86}(109720)
           |                        `-{qemu-system-x86}(109721)
```

* Value reported by `kepler_process_platform_joules_total` for the particular pid `75577` that is `command="CPU 0/KVM"` which is wrong

![Screenshot 2024-04-16 at 1 21 16 PM](https://github.com/sustainable-computing-io/kepler/assets/115542213/f3d14d78-6df4-4fa5-b73f-c454a858885b)

### What did you expect to happen?

Kepler should report the correct command name in the metrics that it exports.

### How can we reproduce it (as minimally and precisely as possible)?

Run Kepler either on Kubernetes or using the docker-compose locally which is present here: https://github.com/sustainable-computing-io/kepler/tree/main/hackdocker-compose

### Kepler image tag

<details>
latest
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
[...]

PR #1628, opened 2024-07-23
feat(bpf): Send events via Ring Buffer

dave-tucker:
This uses a ring buffer to send events from eBPF back to userspace.
Doing so allows for our eBPF probes to complete quickly, and pushes
all delta calculation and summary back into userspace.

Moves the logic that resolves comm from procfs into its own pkg.
In addition, a cache is added to avoid hitting procfs every time
we update process metrics.

Removes the bpftest pkg and includes the test programs in the
main kepler.bpf.c file. This prevents drift in the go generate
flags and generally makes it easier to write tests.

Decreases probe execution time by 40% and uses a lot less CPU cycles/instructions per probe invocation.
Also removes a lot of eBPF maps reducing the overall memory required.

Fixes: #1611, #1607, #1354, #1640


pkg/bpf/exporter.go
```diff
@@ -192,38 +193,221 @@ func (e *exporter) Detach() {
...
+
+func (e *exporter) CollectProcesses() ([]ProcessMetrics, error) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	metrics := make([]ProcessMetrics, len(e.processMetrics))
+	for pid, m := range e.processMetrics {
+		if pid == 2019461 {
+			fmt.Printf("pid: %d, runtime: %d, cpu_cycles: %d, cpu_instructions: %d\n", pid, m.CPUTime.Total/1000, m.CPUCyles.Total, m.CPUInstructions.Total)
+		}
```
sthaha, line 395: debug code?

pkg/bpf/exporter.go
```diff
@@ -192,38 +193,221 @@ func (e *exporter) Detach() {
...
+					return nil
+				}
+				if errors.Is(err, ringbuf.ErrFlushed) {
+					record.RawSample = record.RawSample[:0]
+				}
+				klog.Errorf("reading from reader: %s", err)
+				continue
+			}
+
+			if err := binary.Read(bytes.NewBuffer(record.RawSample), binary.LittleEndian, &event); err != nil {
```
sthaha, line 237: shouldn't we use the host.ByteOrder instead?

pkg/bpf/exporter.go
```diff
@@ -192,38 +193,221 @@ func (e *exporter) Detach() {
...
+func (p *PerCPUCounter) Stop(cpu, taskID uint32, value uint64) error {
+	key := uint64(cpu)<<32 | uint64(taskID)
+
+	if _, ok := p.Values[key]; !ok {
+		return nil
+	}
+
+	if value == 0 {
+		return nil
+	}
```
sthaha, line 311:
```suggestion
	if value == 0 {
		return nil
	}

```

sthaha, line 311:
```suggestion
	if value == 0 {
		return nil
	}
	
	key := uint64(cpu)<<32 | uint64(taskID)
	if _, ok := p.Values[key]; !ok {
		return nil
	}

```

Lets avoid a lookup when possible

pkg/bpf/exporter.go
```diff
@@ -192,38 +193,221 @@ func (e *exporter) Detach() {
...
+
+	var delta uint64
+
+	// Probably a clock issue where the recorded on-CPU event had a
+	// timestamp later than the recorded off-CPU event, or vice versa.
+	if value > p.Values[key] {
+		delta = value - p.Values[key]
+	} else {
+		delta = 0
+	}
```
sthaha, line 321:
```suggestion
	var delta uint64 = 0   

	// Probably a clock issue where the recorded on-CPU event had a
	// timestamp later than the recorded off-CPU event, or vice versa.
	if value > p.Values[key] {
		delta = value - p.Values[key]
	}
```

pkg/bpf/exporter.go
```diff
@@ -192,38 +193,221 @@ func (e *exporter) Detach() {
...
+func (e *exporter) handleEvent(event keplerEvent) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	var p *bpfMetrics
+
+	if _, ok := e.processMetrics[event.Pid]; !ok {
+		e.processMetrics[event.Pid] = newBpfMetrics()
+	}
+	p = e.processMetrics[event.Pid]
```
sthaha, line 342: We can extract better performance by locking only read and write to data. 
Have you considered a RWMutex? 

Is the purpose of Mutex to lock the entirely of the exporter or only the processMetrics map?

pkg/bpf/exporter.go
```diff
@@ -280,13 +465,15 @@ func (h *hardwarePerfEvents) close() {
...
+	fmt.Printf("create hardware perf events")
 	events := &hardwarePerfEvents{
-		cpuCyclesPerfEvents:       []int{},
-		cpuInstructionsPerfEvents: []int{},
-		cacheMissPerfEvents:       []int{},
+		cpuCyclesPerfEvents:       make([]int, 0, numCPU),
+		cpuInstructionsPerfEvents: make([]int, 0, numCPU),
+		cacheMissPerfEvents:       make([]int, 0, numCPU),
 	}
+	fmt.Printf("create hardware perf events: %+v", events)
```
sthaha, line 474: log ?

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
```diff
@@ -87,19 +90,28 @@ func UpdateProcessBPFMetrics(bpfExporter bpf.Exporter, processStats map[uint64]*
...
 	}
 	for _, ct := range processesData {
-		comm := C.GoString((*C.char)(unsafe.Pointer(&ct.Comm)))
+		var comm string
+		if _, err := os.Stat("/proc/" + strconv.Itoa(int(ct.Pid))); os.IsNotExist(err) {
+			// skip process that is not running
+			continue
+		}
+		if commBytes, err := os.ReadFile("/proc/" + strconv.Itoa(int(ct.Pid)) + "/comm"); err == nil {
+			comm = strings.TrimSpace(unix.ByteSliceToString(commBytes))
```
sthaha, line 99: isn't reading from `/proc` expensive ? instead, can't we pass this data from kernel itself as it used to be? 
Also ,should we re-read the proc every time we process data?

sthaha (changes requested):
Looks good overall, some comments though

pkg/bpf/exporter.go
dave-tucker: > Is the purpose of Mutex to lock the entirely of the exporter or only the processMetrics map?

The idea is to lock the entire exporter - so RWMutex won't help.
We want to stop reading new updates from the RingBuffer until after `CollectProcesses` has exited and all the `Total` counters have been reset back to 0.

pkg/bpf/exporter.go
dave-tucker: more debug code - removed

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
dave-tucker: Re: Getting comm from eBPF... See:
https://github.com/sustainable-computing-io/kepler/issues/1354
https://github.com/sustainable-computing-io/kepler/issues/1640

It's "wrong" for 2 reasons:

1. https://github.com/sustainable-computing-io/kepler/blob/19c6afdccaa602d0a619c251f7ea17d9adb2b2c5/bpf/kepler.bpf.h#L281 `current_comm` is the process that is going OFF CPU, not the process going ON CPU 🐛 
2. Even if we were to use this helper correctly, we probably don't want to be getting the comm on every task switch since it never changes.

Not only that, but `comm` from eBPF is limited to 16 bytes (15 bytes + NUL).
The output from procfs is much nicer, see below.

```
kepler_process_cpu_instructions_total{command="postgres",container_id="fbe5d970ec22cf266028f676f98b5670e1193e33d571eb54f4c9f99f991a2d72",pid="1210084",source="bpf",vm_id=""} 971473
kepler_process_cpu_instructions_total{command="postgres: autovacuum launcher",container_id="fbe5d970ec22cf266028f676f98b5670e1193e33d571eb54f4c9f99f991a2d72",pid="1210248",source="bpf",vm_id=""} 73265
kepler_process_cpu_instructions_total{command="postgres: background writer",container_id="fbe5d970ec22cf266028f676f98b5670e1193e33d571eb54f4c9f99f991a2d72",pid="1210245",source="bpf",vm_id=""} 21080
kepler_process_cpu_instructions_total{command="postgres: walwriter",container_id="fbe5d970ec22cf266028f676f98b5670e1193e33d571eb54f4c9f99f991a2d72",pid="1210247",source="bpf",vm_id=""} 20648
```

We actually have access to the full set of commandline args, but this might be too verbose for use as labels, unless we can find a smarter way of adding them without increasing cardinality.

Re: Performance. It is expensive yes and would benefit from a cache since these values never change between collections.

pkg/bpf/exporter.go
sthaha: Thanks for the clarification.  Would you mind moving the mutex to the top of struct ? 
You can find similar patterns in go . The idea is to keep mutex above what it protects.

pkg/bpf/exporter.go
```diff
@@ -192,38 +193,223 @@ func (e *exporter) Detach() {
...
-		)
-		total += count
-		if errors.Is(err, ebpf.ErrKeyNotExist) {
-			break
+		var record *ringbuf.Record
+
+		select {
+		case <-stopChan:
+			if err := rd.Close(); err != nil {
+				return fmt.Errorf("closing ring buffer reader: %v", err)
```
sthaha, line 218:
```suggestion
				return fmt.Errorf("closing ring buffer reader: %w", err)
```

pkg/bpf/exporter.go
```diff
@@ -271,6 +457,7 @@ type hardwarePerfEvents struct {
 }
 
 func (h *hardwarePerfEvents) close() {
+	fmt.Printf("close perf events: %+v", h)
```
sthaha, line 460: klog?

pkg/bpf/exporter.go
```diff
@@ -280,13 +467,14 @@ func (h *hardwarePerfEvents) close() {
 // and updates the corresponding eBPF maps.
 func createHardwarePerfEvents(cpuInstructionsMap, cpuCyclesMap, cacheMissMap *ebpf.Map, numCPU int) (*hardwarePerfEvents, error) {
 	var err error
+	fmt.Printf("create hardware perf events")
```
sthaha, line 470: klog

pkg/comm/resolve_comm.go
```diff
@@ -0,0 +1,89 @@
...
+
+	comm, err := readCommandFromProcFs(pid)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// skip process that is not running
+			r.cacheNotExist[pid] = true
+			return unknownComm, fmt.Errorf("process not running: %w", err)
+		}
+	}
+	r.cacheExist[pid] = comm
```
sthaha, line 43: when do we invalidate both caches?

pkg/comm/resolve_comm.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	comm, err := readCommandFromProcFs(pid)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// skip process that is not running
+			r.cacheNotExist[pid] = true
+			return unknownComm, fmt.Errorf("process not running: %w", err)
+		}
+	}
```
sthaha, line 42:
```suggestion
	if err != nil && os.IsNotExist(err) {
			// skip process that is not running
			r.cacheNotExist[pid] = true
			return unknownComm, fmt.Errorf("process not running: %w", err)
		}
	}
```

pkg/bpf/exporter.go
```diff
@@ -192,38 +193,223 @@ func (e *exporter) Detach() {
...
-	// Get the max number of entries in the map
-	maxEntries := e.bpfObjects.Processes.MaxEntries()
-	total := 0
-	deleteKeys := make([]uint32, maxEntries)
-	deleteValues := make([]ProcessMetrics, maxEntries)
-	var cursor ebpf.MapBatchCursor
+func (e *exporter) Start(stopChan <-chan struct{}) error {
+	rd, err := ringbuf.NewReader(e.bpfObjects.Rb)
+	if err != nil {
+		return fmt.Errorf("failed to create ring buffer reader: %v", err)
```
sthaha, line 208:
```suggestion
		return fmt.Errorf("failed to create ring buffer reader: %w", err)
```

sthaha (commented):
Looks good overall. I think we should address cache invalidation of `comm`.

pkg/bpf/exporter.go
dave-tucker: I've kept it where it was and added code comments. It is protecting processMetrics but also freedPIDs.

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
dave-tucker: Cache added.

pkg/comm/resolve_comm.go
dave-tucker: fixed. we now listen for `sched_process_free` events and evict stale PIDs from both caches every sample period.

dave-tucker:
@sthaha This is now ready for another look. Caching of `comm` from procfs is now solved with another eBPF probe that tells us when PIDs have been freed. I've also marked this as fixing all the issues relating to the process `comm` since these are now accurate:

For example:
```console
$ curl -s localhost:8888/metrics | grep grafana-server | grep cpu_cycles           
kepler_process_cpu_cycles_total{command="grafana-server",container_id="f3e1cb741182257a9fcfb8e71ccf9ef14e044556f6225900323dce7a10cb2ac0",pid="2025102",source="bpf",vm_id=""} 1.593157176e+09
```

``` console
$  sudo ps aux | grep grafana-server
nobody   2025102  0.4  0.1 797072 101700 ?       Ssl  Jul11 117:49 grafana-server --homepath=/usr/share/grafana --config=/etc/grafana/grafana.ini --packaging=docker cfg:default.log.mode=console cfg:default.paths.data=/var/lib/grafana cfg:default.paths.logs=/var/log/grafana cfg:default.paths.plugins=/var/lib/grafana/plugins cfg:default.paths.provisioning=/etc/grafana/provisioning
```

Once this merges I'll open an issue to discuss this further as I think we should probably move this type of data into logs vs adding it as a label - but that would require changes to the operator further down the line.

bpf/kepler.bpf.c
```diff
@@ -3,6 +3,168 @@
 
 #include "kepler.bpf.h"
 
+// Ring buffer sizing
+// 256kB is sufficient to store around 1000 events/sec for 5 seconds
```
sthaha, line 7: ❤️

pkg/bpf/exporter.go
```diff
@@ -35,23 +38,38 @@ import (
...
 	pageWriteLink   link.Link
 	pageReadLink    link.Link
+	processFreeLink link.Link
 
 	perfEvents *hardwarePerfEvents
 
 	enabledHardwareCounters sets.Set[string]
 	enabledSoftwareCounters sets.Set[string]
+
+	// Locks processMetrics and freedPIDs.
```
sthaha, line 54: ❤️

sthaha (approved):
LGTM! Thanks a lot for incorporating the comments 🙇

rootfs:
I did a quick test to evaluate the eBPF latency reduction with regard to 0.7.8 (which has both tracking and sampling modes).

# TL;DR 
In this current implementation, using ringbuffer reduces eBPF tracking mode latency by 4.5x, but still a long shot from sampling mode performance. 

# Result 

| Latest code with this PR, sample rate 1000| 0.7.8 code, tracking (no sampling)|0.7.8 code, sampling rate 1000|
|---|---|---|
| 484 ns | 2043 ns | 78 ns | 

# Test
[asciicast](https://asciinema.org/a/670546)

# Test script
The test script is [here](https://gist.github.com/rootfs/ae1e50cc69bd5d07dc2e0f5df3ea3869)

dave-tucker:
> In this current implementation, using ringbuffer reduces eBPF tracking mode latency by 4.5x, but still a long shot from sampling mode performance.

It's always a compromise between accuracy and performance.
Look at it this way:

| Case | Duration | Accuracy |
|------|----------|----------|
| latest | 484ns | 100% |
| v0.7.8 | 2043ns | 100% |
| v0.7.8 Sample Rate 1000 | 78ns | 0.1% |

The ringbuf case to me seems to be the best of both worlds. Acceptable performance with the best accuracy.

PR #1629, opened 2024-07-23
chore: tidy use of bpfSupportedMetrics for NewStats()

maryamtahhan:
Tidy up the universal passing around of bpfSupportedMetrics for NewStats()

- [x] Introduce a reasonable default set of bpfSupportedMetrics in the stats package so that these can be used for NewStats() whenever it's invoked.
- [x] Remove the `InitAvailableParamAndMetrics()` function and simply initialise the variables in the package at declaration. 
- [x] Remove dependency between bpf and stats packages entirely

### DIFF
--- a/pkg/bpf/exporter.go
+++ b/pkg/bpf/exporter.go
@@ -50,8 +50,8 @@ type exporter struct {
 
 func NewExporter() (Exporter, error) {
 	e := &exporter{
-		enabledHardwareCounters: sets.New[string](),
-		enabledSoftwareCounters: sets.New[string](),
+		enabledHardwareCounters: sets.New[string](config.BPFHwCounters()...),
+		enabledSoftwareCounters: sets.New[string](config.BPFSwCounters()...),
 	}
 	err := e.attach()
 	if err != nil {
@@ -110,7 +110,6 @@ func (e *exporter) attach() error {
 	if err != nil {
 		return fmt.Errorf("error attaching sched_switch tracepoint: %v", err)
 	}
-	e.enabledSoftwareCounters[config.CPUTime] = struct{}{}
 
 	if config.ExposeIRQCounterMetrics() {
 		e.irqLink, err = link.AttachTracing(link.TracingOptions{
@@ -120,9 +119,6 @@ func (e *exporter) attach() error {
 		if err != nil {
 			return fmt.Errorf("could not attach irq/softirq_entry: %w", err)
 		}
-		e.enabledSoftwareCounters[config.IRQNetTXLabel] = struct{}{}
-		e.enabledSoftwareCounters[config.IRQNetRXLabel] = struct{}{}
-		e.enabledSoftwareCounters[config.IRQBlockLabel] = struct{}{}
 	}
 
 	group := "writeback"
@@ -143,8 +139,6 @@ func (e *exporter) attach() error {
 	})
 	if err != nil {
 		klog.Warningf("failed to attach fentry/mark_page_accessed: %v. Kepler will not collect page cache read events. This will affect the DRAM power model estimation on VMs.", err)
-	} else if !e.enabledSoftwareCounters.Has(config.PageCacheHit) {
-		e.enabledSoftwareCounters[config.PageCacheHit] = struct{}{}
 	}
 
 	// Return early if hardware counters are not enabled
@@ -162,9 +156,6 @@ func (e *exporter) attach() error {
 	if err != nil {
 		return nil
 	}
-	e.enabledHardwareCounters[config.CPUCycle] = struct{}{}
-	e.enabledHardwareCounters[config.CPUInstruction] = struct{}{}
-	e.enabledHardwareCounters[config.CacheMiss] = struct{}{}
 
 	return nil
 }
--- a/pkg/collector/stats/stats.go
+++ b/pkg/collector/stats/stats.go
@@ -20,17 +20,11 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/sustainable-computing-io/kepler/pkg/bpf"
 	"github.com/sustainable-computing-io/kepler/pkg/collector/stats/types"
 	"github.com/sustainable-computing-io/kepler/pkg/config"
 	acc "github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator"
 	"k8s.io/klog/v2"
 )
-
-var defaultBPFMetrics = []string{
-	config.CPUCycle, config.CPURefCycle, config.CPUInstruction, config.CacheMiss, config.CPUTime,
-	config.PageCacheHit, config.IRQNetTXLabel, config.IRQNetRXLabel, config.IRQBlockLabel,
-}
 
 // metricSets stores different sets of metrics for energy and resource usage.
 type metricSets struct {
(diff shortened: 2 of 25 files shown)


maryamtahhan:
Hola @dave-tucker @sthaha @rootfs 

Would really appreciate a review if you get the chance.

Thanks in advance
Maryam

cmd/exporter/exporter.go
```diff
@@ -150,7 +148,7 @@ func main() {
 		klog.Fatalf("failed to create eBPF exporter: %v", err)
 	}
 	defer bpfExporter.Detach()
-
+	stats.InitAvailableParamAndMetrics(bpfExporter.SupportedMetrics())
```
dave-tucker, line 151: @maryamtahhan this is very similar to how it used to be before https://github.com/sustainable-computing-io/kepler/commit/0baec479033aaca3d1744ec3010e156afd0075e8

The reason I fixed this by passing it around everywhere was that there is still a risk of racing here.

i.e what is going to prevent the following scenarios:

1. stats.NewStats (or one of the variants) being called **before** `InitAvailableParamAndMetrics`
1. Access to `stats.AvailableBPFMetrics` **before**  `InitAvailableParamAndMetrics`

dave-tucker (changes requested):
Following up from my earlier comment I feel like we're treating a symptom (bpfSupportedMetrics is being passed everywhere) and not the disease (why does bpfSupportMetrics NEED to be plumbed in so many places).

The bigger problem is that `pkg/stats` depending on `pkg/bpf`...
However, `pkg/stats` is regularly being called from parts of the codebase where a bpfExporter object isn't in scope.

I think the `pkg/stats` dependency can be removed entirely by simply populating the hashmap with all the possible eBPF features. We're already filtering them before being posted to prometheus - see: `CollectResUtilizationMetrics`. That should clean up most of the mess.

I don't feel like the cleanup to make `pkg/model` depend on a maybe-unititialized value in `pkg/stats` is a good idea - depending on being provided a `bpfSupportedStats` struct seems fine.

cmd/exporter/exporter.go
maryamtahhan: I'm obviously missing something (sorry); I'm not sure how passing that bpfsupportedmetrics around fixed the risk of racing, when I tracked down the usage of bpfsupportedmetrics the majority of places led to NewStats(), so we were just passing this variable along through n levels of functions so that n stats has a reference to the bpfmetrics. 

From what I can see the model isn't initialised till the Collector is started and in order to create the collector the bpfexporter needs to be created and part of that is that the bpf stats are initialised. we were making a call to `stats.InitAvailableParamAndMetrics(bpfExporter.SupportedMetrics())` directly after bpfexporter creation.

Regardless I will work on this patch to make it so that stats will no longer depend on the bpf package and revert the model to use the bpf package as it did originally and push another patch.

dave-tucker: I don't disagree with your assessment of how it works right now 😄 
The problem is that we're encoding assumptions into the codebase - this will happen, before that.
These are assumptions that no compiler or linter can help ensure remain true, especially as these assumptions seem to cross package/module boundaries.

I'm all for letting the compiler do the hard work. So if thing X needs and instance of thing Y to operate, it has to be given one when it's constructed.

pkg/bpf/exporter.go
```diff
@@ -85,6 +88,11 @@ func (e *exporter) SupportedMetrics() SupportedMetrics {
 	}
 }
 
+func (e *exporter) registerBPFStats() {
```
maryamtahhan, line 91: Not sure if reversing the dependency is Ok - Can remove this altogether if preferred...

pkg/collector/stats/stats.go
```diff
@@ -20,20 +20,35 @@ import (
...
 
-	"github.com/sustainable-computing-io/kepler/pkg/bpf"
 	"github.com/sustainable-computing-io/kepler/pkg/collector/stats/types"
 	"github.com/sustainable-computing-io/kepler/pkg/config"
 	acc "github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator"
 	"k8s.io/klog/v2"
 )
 
 var (
+	// Default metric sets. In addition, each metric set is used by the model
```
maryamtahhan, line 30: For the moment, need to leave these as separate sets as they are used by the model to determine the stats groups to update (dynamic/idle/...)

pkg/model/process_energy.go
```diff
@@ -54,7 +53,7 @@ func createProcessPowerModelConfig(powerSourceTarget string, processFeatureNames
 			pkgUsageMetric := config.CoreUsageMetric
 			coreUsageMetric := config.CoreUsageMetric
 			dramUsageMetric := config.DRAMUsageMetric
-			if !bpfSupportedMetrics.HardwareCounters.Has(config.CPUTime) {
+			if !stats.AvailableBPFMetricsHas(config.CPUTime) {
```
maryamtahhan, line 56: Can reverse if we remove `registerBPFStats()`

pkg/collector/stats/utils.go
```diff
@@ -92,6 +75,10 @@ func GetProcessFeatureNames(bpfSupportedMetrics bpf.SupportedMetrics) []string {
 	return metrics
 }
 
+func AvailableBPFMetricsHas(s string) bool {
```
maryamtahhan, line 78: unused - delete

pkg/collector/stats/stats.go
```diff
@@ -20,224 +20,245 @@ import (
...
 	return val
 }
 
-// ToEstimatorValues return values regarding metricNames.
+// ToEstimatorValues returns values for the specified metric names, normalized if required.
 // The metrics can be related to resource utilization or power consumption.
 // Since Kepler collects metrics at intervals of SamplePeriodSec, which is greater than 1 second, and the power models are trained to estimate power in 1 second interval,
-// it is necessary to normalize the resource utilization by the SamplePeriodSec. Note that this is important because the power curve can be different for higher or lower resource usage within 1 second interval.
-func (m *Stats) ToEstimatorValues(featuresName []string, shouldNormalize bool) []float64 {
+// it is necessary to normalize the resource utilization by the SamplePeriodSec.
```
maryamtahhan, line 194: Re-add note

maryamtahhan:
I've moved the stats globals cleanup to https://github.com/sustainable-computing-io/kepler/pull/1659

rootfs:
@maryamtahhan can you rebase to pick the e2e test diagnostics? thanks

rootfs:
@maryamtahhan can you rebase to pick the e2e test diagnostics? thanks

maryamtahhan:
Hi @dave-tucker 
I think this is ready for review again. Please let me know if there are any other issues

Thanks in advance

pkg/bpf/exporter.go
dave-tucker: > I just thought it might be useful to get the exporter to register what it supports with the stats when it's created to overwrite any assumed default value with the right stats

I'd rather that we remove the dependency between the 2 packages entirely.

The assumptions around defaults aren't awesome either, but they are the lesser of two evils IMO.
It's worse to be mutating global variables at runtime - data races aren't fun, some readers might cache/copy values so never see the "updated" values etc...

If stats has a hard dependency on the bpf metric detection, then the API of the stats package should be modelled that way - it is with the current `bpfSupportedMetrics` but it's not exactly ergonomic

dave-tucker (changes requested):

pkg/bpf/exporter.go
maryamtahhan: ack - I will just get it to stick to the static set of default metrics....

maryamtahhan:
I will rebase this again

maryamtahhan:
This is rebase now :) \o/ yay!

PR #1632, opened 2024-07-25
ACPI validation: Added MSE,MAPE threshold values

vimalk78:


e2e/tools/validator/acpi_validations.yaml
```diff
@@ -10,3 +10,5 @@ validations:
       )
     expected: |
       mock_acpi_power1_average/10^6
+    max_mse: 0.0001
+    # max_mape: 0
```
sthaha, line 14: Commented by mistake?

e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -253,6 +257,17 @@ def report_validation_results(
         report.write(f"  * MAPE: {res.mape}\n")
         report.flush()
 
+        mse_ok = True
+        mape_ok = True
+
+        if v.max_mse is not None and res.mse.error is None and res.mse.value > v.max_mse:
+            click.secho(f"MSE exceeded threshold. mse: {res.mse.value}, max_mse: {v.max_mse}", fg="red")
+            mse_ok = False
```
sthaha, line 265: How about we enforce setting a Max MSE and MAPE for all validations? I.e. This field can't be empty

vimalk78: in my opinion it is better to keep it optional

sthaha (approved):

PR #1652, opened 2024-07-31
fix: mape formula

vimalk78:
- fixes MAPE calculation formula
- fixes associated tests


e2e/tools/validator/src/validator/prometheus/__init__.py
```diff
@@ -127,7 +127,7 @@ def mse(actual: npt.ArrayLike, expected: npt.ArrayLike) -> ValueOrError:
 def mape(actual: npt.ArrayLike, expected: npt.ArrayLike) -> ValueOrError:
     try:
         actual, expected = validate_arrays(actual, expected)
-        return ValueOrError(value=100 * np.abs(np.divide(np.subtract(actual, expected), actual)).mean())
+        return ValueOrError(value=100 * np.abs(np.divide(np.subtract(actual, expected), expected)).mean())
```
sthaha, line 130: I am not sure if this is right.
APE = [abs (Actual – Prediction) / Actual] *100
Actual = Metal values (see below)
Prediction = VM values  (see expected)

```
  - name: package - dynamic
    actual: |
      rate(
        kepler_{level}_package_joules_total{{
          job="{metal_job_name}",
          {vm_selector},
          mode="dynamic",
        }}[{rate_interval}]
      )
    expected: |
      rate(
        kepler_node_package_joules_total{{
          job="{vm_job_name}",
          mode="dynamic",
        }}[{rate_interval}]
      )
```

The current formula seems right to me. We should investigate why values in https://github.com/sustainable-computing-io/kepler-metal-ci/tree/main/docs/validation/2024-07-31/validator-v0.7.11-60-g8d305e7f is giving us `inf` instead.

sthaha (changes requested):
I think the current formula is right. We should use `actual` and `prediction` to make it clearer to the reader.

e2e/tools/validator/src/validator/prometheus/__init__.py
vimalk78: ok, this PR is created as per our earlier discussion

sthaha: True ! The use of expected and actual is quite confusing TBH. We should use actual and predicted or have a mapping to what actual and predicted really is.

PR #1656, opened 2024-07-31
feat(kubernetes): Use workqueues

dave-tucker:
This uses a workqueue as a layer of indirection between:
- Events coming in from the K8s ListWatcher
- The handlers that update ContainerStats

This is necessary for 2 reasons.

1. SharedIndexInformers do not deal well with large backlogs of events
2. ObjListWatcher.Mx is shared 😱 with the Collector.Mx. Meaning that we pause processing K8s events while preparing metrics... and similarly we pause preparing metrics when handling a K8s event.

We should revisit the design of the watcher to avoid sharing both `ContainerStats` and `Mx` as this is not a great design.

### DIFF
--- a/pkg/kubernetes/watcher.go
+++ b/pkg/kubernetes/watcher.go
@@ -22,12 +22,14 @@ import (
 	"sync"
 	"time"
 
-	k8sv1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


sthaha:
@dave-tucker as part of this PR, can we make use of field-selector to watch only pods running on the same node as kepler? 
We can pass the `node-name` that kepler is running on as an cli option to kepler using downward api. 

> We should revisit the design of the watcher to avoid sharing both ContainerStats and Mx as this is not a great design.

💯

sthaha (approved):

Issue #1649, opened 2024-07-30
Validator: synchronize timestamps for better analysis


vprashar2929:
To perform a more accurate comparison between what metal reports vs what VM, which generates values at slightly different timestamps it would be nice to synchronize their data points. For eg: metric A produces values at timestamps t1, t2, t3 and so on, while metric B produces values at t1+2, t2+3 and so on. By merging and aligning the nearest timestamps we can ensure that the data points from both metrics are closely matched in time.

vprashar2929:
closing as it is addressed in #1666

PR #1666, opened 2024-08-04
chore(validator): store test result in an object

sthaha:
All test results are stored in a single TestResult object so that generation of 
various formats (e.g. json) can be done easily. 

Also Fixes: 
 https://github.com/sustainable-computing-io/kepler/issues/1649

To verify values produced by Validator, now the Model Validation dashboard can be used. 
* Fixes Model Dashboard to correctly calculate MAPE and MSE for the selected time range

image

### DIFF
--- a/e2e/tools/validator/.gitignore
+++ b/e2e/tools/validator/.gitignore
@@ -1,3 +1,3 @@
 __pycache__
 validator.yaml
-tmp/
+tmp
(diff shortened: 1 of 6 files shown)


vprashar2929 (approved):

husky-parul (commented):
lgtm

Issue #1670, opened 2024-08-05
High Kepler CPU usage under normal workloads


vimalk78:
Without any load on system, kepler CPU usage goes upto 20%

vimalk78:
on latest [main](https://github.com/sustainable-computing-io/kepler/tree/cdcaf00ba6f33b1c3d7fb76fa077332890b1b0ab), if machine is loaded with stress-ng, the kepler cpu usage spikes. In comparison, the kepler before ringbuffer does not show increase in cpu if machine is loaded.


https://asciinema.org/a/AdY5eXkeysABoXxFWDsFO2iUx

vimalk78:
comparing with old code, some kepler cpu usage spike is understandable since some processing ( 3 map lookup, 2 update, 1 delete) was happening in kernel context and cpu cycles for these were accounted for in the kernel, which now happens in user space and gets counted as kepler cpu.

need to check if we can reduce the cpu spike in kepler when machine is loaded.

dave-tucker:
> need to check if we can reduce the cpu spike in kepler when machine is loaded.

exactly! I'm now able to reproduce with stress-ng and I'm working to keep that CPU spike as low as possible.

vimalk78:
i ran some `perf stat` tests to check impact of kepler on context switch time. idea being  that since kepler traps sched_switch and does some processing, it should have some impact on context switch time.  stress-ng is used in parallel to simulate load.

- without running kepler
```
root@bkr18:~# sudo perf stat -a -e sched:sched_switch --timeout 600000 # with no kepler with load

 Performance counter stats for 'system wide':

        90,480,301      sched:sched_switch                                                    

     600.105927296 seconds time elapsed
```

- with running kepler release-0.7.11
```
root@bkr18:~# sudo perf stat -a -e sched:sched_switch --timeout 600000 # with kepler 0.7.11 with load

 Performance counter stats for 'system wide':

        87,500,721      sched:sched_switch                                                    

     600.100293869 seconds time elapsed
```

- with running kepler latest (with ring buffer )

```
root@bkr18:~# sudo perf stat -a -e sched:sched_switch --timeout 600000 # with kepler latest with load

 Performance counter stats for 'system wide':

        79,620,228      sched:sched_switch                                                    

     600.099929726 seconds time elapsed

```

Observation: with kepler running, the number of context switches goes down, as expected. But with ring-buffer changes, the drop is more than 7-11 release.


[...]

Issue #1660, opened 2024-08-01
bpf: Investigate the *best* value for wakeup_data_size


dave-tucker:
### What would you like to be added?

This constant:
https://github.com/sustainable-computing-io/kepler/blob/main/bpf/kepler.bpf.c#L70

Declares how often we wake up to read the ringbuf.

The current math was as follows:

1. My system (on average) processes around 600-700 context switches per second
1. The sample period in Kepler is once every 3 seconds
1. We need to read at least one batch of ringbuf events within that 3 second interval

So that should have me read every 1.7ish seconds 😄 

### Why is this needed?

When kepler wakes up to read events it consumes CPU. Right now that's between 1-3%.
We should consider whether there is a better formula we could use to compute this magic number of 1000.

It could relate to the sample rate.

e.g `500 * SampleRate` and perhaps even the `500` could come from something better than an educated guess.

rootfs:
The Kepler CPU usage under normal and stress workloads need to be investigated in parallel. The latest stress test results point to a divergence that needs to be fixed.

rootfs:
The current Kepler CPU usage is now 20% without running load. 
[![asciicast](https://asciinema.org/a/670514.svg)](https://asciinema.org/a/670514)

rootfs:
[Test results](https://github.com/sustainable-computing-io/kepler/pull/1628#issuecomment-2266174953) posted on the original PR https://github.com/sustainable-computing-io/kepler/pull/1628

dave-tucker:
> The current Kepler CPU usage is now 20% without running load. [![asciicast](https://camo.githubusercontent.com/8b57338eb6dbf6a69e0d99506b01d91bd5a3e429392fc5c4e1cc9b62ab4d61e2/68747470733a2f2f61736369696e656d612e6f72672f612f3637303531342e737667)](https://asciinema.org/a/670514)

How, and on what machine, can I reproduce this result?

rootfs:
@dave-tucker load the kepler latest image and keep it running for a day.

dave-tucker:
> [Test results](https://github.com/sustainable-computing-io/kepler/pull/1628#issuecomment-2266174953) posted on the original PR #1628

Responded:
https://github.com/sustainable-computing-io/kepler/pull/1628#issuecomment-2269058775

PR #1671, opened 2024-08-05
fix(pkg/bpf): Use channel to process events

dave-tucker:
Processing events in the same goroutine as the ring buffer reader requires acquiring a mutex, which blocks ringbuf event processing causing a backlog. To avoid this, send events via a buffered channel to a dedicated event processing goroutine to ensure that the ringbuf remains unblocked. This has decreased CPU load from 1-3% on my machine to 0-1% CPU load.

Additionally adds some metrics to track events read, events processed and the current depth of the events channel.

Updates: #1670, #1660


dave-tucker:
@vimalk78 @rootfs if you'd like to test this one to see if see a difference on your systems.
Running with `-bpf-debug-metrics` will also help me potentially confirm/deny suspicions on what's causing the CPU usage and help inform any additional work on #1660

rootfs:
let's merge this so we run this on the CI

rootfs (approved):

PR #1676, opened 2024-08-06
chore: cleanup globals in pkg/model

maryamtahhan:


rootfs (approved):

pkg/model/model.go
```diff
@@ -31,14 +31,11 @@ import (
...
-	gauge     = true  // gauge is used to define if the function will update a gauge value
-	counter   = false // gauge is used to define if the function will NOT update a gauge value, but instead a counter value
-)
-
-var (
-	EstimatorSidecarSocket = "/tmp/estimator.sock"
+	idlePower              = true  // idlePower is used to define if the function will update idle power
+	absPower               = false // absPower is used to define if the function will NOT update idle power, but instead an absolute power
+	gauge                  = true  // gauge is used to define if the function will update a gauge value
+	counter                = false // gauge is used to define if the function will NOT update a gauge value, but instead a counter value
```
vimalk78, line 37:
```suggestion
	counter                = false // counter is used to define if the function will update a counter value
```

vimalk78 (commented):
suggest to similarly change
-  variable `NodeComponentPowerModel` in `node_component_energy.go`
- corresponding `Create` methods `CreateProcessPowerEstimatorModel`,  `CreateNodePlatformPoweEstimatorModel`, `CreateNodeComponentPoweEstimatorModel`

pkg/model/model.go
vimalk78: why changed these from `const` to `var` ?

maryamtahhan: I changed `estimatorSidecarSocket = "/tmp/estimator.sock` from var to const because it's value is not updated anywhere that I could see. 
The variables above were always const. I didn't change anything only the alignment of the `=` to match the new const below

vimalk78: aah ! my mistake. sorry

vimalk78:
LGTM

PR #1683, opened 2024-08-09
chore: update git ignore

maryamtahhan:
update gitignore to ignore c object files in the repo.


sthaha (changes requested):
I thought we wanted to checkin these files, didn't we @dave-tucker  ?

dave-tucker:
> I thought we wanted to checkin these files, didn't we @dave-tucker ?

Yes - although I'd rather we didn't.
There were issues with the `make lint` not working if the object files weren't created - even if the `generate` target (which creates these files) was a dependency.
If @maryamtahhan has fixed that, I'd be very happy to merge 😄

dave-tucker:
oh, the other issue is that immediately after checkout, your IDE will get really upset because those files don't exist, which is a bit of an annoying dev experience.

maryamtahhan:
looking into it

maryamtahhan:
so after spending a little bit of time on this, I can see that modifying the generate target to run in a bunch of places is not enough. The main issue is that we would need to deep dive into each github workflow and modify the appropriate locations with make generate... so for now I'm proposing just to update the generate target in the Makefile https://github.com/sustainable-computing-io/kepler/pull/1698 to build everything at once...

PR #1684, opened 2024-08-09
feat: add machine spec generator/reader for model weight request

sunya-ch:
This PR works on the issue https://github.com/sustainable-computing-io/kepler-model-server/issues/231 to have Kepler provides node spec to the model server for model selection. 

This needs new two dependency modules.

```
	"github.com/shirou/gopsutil/cpu"
	"github.com/shirou/gopsutil/mem"
```

Spec can be provided via the configmap. According to @vimalk78 's comment, I add a new manifest build option `MACHINE_SPEC_DEPLOY` to deploy default machine spec config map and patch to select the spec by the label.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kepler-machine-spec
  namespace: system
data:
  m5.metal: |
    {"processor": "intel_xeon_platinum_8259cl", "cores": 96, "chips": 2, "memory": 377, "frequency": 3500}
  i3.metal: |
    {"processor": "intel_xeon_e5_2686v4", "cores": 72, "chips": 2, "memory": 503, "frequency": 3000}
    ...
```

Example output:
```
I0813 06:36:25.599010       1 config.go:260] Read machine spec from config: { intel_xeon_e5_2686v4 72 2 503 3000 0}
```

If not provided, it will be automatically generated.
Example output:
```
I0813 06:43:19.299963       1 config.go:269] Generate machine spec: &{genuineintel intel_xeon_e5_2667v3 32 1 377 3600 2}
```

### Log from validation docker compose (kepler with only model-server)
```
kepler-1        | I0816 06:53:44.621569  202978 config.go:269] Generate machine spec: &{genuineintel intel_core 4 1 0 3200 1}
model-server-1  | INFO:kepler_model.server.model_server:get request /model: {'metrics': ['bpf_cpu_time_ms', 'bpf_net_tx_irq', 'bpf_net_rx_irq', 'bpf_block_irq', 'bpf_page_cache_hit', 'cpu_architecture'], 'output_type': 'AbsPower', 'source': 'acpi', 'node_type': 1, 'weight': True, 'trainer_name': 'SGDRegressorTrainer', 'filter': '', 'spec': {'vendor': 'genuineintel', 'processor': 'intel_core', 'cores': 4, 'chips': 1, 'memory': 0, 'frequency': 3200, 'threads_per_core': 1}}
model-server-1  | WARNING:kepler_model.server.model_server:/mnt/models/specpower-0.7.11/acpi/AbsPower/BPFIRQ has no matched model for node type=1, try all available models
model-server-1  | INFO:kepler_model.server.model_server:response: model SGDRegressorTrainer_1 by ['bpf_cpu_time_ms', 'bpf_page_cache_hit'] with mae=9.85 selected with uncertainty=0, looseness=0
model-server-1  | INFO:werkzeug:172.18.0.3 - - [16/Aug/2024 06:53:45] "POST /model HTTP/1.1" 200 -
kepler-1        | I0816 06:53:45.415611  202978 model.go:120] Machine Spec: &{genuineintel intel_core 4 1 0 3200 1}
kepler-1        | I0816 06:53:45.415947  202978 node_platform_energy.go:52] Using the Regressor/AbsPower Power Model to estimate Node Platform Power
kepler-1        | I0816 06:53:45.438573  202978 config.go:269] Generate machine spec: &{genuineintel intel_core 4 1 0 3200 1}
model-server-1  | INFO:kepler_model.server.model_server:get request /model: {'metrics': ['bpf_cpu_time_ms', 'bpf_net_tx_irq', 'bpf_net_rx_irq', 'bpf_block_irq', 'bpf_page_cache_hit', 'cpu_architecture'], 'output_type': 'AbsPower', 'source': 'intel_rapl', 'node_type': 1, 'weight': True, 'trainer_name': 'SGDRegressorTrainer', 'filter': '', 'spec': {'vendor': 'genuineintel', 'processor': 'intel_core', 'cores': 4, 'chips': 1, 'memory': 0, 'frequency': 3200, 'threads_per_core': 1}}
model-server-1  | INFO:kepler_model.server.model_server:response: model SGDRegressorTrainer_1 by ['bpf_cpu_time_ms', 'bpf_page_cache_hit'] with mae=18.35 selected with uncertainty=0, looseness=0
model-server-1  | INFO:werkzeug:172.18.0.3 - - [16/Aug/2024 06:53:45] "POST /model HTTP/1.1" 200 -
kepler-1        | I0816 06:53:45.709751  202978 model.go:120] Machine Spec: &{genuineintel intel_core 4 1 0 3200 1}
kepler-1        | I0816 06:53:45.713385  202978 node_component_energy.go:56] Using the Regressor/AbsPower Power Model to estimate Node Component Power
```

Next tasks:
- [ ] update kepler doc
- [x] https://github.com/sustainable-computing-io/kepler-model-server/pull/370 -  update kepler-model-server selector logic to find similarity of spec and the trained model


pkg/config/config.go
```diff
@@ -245,6 +247,22 @@ func GetMockACPIPowerPath() string {
 	return MockACPIPowerPath
 }
 
+// GetMachineSpec initializes map of MachineSpecValues from MACHINE_SPEC
+func GetMachineSpec() *MachineSpec {
+	var spec MachineSpec
+	if specStr := getConfig(MachineSpecKey, ""); specStr != "" {
```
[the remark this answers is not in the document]
sunya-ch: @vimalk78 Good idea. I don't know that whether can we set configmap mount name from labels directly but one way I can implement is to have a confimap that contains mapping from label value to the machine spec. Then, in kepler, we read label value from environment (using fieldRef) to select the matched spec if exists. WDYT?

pkg/config/config.go
```diff
@@ -107,6 +109,8 @@ var (
 	redfishSkipSSLVerify          = getBoolConfig("REDFISH_SKIP_SSL_VERIFY", true)
 
 	////////////////////////////////////
+	MachineIDKey        = "MACHINE_ID"
+	MachineSpecPath     = "/etc/machine_spec"
```
sthaha, line 113: Can we keep all kepler related setting under `/etc/kepler/<component>/` ?

pkg/config/config.go
```diff
@@ -245,6 +248,27 @@ func GetMockACPIPowerPath() string {
 	return MockACPIPowerPath
 }
 
+// GetMachineSpec initializes map of MachineSpecValues from MACHINE_SPEC
+func GetMachineSpec() *MachineSpec {
```
sthaha, line 252: Could we write a unit-test for this? 
Could you also please consider refactoring this to avoid deeply nested paths? 


```
// GetMachineSpec initializes a map of MachineSpecValues from MACHINE_SPEC
func GetMachineSpec() *MachineSpec {
	// Retrieve the machine ID from configuration
	machineID := getConfig(MachineIDKey, "")
	if machineID == "" {
		// If machine ID is not found, generate a default machine spec
		return generateSpec()
	}

	specPath := filepath.Join(MachineSpecPath, machineID)
	spec, err := readMachineSpec(specPath)
	if err  != nil {
 		klog.Errorf("failed to read spec file: %s: %v; using default spec", specPath, err);
 		return generateSpec()
	}
 	return spec
}

func readMachineSpec(path string) (MachineSpec, error) {
	...
}
```

pkg/config/spec.go
```diff
@@ -0,0 +1,127 @@
...
+	vendor = strings.ReplaceAll(vendor, ",", "")
+	vendor = strings.ReplaceAll(vendor, "'", "")
+	return vendor
+}
+
+func roundToNearestHundred(value float64) int {
+	return int(math.Round(value/100) * 100)
+}
+
+func generateSpec() *MachineSpec {
```
sthaha, line 87: Can we rewrite test for these as well :) ?

manifests/k8s/config/exporter/patch/patch-machine-spec.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: kepler-exporter
+          env:
+            - name: MACHINE_ID
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.labels['kepler-profile']
```
sthaha, line 15: Would't this set the same profile for all nodes that kepler is deployed to? 
Consider a cluster with different types of nodes. Since label and value on the daemonset is going to be the same, this would force all pods (deployed on each node) to have the same machine spec loaded. 

I think we need to rethink the idea of using daemonsets for kepler since we will want to load node-specific configuration for each pod.

sunya-ch: For the current design, we need one kepler ds for each group of node (with node selector).

sthaha: Isn't that equal to using `Pod` instead of `Daemonset`?

pkg/config/config.go
```diff
@@ -107,6 +107,8 @@ var (
 	redfishSkipSSLVerify          = getBoolConfig("REDFISH_SKIP_SSL_VERIFY", true)
 
 	////////////////////////////////////
+	MachineIDKey        = "MACHINE_ID"
```
vimalk78, line 110: does it make more sense to rename it is `MACHINE_SPEC_ID` ?

pkg/config/spec.go
```diff
@@ -0,0 +1,147 @@
...
+	spec := &MachineSpec{}
+	spec.Cores = -1
+	spec.Memory = -1
+	spec.Frequency = -1
+	spec.Chips = 1
```
vimalk78, line 96: this field is not set. the example in config map shows `chips: 2`

pkg/config/spec.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// MachineSpec defines a machine spec to submit for power model selection
+type MachineSpec struct {
+	Vendor         string `json:"vendor"`
+	Processor      string `json:"processor"`
+	Cores          int    `json:"cores"`
+	Chips          int    `json:"chips"`
+	Memory         int    `json:"memory"`
+	Frequency      int    `json:"frequency"`
```
vimalk78, line 145: along with CPU Freq, the `lscpu` commad also gives these fields
```
    CPU(s) scaling MHz:   19%
    CPU max MHz:          4800.0000
    CPU min MHz:          800.0000
```
Cpu Freq scaling has a direct impact on the power consumed by cpu, and we should consider this in power model. may be in another PR if not in this PR.



perhaps the go module does not give these fields.

vimalk78: please include  `"CPU family"` and `"Model"` in the model 
```
Vendor ID:                GenuineIntel
  Model name:             11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
    CPU family:           6
    Model:                141
```

manifests/k8s/config/exporter/patch/patch-machine-spec.yaml
sthaha: Is this relevant anymore? shouldn't`--machine-profile` be added to the args instead?

pkg/config/spec.go
```diff
@@ -0,0 +1,155 @@
...
+func roundToNearestHundred(value float64) int {
+	return int(math.Round(value/100) * 100)
+}
+
+func generateSpec() *MachineSpec {
+	spec := &MachineSpec{}
+	spec.Cores = -1
+	spec.Memory = -1
+	spec.Frequency = -1
+	spec.Chips = 1
```
sthaha, line 96: I am also wondering if the default 0 value is good enough?

pkg/config/spec.go
```diff
@@ -0,0 +1,155 @@
...
+	if err == nil {
+		if specBytes, err := io.ReadAll(specFile); err == nil {
+			if err = json.Unmarshal(specBytes, &spec); err == nil {
+				klog.Infof("Read machine spec from config: %v", spec)
+				return spec, nil
+			} else {
+				klog.Infof("failed to unmarshal machine spec %s: %v", string(specBytes), err)
+			}
+		}
+	}
```
sthaha, line 142: An idiom in go is to exit early and IMHO, makes the code a bit more readable. 
e.g. this can be reorganised as ... 
Note 
* returning early,
* use of nil
* fmt.Errorf(" ... %w", err) to provide more details to the caller

```
func readMachineSpec(path string) (*MachineSpec, error) {
	specFile, err := os.Open(path)
	if err != nil {
		return nil, fmt.Errorf("failed to open machine spec: %w", err)
	}

	specBytes, err := io.ReadAll(specFile);
	if err != nil {
		return nil, fmt.Errorf("failed to read machine spec: %w", err)
	}

	if err = json.Unmarshal(specBytes, &spec); err != nil {
		return nil, fmt.Errorf("failed to unmarshal machine spec: %v", err")
	}

	klog.V(3).Infof("Read machine spec from config: %v", spec)
	return spec, nil
}
```

cmd/exporter/exporter.go
```diff
@@ -79,6 +79,7 @@ var (
 	apiserverEnabled             = flag.Bool("apiserver", true, "if apiserver is disabled, we collect pod information from kubelet")
 	redfishCredFilePath          = flag.String("redfish-cred-file-path", "", "path to the redfish credential file")
 	exposeEstimatedIdlePower     = flag.Bool("expose-estimated-idle-power", false, "estimated idle power is meaningful only if Kepler is running on bare-metal or when there is only one virtual machine on the node")
+	machineSpecFilePath          = flag.String("machine-profile", "", "path to the machine spec file in json format")
```
sthaha, line 82:
```suggestion
	machineSpecFilePath          = flag.String("machine-profile", "", "path to the machine spec file in json format")
```
Lets use a single terminology ..`machine-spec` or `machine-profile` I am okay with both.  Since you have already called it `machine-spec` and to avoid having to change it everywhere, lets use `--machine-spec` 

sorry about the confusion 👼

pkg/config/spec.go
sunya-ch: Just to make it clear for invalid value. However, zero is also fine as we will never have a zero value for those attributes. Will update.

manifests/k8s/config/exporter/kustomization.yaml
```diff
@@ -6,8 +6,10 @@ resources:
 #  - prometheus_common_service_monitor.yaml
 #  - prometheus_common_rules.yaml
 #  - prometheus_high_granularity_rules.yaml
+# uncomment this line for default machine spec configmap
+#  - machine_spec_configmap.yaml
 
-patchesStrategicMerge: []
+patchesStrategicMerge: # []
```
sthaha, line 12:
```suggestion
patchesStrategicMerge:  []
```

pkg/config/config.go
sunya-ch: Removed.

pkg/config/spec.go
sunya-ch: Updated.

hack/build-manifest.sh
```diff
@@ -203,6 +204,16 @@ deploy_habana() {
...
 	ok "Habana deployment configured"
 }
+deploy_machine_spec() {
+	header "Machine Spec Deployment"
+	$MACHINE_SPEC_DEPLOY || {
+		skip "skipping machine spec deployment"
+		return 0
+	}
+	uncomment machine_spec_configmap "${MANIFESTS_OUT_DIR}"/exporter/kustomization.yaml
+	uncomment_patch machine-spec "${MANIFESTS_OUT_DIR}"/exporter/kustomization.yaml
```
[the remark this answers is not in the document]
sunya-ch: Have this only with the option "MACHINE_SPEC_DEPLOY".

pkg/config/spec.go
sunya-ch: Need update in both kepler-model-server. Need investigation and discussion later.

pkg/config/spec.go
sunya-ch: Removed.

sunya-ch:
Note:
notice duplication of MachineSpec definition in validator.

https://github.com/sustainable-computing-io/kepler/blob/de1aad9a9ba9dda3983a59df2c6f0d8e9cdb4769/e2e/tools/validator/src/validator/specs/__init__.py#L79

Need to make convergence.

pkg/model/estimator/local/regressor/regressor.go
```diff
@@ -32,20 +32,16 @@ import (
...
-	TrainerName  string   `json:"trainer_name"`
-	SelectFilter string   `json:"filter"`
+	MetricNames  []string           `json:"metrics"`
+	OutputType   string             `json:"output_type"`
+	EnergySource string             `json:"source"`
+	NodeType     int                `json:"node_type"`
+	Weight       bool               `json:"weight"`
+	TrainerName  string             `json:"trainer_name"`
+	SelectFilter string             `json:"filter"`
+	Spec         config.MachineSpec `json:"spec"`
```
sthaha, line 44: minor: spec too generic imo 

```suggestion
	Spec         config.MachineSpec `json:"machine_spec"`
```

go.mod
```diff
@@ -18,8 +18,9 @@ require (
 	github.com/onsi/ginkgo/v2 v2.20.0
 	github.com/onsi/gomega v1.34.1
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.20.0
-	github.com/prometheus/prometheus v0.54.0
+	github.com/prometheus/client_golang v1.19.1
+	github.com/prometheus/prometheus v0.53.1
```
sthaha, line 22: unrelated changes?

pkg/config/config.go
```diff
@@ -245,6 +247,22 @@ func GetMockACPIPowerPath() string {
...
+}
+
+// GetMachineSpec initializes a map of MachineSpecValues from MACHINE_SPEC
+func GetMachineSpec() *MachineSpec {
+	if machineSpecFilePath != "" {
+		if spec, err := readMachineSpec(machineSpecFilePath); err == nil {
+			return spec
+		} else {
+			klog.Errorf("failed to read spec from %s: %v", machineSpecFilePath, err)
+		}
```
sthaha, line 261: minor: 
```suggestion
		}
		klog.Errorf("failed to read spec from %s: %v", machineSpecFilePath, err)
```

In this case, shouldn't we return an error and fail?

pkg/model/estimator/local/regressor/regressor.go
sunya-ch: After the PR in kepler-model-server https://github.com/sustainable-computing-io/kepler-model-server/pull/411 merged, I will update.

pkg/config/config.go
sunya-ch: Updated to warning.

go.mod
sunya-ch: Removed the conflict.

rootfs:
@sthaha @vimalk78 is it ready to merge?

sthaha (approved):

PR #1686, opened 2024-08-09
chore: cleanup watcher

maryamtahhan:


pkg/kubernetes/watcher.go
```diff
@@ -107,15 +107,15 @@ func NewObjListWatcher(bpfSupportedMetrics bpf.SupportedMetrics) *ObjListWatcher
...
+		options.FieldSelector = fields.Set{"spec.nodeName": stats.NodeName}.AsSelector().String() // to filter events per node
 	}
 	objListWatcher := cache.NewFilteredListWatchFromClient(
 		w.k8sCli.CoreV1().RESTClient(),
 		w.ResourceKind,
 		metav1.NamespaceAll,
 		optionsModifier,
 	)
-	w.informer = cache.NewSharedIndexInformer(objListWatcher, &corev1.Pod{}, 0, cache.Indexers{})
+	w.informer = cache.NewSharedIndexInformer(objListWatcher, &corev1.Pod{}, time.Minute*5, cache.Indexers{})
```
dave-tucker, line 118: This forces a full resync every 5 mins? Why?

pkg/kubernetes/watcher.go
```diff
@@ -240,26 +241,25 @@ func (w *ObjListWatcher) handleAdd(obj interface{}) {
 			klog.V(5).Infof("Pod %s %s is ready with %d container statuses, %d init container status, %d ephemeral statues",
 				pod.Name, pod.Namespace, len(pod.Status.ContainerStatuses), len(pod.Status.InitContainerStatuses), len(pod.Status.EphemeralContainerStatuses))
 			w.Mx.Lock()
-			err1 := w.fillInfo(pod, pod.Status.ContainerStatuses)
-			err2 := w.fillInfo(pod, pod.Status.InitContainerStatuses)
-			err3 := w.fillInfo(pod, pod.Status.EphemeralContainerStatuses)
+			w.fillInfo(pod, pod.Status.ContainerStatuses)
```
dave-tucker, line 244: It's probably better to return an error here?
If we bubble that error back up to `handleEvent` then the item will get requeued (after backoff) so we can attempt to get the necessary ID's again.

dave-tucker: Also note the `continue` on line 238 above, that needs fixing too 😄

dave-tucker (changes requested):
Thanks for digging into this @maryamtahhan

These issues are likely my fault in the #1656 PR.
We moved from being edge-driven to level-driven.

In the previous `edge-driven' logic we'd do stuff like - "oh well, we couldn't find the info now so just skip it... we'll get an update event some time and the info will be there at that point".

In the new 'level-driven' logic the workqueue doesn't differentiate between an `add` or `update` event. We just know that we have "work" to do for a given key.
In that case, if the pod object isn't in the right state we need to error out - which will cause that item to get re-queued so we can try again (after a backoff).

pkg/kubernetes/watcher.go
maryamtahhan: Cool, I will fix this up. Thanks for the suggestion

pkg/kubernetes/watcher.go
maryamtahhan: I thought it might be good to resync periodically - but I can set this back to 0

maryamtahhan:
Still tidying please - pls ignore for the minute

maryamtahhan:
Ok - this is ready now

pkg/kubernetes/watcher.go
```diff
@@ -225,39 +230,46 @@ func (w *ObjListWatcher) Stop() {
...
 	}
+	return errors.New("pod not ready")
 }
 
-func (w *ObjListWatcher) fillInfo(pod *corev1.Pod, containers []corev1.ContainerStatus) error {
-	var err error
+func (w *ObjListWatcher) fillInfo(pod *corev1.Pod, containers []corev1.ContainerStatus, mx *sync.Mutex) error {
 	var exist bool
+	var err error
+	mx.Lock()
```
maryamtahhan, line 269: I can move this back out - but the code did look a bit ugly:

```
func (w *ObjListWatcher) handleAdd(obj interface{}) error {
	switch w.ResourceKind {
	case podResourceType:
		pod, ok := obj.(*corev1.Pod)
		if !ok {
			err := fmt.Errorf("could not convert obj: %v", w.ResourceKind)
			return err
		}
		for _, condition := range pod.Status.Conditions {
			if condition.Type != corev1.ContainersReady {
				err := fmt.Errorf("pod not ready yet: %v", pod.Name)
				return err
			}
			klog.V(5).Infof("Pod %s %s is ready with %d container statuses, %d init container status, %d ephemeral statues",
				pod.Name, pod.Namespace, len(pod.Status.ContainerStatuses), len(pod.Status.InitContainerStatuses), len(pod.Status.EphemeralContainerStatuses))
			w.Mx.Lock()
			if err := w.fillInfo(pod, pod.Status.ContainerStatuses); err != nil {
				w.Mx.Unlock()
				return err
			}
			if err := w.fillInfo(pod, pod.Status.InitContainerStatuses); err != nil {
				w.Mx.Unlock()
				return err
			}
			if err := w.fillInfo(pod, pod.Status.EphemeralContainerStatuses); err != nil {
				w.Mx.Unlock()
				return err
			}
			w.Mx.Unlock()
			klog.V(5).Infof("parsing pod %s %s status: %v %v %v", pod.Name, pod.Namespace, pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses, pod.Status.EphemeralContainerStatuses)
		}
	default:
		err := fmt.Errorf("watcher does not support object type %s", w.ResourceKind)
		return err
	}
	return nil
}
```

pkg/kubernetes/watcher.go
```diff
@@ -225,39 +230,46 @@ func (w *ObjListWatcher) Stop() {
...
 			}
 			klog.V(5).Infof("Pod %s %s is ready with %d container statuses, %d init container status, %d ephemeral statues",
 				pod.Name, pod.Namespace, len(pod.Status.ContainerStatuses), len(pod.Status.InitContainerStatuses), len(pod.Status.EphemeralContainerStatuses))
-			w.Mx.Lock()
-			err1 := w.fillInfo(pod, pod.Status.ContainerStatuses)
-			err2 := w.fillInfo(pod, pod.Status.InitContainerStatuses)
-			err3 := w.fillInfo(pod, pod.Status.EphemeralContainerStatuses)
-			w.Mx.Unlock()
-			klog.V(5).Infof("parsing pod %s %s status: %v %v %v", pod.Name, pod.Namespace, err1, err2, err3)
+			if err := w.fillInfo(pod, pod.Status.ContainerStatuses, w.Mx); err != nil {
```
dave-tucker, line 247:
```go
for statuses := range [pod.Status.ContainersStatuses, pod.Status.InitContainerStatuses, pod.Status.EphemeralContainerStatuses] {
    if err := if err := w.fillInfo(pod, statuses, w.Mx); err !=nil {
        return err
    }
}
```

Actually we could go one better and just concatenate those arrays to avoid iterating again inside `fillInfo`

maryamtahhan: I will tidy it up on Monday morning

maryamtahhan:
It seems like the requeuing logic is not working as is. When we bubble up the error to readd the item to the workqueue, I never see the item processed again even though the code path is a requeuing path...

rootfs:
@maryamtahhan if this pass the e2e multiple times without error, it is good to go. cc @vprashar2929

maryamtahhan:
> @maryamtahhan if this pass the e2e multiple times without error, it is good to go. cc @vprashar2929

I'm still debugging some issues with the requeuing part... but as soon as it passes multiple times I will flag it as ready

pkg/kubernetes/watcher.go
```diff
@@ -159,41 +162,50 @@ func (w *ObjListWatcher) processNextItem() bool {
...
 func (w *ObjListWatcher) handleEvent(key string) error {
 	obj, exists, err := w.informer.GetIndexer().GetByKey(key)
 	if err != nil {
-		klog.Errorf("Fetching object with key %s from store failed with %v", key, err)
+		klog.Errorf("fetching object with key %s from store failed with %v", key, err)
 		return err
 	}
 	if !exists {
-		w.handleDeleted(obj)
+		if err := w.handleDeleted(obj); err != nil {
```
maryamtahhan, line 193:
```suggestion
		return w.handleDeleted(obj); 
```

pkg/kubernetes/watcher.go
```diff
@@ -159,41 +162,50 @@ func (w *ObjListWatcher) processNextItem() bool {
...
+			return err
+		}
 	} else {
-		w.handleAdd(obj)
+		if err := w.handleAdd(obj); err != nil {
```
maryamtahhan, line 197:
```suggestion
		return w.handleAdd(obj); 
```

maryamtahhan:
@rootfs can you retrigger the workflow verification. It passed once. I just want to make sure it passes a second time.

hack/verify.sh
```diff
@@ -76,8 +77,6 @@ watch_service() {
 
 intergration_test() {
 	log_kepler &
-	# dump metrics before running tests
-	exec_kepler &
```
sthaha, line 80: can we restore this?

pkg/kubernetes/watcher.go
```diff
@@ -138,7 +140,8 @@ func NewObjListWatcher(bpfSupportedMetrics bpf.SupportedMetrics) *ObjListWatcher
 		},
 	})
 	if err != nil {
-		klog.Fatalf("%v", err)
+		klog.Errorf("%v", err)
+		return nil
```
sthaha, line 144:
```suggestion
		return err
```

pkg/kubernetes/watcher.go
```diff
@@ -225,42 +234,46 @@ func (w *ObjListWatcher) Stop() {
...
 			w.Mx.Lock()
-			err1 := w.fillInfo(pod, pod.Status.ContainerStatuses)
-			err2 := w.fillInfo(pod, pod.Status.InitContainerStatuses)
-			err3 := w.fillInfo(pod, pod.Status.EphemeralContainerStatuses)
+			for _, statuses := range [][]corev1.ContainerStatus{pod.Status.InitContainerStatuses, pod.Status.EphemeralContainerStatuses, pod.Status.ContainerStatuses} {
+				if err = w.fillInfo(pod, statuses); err != nil {
+					w.Mx.Unlock()
+					return err
+				}
+			}
```
sthaha, line 259: The previous implementation used to 
1. lock
2. try and fill Containers
3. try and fill InitContainers
4. try and fill EphemeralContainers
5. unlock

Whereas the current implementation 

1. lock
2. try and fill InitContainers
      - unlock and exit on error
3. try and fill EphemeralContainers
    - unlock and exit on error
4. try and fill Containers
    - unlock and exit on error
8. unlock on no errors

While I like the change of order, (Init, Emphermal, Containers), I think we shouldn't exit early. Rather preserve the old behaviour which is to continue to process the rest unless there is a good reason not to.

pkg/manager/manager.go
```diff
@@ -53,9 +54,14 @@ func New(bpfExporter bpf.Exporter) *CollectorManager {
...
 	// configure the watcher
 	manager.Watcher = kubernetes.NewObjListWatcher(supportedMetrics)
+	if manager.Watcher == nil {
+		klog.Error("could not create the watcher")
+	}
 	manager.Watcher.Mx = &manager.PrometheusCollector.Mx
 	manager.Watcher.ContainerStats = manager.StatsCollector.ContainerStats
-	manager.Watcher.Run()
+	if manager.Watcher.Run() != nil {
+		klog.Error("could not run the watcher")
```
sthaha, line 63: can we also print the error?

pkg/kubernetes/watcher.go
maryamtahhan: I didn't change the function to return errors:

```
func NewObjListWatcher(bpfSupportedMetrics bpf.SupportedMetrics) *ObjListWatcher
```

which is why I was returning nil, I can update the function to also return the error, WDYT?

pkg/manager/manager.go
maryamtahhan: klog.Error will print the error

maryamtahhan:
Looking @ https://github.com/submariner-io/admiral/blob/cb27a3b8a9a1e37affcb8f1f695daf9fd30c3b16/pkg/workqueue/queue.go#L39 and https://github.com/submariner-io/admiral/blob/cb27a3b8a9a1e37affcb8f1f695daf9fd30c3b16/pkg/syncer/resource_syncer.go#L120 to see how we can fixup the requeuing issues we are seeing

sthaha:
@maryamtahhan , the last run failed with the following error 
```
created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start in goroutine 9
	/workspace/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:70 +0x73
panic: interface conversion: interface {} is *v1.Pod, not *v1.Event [recovered]
	panic: interface conversion: interface {} is *v1.Pod, not *v1.Event
```
https://github.com/sustainable-computing-io/kepler/actions/runs/10390788824/job/28782102450?pr=1686#step:11:135

maryamtahhan:
> @maryamtahhan , the last run failed with the following error 
> ```
> created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start in goroutine 9
> 	/workspace/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:70 +0x73
> panic: interface conversion: interface {} is *v1.Pod, not *v1.Event [recovered]
> 	panic: interface conversion: interface {} is *v1.Pod, not *v1.Event
> ```
> https://github.com/sustainable-computing-io/kepler/actions/runs/10390788824/job/28782102450?pr=1686#step:11:135
> 
> On a related note, the way I wanted to solve this issue was to make use of controller-runtime/ package, use manager  reconciler pattern and let the framework deal with setting up the work queue and shared-informer factory than handcode it ourselves. 
> 

Ok, I will timebox fixing this to today as it stands. If I can't progress it we can close the PR and move to the controller-runtime manager... 

I guess the 2 main issues we're running into now are:
- the mismatch in the criteria for pod selection for the e2e test vs where kepler is deployed (as @vprashar2929 suggested this will be resolved by modifying the kepler toleration to run on the controller nodes also). 
- The watcher processing of requeued pods whose containers don't have a containerID. 

Regarding the containerID issue I guess the part I find a little odd from the logs that we print for the kepler pod/container is that the kepler watcher is subscribed to pod events for its node. So when it receives a pod event for the kepler pod on its node, then why would the kepler ContainerID be empty... in order to have the logs in the first place the kepler container must be running, so it should have a ContainerID? I will dig into this a little more today during the time box...

maryamtahhan:
Ok - I found the end of the issues (workqueue was shuttingdown early, hence the lack of updates after requeuing). It's passed the CI once. 

I'm going to collapse the commits into one which will force another check...

dave-tucker (approved):

PR #1687, opened 2024-08-09
[DO NOT MERGE] Ringbuf redux

dave-tucker:
Brings back the ringbuf work, just to see if it's something we want to continue with or not.


vimalk78:
should we start a new branch for this change?

Issue #1640, opened 2024-07-25
Kepler latest reporting wrong command in process metrics


vprashar2929:
### What happened?

When Kepler using `latest` is deployed it reports the wrong command in the `kepler_process` metrics. 
Attaching screenshot for a comparison between what Kepler's latest and Scaphendre reports:

![Screenshot 2024-07-25 at 11 13 34 PM](https://github.com/user-attachments/assets/6f673294-3a5a-4bef-aee9-57012813f40b)

Here pid=`503764` which is actually pid for qemu process running on the machine
```
❯ pstree -p | grep 503764
           |-qemu-system-x86(503764)-+-{qemu-system-x86}(503771)
```

### What did you expect to happen?

Kepler should report the command name correctly in the metrics

### How can we reproduce it (as minimally and precisely as possible)?

Deploy Kepler latest using Docker compose and then check for commands in the Kepler process metrics for one of the pid

### Kepler image tag

<details>
latest
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE}
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE}
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

vprashar2929:
Verified locally and now Kepler latest is producing the correct command name for a particular pid

![Screenshot 2024-07-31 at 12 29 28 PM](https://github.com/user-attachments/assets/422bad72-1a48-42b8-9cfd-ae648c9ad9fe)

Thanks @dave-tucker for the fix!

vimalk78:
fixed by #1693

PR #1693, opened 2024-08-14
fix(bpf): use prev_tgid to register process

vimalk78:
In release 7.11 `bpf_get_current_pid_tgid()` was used to get the `tgid` to register new process, which refers to the task being switched out. reverting it back to same behavior using `prev_tgid`.


vprashar2929:
Compared locally on BM against `release-0.7.11` and `latest`. Kepler with `release-0.7.11` and with this patch reporting similar value. It also address #1640 

screencapture-192-168-0-110-9090-graph-2024-08-14-12_46_44

In the above:
* `job="latest"` is using `release-0.7.11`
* `job="dev"` is using this patch
* `job="new"` is using `latest`

vprashar2929 (approved):

PR #1698, opened 2024-08-15
chore: remove go generate build targets from the repo

maryamtahhan:
This PR removes all the bpf2go generated files from the codebase and updates the makefile `generate` target 
it also introduces a go tag `bpf` and a mock bpf file to keep golinterci happy.

Fixes: https://github.com/sustainable-computing-io/kepler/pull/1683#issuecomment-2291242515


pkg/bpf/mockbpf.go
```diff
@@ -1,17 +1,31 @@
...
 import (
 	"bytes"
-	_ "embed"
 	"fmt"
 	"io"
 
 	"github.com/cilium/ebpf"
 )
 
+//nolint:stylecheck // This is just a dummy file
```
dave-tucker, line 28: Does this mean that we need to update the mock when the struct changes?
bpf2go autogenerates this for us.

pkg/bpf/mockbpf.go
```diff
@@ -94,16 +60,11 @@ type keplerObjects struct {
...
-		&o.keplerPrograms,
-		&o.keplerMaps,
-	)
-}
-
 // keplerMaps contains all maps after they have been loaded into the kernel.
 //
 // It can be passed to loadKeplerObjects or ebpf.CollectionSpec.LoadAndAssign.
+//
+//nolint:stylecheck // This is just a dummy file
```
dave-tucker, line 67: Same comment here re: Maps as this is autogenerated too

pkg/sensors/platform/source/redfish.go
```diff
@@ -270,7 +270,7 @@ func (rf *RedFishClient) IsSystemCollectionSupported() bool {
 			}
 		}
 	}()
-	return rf.systems != nil && len(rf.systems) > 0
+	return len(rf.systems) > 0
```
dave-tucker, line 273: 👀 is this related?

Makefile
```diff
@@ -41,7 +41,7 @@ export CTR_CMD     ?= $(or $(shell command -v podman), $(shell command -v docker
 # E.g. --tls-verify=false for local develop when using podman
 CTR_CMD_PUSH_OPTIONS ?=
 
-GENERAL_TAGS := include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo
+GENERAL_TAGS := include_gcs include_oss containers_image_openpgp gssapi providerless netgo osusergo bpf
```
dave-tucker, line 44: nooo! we've only just escaped the build tag

pkg/bpf/kepler.bpf.c
```diff
@@ -1,3 +1,5 @@
+//go:build ignore
```
dave-tucker, line 1: Why move this file?

pkg/bpf/mockbpf.go
```diff
@@ -1,17 +1,31 @@
-// Code generated by bpf2go; DO NOT EDIT.
-//go:build mips || mips64 || ppc64 || s390x
-
+//go:build !bpf
```
dave-tucker, line 1: The issue with this is that I'm pretty sure this will break the Go language server.

Currently, if I add a new map to the C code, run `make generate`, the language server will gladly autocomplete for me when I'm looking at `keplerObjects`, `keplerMaps` etc...
Given this is now the "default" behaviour, things like map renames etc.. will still compile just fine (even though it shouldn't) and you won't see the warnings in your editor 😢

dave-tucker (changes requested):
I hate the generated files being checked in too - but I also like the Go language server in my IDE working properly so it catches errors and autocomplete works etc...
Introducing a `mockBpf` inbetween will break this - and it's fragile as we'll need to keep it manually up-to-date.

pkg/bpf/kepler.bpf.c
maryamtahhan: honestly - this was more of a logical move (at least to me). The file is used to generate the go implementation in `pkg/bpf/` and for that reason I moved it to reside in the package where the generated code resides

pkg/bpf/mockbpf.go
maryamtahhan: so yeah, this file would need to be manually updated for the structs to match... let me think on this a little more to see if there's anything else we can do here.

pkg/sensors/platform/source/redfish.go
maryamtahhan: no I can pull it out into a separate PR... golangci-lint doesn't like it

Makefile
maryamtahhan: yeah - I know. 
I guess I can't see a way around this without build tags, but I do think this is a valid use case for the build tag if we don't want to check in generated objects into the codebase

maryamtahhan:
> I hate the generated files being checked in too - but I also like the Go language server in my IDE working properly so it catches errors and autocomplete works etc... Introducing a `mockBpf` inbetween will break this - and it's fragile as we'll need to keep it manually up-to-date.

Yeah - I will see if there's a way we can keep that part synced without manual updates (ideally with go generate)... let me look into this and will post back soon

maryamtahhan:
In this latest version: 
I've automated the generation of the mock files so that they are essentially generated via bpf2go. The main difference is that these mock files don't have the go embed directive and different go tags to the bpf generated files when they exist and no build flags when then don't exist.

I've gotten rid of the bpf build flag I've introduce in the makefile also

I've also modified the golanglintci configuration file to ignore the bpf2go generated files. 

These things mean that unfortunately in order to pass the existing CI we at least need the mock files checked in - but everything else can be ignored. The mock files will always track the real bpf2go structs and types as they are autogenerated and will need to be checked in if the c bpf implementation is ever updated.

maryamtahhan:
@dave-tucker I can probably close this PR - but it's one way of getting the generated files out of the current code base... WDYT?

PR #1699, opened 2024-08-16
feat: add model_name attribute to ComponentModelWeights

sunya-ch:
For updating model selection logic, I would like to add more information on the model weight file started with the model_name in PR https://github.com/sustainable-computing-io/kepler-model-server/pull/370. 

This PR update the ComponentModelWeights in Kepler in order to unmarshal the new weight file. 
The new structure is also support the old version of weight file.

### DIFF
--- a/pkg/model/estimator/local/regressor/model_weights.go
+++ b/pkg/model/estimator/local/regressor/model_weights.go
@@ -92,4 +92,11 @@ type NormalizedNumericalFeature struct {
 	Weight float64 `json:"weight,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


rootfs:
This PR helps pick up the model when needed. From the code change, I don't think this will impact the model validation result. @KaiyiLiu1234 @sthaha can you double check?

A separate note on using models. When running on bare metal, it is likely certain RAPL domains are missing (e.g. core or dram), would it be possible to use estimator in this case?

rootfs (approved):

KaiyiLiu1234 (approved):
LGTM

PR #1708, opened 2024-08-20
chore: move model_weights to its own directory

sthaha:
This commit moves model_weights from/var/lib/kepler/data/ to its own directory - var/lib/kepler/data/model_weights/ this allows additional data like machine-spec to be stored its own directory.

Additionally this change fixes the blank cpu.yaml that gets created when running compose files. Previously the VM compose  used to blank `cpu.yaml` file instead of the one in `data/`


sunya-ch (approved):
LGTM

vprashar2929 (approved):

Issue #1709, opened 2024-08-20
Mock ACPI CI doesn't checkout the PR code base


vprashar2929:
### What happened?

When mock-acpi workflow is called the checkout code step of the GH workflow always points to main branch commit. Any changes made to ansible playbook in the PR will not be present. Only once the PR is merged then only ansible changes will be reflected.  

Attaching the CI logs for reference: https://github.com/sustainable-computing-io/kepler/actions/runs/10466469721/job/28983459922#step:4:440

### What did you expect to happen?

Ideally, this shouldn't happen. Like other workflows that run on PR which checks out the current code base the mock-acpi workflow should also do the same

### How can we reproduce it (as minimally and precisely as possible)?

Create a PR with a change to any ansible playbook step and run the CI by commenting `/test-acpi` 

### Kepler image tag

<details>

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or bare metal

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE}
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE}
```

For standalone:
# put your Kepler command argument here

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

vprashar2929:
closing as its being resolved in latest workflow commits

PR #1712, opened 2024-08-20
fix(ci): ensure proper status reporting for mock-acpi workflow

vprashar2929:
This commit resolves two key issues with the mock-acpi workflow:
- Checkout correct branch: The workflow previously checkout out the default branch when triggered by a pull request. This fix ensures that the correct pull request branch is checked out during CI run.
- Attach workflow to pull request checks: The workflow was not being reflected under pull request checks. With this fix, the workflow will be correctly attached, ensuring its status visible and reported under pull request checks.

Also addresses: #1709


sunya-ch (approved):
LGTM

PR #1718, opened 2024-08-21
chore: cleanup pkg/config globals

maryamtahhan:
Cleanup the global variables in pkg/config

### DIFF
--- a/pkg/bpf/test_utils.go
+++ b/pkg/bpf/test_utils.go
@@ -23,7 +23,7 @@ func defaultHardwareCounters() sets.Set[string] {
 
 func defaultSoftwareCounters() sets.Set[string] {
 	swCounters := sets.New(config.CPUTime, config.PageCacheHit)
-	if config.ExposeIRQCounterMetrics {
+	if config.ExposeIRQCounterMetrics() {
 		swCounters.Insert(config.IRQNetTXLabel, config.IRQNetRXLabel, config.IRQBlockLabel)
 	}
 	return swCounters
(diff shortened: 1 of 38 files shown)


maryamtahhan:
Double check if we need to introduce locking

maryamtahhan:
> Double check if we need to introduce locking

It looks like the set functions are only called in one place. all the rest are gets - so this should be fine without locking.

rootfs:
@sthaha @sunya-ch can you take a look too? thanks

pkg/collector/stats/node_stats.go
```diff
@@ -25,24 +25,38 @@ import (
...
 type NodeStats struct {
 	Stats
 
 	// IdleResUtilization is used to determine idle pmap[string]eriods
 	IdleResUtilization map[string]uint64
 }
 
+// NodeName returns the node name
+func NodeName() string {
+	return GetNodeName()
```
sthaha, line 37: How about we rename GetNodeName() to NodeName() ?

pkg/bpf/test_utils.go
```diff
@@ -11,6 +11,7 @@ type mockExporter struct {
 }
 
 func DefaultSupportedMetrics() SupportedMetrics {
+	config.GetConfig()
```
sthaha, line 14: Is this needed here?

maryamtahhan:
Hi @sthaha 
I incorporated the suggested changes. Please let me know if there's anything else.

Thanks
Maryam

sthaha (approved):

PR #1722, opened 2024-08-21
fix(ci): consolidate mock-acpi workflow into single job

vprashar2929:
This commit addresses the issue of multiple jobs defined in the `mock-acpi` workflow, which were unintentionally executing in parallel due to sequence constraints. By consolidating the workflow into a single job, we ensure that the tasks are executed sequentially


vimalk78 (approved):
lgtm

PR #1726, opened 2024-08-22
chore: cleanup globals in pkg/cgroup

maryamtahhan:
cleanup globals in pkg/cgroup

### DIFF
--- a/pkg/cgroup/resolve_container.go
+++ b/pkg/cgroup/resolve_container.go
@@ -18,13 +18,13 @@ package cgroup
 
 import (
 	"bufio"
-	"encoding/binary"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/sustainable-computing-io/kepler/pkg/kubelet"
 	"github.com/sustainable-computing-io/kepler/pkg/utils"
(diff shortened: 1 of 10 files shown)


maryamtahhan:
Double check if we need to introduce locking

rootfs:
@sthaha @sunya-ch can you take a look? thanks

pkg/cgroup/resolve_container.go
```diff
@@ -193,26 +242,27 @@ func getPathFromcGroupID(cgroupID uint64) (string, error) {
 		if !dentry.IsDir() {
 			return nil
 		}
-		getCgroupID, err := utils.GetCgroupIDFromPath(byteOrder, path)
+		getCgroupID, err := utils.GetCgroupIDFromPath(utils.DetermineHostByteOrder(), path)
```
sunya-ch, line 245: Why do we need to call DetermineHostByteOrder every call here?

pkg/cgroup/resolve_container.go
```diff
@@ -40,30 +40,59 @@ type ContainerInfo struct {
...
 
+// GetConfig returns the singleton Config instance, creating it if necessary.
+func InitCache() {
+	once.Do(func() {
+		instance = newCache()
+	})
+}
+
+// SetConfig replaces the global instance
+func SetConfig(c *cache) {
```
sunya-ch, line 73: Is this function expected to be used later?

sunya-ch (commented):
refactoring by moving all variable to cache struct is fine to me. Just leave some comments.

pkg/cgroup/resolve_container.go
```diff
@@ -90,16 +120,21 @@ func GetContainerInfo(cGroupID, pid uint64, withCGroupID bool) (*ContainerInfo,
 		return info, err
 	}
 
-	if i, ok := containerIDToContainerInfo[containerID]; ok {
+	instance.mu.Lock()
+	defer instance.mu.Unlock()
```
sthaha, line 124:
```suggestion
	instance.mu.Lock()
	defer instance.mu.Unlock()
```
We should let the callee handle Mutex than the caller.

pkg/cgroup/resolve_container.go
```diff
@@ -40,30 +40,59 @@ type ContainerInfo struct {
...
 	pods := []corev1.Pod{}
 	return &pods, nil
 }
 
+// GetConfig returns the singleton Config instance, creating it if necessary.
+func InitCache() {
+	once.Do(func() {
+		instance = newCache()
+	})
+}
```
sthaha, line 70: instead of `once.Do` how about we use `func init()` to initialise the cache? That will make `ensureCache...` unnecessary.

sthaha (changes requested):
requesting changes to consider 
* use of init
* remove `ensureCacheInitialised` 
* let `cache` handle when Mutex should be locked / unlocked

pkg/cgroup/resolve_container.go
maryamtahhan: Yes l, thought it would be useful for future tests but can remove it in the meantime

pkg/cgroup/resolve_container.go
maryamtahhan: I will have a look

pkg/cgroup/resolve_container.go
maryamtahhan: I'm sorry I don't understand the leaky abstraction here. These are the functions in the same file that are accessing the cache. How does one avoid the locking here? Or is the suggestion to have more functions for the cache structure in terms of setters and getters that these functions call essentially making them wrappers around these setters and getters?

pkg/cgroup/resolve_container.go
maryamtahhan: No. I will set this back

maryamtahhan: I will update this

maryamtahhan:
> requesting changes to consider
> 
> * use of init
> * remove `ensureCacheInitialised`
> * let `cache` handle when Mutex should be locked / unlocked

Hi @sthaha 
I incorporated the suggested changes. Please let me know if there's anything else.

Thanks
Maryam

maryamtahhan:
@sthaha sorry to poke again, can you have a look at the updates? 

Thanks again

sthaha:
> @sthaha sorry to poke again, can you have a look at the updates?
> 
> Thanks again

Sorry, this completely slipped my radar :( ..

pkg/cgroup/resolve_container.go
```diff
@@ -40,29 +41,196 @@ type ContainerInfo struct {
...
+
+// newConfig creates and returns a new Config instance.
+func newCache() *cache {
+	return &cache{
+		containerIDCache:           map[uint64]string{},
+		containerIDToContainerInfo: map[string]*ContainerInfo{},
+		cGroupIDToPath:             map[uint64]string{},
+		byteOrder:                  utils.DetermineHostByteOrder(),
+	}
+}
```
sthaha, line 86: Given that the usage of the package is to use the `Instance` which in the current implementation should be initialised by calling `InitCache`, can we make use of `func init` to initialise the `Instance` without requiring the caller to `InitCache` ? 

```
func init() {
  instance = newCache()
}
```

pkg/cgroup/resolve_container.go
```diff
@@ -90,16 +258,18 @@ func GetContainerInfo(cGroupID, pid uint64, withCGroupID bool) (*ContainerInfo,
 		return info, err
 	}
 
-	if i, ok := containerIDToContainerInfo[containerID]; ok {
-		return i, nil
+	if instance.checkContainerID(containerID) {
```
sthaha, line 261: `nit`: 

```suggestion
	if instance.hasContainerID(containerID) {
```

sthaha (approved):
Thanks @maryamtahhan

PR #1727, opened 2024-08-22
feat(compose): add node-exporter for rapl metrics

vimalk78:
node-exporter provides rapl power metrics
    - node_rapl_core_joules_total
    - node_rapl_uncore_joules_total
    - node_rapl_package_joules_total
    - node_rapl_psys_joules_total

removed unwanted dependency on kepler-latest in default override.yaml

### DIFF
--- a/manifests/compose/dev/compose.yaml
+++ b/manifests/compose/dev/compose.yaml
@@ -74,6 +74,34 @@ services:
     networks:
       - scaph-network
 
+  node-exporter:
+    image: quay.io/prometheus/node-exporter:latest
+    pid: host
+    ports:
+      - 9100:9100
+    volumes:
+      - type: bind
+        source: /proc
+        target: /host/proc
+      - type: bind
+        source: /sys
+        target: /host/sys
+      - type: bind
+        source: /
+        target: /rootfs
+    command:
+      - --path.procfs=/host/proc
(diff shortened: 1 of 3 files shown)


vimalk78:
cc: @sthaha @sunya-ch

manifests/compose/dev/compose.yaml
```diff
@@ -74,6 +74,34 @@ services:
...
+    command:
+      - --path.procfs=/host/proc
+      - --path.sysfs=/host/sys
+      - --path.rootfs=/rootfs
+      - --collector.rapl  # Enable RAPL collector
+      - --collector.filesystem.mount-points-exclude=^/(sys|proc|dev|host|etc)($$|/)
+    privileged: true
+    user: root
+    cap_add:
+      - ALL
```
sthaha, line 101: This may not be needed and may not require privileged since we are mounting to a different path.

manifests/compose/dev/compose.yaml
```diff
@@ -74,6 +74,34 @@ services:
...
+    networks:
+      - kepler-network
```
sthaha, line 103: may be create a different network for node-exporter and add that to prometheus since only prometheus needs to access node-exporter.

sthaha (approved):

PR #1728, opened 2024-08-23
feat: update static models to use 0.7.11 models

sthaha:
NOTE: `intel_rapl_DynPower` has not been updated since 
SGDRegressor for Dynamic Power Node Type 0 is missing  https://github.com/sustainable-computing-io/kepler-model-db/tree/main/models/v0.7/ec2-0.7.11/rapl-sysfs/DynPower/BPFOnly 

After updating the models to the latest, here are the changes on **my** machine 
```

kepler_node_info{components_power_source="rapl-sysfs", cpu_architecture="Skylake", instance="kepler-latest:8888", job="latest", platform_power_source="none", source="os"} | 1
-- | --
kepler_node_info{components_power_source="estimator", cpu_architecture="Skylake", instance="kepler-dev:8888", job="dev", platform_power_source="none", source="os"}

```
NOTE: `kepler-latest` reports power_sources incorrectly (fixed in this pr; see "dev") 

### Platform / Idle
* seems like an improvement 

image

### Platform / Dynamic 
image

### Package / Idle
* Idle power is much higher than the old 
image

### Package / dynamic 
* higher but closer to real usage

image

### DIFF
--- a/pkg/model/node_component_energy.go
+++ b/pkg/model/node_component_energy.go
@@ -55,11 +55,14 @@ func CreateNodeComponentPowerEstimatorModel(nodeFeatureNames, systemMetaDataFeat
 	var err error
 	nodeComponentPowerModel, err = createPowerModelEstimator(modelConfig)
 	if err != nil {
-		klog.Errorf("Failed to create %s/%s Model to estimate Node Component Power: %v", modelConfig.ModelType, modelConfig.ModelOutputType, err)
+		klog.Errorf("Failed to create %s/%s Model from %s to estimate Node Component Power: %v",
+			modelConfig.ModelType, modelConfig.ModelOutputType,
+			modelConfig.SourceURL(), err)
 		return
 	}
 
-	klog.V(1).Infof("Using the %s/%s Model to estimate Node Component Power", modelConfig.ModelType, modelConfig.ModelOutputType)
+	klog.V(1).Infof("Using the %s/%s Model from %s to estimate Node Component Power",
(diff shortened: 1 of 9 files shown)


pkg/model/node_component_energy.go
```diff
@@ -27,37 +27,46 @@ import (
...
+// the absolute power model includes both the absolute and idle power consumption
+var nodeComponentPowerModel PowerModelInterface
 
 // createNodeComponentPowerModelConfig: the node component power model url must be set by default.
 func createNodeComponentPowerModelConfig(nodeFeatureNames, systemMetaDataFeatureNames, systemMetaDataFeatureValues []string) *types.ModelConfig {
 	modelConfig := CreatePowerModelConfig(config.NodeComponentsPowerKey)
 	if modelConfig.InitModelURL == "" {
 		modelConfig.InitModelFilepath = config.GetDefaultPowerModelURL(modelConfig.ModelOutputType.String(), types.ComponentEnergySource)
+		// HACK: this is hardcoded until model metadata has the trainer information
+		modelConfig.TrainerName = "LogarithmicRegressionTrainer"
```
sunya-ch, line 39: I don't think we should have this hardcode. Prefer to extract model from the file path by having an error check when the file path doesn't contain the trainer name in expected list.

sthaha: The name of the file in question is `intel_rapl_AbsPower.json` and `acpi_AbsPower.json` . is the suggestion to modify the filename to something like `intel_rapl_AbsPower_LogisticRegressionTrainer.json` ?

sunya-ch (approved):
/lgtm Thank you so much!

PR #1731, opened 2024-08-25
fix(ci): checkout code base earlier

vprashar2929:
This commit addresses an issue where the codebase was not being checked out before initiating the runner and setting up the test environment


vprashar2929:
Ignore Mock ACPI test run as it will use v0.2.0 until the workflow is merged to the main branch

sthaha (approved):

PR #1732, opened 2024-08-26
feat: compute idle power with core ratio

sunya-ch:
This patch is for adding an idle power based on ratio of the machine cores over the cores of the machine that is used for training the power model. 

Need the following Kepler PR to solve the TODO.
- [ ] https://github.com/sustainable-computing-io/kepler/pull/1684

Need the following PR in kepler-model-server to have estimator provides machine spec and server API to put machine spec to weight response.
- [ ] https://github.com/sustainable-computing-io/kepler-model-server/pull/405


rootfs (approved):

rootfs:
per Pang's instruction, this PR can be merged independently. Let's have this one first and wait for #1684

PR #1738, opened 2024-08-27
fix(ci): switch to custom metal-runner-action in mock-acpi workflow

vprashar2929:
This commit updates the mock-acpi workflow to use the custom `metal-runner-action` instead of the `equinix-labs`. This change is necessary because the workflow does not expose the `github.organization` value when triggered by the `issue_comment` event

Ref: https://github.com/sustainable-computing-io/kepler/actions/runs/10572316346/job/29289799353#step:6:14

### DIFF
--- a/.github/workflows/mock_acpi.yml
+++ b/.github/workflows/mock_acpi.yml
@@ -34,7 +34,8 @@ jobs:
           ref: ${{ steps.pr_branch.outputs.head_sha }}
 
       - name: metal-runner-action
-        uses: equinix-labs/metal-runner-action@v0.3.0
+        uses: vprashar2929/metal-runner-action@custom-action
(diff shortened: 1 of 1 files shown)


.github/workflows/mock_acpi.yml
```diff
@@ -34,14 +34,16 @@ jobs:
           ref: ${{ steps.pr_branch.outputs.head_sha }}
 
       - name: metal-runner-action
-        uses: equinix-labs/metal-runner-action@v0.3.0
+        uses: vprashar2929/metal-runner-action@custom-action
+        # uses: equinix-labs/metal-runner-action@v0.3.0
```
vimalk78, line 38: it would be good if we have the custom `metal-runner-action@custom-action` in sustainable-computing org

vimalk78 (approved):

PR #1740, opened 2024-08-28
fix: limit max core ratio to 1

sunya-ch:
This is a fix to avoid core ratio > 1 (trained machine is smaller than the target machine itself). 
Idle power should not affect DynEnergy.

https://github.com/sustainable-computing-io/kepler/blob/777c733dfc4a696c17d06494fa7d783e4b3af6e5/pkg/collector/stats/stats.go#L140

### DIFF
--- a/pkg/model/utils/utils.go
+++ b/pkg/model/utils/utils.go
@@ -17,6 +17,8 @@ limitations under the License.
 package utils
 
 import (
+	"math"
+
 	"github.com/sustainable-computing-io/kepler/pkg/sensors/components/source"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/model/utils/utils.go
```diff
@@ -63,7 +65,7 @@ func FillNodeComponentsPower(pkgPower, corePower, uncorePower, dramPower uint64)
 func GetCoreRatio(isIdlePower bool, inCoreRatio float64) float64 {
 	var coreRatio float64 = 1
 	if isIdlePower && inCoreRatio > 0 {
-		coreRatio = inCoreRatio
+		coreRatio = math.Min(inCoreRatio, 1)
```
sthaha, line 68: minor: can we add a unit-test for this?

sthaha (approved):

pkg/model/utils/utils.go
sunya-ch: Thank you. Added.

PR #1744, opened 2024-08-28
feat: get trainer from model_name in weight

sunya-ch:
Related to the comment in https://github.com/sustainable-computing-io/kepler/pull/1728#discussion_r1733759844 and https://github.com/sustainable-computing-io/kepler-model-server/issues/414, for now, we can extract the trainer name from the model name. 


rootfs (approved):

rootfs:
@sunya-ch can you rebase? thanks

data/model_weight/acpi_AbsPowerModel.json
```diff
@@ -1 +1 @@
-{"platform": {"All_Weights": {"Bias_Weight": 220.9079278650894, "Categorical_Variables": {}, "Numerical_Variables": {"bpf_cpu_time_ms": {"scale": 5911.969193263386, "mean": 0, "variance": 0, "weight": 29.028228361462897}}}}}
+{"model_name": "SGDRegressorTrainer_0", "platform": {"All_Weights": {"Bias_Weight": 220.9079278650894, "Categorical_Variables": {}, "Numerical_Variables": {"bpf_cpu_time_ms": {"scale": 5911.969193263386, "mean": 0, "variance": 0, "weight": 29.028228361462897}}}}}
```
sthaha, line 1: can we have the trainer_name instead? 
I am fine with model_name but I think we shouldn't require parsing model_name to get trainer name.

pkg/model/estimator/local/regressor/model_weights.go
```diff
@@ -122,3 +125,24 @@ func (w ComponentModelWeights) String() string {
...
+	if w.ModelName == "" {
+		return ""
+	}
+	modelNameSplits := strings.Split(w.ModelName, "_")
+	splitTrainer := strings.Join(modelNameSplits[0:len(modelNameSplits)-1], "_")
+	if isSupportedTrainer(splitTrainer) {
+		return splitTrainer
+	}
+	return ""
+}
```
sthaha, line 139: IMHO, models should have the trainer-name metadata embedded in the json itself so that this isn't needed.

sunya-ch: model_name has more complete information since it also embed the node_type and linked to model db.
I think we can add trainer_name additionally if needed.

sthaha (approved):

PR #1745, opened 2024-08-29
chore(compose/dev): allow power meters to be disabled

sthaha:
This pr allows dev compose to turn off power-meters (on BM) and expose idle metrics.


sunya-ch (approved):
/lgtm

PR #1746, opened 2024-08-29
fix(models): log name of regressor

sthaha:
The PR modifies the models/regressor 
* to add a `name` method to predictor so the name of of the predictor can be used in logs or elsewhere
* adds predictor name to the error returned


sunya-ch (approved):
/lgtm

Issue #1748, opened 2024-08-29
Update DynPower Model to 0.7.11


sthaha:
https://github.com/sustainable-computing-io/kepler/pull/1728 wasn't able to update the `intel_rapl_DynPower` model since the model is missing in model-db (see: https://github.com/sustainable-computing-io/kepler-model-db/issues/27).

The task is to update the models when the linked bug is fixed.

sunya-ch:
We set upper bound of MAE to 10 and MAPE to 20%. The SGDTrainer is much worse due to sparse data collected from latest Kepler as below.

```bash
MAE =  65.88490729951363
         MAE          MSE       MAPE    n energy_component energy_source                  Model Feature Group
0  65.884907  5580.690108  55.722063  172          package    rapl-sysfs  SGDRegressorTrainer_0       BPFOnly
1   0.000000     0.000000  -1.000000  172             core    rapl-sysfs  SGDRegressorTrainer_0       BPFOnly
2   0.000000     0.000000  -1.000000  172           uncore    rapl-sysfs  SGDRegressorTrainer_0       BPFOnly
3   2.876924    11.362931  54.496488  172             dram    rapl-sysfs  SGDRegressorTrainer_0       BPFOnly
```

![estimate_dyn_default_min_rapl-sysfs_BPFOnly_SGDRegressorTrainer_0 using BPFOnly_corr](https://github.com/user-attachments/assets/73ef62a9-d65e-40f4-aeef-a598717e77be)

I can upload the weight for DynPower with the remark of this potential model error.

vprashar2929:
@sunya-ch Are we good to close this?

PR #1749, opened 2024-08-30
chore: update intel_rapl_DynPowerModel from ec2-0.7.11

sunya-ch:
As mentioned in https://github.com/sustainable-computing-io/kepler/issues/1748#issuecomment-2319703608, this PR adds the default power model for DynPower for rapl-sysfs but with the remark of high training MAE against linear regressor.


sthaha:
@sunya-ch how about we use the Linear Regression Model but updated to 0.7.11 ?

sthaha (approved):

PR #1756, opened 2024-09-03
fix(dockerfile): remove redundant habanalabs installation steps

vprashar2929:
This commit resolves an issue where the habanalabs installation step was unnecessary repeated multiple times in the Dockerfile.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -5,11 +5,6 @@ WORKDIR /workspace
 COPY . .
 
 RUN make tidy-vendor format
-RUN if [[ "$INSTALL_HABANA" == "true" ]]; then \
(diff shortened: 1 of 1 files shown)


maryamtahhan:
Will this not break the kepler: `latest-habana` image?
If the library is not installed then the implementation will fail to initialise and render the image useless

vprashar2929:
We are already installing in the later step: https://github.com/sustainable-computing-io/kepler/blob/71b7c63b5c3f1c145f82b61ff7600ac2b93eec30/build/Dockerfile#L44

maryamtahhan (approved):

maryamtahhan:
I think my commit broke the Habana image build: 
Screenshot 2024-09-03 at 13 56 28

I have a fix here: https://github.com/sustainable-computing-io/kepler/pull/1758

rootfs (approved):

sthaha (approved):

PR #1767, opened 2024-09-06
chore(validator): fix graphs rendered incorrectly

sthaha:
This fixes the graphs that were flipped leading to confusion in how MAPE is calculated!
See: https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/docs/train-validate-e2e/2024-09-05_02-22-45/AbsPower/BPFOnly/SGDRegressorTrainer/report-v0.7.11-196-g6a76bc35.md#platform---dynamic 

The graph should have caused MAPE to be 100% where as the MAPE is `inf`  as the [dynamic data](https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/docs/train-validate-e2e/2024-09-05_02-22-45/AbsPower/BPFOnly/SGDRegressorTrainer/artifacts/kepler_vm_platform_joules_total--dynamic.json) is all 0

### DIFF
--- a/e2e/tools/validator/src/validator/cli/__init__.py
+++ b/e2e/tools/validator/src/validator/cli/__init__.py
@@ -533,8 +533,8 @@ def run_validation(
         if not result.mape_passed:
             click.secho(f"MAPE exceeded threshold. mape: {cmp.mape}, max_mape: {v.max_mape}", fg="red")
 
-        result.actual_filepath = dump_query_result(results_dir, v.expected, cmp.expected_series)
-        result.expected_filepath = dump_query_result(results_dir, v.actual, cmp.actual_series)
+        result.actual_filepath = dump_query_result(results_dir, v.expected, cmp.actual_series)
+        result.expected_filepath = dump_query_result(results_dir, v.actual, cmp.expected_series)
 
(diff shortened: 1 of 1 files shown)


KaiyiLiu1234 (approved):
LGTM. I think we should try and clarify in future what expected and what actual means. If we are assuming metal is the "correct" one, we should treat that as the actual and the vm as forecasted

sthaha:
> If we are assuming metal is the "correct" one, we should treat that as the actual and the vm as forecasted

Indeed! .. I think we should use `actual` and `predicted` for all comparisons 
I am thinking of modifying validation.yaml as 

```yaml
config:
  map:
    actual: metal
    predicted: vm
  unit: watts

validations:
- metal:  <metal-query>
   vm: <vm-query>

- metal:  kepler_process_bpf_cpu_time
   vm:  ... 
   unit: ms
```

KaiyiLiu1234:
I agree. We can also consider using percentage difference as well. That is dividing the absolute value of the difference between two numbers by the average of those two numbers. Could actually address the divide by 0 issue (assuming both do not produce 0 - but this is an edge case where the difference is 0%).

PR #1773, opened 2024-09-09
chore(validator): support validation config to map actual and predicted

sthaha:
This PR introduces the following changes to validator 
* Use `predicted` instead of `expected` terminology to align with MAPE and MSE teminology
* Add ability map `actual` and `predicted` in validations. This allwos for 
```
config:
  mapping:
     actual: metal
     predicted: vm

validations:
  - metal: <metal-query>
    vm: <vm-query>
```

* Generated reports and artifacts now make use of these label
* Add support for Units

image

### DIFF
--- a/e2e/tools/validator/acpi_validations.yaml
+++ b/e2e/tools/validator/acpi_validations.yaml
@@ -1,6 +1,8 @@
 validations:
   - name: mock - node
     actual: |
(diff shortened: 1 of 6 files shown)


KaiyiLiu1234 (approved):
lgtm

PR #1777, opened 2024-09-10
chore(validator): disable node-rapl validation

sthaha:
Metal CI seems to not have node-exporter properly configured. Hence disabling node-rapl validation until this is fixed.


vimalk78:
Metal-CI issue should be fixed now

PR #1789, opened 2024-09-18
ci: add attestation to container images and SBOMs

arthurus-rex:
Add in-toto attestation to all image-building ci workflows, for both SBOMs (image.yml) and images themselves (all 3 workflows). Allow necessary permissions to attest (attestations, packages).

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -21,6 +21,8 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      attestations: write
+      packages: write
     strategy:
       fail-fast: false
       matrix:
@@ -96,6 +98,14 @@ jobs:
         run: |
           cosign sign -y quay.io/sustainable_computing_io/${{matrix.IMAGE_NAME}}:${{matrix.LABEL}}@${{ steps.build-push-image.outputs.digest }}
 
+      - name: Generate image attestation
+        if: ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch')}}
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-name: quay.io/sustainable_computing_io/${{matrix.IMAGE_NAME}}
+          subject-digest: ${{ steps.build-push-image.outputs.digest }}
+          push-to-registry: ${{ inputs.pushImage }}
+
       - name: Generate SBOM
         uses: anchore/sbom-action@v0.17.2
         with:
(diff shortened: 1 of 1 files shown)


SamYuan1990:
> Attestations bind some subject (a named artifact along with its digest) to a [SLSA build provenance](https://slsa.dev/spec/v1.0/provenance) predicate using the [in-toto](https://github.com/in-toto/attestation/tree/main/spec/v1) format.

> A verifiable signature is generated for the attestation using a short-lived [Sigstore](https://www.sigstore.dev/)-issued signing certificate. If the repository initiating the GitHub Actions workflow is public, the public-good instance of Sigstore will be used to generate the attestation signature. If the repository is private/internal, it will use the GitHub private Sigstore instance.

ref to https://github.com/actions/attest-build-provenance
well, I am open for SLSA, sigstore or any openSSF kind structure in CI approach.
But, or well...
https://github.com/sustainable-computing-io/kepler/actions/runs/10931316693/job/30346044739?pr=1789
it should not in PR image build, IMO, it's valuable as user will use release image or daily image(which is latest tag image).

I am not sure if we need additional project configuration for sigstore, @rootfs , could you please advise?
at least within this PR, it seems we need a github token. 
https://github.com/actions/attest-build-provenance?tab=readme-ov-file#inputs
or @arthurus-rex , ref to github token settings, it seems PR(as PR owner is from a fork repo) can't use upstream repo's github token setting as security reason, please update code, remove/update settings for this github action. I suppose we don't need push to reg for each PR.

arthurus-rex:
@SamYuan1990 sounds good, I think removing the attestation from `image_pr.yml` makes sense.  The attestation in general was suggested to me by @dave-tucker (tagged in case he has any input/suggestions here).  I will update this PR over the next 1-2 days.

SamYuan1990:
> @SamYuan1990 sounds good, I think removing the attestation from `image_pr.yml` makes sense. The attestation in general was suggested to me by @dave-tucker (tagged in case he has any input/suggestions here). I will update this PR over the next 1-2 days.
> 
> Follow-up question: Is there a reason we are pushing to registry at all for PRs?

IMO for PR level, 
https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/pull_request.yml#L50
the push image is set to false, so for each PR we should not push image.

https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/image_pr.yml#L4 is working with `workflow_dispatch`. so when we need to debug/test a pr image/pr version, someone with manual build the image and upload it.

https://quay.io/repository/sustainable_computing_io/kepler?tab=tags&tag=latest which same as image tags you see on quay io.

arthurus-rex:
I removed the attestation on PR and restricted permissions where possible.

SamYuan1990 (approved):
LGTM

SamYuan1990:
the code change look good to me, @arthurus-rex , as `push-to-registry` is set, I am not sure if @rootfs, @vprashar2929  or other kepler maintainer can have a link to preview the output?

PR #1797, opened 2024-09-25
chore: update gosec to use master

maryamtahhan:
As suggested by https://github.com/securego/gosec/issues/1234


dave-tucker (approved):
LGTM

PR #1798, opened 2024-09-25
fix(bpf_collector): fix command for kernel processes

vimalk78:
-  fix command name in case of kernel processes

if the cgroupID for the pid is 1, it is considered as kernel process
and bpf cpu time for each such process is aggregated under the same  ProcessStat

But the command which shows up in metric is from the first pid which  created the entry in ProcessStats. It could be a `kworker`, or  `rcu_preempt`, or some other kernel process. Which is misleading because it shows that process consuming most of bpf cpu time.

  This commit:
  - uses "kernel_processes" as the command for pid 1 kernel processes
  - sets the pid as 1 instead of the pid of the process that created the  entry

after change:

image

### DIFF
--- a/pkg/collector/resourceutilization/bpf/process_bpf_collector.go
+++ b/pkg/collector/resourceutilization/bpf/process_bpf_collector.go
@@ -110,17 +110,19 @@ func UpdateProcessBPFMetrics(bpfExporter bpf.Exporter, processStats map[uint64]*
 		}
 
 		mapKey := ct.Pid
+		process := comm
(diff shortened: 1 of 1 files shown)


sthaha (commented):
lgtm
@aharivel do you want to have a look at this as well?

aharivel:
Indeed @sthaha thanks ! 
I'll review the patch today.

vimalk78:
@sthaha updated PR for aggregating kernel processes only

sthaha (approved):

PR #1800, opened 2024-09-26
chore: load cpu architecture dependent models

rootfs:
This is to prepare for coming CPU dependent models


pkg/config/config.go
```diff
@@ -247,8 +247,18 @@ func setModelServerReqEndpoint() string {
...
 // e.g., /var/lib/kepler/data/model_weight/acpi_AbsPowerModel.json
-func GetDefaultPowerModelURL(modelOutputType, energySource string) string {
-	return fmt.Sprintf(`/var/lib/kepler/data/model_weight/%s_%sModel.json`, energySource, modelOutputType)
+func GetDefaultPowerModelURL(modelOutputType, energySource, cpuArch string) string {
+	// strip white space or new line from cpuArch
+	cpuArch = strings.TrimSuffix(cpuArch, "\n")
+	cpuArch = strings.TrimSpace(cpuArch)
+	fullPath := fmt.Sprintf(`/var/lib/kepler/data/%s/model_weight/%s_%sModel.json`, cpuArch, energySource, modelOutputType)
+	// if the model does not exist, return the default model
+	if _, err := os.Stat(fullPath); os.IsNotExist(err) {
```
KaiyiLiu1234, line 256: Won't this result in a possible race condition? If you can guarantee that after the os.Stat the file exists, then ignore this comment.

pkg/config/config.go
```diff
@@ -247,8 +247,18 @@ func setModelServerReqEndpoint() string {
 
 // return local path to power model weight
 // e.g., /var/lib/kepler/data/model_weight/acpi_AbsPowerModel.json
-func GetDefaultPowerModelURL(modelOutputType, energySource string) string {
-	return fmt.Sprintf(`/var/lib/kepler/data/model_weight/%s_%sModel.json`, energySource, modelOutputType)
+func GetDefaultPowerModelURL(modelOutputType, energySource, cpuArch string) string {
+	// strip white space or new line from cpuArch
+	cpuArch = strings.TrimSuffix(cpuArch, "\n")
```
KaiyiLiu1234, line 252: TrimSpace should remove \n characters and white space, so you can remove this line

pkg/config/config.go
KaiyiLiu1234, line 256: On a minor note, os.isNotExist(err) might be misleading as os.Stat can return other types of errors which is not the same as a file/model not existing.

pkg/bpf/kepler_bpfeb.o
sthaha, line 1: these files shouldn't change at all 🤔

sthaha (changes requested):
Requesting change to remove the ebpf file changes.

PR #1804, opened 2024-10-09
chore: unit test regressors to match estimator results

rootfs:


pkg/model/estimator/local/regressor/local_test.go
```diff
@@ -0,0 +1,67 @@
...
+	Expect(powers[0].DRAM).Should(BeEquivalentTo(dram))
+	Expect(powers[0].Pkg).Should(BeEquivalentTo(pkg))
+}
+
+var _ = Describe("Test Regressor Weight Unit (models from URL)", func() {
+	It("Get Node Components Power By SGD Regression with model from URL", func() {
+		// Test power calculation. The results should match those from estimator
+		// https://github.com/sustainable-computing-io/kepler-model-server/pull/493#discussion_r1795610556
+		testModel("https://raw.githubusercontent.com/sustainable-computing-io/kepler-model-db/refs/heads/main/models/v0.7/ec2-0.7.11/rapl-sysfs/AbsPower/BPFOnly/SGDRegressorTrainer_0.json",
+			types.LinearRegressionTrainer, 146994, 18704, 146994)
```
sunya-ch, line 57: We may make the note that the expected value must be updated if when the model at the target URL has changed.

sunya-ch (approved):
Mostly looks good to me. Just leave a comment. 
Thank you.

PR #1810, opened 2024-10-16
feat(metrics): refactor idle power exposure

rootfs:
As discussed in the Oct 15 meeting, we make the idle power classification optional, until we have an idle power regression model that can predict idle power by system utilization.

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -79,7 +79,7 @@ func newAppConfig() *AppConfig {
 	flag.StringVar(&_config.Kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file, if empty we use the in-cluster configuration")
 	flag.BoolVar(&_config.ApiserverEnabled, "apiserver", true, "if apiserver is disabled, we collect pod information from kubelet")
 	flag.StringVar(&_config.RedfishCredFilePath, "redfish-cred-file-path", "", "path to the redfish credential file")
-	flag.BoolVar(&_config.ExposeEstimatedIdlePower, "expose-estimated-idle-power", false, "estimated idle power is meaningful only if Kepler is running on bare-metal or when there is only one virtual machine on the node")
+	flag.BoolVar(&_config.ExposeEstimatedIdlePower, "expose-estimated-idle-power", false, "Whether to expose the estimated idle power as a metric")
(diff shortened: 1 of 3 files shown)


KaiyiLiu1234 (approved):
Code lgtm. So this ensures idle Energy is always 0 now? Dynpower will be reflective of the total energy? This means if I validate kepler's energy metrics with node exporter, they should be essentially identical.

cmd/exporter/exporter.go
```diff
@@ -79,7 +79,7 @@ func newAppConfig() *AppConfig {
 	flag.StringVar(&_config.Kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file, if empty we use the in-cluster configuration")
 	flag.BoolVar(&_config.ApiserverEnabled, "apiserver", true, "if apiserver is disabled, we collect pod information from kubelet")
 	flag.StringVar(&_config.RedfishCredFilePath, "redfish-cred-file-path", "", "path to the redfish credential file")
-	flag.BoolVar(&_config.ExposeEstimatedIdlePower, "expose-estimated-idle-power", false, "estimated idle power is meaningful only if Kepler is running on bare-metal or when there is only one virtual machine on the node")
+	flag.BoolVar(&_config.ExposeEstimatedIdlePower, "expose-estimated-idle-power", false, "Whether to expose the estimated idle power as a metric")
```
sunya-ch, line 82: We may change default here to true to make sure that it is disabled by default for now.

pkg/collector/stats/stats.go
```diff
@@ -174,7 +174,7 @@ func (s *Stats) CalcDynEnergy(absM, idleM, dynM, id string) {
 
 // calcDynEnergy calculates the dynamic energy.
 func calcDynEnergy(totalE, idleE uint64) uint64 {
-	if (totalE == 0) || (idleE == 0) || (totalE < idleE) {
+	if (totalE == 0) || (totalE < idleE) {
```
sthaha, line 177: I think we should fix this bit as well .. 
If the totalE < the idle, then we have a new idle, IOW kepler over time will continue to improve its accuracy of idle by measuring the least amount of power consumed by the system.

sunya-ch (approved):
/lgtm thank you!

pkg/collector/stats/stats.go
sunya-ch: Agree but I think should note and fix this in the different issue and PR.

sthaha (approved):

PR #1815, opened 2024-10-17
feat(validator): incorporate MAE in validations

vprashar2929:
This commit adds MAE to the computation. The validator runs will now include MAE results in the output.

The MSE, MAE and MAPE are now computed using the scikit-learn library

### DIFF
--- a/e2e/tools/validator/src/validator/prometheus/__init__.py
+++ b/e2e/tools/validator/src/validator/prometheus/__init__.py
@@ -6,6 +6,7 @@ from typing import NamedTuple, Protocol
 import numpy as np
 import numpy.typing as npt
 from prometheus_api_client import PrometheusConnect
+from sklearn.metrics import mean_absolute_error, mean_absolute_percentage_error, mean_squared_error
 
 from validator.config import Prometheus as PromConfig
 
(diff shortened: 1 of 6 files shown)


e2e/tools/validator/src/validator/prometheus/__init__.py
```diff
@@ -123,6 +124,16 @@ def mape(actual: npt.ArrayLike, predicted: npt.ArrayLike) -> ValueOrError:
         return ValueOrError(value=0, error=str(e))
 
 
+def mae(actual: npt.ArrayLike, predicted: npt.ArrayLike) -> ValueOrError:
+    try:
+        actual, predicted = validate_arrays(actual, predicted)
+        return ValueOrError(value=np.abs(np.subtract(actual, predicted)).mean())
```
sthaha, line 130: IIUC,  https://www.statisticshowto.com/absolute-error/ 
the `abs` should be applied to `subtract`  and not to the mean 
```
|(1  + - 1 ) / 2 | != ( |1|  + |-1| ) / 2 
```
Hence 
```
def mean_absolute_error(actual, predicted):
    return np.mean(np.abs(actual - predicted))
```

sthaha (changes requested):
Can you please check the mae calculation ? 
Alternatively, how about we use the functions in scikit.metrics ? 

See usage: https://github.com/sustainable-computing-io/kepler-experimental/blob/main/st/power-model/src/power_model/trainer/runner.py#L14

rootfs (approved):

sthaha (approved):

PR #1814, opened 2024-10-17
chore(validator): remove idle power estimator in validator

rootfs:
Disabling idle power estimator caused validation test fail in [this round](https://github.com/sustainable-computing-io/kepler-metal-ci/actions/runs/11388766526/job/31686722838). So we have to remove it from validator for now.

PR #1816, opened 2024-10-20
fix(remove_idle): Remove Idle Validations

KaiyiLiu1234:
Given current issues with idle metrics, it should not be validated upon on the metal ci. Current validations should only use absolute power (no dynamic/idle separation).

### DIFF
--- a/e2e/tools/validator/validations.yaml
+++ b/e2e/tools/validator/validations.yaml
@@ -183,74 +183,74 @@ validations:
 
   # idle power comparison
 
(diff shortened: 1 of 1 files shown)


KaiyiLiu1234:
@rootfs. This will fix the errors in metal ci after the PR which removed idle metrics.

vprashar2929:
It is addressed in https://github.com/sustainable-computing-io/kepler/pull/1814

PR #1820, opened 2024-10-23
fix(validator): include scikit-learn in dependencies

vprashar2929:
This commit adds the missing scikit-learn dependency to the validator as pip install fails currently due to missing dependency.


vprashar2929:
This should fix the failing CI run on metal-ci: https://github.com/sustainable-computing-io/kepler-metal-ci/actions/runs/11474006398/job/31929494370

rootfs (approved):

PR #1825, opened 2024-10-24
feat: move from ubi to ubi-minimal

maichouni-mitek:
This PR should close https://github.com/sustainable-computing-io/kepler/issues/1826.
Using a smaller base image has several benefits:
  - Reducing the storage footprint. Quicker deployments (less and/or smaller layers to pull).
  - Reducing the potential attack surface.
  - Reducing the amount of vulnerabilities that will arise between one `kepler` release and another, simply because there are less items installed.

A picture is worth a thousand words:
image
image
image

The vulnerabilities we see above (as of 2024/10/24, in `kepler:release-0.7.12`) are inherited from the base image. They are in the python namespace, which is not needed at all in the `kepler` image, and which is why https://github.com/sustainable-computing-io/kepler/pull/1361 cannot get rid of them.

Thank you.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -19,7 +19,7 @@ RUN make tidy-vendor format
 
 RUN make build
 
-FROM registry.access.redhat.com/ubi9:latest
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 # NOTE: use bash instead of sh
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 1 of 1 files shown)


maichouni-mitek:
@sthaha , would you please start the GHAs?

rootfs:
@vimalk78 can you take a look? thanks

SamYuan1990 (approved):
LGTM, but we need test image at PR level. @rootfs, @marceloamaral , @sthaha  could you please help?

maichouni-mitek:
@rootfs, @marceloamaral, @sthaha, can you please help with the image test?
Thank you very much.

sthaha:
@SamYuan1990 @maichouni-mitek could you please elaborate what you meant by `image-test` ?

SamYuan1990:
> @SamYuan1990 @maichouni-mitek could you please elaborate what you meant by `image-test` ?

@rootfs and I once made a CI job https://github.com/sustainable-computing-io/kepler/actions/workflows/image_pr.yml. The job can build an temp image with a specific PR as code base.

As our PR level testing almost running on GHA VM, which is not a BM instance, at meanwhile,  this PR has base image change, to ensure the change does not harmful, we can use this PR level CI to build a PR level image and tested on a BM instance if necessary.

vimalk78:
LGTM.

PR #1830, opened 2024-10-29
fix(bpf): exclude swapper process bpf_cpu_time

vimalk78:
when cpu is idle, swapper process is scheduled, and kepler adds its bpf_cpu_time to kernel processes

adding a configuration to exclude swapper process

with-swapper
with-swapper

without-swapper
without-swapper

### DIFF
--- a/pkg/collector/resourceutilization/bpf/process_bpf_collector.go
+++ b/pkg/collector/resourceutilization/bpf/process_bpf_collector.go
@@ -89,6 +89,11 @@ func UpdateProcessBPFMetrics(bpfExporter bpf.Exporter, processStats map[uint64]*
 	for _, ct := range processesData {
 		comm := C.GoString((*C.char)(unsafe.Pointer(&ct.Comm)))
 
+		if ct.Pid == 0 && config.ExcludeSwapperProcess() {
+			// exclude swapper process
+			continue
+		}
+
 		if ct.Pid != 0 {
 			klog.V(6).Infof("process %s (pid=%d, cgroup=%d) has %d process run time, %d CPU cycles, %d instructions, %d cache misses, %d page cache hits",
 				comm, ct.Pid, ct.CgroupId, ct.ProcessRunTime, ct.CpuCycles, ct.CpuInstr, ct.CacheMiss, ct.PageCacheHit)
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -90,11 +90,12 @@ const (
 	// MaxIRQ is the maximum number of IRQs to be monitored
 	MaxIRQ = 10
 	// defaultSamplePeriodSec is the time in seconds that the reader will wait before reading the metrics again
-	defaultSamplePeriodSec = 3
-	configDir              = "/etc/kepler/kepler.config"
-	defaultKubeConfig      = ""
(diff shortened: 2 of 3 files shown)


pkg/config/types.go
```diff
@@ -95,6 +95,7 @@ const (
 	defaultKubeConfig      = ""
 	defaultBPFSampleRate   = 0
 	defaultCPUArchOverride = ""
+	defaultExcludeSwapperProcess = true
```
rootfs, line 98: This is a new option, let's make the default false first.

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
```diff
@@ -89,6 +89,11 @@ func UpdateProcessBPFMetrics(bpfExporter bpf.Exporter, processStats map[uint64]*
 	for _, ct := range processesData {
 		comm := C.GoString((*C.char)(unsafe.Pointer(&ct.Comm)))
 
+		if ct.Pid == 0 && config.ExcludeSwapperProcess() {
```
rootfs, line 92: how about add a new metrics called idle cpu time?

pkg/config/types.go
vimalk78: updated, to change default behavior backward compatible.

pkg/collector/resourceutilization/bpf/process_bpf_collector.go
sthaha: Yes, this is a good idea and aligns with  user / system / idle time in https://man7.org/linux/man-pages/man5/proc_stat.5.html

sthaha: @vimalk78 with this patch are you able to get sum(rate(kepler_process_bpf_cpu_time_ms_total[20s]))) == sum(rate(node_cpu_seconds_total{mode=~"user|system"}[20s]))

vimalk78: with this change, under load condition
image

rootfs:
I am merging this now and will come up with a new idle time metrics PR.

rootfs:
@vimalk78 @sthaha @KaiyiLiu1234  @sunya-ch 

I tested this idle time exclusion in metal CI. There is improvement on SGD trainer but not much on exponential or logarithmic trainer.

Please find the new config parameter in kepler [metal](https://sustainable-computing-io.github.io/kepler-metal-ci/train-validate-e2e/2024-11-04_18-53-25/AbsPower/BPFOnly/SGDRegressorTrainer/report-v0.7.12-27-g75b95332.html) and [VM](https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/ansible/roles/kepler_systemd/tasks/vm.yml#L28) systemd.

In the trained model validation [report](https://sustainable-computing-io.github.io/kepler-metal-ci/kepler-model-train-validate.html), those in Nov 01 are without this change, and those in Nov 04 are with this new config param. The SGD trainer MAPE sees  improvement from 51% to 21%

sthaha:
We should see improvement in all models. @sunya-ch is there a way to plot or export all metrics used for training the model ?

PR #1832, opened 2024-11-01
chore(compose): switch to stable Prometheus version

vprashar2929:
This commit updates the compose config to use a stable Prometheus version instead of the main development version.
Additionally, it resolves an issue where Prometheus could not scrape metrics from Scaphandre, which was failing due to an invalid `Content-Type` header in Scaphandre's response.


vprashar2929:
Ref:

Screenshot 2024-11-01 at 7 26 36 PM

sthaha (changes requested):
Lets use the latest as it point to the latest stable release 
https://quay.io/repository/prometheus/prometheus?tab=tags&tag=latest

rootfs (approved):

PR #1834, opened 2024-11-04
feat(validator): add support to validate kepler metrics

vprashar2929:


e2e/tools/validator/metric_validations.yaml
```diff
@@ -0,0 +1,319 @@
...
+    dev: |
+      sum(
+        rate(
+          kepler_process_joules_total{{
+            job="{dev_job_name}",
+          }}[{rate_interval}]
+        )
+      )
+
+    max_mae: 2.0
```
sthaha, line 26: max_mae of 2 (watts) is a large error. Lets 0.2 instead.

e2e/tools/validator/src/validator/config/__init__.py
```diff
@@ -31,6 +31,8 @@ class Metal(NamedTuple):
 class PrometheusJob(NamedTuple):
     metal: str
     vm: str
+    dev: str
```
sthaha, line 34: we don't need this if we just hard code "latest" and "dev" in the queries.

e2e/tools/validator/metric_validations.yaml
```diff
@@ -0,0 +1,319 @@
+config:
+  mapping:
+    actual: latest
+    predicted: dev
+
+validations:
+  # absolute power comparison
+  - name: Total - absolute
```
sthaha, line 8: Can we also validate these invariants in the same version of `dev` 

* kepler_node_<pkg|core|uncore|dram|other>{dev} = sum of ( process_<pkg|core|uncore|dram|other>{dev} ) 
* kepler_node_<pkg|core|..> = node_exporter_rapl_<pkg|core...>
*sum( kepler_process_bpf_cpu ) = node_exporter_cpu_time

e2e/tools/validator/src/validator/cli/__init__.py
```diff
@@ -610,6 +611,34 @@ def validate_acpi(cfg: config.Validator, duration: datetime.timedelta, report_di
...
+    click.secho("  * Generating build and node info ...", fg="green")
+    res.build_info, res.node_info = get_build_and_node_info(cfg.prometheus)
+    click.secho("  * Generating spec report ...", fg="green")
+    res.host_spec = get_host_spec()
+    script_result = ScriptResult(res.start_time, res.end_time)
+    res.validations = run_validations(cfg, script_result, results_dir)
+    click.secho("  * Generating validate metrics report file and dir", fg="green")
+    write_md_report(results_dir, res)
+
+    raise Exit(1) if not res.validations.passed else Exit(0)
```
sthaha, line 639: can't we just return 0 or 1 ?

vprashar2929: by default when a Click script is invoked as command line application (through BaseCommand.main()) the return value is ignored unless the standalone_mode is disabled in which case it’s bubbled through.
https://click.palletsprojects.com/en/stable/commands/#command-return-values

e2e/tools/validator/metric_validations.yaml
vprashar2929: for `kepler_node_<pkg|core|dram...>{dev} = sum of (process_<pkg|core|dram....>){dev}` do you mean 
MAE of `sum(rate(kepler_node<pkg|core|dram>){dev}[20s]) and sum(rate(process_<pkg|core|dram>{dev}[20s]))`?

e2e/tools/validator/validator-regression.yaml
```diff
@@ -0,0 +1,30 @@
...
+  vm:
+    pid: 123456 # Process ID for the KVM process running on metal
+
+prometheus:
+  job:
+    vm: vm  # Job name for virtual machine metrics, default is vm
+    metal: metal  # Job name for metal metrics, default is metal
+
+  url: http://localhost:9090 # Prometheus server URL
+  rate_interval: 60s  # Rate interval for Promql, default is 20s, typically 4 x $scrape_interval
```
vprashar2929, line 20: Explicitly using rate interval as 60s because:
```
Prometheus scrape Interval = 3s
Data points for 12s Interval(i.e 4* scrape interval) = 12/3 = 4 data points
Data points for 60s interval = 60/3 = 20 data points
```
With 20 data points, we get a smoother and more reliable estimate. When comparing two `sum(rate(...))` a stable rate reduces the variability in MAE calculations leading to more accurate assessments.

manifests/compose/monitoring/prometheus/prometheus.yml
```diff
@@ -1,5 +1,5 @@
 global:
-  scrape_interval: 5s # Set the scrape interval to every 5 seconds. Default is every 1 minute.
+  scrape_interval: 3s # Set the scrape interval to every 5 seconds. Default is every 1 minute.
```
vimalk78, line 2: check why changed scrape interval, and update comment accordingly

vprashar2929: setting scrape every 3 seconds rather than every 5 seconds, over a typical time window will collect significantly more data

vprashar2929:
Here is sample CI run that would look like for reference once we have this merged: https://github.com/sustainable-computing-io/kepler-metal-ci/actions/runs/12366281744/job/34512777104

My idea is to use the equinix runners on demand on PR's. Reviewers or authors can add a comment in the PR something like `/test-regression` which will trigger a workflow like this which can test if metrics produced by PR code base Kepler are off to what is already present in `latest`

sthaha:
@vprashar2929 can we bring this into `reboot` ? Perhaps validator needs a bit of rewrite to only test baremetal

PR #1847, opened 2024-11-15
chore(compose): use fixed version of prometheus

vprashar2929:
This commit fixes the issue where prometheus using `latest` is unable to scrape scaphandre metrics due to issue in scaphandre of having an invalid `Content-Type` header in Scaphandre's response.

### DIFF
--- a/manifests/compose/monitoring/prometheus/Dockerfile
+++ b/manifests/compose/monitoring/prometheus/Dockerfile
@@ -1,3 +1,3 @@
-FROM quay.io/prometheus/prometheus:latest
+FROM quay.io/prometheus/prometheus:v2.55.0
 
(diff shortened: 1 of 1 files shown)


vprashar2929:
Ref:
Screenshot 2024-11-15 at 11 14 08 PM

manifests/compose/monitoring/prometheus/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM quay.io/prometheus/prometheus:latest
+FROM quay.io/prometheus/prometheus:v2.55.0
```
sthaha, line 1: Could you please add a comment for pinning the version down?

sthaha (approved):

PR #1858, opened 2024-11-21
fix(cpu): fixed reading cpus.yaml

vimalk78:
- fixed reading cpus.yaml file
- added unit tests

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -34,6 +34,7 @@ import (
 const (
 	cpuModelDataPath = "/var/lib/kepler/data/cpus.yaml"
 	cpuPmuNamePath   = "/sys/devices/cpu/caps/pmu_name"
+	unknownCPUArch   = "unknown"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/node/node.go
```diff
@@ -215,44 +216,52 @@ func readCPUModelData() ([]byte, error) {
...
+func cpuMicroArchitecture(family, model, stepping string) (string, error) {
+	yamlBytes, err := readCPUModelData()
+	if err != nil {
+		klog.Errorf("failed to read cpus.yaml: %v", err)
+		return "", err
+	}
+	arch, err := cpuMicroArchitectureFromModel(yamlBytes, family, model, stepping)
+	if err != nil {
+		return arch, nil
+	}
```
sthaha, line 264:
```suggestion
	if err == nil {
		return arch, nil
	}
```

sthaha (approved):
lgtm

PR #1859, opened 2024-11-22
[fix]: base image elfutils version issue

SamYuan1990:
ref https://github.com/sustainable-computing-io/kepler/actions/runs/11814219047/job/32912842046
it seems with elfutils version upgrade, our base image code need to be update accordingly.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -55,13 +55,23 @@ jobs:
   changes_baseimage:
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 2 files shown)


SamYuan1990:
it seems build success and fail at image sign process without secret available as PR from fork.

SamYuan1990:
it seems build success and fail at image sign process without secret available as PR from fork.

vprashar2929 (approved):

SamYuan1990:
but should we consider to bump up base image tag to libbpf 1.4?

PR #1867, opened 2024-11-29
[fix]: adjust gosec permissions settings according to GHA ref

### LINKED ISSUE #1865: gosec workflow fails due to resource not found
After the merge of #1864 the gosec workflow which gets triggered at the commit to main branch fails due to resource not found. Could be due to read permission set for the workflow. 

CI failure ref: https://github.com/sustainable-computing-io/kepler/actions/runs/12072619861/job/33667898660

SamYuan1990:
fix #1865
ref https://github.com/github/codeql-action?tab=readme-ov-file#workflow-permissions

### DIFF
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -6,7 +6,8 @@ on: # yamllint disable-line rule:truthy
   workflow_call:
 
 permissions:
-  pull-requests: read
+  security-events: write
+  contents: read
 
 jobs:
   tests:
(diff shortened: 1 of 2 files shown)


.github/workflows/gosec.yml
```diff
@@ -6,7 +6,8 @@ on: # yamllint disable-line rule:truthy
   workflow_call:
 
 permissions:
-  pull-requests: read
+  security-events: write
```
SamYuan1990, line 9: ref https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token which is interesting, as a PR level permission impacts security events. but meanwhile https://github.com/sustainable-computing-io/kepler/actions/runs/12065944361 daily job works, I suppose that at the merge event, the PR info is used. Anyway, in this PR updated both gosec and daily job with same permission.

vprashar2929 (approved):

PR #1868, opened 2024-11-29
[test]: add test case on package cgroup

SamYuan1990:
add test case on package cgroup
remove duplicate function
fix a cache set logic when error happen on code logic

### DIFF
--- a/pkg/cgroup/resolve_container.go
+++ b/pkg/cgroup/resolve_container.go
@@ -119,12 +119,9 @@ func (c *cache) getContainerIDFromCache(pid uint64) (string, bool) {
 }
 
 func (c *cache) getGetContainerIDFromPID(pid uint64) (string, error) {
-	p, ok := instance.containerIDCache.Load(pid)
+	containerID, ok := instance.getContainerIDFromCache(pid)
 	if ok {
-		containerID, ok := p.(string)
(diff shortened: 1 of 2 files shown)


pkg/cgroup/resolve_container.go
```diff
@@ -178,6 +175,7 @@ func (c *cache) getPathFromcGroupID(cgroupID uint64) (string, error) {
 
 		getCgroupID, err := utils.GetCgroupIDFromPath(instance.byteOrder, path)
 		if err != nil {
+			instance.cGroupIDToPath.Store(cgroupID, unknownPath)
```
SamYuan1990, line 178: @sunya-ch , it seems we want to set cache with key value as cgroupID, unknownPath when error happen, but in previous, the cache is never set, as the code will return before cache setting?

sunya-ch: @SamYuan1990 we should not store cgroupID and unknownPath here.
This is inside the loop to set cGroupIDToPath values for all traversed path in cgroupPath path.
If the path doesn't provide cgroup (getCgroupID) value, we will just ignore and continue checking the next path (return err in func parameter of WalkDir).

sunya-ch (approved):
LGTM!

PR #1871, opened 2024-12-02
[fix]: fix CI permissions according to failed CI jobs

SamYuan1990:
fix CI permissions according to failed CI jobs

ref 
https://github.com/sustainable-computing-io/kepler/actions/runs/12104149942
https://github.com/sustainable-computing-io/kepler/actions/runs/12100596870

### DIFF
--- a/.github/workflows/pre-commit-auto-update.yml
+++ b/.github/workflows/pre-commit-auto-update.yml
@@ -4,8 +4,6 @@ on: # yamllint disable-line rule:truthy
   schedule:
     - cron: 0 0 1 * *
 
-permissions: read-all
-
 jobs:
   auto-update:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-commit-auto-update.yml
```diff
@@ -4,8 +4,6 @@ on: # yamllint disable-line rule:truthy
   schedule:
     - cron: 0 0 1 * *
 
-permissions: read-all
```
maryamtahhan, line 7: I think this needs to be 

```
permissions:
  contents: write
  pull-requests: write
```

SamYuan1990: cool, updated!

maryamtahhan (approved):

PR #1876, opened 2024-12-02
chore(config): if the config dir doesn't exist, create it

rootfs:
fixing the CI error


KaiyiLiu1234 (approved):
I will preapprove. Will Kepler still look at the env variables?

SamYuan1990 (approved):

maryamtahhan (approved):

rootfs:
@KaiyiLiu1234 yes, the env var are still valid if the config file is empty

PR #1878, opened 2024-12-03
feat(baremetal_validator): Add Validator for Process, Container Metrics

KaiyiLiu1234:
This commit includes validation setups for Process and Container RAPL related Baremetal power metrics (node, process, container).

### DIFF
--- a/e2e/tools/validator/scripts/targeted_stresser.sh
+++ b/e2e/tools/validator/scripts/targeted_stresser.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+trap exit_all INT
+exit_all() {
+	pkill -P $$
+}
+
+run() {
+	echo "❯ $*"
+	"$@"
+	echo "      ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾"
+}
+
+usage() {
+    echo "Usage: $0 -g <general_mode> -r <cpu_range> -c <cpus> -d <mount_dir> -t <time_interval_log_name> -l <load_curve> -n <iterations>"
+    echo "  -g <general_mode> If set, <cpu_range> and <cpus> are ignored."
+    echo "  -r <cpu_range>   CPU range for stress-ng taskset (Default: '15')"
+    echo "  -c <cpus>    Number of CPUs to use for stress-ng (Default: '1')"
+    echo "  -d <mount_dir>   Directory to mount for logging (Default: '/tmp')"
+    echo "  -t <time_interval_log_name> Filename for start and end time file log (Default: 'time_interval.log')"
+    echo "  -l <load_curve>  Load curve as a comma-separated list (Default: '0:5,50:20,75:20,100:20,75:20,50:20')"
+    echo "  -n <iterations> Number of times to iterate the Load curve (Default: '1')"
+    exit 1
+}
+
+main() {
+
+    set_general_mode=false
+    DEFAULT_CPU_RANGE="15"
+    DEFAULT_CPUS="1"
+    DEFAULT_MOUNT_DIR="/tmp"
+    DEFAULT_LOAD_CURVE_STR="0:5,50:20,75:20,100:20,75:20,50:20"
+    DEFAULT_TIME_INTERVAL_LOG_NAME="time_interval.log"
+    DEFAULT_ITERATIONS="1"
+
+    # Parse command-line options
+    while getopts "g:r:c:d:t:l:n:" opt; do
+        case "$opt" in
+            g) set_general_mode=true ;;
+            r) cpu_range="$OPTARG" ;;
+            c) cpus="$OPTARG" ;;
+            d) mount_dir="$OPTARG" ;;
+            t) time_interval_log_name="$OPTARG" ;;
+            l) load_curve_str="$OPTARG" ;;
+            n) iterations="$OPTARG" ;;
+            *) usage ;;
+        esac
+    done
+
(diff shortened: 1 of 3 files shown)


e2e/tools/validator/scripts/targeted_stresser.sh
vprashar2929, line 1: Couldn't we modify the existing stressor script instead of adding new ones?

e2e/tools/validator/src/validator/config/__init__.py
```diff
@@ -57,6 +57,113 @@ def __repr__(self):
...
+    node_config = config["node"]
+    process_config = config["process"]
+    container_config = config["container"]
+    # node config
+    if not node_config:
+        node_config = {}
+    node = Local(
+        load_curve=node_config.get("load_curve", default_config["load_curve"]),
+        iterations=node_config.get("iterations", default_config["iterations"]),
+        mount_dir=os.path.expanduser(node_config.get("mount_dir", default_config["mount_dir"]))
```
vprashar2929, line 133: Why do we need to mount the directory? Are we deploying Kepler on k8s or via docker compose on BM?

e2e/tools/validator/src/validator/config/__init__.py
```diff
@@ -57,6 +57,113 @@ def __repr__(self):
...
+    )
+    print(node)
+    if not process_config:
+        process_config = {}
+    process = LocalProcess(
+        isolated_cpu=process_config.get("isolated_cpu", default_config["isolated_cpu"]),
+        load_curve=process_config.get("load_curve", default_config["load_curve"]),
```
vprashar2929, line 140: What is the difference in the load curve between the process and the container?

e2e/tools/validator/scripts/targeted_stresser.sh
```diff
@@ -0,0 +1,85 @@
...
+    iterations="${iterations:-$DEFAULT_ITERATIONS}"
+
+    IFS=',' read -r -a load_curve <<< "$load_curve_str"
+
+    TIME_INTERVAL_LOG="${mount_dir}/${time_interval_log_name}"
+
+    > "$TIME_INTERVAL_LOG"
+
+    start_time=$(date +%s)
+    echo "Stress Start Time: $start_time" >> "$TIME_INTERVAL_LOG"
```
vprashar2929, line 66: AFAIK we do something similar in our existing validator source by checking the [start and end time](https://github.com/sustainable-computing-io/kepler/blob/878599d33cd0cd2eab49a8936890cf9e4ba06c54/e2e/tools/validator/src/validator/cli/__init__.py#L439) of the `stressor.sh` script. I think we can do similar in this case as well instead of computing start and end time in the bash script itself

e2e/tools/validator/src/validator/config/__init__.py
KaiyiLiu1234: No difference. the logic is the same

e2e/tools/validator/scripts/targeted_stresser.sh
KaiyiLiu1234: The reason this is necessary is primarily for container validation and in future if we attempt pods (which should fall under container still anyway). Container setup requires an image installation (in this case, fedora and stressng need to be installed). If we get start and end time outside we will include the time for installing which will skew the results. The easiest solution I can think of was to actually compute it in the bash script so we do not include the installation time.

e2e/tools/validator/scripts/targeted_stresser.sh
KaiyiLiu1234: I did not want to modify the other scripts because it might break the metal ci. I think this script is the best to replace the other stressor scripts (as it enables a custom load curve). A future PR should be made for modifying validator to have just one goto stressor script.

e2e/tools/validator/src/validator/config/__init__.py
KaiyiLiu1234: Currently, kepler is deployed with docker compose for simplicity. however, this should still work the same even in a kubernetes environment. The mount directory is specifically to get the start and end times of the stressor script as a log file from the container. In a non container, the log file will be saved at the mount directory.

vprashar2929: Shouldn't we target a single environment for now and focus on expanding it further? Currently, we use docker compose for our development and validations. Let's stick with that for now and afterwards add support for k8s 
Wdyt?

e2e/tools/validator/src/validator/config/__init__.py
vprashar2929: Then why we require separate configs for both?

e2e/tools/validator/validator.bm.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+prometheus:
+  job: metal # Job name for baremetal metrics
+  url: http://localhost:9091 # Prometheus server URL
+  rate_interval: 21s  # Rate interval for Promql, default is 20s, typically 4 x $scrape_interval
+  step: 3s  # Step duration for Prometheus range queries
+
+config: # default settings
+  isolated_cpu: "15" # Logical processor that is fully isolated from scheduler (ex. isolcpus)
+  load_curve: "0:15,10:20,25:20,50:20,75:20,100:30,75:20,50:20,25:20,10:20,0:15"
```
vprashar2929, line 11: Why can't we add a new load curve in the existing stressor and not make it part of the config as it overcomplicated the config part How's this different from what [default load curve](https://github.com/sustainable-computing-io/kepler/blob/021b5447285065b7d34b6901da57c6b8c82624bd/e2e/tools/validator/scripts/stressor.sh#L32) gives us?

vprashar2929:
@KaiyiLiu1234 Can you take a look at [this](https://github.com/sustainable-computing-io/kepler/pull/1834)? I believe this somewhat tries to achieve what we want from BM validations. Also please give it a go as well

KaiyiLiu1234:
Sure I can take a look. We should probably organize tasks and merge the work between our PRs (I have no issue with directly merging your work and then adding what I did here on to yours - for container validation). In the meantime, I am going to just see if i can get some results asap from the metal ci.

PR #1882, opened 2024-12-04
chore(ci): migrate mock-acpi workflow to GH runner

vprashar2929:
This commit migrates the mock-acpi workflow to use the GitHub runner instead of the Equinix self-hosted runner. Since the workflow is designed for testing ACPI functionality, using a mock, a self-hosted runner is not required.

Running the workflow on the GitHub runner, which operates as a VM, enables execution on every pull request, ensuring consistent validation of ACPI functionality for Kepler.

### DIFF
--- a/.github/workflows/acpi-validation.yml
+++ b/.github/workflows/acpi-validation.yml
@@ -1,0 +1,39 @@
+# This workflow will run mock-acpi compose on GH runner and run validator tool
+
+name: Validate ACPI
+
+on: # yamllint disable-line rule:truthy
+  workflow_call:
+
+jobs:
+  validate-acpi:
+    name: Validate ACPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: install hatch
+        run: |
+          python -m pip install --upgrade pip
+          pip install hatch
+
+      - name: Run mock-acpi compose
+        working-directory: manifests/compose/mock-acpi
+        run: |
+          docker compose up -d
+
+      - name: Warm up for 60 seconds
+        run: |
+          echo "Warm up for 60 seconds before running validator"
+          sleep 60
+
+      - name: Run mock-acpi validator tool
+        working-directory: e2e/tools/validator
+        run: |
+          hatch run validator -f ./validator.mock-acpi.yaml validate-acpi -d 1m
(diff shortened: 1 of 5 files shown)


vprashar2929:
The existing mock-acpi workflow file and ansible code can kept as they will be reused when adding #1834

.github/workflows/acpi-validation.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: Validate ACPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
SamYuan1990, line 19: do we need specific python version?
if so, we may have to upgrade manually when GHA bump up runner OS version.

.github/workflows/acpi-validation.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Run mock-acpi compose
+        working-directory: manifests/compose/mock-acpi
+        run: |
+          docker compose up -d
+
+      - name: Wait for services to be ready
+        run: |
+          echo "Waiting for services to start..."
+          sleep 60
```
SamYuan1990, line 34: from robust part of view, can we make a timeout? as a loop with condition to wait for compose up and timeout break as failure?

.github/workflows/pull_request.yml
```diff
@@ -26,6 +26,9 @@ jobs:
   # for each PR run integration test
   integration_test:
     uses: ./.github/workflows/integration_test.yml
+  # for each PR run ACPI validation
+  acpi_validation:
+    uses: ./.github/workflows/acpi-validation.yml
```
SamYuan1990, line 31: as it's a PR level job, how do we ensure it build latest container image per PR? if the compose file just latest, so... the PR will use latest kepler image(which build in last night) and nothing checked.

.github/workflows/acpi-validation.yml
vprashar2929: Sure

.github/workflows/acpi-validation.yml
vprashar2929: Ideally, it should be warm-up time for Kepler to start producing metrics. I have updated this step

.github/workflows/pull_request.yml
vprashar2929: We are building Kepler as part of our compose: https://github.com/sustainable-computing-io/kepler/blob/878599d33cd0cd2eab49a8936890cf9e4ba06c54/manifests/compose/mock-acpi/compose.yaml#L11

SamYuan1990 (approved):
LGTM from CI part of view.

PR #1888, opened 2024-12-09
chore: update watcher

maryamtahhan:


SamYuan1990 (approved):
LGTM, but please hold on for merge as it contains k8s api version bump up. @rootfs, as far as I can remember, we take care for any dep version bump up as it  contains k8s api?

PR #1894, opened 2024-12-12
[fix]: enhance kubelet package impl

SamYuan1990:
keep kubelet token in memory to reduce os.file read for each time to get auth token.
if token expired, try again.
refactor with var injection for test case impl.


sunya-ch (approved):
LGTM. Thank you.

PR #1895, opened 2024-12-14
[fix]: remove unused func

SamYuan1990:
it seems we don't need updateProcessResourceUtilizationMetrics function any more as it just done a wait group.
and if so, we can remove that wait group, as the wait group has 2 funcs to wait, and we removed one.

### DIFF
--- a/pkg/collector/metric_collector.go
+++ b/pkg/collector/metric_collector.go
@@ -18,7 +18,6 @@ package collector
 
 import (
 	"os"
-	"sync"
 	"syscall"
 	"time"
 
@@ -63,7 +62,7 @@ type Collector struct {
 
 func NewCollector(bpfExporter bpf.Exporter) *Collector {
 	bpfSupportedMetrics := bpfExporter.SupportedMetrics()
-	c := &Collector{
+	c := Collector{
 		NodeStats:           *stats.NewNodeStats(),
 		ContainerStats:      map[string]*stats.ContainerStats{},
 		ProcessStats:        map[uint64]*stats.ProcessStats{},
@@ -71,7 +70,7 @@ func NewCollector(bpfExporter bpf.Exporter) *Collector {
 		bpfExporter:         bpfExporter,
 		bpfSupportedMetrics: bpfSupportedMetrics,
 	}
-	return c
+	return &c
 }
 
 func (c *Collector) Initialize() error {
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/watcher.go
```diff
@@ -99,15 +99,15 @@ func newK8sClient() *kubernetes.Clientset {
 }
 
 func NewObjListWatcher(bpfSupportedMetrics bpf.SupportedMetrics) (*ObjListWatcher, error) {
-	w := &ObjListWatcher{
+	w := ObjListWatcher{
```
SamYuan1990, line 102: fix https://github.com/sustainable-computing-io/kepler/actions/runs/12288498261/job/34292384684#step:4:981

pkg/collector/metric_collector.go
```diff
@@ -63,15 +62,15 @@ type Collector struct {
 
 func NewCollector(bpfExporter bpf.Exporter) *Collector {
 	bpfSupportedMetrics := bpfExporter.SupportedMetrics()
-	c := &Collector{
+	c := Collector{
```
SamYuan1990, line 65: fix https://github.com/sustainable-computing-io/kepler/actions/runs/12288498261/job/34292384684#step:4:1041

pkg/collector/metric_collector.go
```diff
@@ -139,22 +138,12 @@ func (c *Collector) UpdateProcessEnergyUtilizationMetrics() {
...
-	go c.updateNodeResourceUtilizationMetrics(wg)
-	go c.updateProcessResourceUtilizationMetrics(wg)
-	wg.Wait()
+	c.updateProcessResourceUtilizationMetrics()
 	// aggregate processes' resource utilization metrics to containers, virtual machines and nodes
 	c.AggregateProcessResourceUtilizationMetrics()
 }
 
-// update the node metrics that are not related to aggregated resource utilization of processes
-func (c *Collector) updateNodeResourceUtilizationMetrics(wg *sync.WaitGroup) {
```
sunya-ch, line 152: Do we want to keep this for TODO list when we have metrics those are not related to aggregated resource utilization of processes. @rootfs

sunya-ch (commented):
The fix on heap escapes are good to me. 
However, I'm not sure about the removal of unimplemented `updateNodeResourceUtilizationMetrics`.

PR #1899, opened 2024-12-20
feat: add TLS endpoint to kepler exporter

aharivel:
Add TLS support via the new web configuration file, following the Prometheus Exporter Toolkit style for consistency across exporters.

* Usage:

kepler --web.config.file=web-config.yml

* Content of web-config.yml:

tls_server_config:
  cert_file: /path/to/server.crt
  key_file: /path/to/server.key

### DIFF
--- a/cmd/exporter/exporter.go
+++ b/cmd/exporter/exporter.go
@@ -37,6 +37,7 @@ import (
 	"github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator"
 	"github.com/sustainable-computing-io/kepler/pkg/sensors/components"
 	"github.com/sustainable-computing-io/kepler/pkg/sensors/platform"
+	"gopkg.in/yaml.v3"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -50,6 +51,15 @@ const (
 	finishingMsg = "Exiting..."
 	startedMsg   = "Started Kepler in %s"
 )
+
+type TLSConfig struct {
+	CertFile string `yaml:"cert_file"`
+	KeyFile  string `yaml:"key_file"`
+}
+
+type TLSServerConfig struct {
+	TLSConfig TLSConfig `yaml:"tls_server_config"`
+}
 
 // AppConfig holds the configuration info for the application.
 type AppConfig struct {
@@ -66,6 +76,7 @@ type AppConfig struct {
 	ExposeEstimatedIdlePower     bool
 	MachineSpecFilePath          string
 	DisablePowerMeter            bool
+	TLSFilePath                  string
 }
 
 func newAppConfig() *AppConfig {
@@ -84,6 +95,7 @@ func newAppConfig() *AppConfig {
 	flag.BoolVar(&cfg.ExposeEstimatedIdlePower, "expose-estimated-idle-power", false, "Whether to expose the estimated idle power as a metric")
 	flag.StringVar(&cfg.MachineSpecFilePath, "machine-spec", "", "path to the machine spec file in json format")
 	flag.BoolVar(&cfg.DisablePowerMeter, "disable-power-meter", false, "whether manually disable power meter read and forcefully apply the estimator for node powers")
+	flag.StringVar(&cfg.TLSFilePath, "web.config.file", "", "path to TLS web config file")
 
 	return cfg
 }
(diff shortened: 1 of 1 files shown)


SamYuan1990 (changes requested):
https://prometheus.io/docs/introduction/roadmap/#tls-and-authentication-in-http-serving-endpoints I don't mind for TLS endpoint, but considering with prometheus TLS note for official exporters...
> TLS and authentication in HTTP serving endpoints
> > TLS and authentication are currently being rolled out to the Prometheus, Alertmanager, and the official exporters. Adding this support will make it easier for people to deploy Prometheus components securely without requiring a reverse proxy to add those features externally.

at meanwhile, if we are going to enable TLS for kepler, considering in a k8s cluster with node A,B,Cwhat will be the hostname in cert for the TLS cert?what will be the changes for kepler pod? as for example if we are going to use kepler.svc.nodeX.com as host name in k8s cluster, and the key for each node are different.so, somehow will it impact/it will change our deployment yaml? as we have to inject/put a specific node cert-key pair as configuration?

aharivel:
Hi @SamYuan1990 , 
I wanted to clarify something regarding your comments about Prometheus official support. Could it be that there was a mix-up between the terms "rolled out" and "rules out"?

While I don't claim expertise in Kubernetes security or managing certificates in multi-node deployments, I can point out that some IaaS platforms, like Red Hat OpenStack (RHOSO), are already using node_exporter with TLS. Here are a couple of links to highlight this:

[Node Exporter Configuration](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/templates/node_exporter.yaml.j2)
[Node Exporter Setup Tasks
](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/tasks/configure.yml)
Furthermore, they are also using Kepler, as shown here:
[Kepler Integration Tasks](https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_telemetry_power_monitoring/tasks)

This leads me to my point: if someone knows how to manage their certificates effectively, Kepler should be capable of supporting TLS endpoints. Why should Kepler be the only exporter in an IaaS setup like OpenStack that does not support TLS, especially when node_exporter, ipmi_exporter, and others do?

Additionally, Kepler is handling sensitive data. For instance, consider the CVEs related to RAPL and side-channel attacks, which I’m sure you’re familiar with. Would organizations that prioritize security really allow this data to be transmitted in plaintext over the network?

Finally, I want to stress that this is an optional feature. For users content with unencrypted data, nothing changes—they can continue as before. But adding this option would present Kepler as a more robust and serious tool for those who care about security.

SamYuan1990:
> Hi @SamYuan1990 , I wanted to clarify something regarding your comments about Prometheus official support. Could it be that there was a mix-up between the terms "rolled out" and "rules out"?
> 
> While I don't claim expertise in Kubernetes security or managing certificates in multi-node deployments, I can point out that some IaaS platforms, like Red Hat OpenStack (RHOSO), are already using node_exporter with TLS. Here are a couple of links to highlight this:
> 
> [Node Exporter Configuration](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/templates/node_exporter.yaml.j2) [Node Exporter Setup Tasks ](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/tasks/configure.yml) Furthermore, they are also using Kepler, as shown here: [Kepler Integration Tasks](https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_telemetry_power_monitoring/tasks)
> 
> This leads me to my point: if someone knows how to manage their certificates effectively, Kepler should be capable of supporting TLS endpoints. Why should Kepler be the only exporter in an IaaS setup like OpenStack that does not support TLS, especially when node_exporter, ipmi_exporter, and others do?
> 
> Additionally, Kepler is handling sensitive data. For instance, consider the CVEs related to RAPL and side-channel attacks, which I’m sure you’re familiar with. Would organizations that prioritize security really allow this data to be transmitted in plaintext over the network?
> 
> Finally, I want to stress that this is an optional feature. For users content with unencrypted data, nothing changes—they can continue as before. But adding this option would present Kepler as a more robust and serious tool for those who care about security.

From security point of view, yes. We can make kepler protected by TLS to improve our baseline for sensitive data protection.
Once more, this PR is a great job and I just "mark with change request for more information and discussion."

### So from deployment point of view, considering with kepler community operator and kepler community helm chat as deployment approaches.
- [ ] do we have issue open on kepler operator or kepler helm chat with TLS support?
- [ ] are we going to integrate with cert mgr or sidecar as istio? by document supporting or we just leave it to user?
- [ ] what if kepler running as a static pod?

and btw, any impacts to kepler RPM service?

### from security point:
what's specific level for kepler's sensitive data? what are they?
> Additionally, Kepler is handling sensitive data.

how to protect those data according to it's level? to clear/clarify our security boundary.
- [ ] do we need considering a key/a cert with password?
- [ ] do we recommended user to enable tls verify when enable tls?
- [ ] do we need considering mtls?

> for instance:
> if we considering pod list is a sensitive data, and which should not listed by plaintext through network, as kubelet having TLS to protect those info. we should add TLS to protect those info as well. I suppose this is from node mgr's points of view.
> if we don't want user A to see user B's pod on the kepler metric... should we considering username and password? I suppose this is from user's point of view.

### from impl points of view:
is there any chance for us to reuse https://github.com/prometheus/exporter-toolkit
ref
https://github.com/prometheus/node_exporter/blob/d0c1d00d18867297326bc7b1c21de36ae50e6a92/README.mdplain=1#L383
reuse https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md
https://github.com/prometheus/exporter-toolkit/tree/master/web 
https://github.com/prometheus/node_exporter/blob/d0c1d00d18867297326bc7b1c21de36ae50e6a92/go.mod#L28

SamYuan1990 (approved):
In this case, I will change my review to LGTM, we can have a TLS impl in short term and for long term plan, move to prometheus community solution. I will leave it to @rootfs , @vprashar2929 , @sthaha or @vimalk78 to merge this PR.
if we are going to merge current, do we need a new issue for prometheus tookit?

PR #1809, opened 2024-10-16
fix: remove .dockerignore to ensure clean version computation

vprashar2929:
This commit removes the `.dockerignore` file from the repository. The current build process relies on `git describe` to compute the Kepler version. However, the `.dockerignore` file was causing certain Git-tracked files to be excluded from the Docker build context, leading to `dirty` version state. By removing the `.dockerignore` file, we ensure that all files are included, allowing Git to recognize a clean working directory during the image build process

Address: #1808

PR #1913, opened 2025-01-21
Fix accelerator device map

vimalk78:
- brings back `.dockerignore` file
- fix the logic of adding accelerator device map
- added few log statements


rootfs:
can you sign DCO?

vimalk78:
> can you sign DCO?

@rootfs signed DCO

.dockerignore
vprashar2929, line 1: Adding .dockerignore will lead to incorrect version computation https://github.com/sustainable-computing-io/kepler/pull/1809

pkg/sensors/accelerator/devices/dcgm.go
```diff
@@ -138,6 +139,7 @@ func (d *gpuDcgm) InitLib() (err error) {
 	if err != nil {
 		klog.Infof("There is no DCGM daemon running in the host: %s", err)
 		// embedded mode is not recommended for production per https://github.com/NVIDIA/dcgm-exporter/issues/22#issuecomment-1321521995
+		klog.Info("Attempting to inilialize dcgm in Embedded mode.")
```
sthaha, line 142: nit typo: initialize

rootfs (approved):

pkg/sensors/accelerator/accelerator.go
```diff
@@ -130,7 +131,10 @@ func New(atype string, sleep bool) (Accelerator, error) {
 
 	// Init the available devices.
 
-	devs := devices.GetRegistry().GetAllDeviceTypes()
+	r := devices.GetRegistry()
+	j, _ := json.Marshal(r.GetAllDevices())
+	klog.V(5).Infof("Accelerator Registry AllDevices: %s", string(j))
```
sthaha, line 136: how about we marshal only if klog.V(5).IsEnabled() ?

pkg/sensors/accelerator/devices/device.go
```diff
@@ -121,8 +121,13 @@ func (r *Registry) MustRegister(a string, d DeviceType, deviceStartup deviceStar
...
 	klog.V(5).Infof("Adding the device to the registry [%s][%s]", a, d.String())
-	r.Registry[a] = map[DeviceType]deviceStartupFunc{
-		d: deviceStartup,
+	m, ok := r.Registry[a]
+	if !ok {
+		r.Registry[a] = map[DeviceType]deviceStartupFunc{
+			d: deviceStartup,
+		}
+	} else {
+		m[d] = deviceStartup
```
sthaha, line 130: shouldn't we panic here instead? The accelerator has ready been defined and there is now an attempt to overwrite it. IMHO, we should panic and aligns with the `Must` in MustRegister.

Issue #1837, opened 2024-11-06
kepler node_package does not equal total of kepler_process_package


sthaha:
Steps to reproduce on a Baremetal 
* deploy kepler 
*  curl <kepler-host>/metrics
*  grep for `node_package_joules`
* grep for kepler_process_package_joules
* sum the values of process
* see if the total is different to node_package 

Expected: there shouldn't be any significant difference 
Actual: The difference quite large and grows over time 

Using Prometheus 
* `kepler_node_package_joules_total{job="metal"} - on() sum(kepler_process_package_joules_total{job="metal"})`

![image](https://github.com/user-attachments/assets/db8eb862-a07f-443b-be10-1e35251b5f22)

marvin-steinke:
do you think this is related to #1833 ?

PR #1927, opened 2025-02-20
fix(ratio-model): reduce round off error when attributing power to process

sthaha:
Partially fixes: https://github.com/sustainable-computing-io/kepler/issues/1837

### DIFF
--- a/pkg/collector/metric_collector_test.go
+++ b/pkg/collector/metric_collector_test.go
@@ -39,14 +39,12 @@ func newMockCollector(mockAttacher bpf.Exporter) *Collector {
 }
 
 var _ = Describe("Test Collector Unit", func() {
-
 	BeforeEach(func() {
 		_, err := config.Initialize(".")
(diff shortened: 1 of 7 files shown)


pkg/collector/metric_collector.go
```diff
@@ -138,120 +136,201 @@ func (c *Collector) UpdateProcessEnergyUtilizationMetrics() {
 }
 
 func (c *Collector) updateResourceUtilizationMetrics() {
-	wg := &sync.WaitGroup{}
-	wg.Add(2)
-	go c.updateNodeResourceUtilizationMetrics(wg)
-	go c.updateProcessResourceUtilizationMetrics(wg)
-	wg.Wait()
```
vimalk78, line 145: move `sync.WaitGroup` changes in a separate commit

vprashar2929:
validated against OpenShift

vimalk78:
@sthaha can you please add the key ideas of this change in the PR description?

PR #1929, opened 2025-02-21
fix: Redfish: Use Chassis instead of Systems to obtain chassis ID

YaSuenag:
I'm trying to evaluate Kepler with [dmtf/redfish-interface-emulator](https://hub.docker.com/r/dmtf/redfish-interface-emulator) on my laptop, but it does not seem to work because `/redfish/v1/Systems` endpoint returns system ID, not chassis ID which we  expect.

Redfish client in Kepler attempts to get system ID from `/redfish/v1/Systems`, and it is used in the request to `/redfish/v1/Chassis` to get `PowerControl`. However, according to [Redfish Resource and Schema Guide](https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2023.1.pdf), system ID is different from chassis ID. `PowerControl` is under `Chassis`, so we need to get chassis ID.

**System ID (6.24 ComputerSystem)**
> The ComputerSystem schema represents a computer or system instance and the software-visible resources, or items within the data plane, such as memory, CPU, and other devices that it can access.

**Chassis ID (6.19 Chassis)**
> The Chassis schema represents the physical components of a system. This resource represents the sheet-metal confined spaces and logical zones such as racks, enclosures, chassis and all other containers. Subsystems, such as sensors, that operate outside of a system's data plane are linked either directly or indirectly through this resource.

And also
 Redfish spec does not require to authentication, so this PR allows Redfish configuration without credentials (user/passphrase).

### DIFF
--- a/pkg/nodecred/csv_cred.go
+++ b/pkg/nodecred/csv_cred.go
@@ -46,7 +46,7 @@ func (c csvNodeCred) GetNodeCredByNodeName(nodeName, target string) (map[string]
 		cred["redfish_username"] = credMap["redfish_username"]
 		cred["redfish_password"] = credMap["redfish_password"]
 		cred["redfish_host"] = credMap["redfish_host"]
-		if cred["redfish_username"] == "" || cred["redfish_password"] == "" || cred["redfish_host"] == "" {
+		if cred["redfish_host"] == "" {
 			return nil, fmt.Errorf("no credential found")
 		}
 		return cred, nil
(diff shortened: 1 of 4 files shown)


sunya-ch (commented):
The PR looks good to me in terms of the changes (remove authentication information and use chassis ID instead of system ID). However, I don't think I could approve this PR until the author of this module @rootfs approve.

PR #1934, opened 2025-03-03
feat(idle-power): Update Idle Energy Estimation Calculation

KaiyiLiu1234:
Changes:

- Idle Energy Calculation has been changed to track the minimum known cpu time and absolute energy and maximum known cpu time and absolute energy. Using these two points, kepler calculates the slope and y intercept (idle estimated energy or energy when cpu time == 0). 
- Idle Energy is only exposed when kepler has a sufficiently wide difference between the minimum and maximum known cpu times, the calculated idle energy converge, and negative idle energy estimations are ignored.

Pending Issue (must address):
- CPU Time is provided per node, but idle power and absolute power is provided in per socket. 

Before this PR is ready to merge, the above Pending Issue needs to be addressed.

### DIFF
--- a/pkg/collector/stats/node_stats.go
+++ b/pkg/collector/stats/node_stats.go
@@ -28,8 +28,17 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type Coordinate struct {
-	X, Y float64
+var (
+	minUpperLowerBoundDiff = 0.5
+	historyLength          = 10
+)
+
+type IdleEnergyResult struct {
+	calculatedIdleEnergy float64
+	diff                 float64
+	history              []float64
+	averageIdlePower     float64
+	totalDatapoints      float64
 }
 
 type IdleEnergyCalculator struct {
@@ -38,16 +47,21 @@ type IdleEnergyCalculator struct {
 	// Max CPU Time
 	maxUtilization *Coordinate
 
-	calculatedIdleEnergy float64
+	result *IdleEnergyResult
 }
 
-func (ic *IdleEnergyCalculator) UpdateIdleEnergy(newResutilization float64, newEnergyDelta float64) float64 {
+func (ic *IdleEnergyCalculator) UpdateIdleEnergy(newResutilization float64, newEnergyDelta float64, maxTheoreticalCPUTime float64) {
 	klog.V(5).Infof("New Datapoint Candidate: (%f, %f)", newResutilization, newEnergyDelta)
 	klog.V(5).Infof("Current Min Utilization Datapoint: (%f, %f)", ic.minUtilization.X, ic.minUtilization.Y)
 	klog.V(5).Infof("Current Max Utilization Datapoint: (%f, %f)", ic.maxUtilization.X, ic.maxUtilization.Y)
+	klog.V(5).Infof("Current History: (%v)", ic.result.history)
+	klog.V(5).Infof("Current Spread: (%v)", ic.result.diff)
 	if newResutilization > ic.minUtilization.X && newResutilization < ic.maxUtilization.X {
 		klog.V(5).Infof("Excess Datapoint: (%f, %f)", newResutilization, newEnergyDelta)
-		return ic.calculatedIdleEnergy
+		// Record History
+		klog.V(5).Infof("Push idleEnergy to history")
+		appendToSliceWithSizeRestriction(&ic.result.history, historyLength, ic.result.calculatedIdleEnergy)
+		return
 	}
 
 	if newResutilization <= ic.minUtilization.X {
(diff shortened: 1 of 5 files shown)


KaiyiLiu1234:
@sthaha @sunya-ch @rootfs Please take a look before the upcoming Community Meeting. Thanks!

pkg/collector/stats/node_stats.go
```diff
@@ -19,18 +19,175 @@ package stats
...
 	acc "github.com/sustainable-computing-io/kepler/pkg/sensors/accelerator"
 	"github.com/sustainable-computing-io/kepler/pkg/utils"
+	"k8s.io/klog/v2"
 )
 
+var (
+	// Modify Manually
+	spreadDiff           = 0.3
+	historyLength        = 10
+	energyTypeToMinSlope = map[string]float64{
```
sthaha, line 35: Could you please add a comment on how these are used?

pkg/collector/stats/node_stats.go
```diff
@@ -19,18 +19,175 @@ package stats
...
+	maxUtilization *Coordinate
+	// Min y-intercept
+	minIntercept float64
+	// Min slope
+	minSlope float64
+
+	result *IdleEnergyResult
+}
+
+func (ic *IdleEnergyCalculator) UpdateIdleEnergy(newResutilization float64, newEnergyDelta float64, maxTheoreticalCPUTime float64) {
```
sthaha, line 68:
```suggestion
func (ic *IdleEnergyCalculator) UpdateIdleEnergy(newResUtilization float64, newEnergyDelta float64, maxTheoreticalCPUTime float64) {
```

pkg/collector/stats/node_stats.go
```diff
@@ -19,18 +19,175 @@ package stats
...
+	klog.V(5).Infof("New Datapoint Candidate: (%f, %f)", newResutilization, newEnergyDelta)
+	klog.V(5).Infof("Current Min Utilization Datapoint: (%f, %f)", ic.minUtilization.X, ic.minUtilization.Y)
+	klog.V(5).Infof("Current Max Utilization Datapoint: (%f, %f)", ic.maxUtilization.X, ic.maxUtilization.Y)
+	klog.V(5).Infof("Current History: (%v)", ic.result.history)
+	klog.V(5).Infof("Current Spread: (%v)", ic.result.diff)
+	if newResutilization > ic.minUtilization.X && newResutilization < ic.maxUtilization.X {
+		klog.V(5).Infof("Excess Datapoint: (%f, %f)", newResutilization, newEnergyDelta)
+		// Record History
+		klog.V(5).Infof("Push Idle Energy to history")
+		appendToSliceWithSizeRestriction(&ic.result.history, historyLength, ic.result.calculatedIdleEnergy)
```
sthaha, line 78: could you please explain why this is needed? Would this be removed in the final draft ?

pkg/collector/stats/node_stats.go
```diff
@@ -19,18 +19,175 @@ package stats
...
+	}
+)
+
+type IdleEnergyResult struct {
+	calculatedIdleEnergy float64
+	diff                 float64
+	history              []float64
+}
+
+type IdleEnergyCalculator struct {
```
sthaha, line 55: @KaiyiLiu1234 , could you please add a unit test for this? It will help us understand the behaviour better.

pkg/collector/stats/node_stats.go
KaiyiLiu1234: This tracks consistency. One of the requirements is to make sure the idle power is consistent. So keeping track of the recent history (say the past 10 calculations) and comparing the average of the history with the current idle power can indicate if idle power is consistent (ex. percentage error difference should be less than 0.1). Note, spread and history are tracked and once the idle power passes both, the idle power is freely reported without restrictions (let me know if this should be changed)

pkg/collector/stats/node_stats.go
KaiyiLiu1234: Understood. I can add this unit test in node_stats_test.go

KaiyiLiu1234:
These images show the idle power calculation between kepler latest and kepler dev. kepler dev contains the new way of calculating idle and dynamic power. 
Screenshot From 2025-03-03 19-07-51
Screenshot From 2025-03-03 19-08-26

pkg/collector/stats/node_stats.go
```diff
@@ -19,18 +19,175 @@ package stats
...
+	}
+
+	// log candidates
+	klog.V(5).Infof("Candidate Min Utilization Datapoint: (%f, %f)", newMinUtilizationX, newMinUtilizationY)
+	klog.V(5).Infof("Candidate Max Utilization Datapoint: (%f, %f)", newMaxUtilizationX, newMaxUtilizationY)
+
+	// note minutilization == maxutilization only occurs when we only have one value at the very beginning
+	// in that case, we can rely on the default values provided by NewIdleEnergyCalculator
+	//if ic.minUtilization.X < ic.maxUtilization.X {
+	if newMinUtilizationX < newMaxUtilizationX {
```
sunya-ch, line 124: I think we can skip calculating intercept by checking 

```
diff := math.Abs(newMaxUtilization.X/maxTheoreticalCPUTime-newMinUtilization.X/maxTheoreticalCPUTime )
if diff < spreadDiff {
   klog.V(5).Infof("Sample Size is large enough!")
   return 
}
And then we can use only when ic.result.calculatedIdleEnergy is set to >= ic.minIntercept (which it seems to set to 0).

**Review by sunya-ch — COMMENTED** [review]
The logic to get intercept from two points which spread more than x(50)% relative to the theoretical CPU time makes sense to me. 
Just leave some comment that we can reduce some unused calculation.

**Comment by KaiyiLiu1234** [comment]
Note Unit test fails due to the removal of CalcDynEnergy from stats.go. I believe it's best to in a separate PR remove this functionality and modify follow up unit tests. This means this PR will not fix the bug of idle energy influencing dynamic energy. Additionally, UpdateIdleEnergy can also be divided into multiple functions to satisfy SRP.

**Comment by sthaha** [comment]
@KaiyiLiu1234 Could you also please report the result of the following test

* have your machine run minimal amount of processes (init 3?)
* measure the power .. (min W)
* Increase the load (stress-ng )
* start kepler
* see if the idle estimator estimates **lower** than the `min W`

File: manifests/compose/compose.yaml [hunk 6]
```diff
@@ -42,8 +42,9 @@ services:
         /usr/bin/kepler \
           -address "0.0.0.0:8888" \
           -disable-power-meter=$$DISABLE_POWER_METER \
-          -v "8" \
-          -enable-gpu=false
+          -v "6" \
+          -enable-gpu=false \
+          -expose-estimated-idle-power=true
```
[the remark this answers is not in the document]
KaiyiLiu1234: I did not mean to push the compose files. I will revert them. That said, I don't think it works actually. It's bad that we have two sections where config flags are defined. In the case of the expose estimated idle power, the env variable (pkg/config) gets overriden by cmd/exporter (aka expose-estimated-idle-power=true).

PR #1941, opened 2025-03-07
feat(config): Make /proc and /sys paths configurable

sthaha:
This patch replaces hardcoded /proc and /sys paths just like in other exporters like node-exporter with configurable ones using the config package. E.g., config.ProcDir() and config.SysDir().
These paths can now be set via SYS_DIR and PROC_DIR config variables, defaulting to "/proc" and "/sys" if unspecified.

Mounting host /proc and /sys into containers is discouraged because it compromises security and isolation.

### DIFF
--- a/pkg/libvirt/resolve_vm.go
+++ b/pkg/libvirt/resolve_vm.go
@@ -31,9 +31,8 @@ import (
 )
 
 const (
-	procPath            string = "/proc/%d/cgroup"
-	maxCacheSize        int    = 1000
-	cacheRemoveElements int    = 100
+	maxCacheSize        int = 1000
+	cacheRemoveElements int = 100
 )
 
 var (
@@ -43,8 +42,8 @@ var (
 )
 
 func GetVMID(pid uint64) (string, error) {
-	fileName := fmt.Sprintf(procPath, pid)
-	return getVMID(pid, fileName)
+	filePath := fmt.Sprintf("%s/%d/cgroup", config.ProcDir(), pid)
+	return getVMID(pid, filePath)
 }
 
 func getVMID(pid uint64, fileName string) (string, error) {
--- a/pkg/sensors/components/source/rapl_sysfs.go
+++ b/pkg/sensors/components/source/rapl_sysfs.go
@@ -23,15 +23,14 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/sustainable-computing-io/kepler/pkg/config"
 	"k8s.io/klog/v2"
 )
 
 const (
 	// sysfs path templates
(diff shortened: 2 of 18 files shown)


pkg/libvirt/resolve_vm.go
```diff
@@ -43,6 +42,7 @@ var (
 )
 
 func GetVMID(pid uint64) (string, error) {
+	procPath := config.ProcDir() + "/%d/cgroup"
 	fileName := fmt.Sprintf(procPath, pid)
```
vimalk78, line 46: combine the two statements in to same sprintf ?

sthaha: 💯  ... I will clean this up in a followup patch which addresses a bigger issue.

pkg/sensors/components/source/rapl_sysfs.go
```diff
@@ -51,7 +50,6 @@ var (
 
 func init() {
 	eventPaths = map[string]map[string]string{}
-	detectEventPaths()
```
vimalk78, line 54: removing `detectEventPaths` from init will leave `eventPaths` uninitialized. can we call `detectEventPaths` in `once.Do` inside the functions which reads `detectEventPaths`

pkg/sensors/platform/source/acpi.go
```diff
@@ -109,12 +107,14 @@ func (a *ACPI) StopPower() {
...
 
 func (a *ACPI) GetCPUCoreFrequency() map[int32]uint64 {
+	freqPathDir := config.SysDir() + "/devices/system/cpu/cpufreq/"
 	files, err := os.ReadDir(freqPathDir)
 	cpuCoreFrequency := map[int32]uint64{}
 	if err != nil {
 		klog.Warning(err)
 		return cpuCoreFrequency
 	}
+	freqPath := config.SysDir() + "/devices/system/cpu/cpufreq/policy%d/scaling_cur_freq"
```
vimalk78, line 117: keep the `/devices/system/cpu/cpufreq/policy%d/scaling_cur_freq` as constants and use the constants here.

pkg/sensors/components/source/rapl_sysfs.go
vimalk78: same comment for other places.

pkg/sensors/platform/source/acpi.go
vimalk78: same comment for other places.

pkg/libvirt/resolve_vm.go
sthaha: Done

vimalk78 (approved):

PR #1944, opened 2025-03-10
chore(compose): allow scraping from kepler on host

sthaha:
This change allows running kepler on host and scrape it, making it easier during
development to test changes quickly without building an image.

### DIFF
--- a/manifests/compose/dev/prometheus/scrape-configs/dev.yaml
+++ b/manifests/compose/dev/prometheus/scrape-configs/dev.yaml
@@ -3,6 +3,10 @@ scrape_configs:
     static_configs:
       - targets: [kepler-dev:8888]
 
+  - job_name: host
+    static_configs:
+      - targets: [host.docker.internal:8888]
+
   - job_name: scaphandre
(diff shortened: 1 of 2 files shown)


manifests/compose/dev/prometheus/scrape-configs/dev.yaml
```diff
@@ -3,6 +3,10 @@ scrape_configs:
     static_configs:
       - targets: [kepler-dev:8888]
 
+  - job_name: host
+    static_configs:
+      - targets: [host.docker.internal:8888]
```
vprashar2929, line 8: We should mention that it will require Kepler, I.e running on host to be running on port 8888

vprashar2929 (approved):

manifests/compose/dev/prometheus/scrape-configs/dev.yaml
sthaha: Not sure I agree, imho it should be the same as not mentioning that kepler-dev:8888 should be running on 8888

PR #1961, opened 2025-04-01
feat: add initial skeleton

vprashar2929:
This commit adds the Initial skeleton for Kepler reboot
* Basic main.go file
* Makefile
* Dockerfile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,90 @@
+.PHONY: all build clean test lint vet fmt image push
+
+# Go parameters
+GOCMD=go
+GOBUILD=$(GOCMD) build
+GOCLEAN=$(GOCMD) clean
+GOTEST=$(GOCMD) test
+GOGET=$(GOCMD) get
+GOMOD=$(GOCMD) mod
+GOLINT=golangci-lint
+GOVET=$(GOCMD) vet
+GOFMT=$(GOCMD) fmt
+
+GOOS=$(shell go env GOOS)
+GOARCH=$(shell go env GOARCH)
+
+# Project parameters
+BINARY_NAME=kepler
+BINARY_DIR=bin
+MAIN_GO_PATH=./cmd/kepler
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,90 @@
...
+GOFMT=$(GOCMD) fmt
+
+GOOS=$(shell go env GOOS)
+GOARCH=$(shell go env GOARCH)
+
+# Project parameters
+BINARY_NAME=kepler
+BINARY_DIR=bin
+MAIN_GO_PATH=./cmd/kepler
+VERSION=$(shell git describe --tags --always --dirty || echo "dev")
```
sthaha, line 21: @vprashar2929 we want the version to have a constant prefix, say k2-$(VERSION) ?

Makefile
```diff
@@ -0,0 +1,90 @@
...
+BUILD_TIME=$(shell date -u '+%Y-%m-%d_%H:%M:%S')
+GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
+GIT_COMMIT=$(shell git rev-parse --short HEAD 2>/dev/null)
+
+# Docker parameters
+IMG_BASE ?= quay.io/sustainable_computing_io
+
+KEPLER_IMAGE ?= $(IMG_BASE)/kepler:$(VERSION)
```
sthaha, line 29: should we create a different image instead? of `kepler` ? say `kepler-reboot` ?

vprashar2929: good question

sthaha (approved):

Makefile
vimalk78: better to keep same image and add a tag prefix `kepler:reboot-$(VERSION)`

vimalk78:
didnt we want to implement a 2 LGTM rule?

sthaha:
No, that isn't turned on yet. How about we do that as soon as we are able to publish images? If we can get the CI to publish images, then we have achieved a crucial step.

PR #1963, opened 2025-04-01
chore(main): use run.Group to run multiple services

sthaha:
This commit provide a skeleton to run multiple services in separate go threads and orchestrate clean up if any of these services exit with an error during its startup phase.

NOTE: interrupting (SIGINT) kepler does not result in exit code 0 rather it initiates the shutdown process.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/sustainable-computing-io/kepler
 
-go 1.23.0
+go 1.22
+
+require github.com/oklog/run v1.1.0
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -21,47 +21,73 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"syscall"
-	"time"
+
+	"github.com/oklog/run"
 )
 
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/sustainable-computing-io/kepler
 
-go 1.23.0
+go 1.24.1
```
vimalk78, line 3: downstream does not support more than 1.22, yet.

vimalk78:
can we defer use of this package `github.com/oklog/run` till the point we actually need it for multiple services ?

sthaha:
We know we are going run multliple services like http, and the power-monitor at the very least, so having the stub here makes it easier to write these services without starting another go-routine with (wg. etc) . Those services can be simply written as if running in a single thread.

vimalk78:
> We know we are going run multliple services like http, and the power-monitor at the very least, so having the stub here makes it easier to write these services without starting another go-routine with (wg. etc) . Those services can be simply written as if running in a single thread.

ok

vimalk78 (approved):

PR #1964, opened 2025-04-01
feat(logger): use slog for logging

sthaha:
This PR sets up slog to use logger instead of `fmt.Println` to log all information.

### DIFF
--- a/cmd/kepler/logger.go
+++ b/cmd/kepler/logger.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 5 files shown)


vprashar2929 (approved):

cmd/kepler/logger.go
```diff
@@ -0,0 +1,82 @@
...
+	"path/filepath"
+	"strings"
+)
+
+func setupLogger(level, format string) *slog.Logger {
+	logLevel := parseLogLevel(level)
+	return slog.New(handerForFormat(format, logLevel))
+}
+
+func handerForFormat(format string, logLevel slog.Level) slog.Handler {
```
KaiyiLiu1234, line 32: do you mean handler?

cmd/kepler/logger.go
```diff
@@ -0,0 +1,82 @@
...
+			AddSource: true,
+			ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
+				if a.Key == slog.SourceKey {
+					if src, ok := a.Value.Any().(*slog.Source); ok {
+
+						// Split the path into components
+						parts := strings.Split(filepath.ToSlash(src.File), "/")
+
+						// If we have enough components, take the last 3 -> 2 dirs + filename
+						if len(parts) > 2 {
```
KaiyiLiu1234, line 52: This might be excessive but is it worth adding tests for the logger to test this section of code

vprashar2929 (dismissed):

cmd/kepler/logger.go
sthaha: fixed.

PR #1965, opened 2025-04-01
feat(version): add version information to build

sthaha:
This commits adds version information to kepler. This infomation get
logged in the startup as follows

```
time=2025-04-01T10:21:52.529-04:00 level=INFO source=cmd/kepler/main.go:82 \
msg="Kepler version information"
version=074c9fe5 \
buildTime=2025-04-01_14:21:52 \
gitBranch=reboot \
gitCommit=074c9fe5 \
goVersion=go1.22.12 \
goOS=linux \
goArch=amd64
```
NOTE: output above is broken into multiple lines for better readability.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,13 @@ BUILD_TIME=$(shell date -u '+%Y-%m-%d_%H:%M:%S')
 GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_COMMIT=$(shell git rev-parse --short HEAD 2>/dev/null)
 
+LDFLAGS=-ldflags "\
+	-X github.com/sustainable-computing-io/kepler/internal/version.version=$(VERSION) \
+	-X github.com/sustainable-computing-io/kepler/internal/version.buildTime=$(BUILD_TIME) \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -23,6 +23,13 @@ BUILD_TIME=$(shell date -u '+%Y-%m-%d_%H:%M:%S')
 GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_COMMIT=$(shell git rev-parse --short HEAD 2>/dev/null)
 
+LDFLAGS=-ldflags "\
+	-X github.com/sustainable-computing-io/kepler/internal/version.version=$(VERSION) \
```
vimalk78, line 27: `version` and `gitCommit` have same value. Version can also have `v.X.Y.Z` information

sthaha:
```
VERSION=$(shell git describe --tags --always --dirty || echo "dev")
```
whereas
```
GIT_COMMIT=$(shell git rev-parse --short HEAD 2>/dev/null)
```
they could be same but it will change with release tags.

vprashar2929 (dismissed):

vprashar2929 (approved):

PR #1966, opened 2025-04-01
feat(ci): add workflow to build, push image to quay

vprashar2929:
This commit adds a basic workflow which builds and pushes Kepler image to quay.io registry on push to `reboot` branch

It will push the image to `kepler-reboot` repo under sustainable_computing_io org. The tag will be pointing to current commit SHA

**Note: BOT_NAME and BOT_TOKEN secrets must be defined under repo's settings**

Ref: https://github.com/sustainable-computing-io/kepler/blob/85d017bc23bb6be33adf471414870d874357f532/.github/workflows/push.yml#L16

Signed-off-by: vprashar2929 <vibhu.sharma2929@gmail.com>

### DIFF
--- a/.github/publish-image/action.yaml
+++ b/.github/publish-image/action.yaml
@@ -1,0 +1,51 @@
+name: Build and Publish Images
+description: "Publishes Kepler image to an Image Registry"
+inputs:
+  registry:
+    description: "image registry"
+    required: true
+  username:
+    description: "registry username"
+    required: true
+  password:
+    description: "registry password"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout source
+      uses: actions/checkout@v4
+      with:
+        # NOTE: setting fetch-depth to 0 to retrieve the entire history
+        # intead of a shallow -clone so that all tags are fetched as well.
+        # This is necessary for computing the VERSION using `git describe`
+        fetch-depth: 0
+
+    - uses: actions/setup-go@main
+      with:
+        go-version-file: go.mod
+        check-latest: true
+        cache: true
+
+    - name: Login to Image Registry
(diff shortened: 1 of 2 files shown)


vprashar2929:
Will refine it further in follow up PR

vimalk78 (commented):
can we revert to use image name as kepler rather than kepler-reboot ? and have reboot as part/prefix of the tag?
we dont want to change workflows later.

sthaha:
I would rather that we keep the `kepler-reboot` to indicate that this is totally separate to kepler and make use of variables for `image_name`. This way we don't run the risk of overwriting any of the kepler images even if the `version` (which is the `tag`) is somehow the same as kepler.

sthaha (approved):

KaiyiLiu1234:
lgtm. I am ok with kepler-reboot to differentiate from old kepler, but I think we can select a better name for this new kepler.

sthaha:
> we can select a better name for this new kepler. 

sure this name is emphermral ... , we ultimately want to rename that "kepler-<X>" back to kepler

.github/publish-image/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - uses: actions/setup-go@main
+      with:
+        go-version-file: go.mod
+        check-latest: true
+        cache: true
+
+    - name: Login to Image Registry
+      uses: docker/login-action@v2
+      with:
+        registry: ${{ inputs.registry }}
```
KaiyiLiu1234, line 34: Just wondering, if someone were to accidentally or intentionally print out inputs.password and inputs.username in this workflow, wouldn't it expose important information? For instance, can't I see what secrets.BOT_NAME and secrets.BOT_TOKEN?

PR #1967, opened 2025-04-02
feat(compose): add compose manifests

vprashar2929:
This commit adds the compose manifests for local development. It is responsible for:
* Building kepler-dev from the source
* Spinning up kepler-latest, scaphandre, node-exporter and process-exporter for comparisons
* Setup Grafana and Prometheus and configure them accordingly

Note: The Grafana dashboards are yet to be updated according to kepler-dev


sthaha (approved):

KaiyiLiu1234 (approved):
lgtm as it is the same as it is the same design as in kepler/main. Only thing is it would be important to keep the configuration for reboot separate from old kepler. I assume changes will be made to reboot's configuration fields and how it will accept configuration fields (via configmap, env, or command line).

PR #1968, opened 2025-04-02
feat(ci): add workflow to pr checks

vprashar2929:
This commit adds basic workflow to run PR checks on kepler reboot Currently it performs the following checks:
* Runs `make fmt` to check if the code is formatted correctly
* Runs `make test` to check if the code is unit tested
* Runs `golangci-lint` to check if the code is linted correctly
* Runs `build-image` to check if the image can be built successfully

### DIFF
--- a/.github/publish-image/action.yaml
+++ b/.github/publish-image/action.yaml
@@ -29,11 +29,16 @@ runs:
         cache: true
 
     - name: Login to Image Registry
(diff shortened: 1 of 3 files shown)


sthaha (approved):

KaiyiLiu1234 (approved):
lgtm

PR #1969, opened 2025-04-02
feat(ci): add additional tags when pushing image

vprashar2929:
This commit adds additional tags to the image when pushing to quay.io. Now we will have tags like `latest`, `SHA` and `SHA-timestamp` pushed to quay.io.

### DIFF
--- a/.github/workflows/publish-image.yaml
+++ b/.github/workflows/publish-image.yaml
@@ -17,10 +17,17 @@ jobs:
           go-version-file: go.mod
           cache: true
 
+      - name: additional tags
+        id: additional_tags
+        shell: bash
+        run: |
+         echo "result=$(git rev-parse --short HEAD)-$(date +%Y%m%d%H%M%S),latest" >> $GITHUB_OUTPUT
+
       - name: Build and Publish image to external registry
         uses: ./.github/publish-image
         with:
           registry: quay.io/sustainable_computing_io
           username: ${{ secrets.BOT_NAME }}
           password: ${{ secrets.BOT_TOKEN }}
+          additional_tags: ${{ steps.additional_tags.outputs.result }}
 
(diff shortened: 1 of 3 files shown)


sthaha (approved):

.github/workflows/publish-image.yaml
```diff
@@ -17,10 +17,17 @@ jobs:
...
+        shell: bash
+        run: |
+         echo "result=$(git rev-parse --short HEAD)-$(date +%Y%m%d%H%M%S),latest" >> $GITHUB_OUTPUT
+
       - name: Build and Publish image to external registry
         uses: ./.github/publish-image
         with:
           registry: quay.io/sustainable_computing_io
           username: ${{ secrets.BOT_NAME }}
           password: ${{ secrets.BOT_TOKEN }}
```
KaiyiLiu1234, line 31: I wonder if it is safer to pass this as an env variable or explicitly mask this password in publish-image/action.yaml.

Makefile
```diff
@@ -86,12 +86,48 @@ image: test deps
...
+# E.g. given foo:bar, a,b,c, it will tag foo:bar as foo:a, foo:b, foo:c
+define docker_tag
+@{ \
+	set -eu ;\
+	img="$(1)" ;\
+	tags="$(2)" ;\
+	echo "tagging container image $$img with additional tags: '$$tags'" ;\
+	\
+	img_path=$${img%:*} ;\
+	for tag in $$(echo $$tags | tr -s , ' ' ); do \
```
KaiyiLiu1234, line 110: In bash, it should handle an empty string for tags gracefully, I think it's good idea to include a check to handle whitespaced (and empty) strings because that might cause the loop to still execute and push an image with a tag that is a space.

.github/workflows/publish-image.yaml
sthaha: IIUC, GH masks all `secrets.*`

Makefile
sthaha: Good suggestion. but ROI is low IMHO. 
How about as a follow up PR so that we this functionality can go in?

KaiyiLiu1234 (approved):
lgtm

PR #1970, opened 2025-04-03
feat(config): config and arg parsing using kingpin

sthaha:
This commit adds the feature to load application configuration from a config file using `--config.file=/path/to/config.yaml`. It feature also allows values in the configuration file such as log-level or log-format to be overridden (if needed) using flags such as --log.level and --log.format.


cmd/kepler/main.go
```diff
@@ -105,3 +112,33 @@ func waitForInterrupt(ctx context.Context, signals ...os.Signal) (func() error,
 			cancel()
 		}
 }
+
+func parseArgsAndConfig() (*config.Config, error) {
+	app := kingpin.New("power-monitor", "Power consumption monitoring exporter for Prometheus.")
```
vimalk78, line 117: "kepler" not  "power-monitor"

cmd/kepler/main.go
```diff
@@ -105,3 +112,33 @@ func waitForInterrupt(ctx context.Context, signals ...os.Signal) (func() error,
...
+
+	configFile := app.Flag("config.file", "Path to YAML configuration file").String()
```
vimalk78, line 119: add a default path same as what was for original kepler

cmd/kepler/main.go
```diff
@@ -105,3 +112,33 @@ func waitForInterrupt(ctx context.Context, signals ...os.Signal) (func() error,
...
+	updateConfig := config.RegisterFlags(app)
```
vimalk78, line 120: suggestion: change variable name to `updateConfigFn` to indicate this is a function

cmd/kepler/main.go
sthaha: Default path indicates that the user has provided a file even when they haven't. I think that is a wrong assumption to make.

cmd/kepler/main.go
sthaha: updateConfig is a `verbNoun` and should mean its a func

cmd/kepler/main.go
sthaha: Expanding on the above ... say the invocation is 

```
1. kepler
2. kepler --config.file=/path/to/a/config.file.yaml
3. kepler --config.file=/path/to/default/config.yaml
```

What should the result be if these files are missing?

cmd/kepler/main.go
vimalk78: also the description should be same as for Kepler "Kubernetes Efficient Power Level Exporter"

internal/config/config.go
```diff
@@ -0,0 +1,167 @@
...
+// Validate checks for configuration errors
+func (c *Config) Validate() error {
+	var errs []string
+	{ // log level
+
+		validLogLevels := map[string]bool{
+			"debug": true,
+			"info":  true,
+			"warn":  true,
+			"error": true,
```
vimalk78, line 144: doesn't kingpin already perform the same validation?

internal/config/config.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		// Validate logging settings
+		if _, valid := validLogLevels[c.Log.Level]; !valid {
+			errs = append(errs, "invalid log level")
```
vimalk78, line 149: include invlid log level in error message

cmd/kepler/main.go
vimalk78: operator will be providing a config always, the path can be fixed there.  this is just a suggestion, you can ignore it.

cmd/kepler/main.go
```diff
@@ -22,12 +22,19 @@ import (
...
+	"github.com/alecthomas/kingpin/v2"
 	"github.com/oklog/run"
+	"github.com/sustainable-computing-io/kepler/internal/config"
 	"github.com/sustainable-computing-io/kepler/internal/version"
 )
 
 func main() {
-	logger := setupLogger("info", "text")
+	// parse args and config and exit with error if there is an error
+	cfg, err := parseArgsAndConfig()
```
vimalk78, line 33: depending on the log level, dump the config in logs after parsing.

internal/config/config.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/alecthomas/kingpin/v2"
+	"gopkg.in/yaml.v3"
+)
+
+// Config represents the complete application configuration
+type (
+	Log struct {
+		Level  string `yaml:"level"`
+		Format string `yaml:"format"`
```
vimalk78, line 33: should config default values be supported using go struct tags? or ALL future config default values should be defined in `DefaultConfig()` only?

internal/config/config.go
sthaha: but we also want the config read from yaml to be validated as well.

internal/config/config.go
```diff
@@ -0,0 +1,167 @@
...
+	LogLevelFlag  = "log.level"
+	LogFormatFlag = "log.format"
+)
+
+// DefaultConfig returns a Config with default values
+func DefaultConfig() *Config {
+	cfg := &Config{}
+
+	cfg.Log.Level = "info"
+	cfg.Log.Format = "text"
```
vimalk78, line 52: suggestion: return a struct with default values
```
return &Config{
  Log: LogConfig{
    Level: "info",
    Format: "text",
  }
}
```

IMHO, makes it more clear what the default values are.

cmd/kepler/main.go
sthaha: good idea!

internal/config/config.go
vimalk78: or a config parsed from a default yaml file is also file.

internal/config/config.go
sthaha: > should config default values be supported using go struct tags?

Not sure if that works ? https://pkg.go.dev/gopkg.in/yaml.v3#section-readme 

> ALL future config default values should be defined in DefaultConfig() only?

yes, that's the idea, put all Defaults (including those that may need to be computed) in this function.

internal/config/config.go
sthaha: We can go with your suggestion for now but when the config grows quite long, we will have to revert to the current version for better readability.

vimalk78: i am also asking from better readability point of view.

don't accept a suggestion "for now" , no point in accepting if you think it will change very soon.

internal/config/config.go
vimalk78: > Not sure if that works ? https://pkg.go.dev/gopkg.in/yaml.v3#section-readme

we will need something more than just yaml package.

cmd/kepler/main.go
sthaha: implemented 

```
time=2025-04-03T01:39:20.984-04:00 level=INFO source=cmd/kepler/main.go:91 msg="Kepler version information" version=bb4aeacd-dirty buildTime=2025-04-03_05:39:20 gitBranch=feat-config gitCommit=bb4aeacd goVersion=go1.22.12 goOS=linux goArch=amd64

Configuration
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
log:
    level: info
    format: text

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

internal/config/config.go
sthaha: Resolving this, please reopen to discuss further.

cmd/kepler/main.go
sthaha: Marking it a resolved.

internal/config/config.go
sthaha: implemented now

cmd/kepler/main.go
sthaha: If the logging is text, and the level is info, we print the config in yaml format if possible (green path) or else manually build the configuration string.

cmd/kepler/main.go
vimalk78: @sthaha you want to change the help string  to "Kubernetes Efficient Power Level Exporter" ?

vimalk78 (approved):

PR #1972, opened 2025-04-03
feat(stubs) stubs for all services

sthaha:
This commit introduces the following changes:
* Adds a service interfaces that all services (currently monitor)
  implements
* Refactors application startup to use services and starts them up,
  terminating if any of the service fails to start.
* Adds monitor service (with unit-tests)
* Modifies Device startup to accept a context so that the startup can be
  aborted if context is canceled.

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -14,6 +14,8 @@ import (
 	"github.com/oklog/run"
 	"github.com/sustainable-computing-io/kepler/internal/config"
 	"github.com/sustainable-computing-io/kepler/internal/logger"
+	"github.com/sustainable-computing-io/kepler/internal/monitor"
+	"github.com/sustainable-computing-io/kepler/internal/service"
 	"github.com/sustainable-computing-io/kepler/internal/version"
 )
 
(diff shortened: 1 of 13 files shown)


sthaha:
@vimalk78 @KaiyiLiu1234 @vprashar2929 , I am fixing the code-coverage. Meanwhile, could you please review the change?

cmd/kepler/main.go
```diff
@@ -27,46 +29,29 @@ func main() {
 	logVersionInfo(logger)
 	printConfigInfo(logger, cfg)
 
-	var g run.Group
+	logger.Info("Staring Kepler services")
```
vimalk78, line 32: nit: redundant log?

cmd/kepler/main.go
```diff
@@ -27,46 +29,29 @@ func main() {
...
 			func() error {
-				logger.Info("HTTP server is running. Press Ctrl+C to stop.")
-				<-ctx.Done()
-				return nil
+				return s.Start(ctx)
 			},
 			func(err error) {
-				logger.Info("HTTP Server: Shutting down...:", "error", err)
+				if cleanupErr := s.Stop(); cleanupErr != nil {
+					logger.Warn("service terminated with error: %v", "service", s.Name(), "error", err)
```
vimalk78, line 45: %v for error is redundant

cmd/kepler/main.go
```diff
@@ -27,46 +29,29 @@ func main() {
...
 		g.Add(
 			func() error {
-				logger.Info("HTTP server is running. Press Ctrl+C to stop.")
-				<-ctx.Done()
-				return nil
+				return s.Start(ctx)
 			},
 			func(err error) {
-				logger.Info("HTTP Server: Shutting down...:", "error", err)
+				if cleanupErr := s.Stop(); cleanupErr != nil {
```
vimalk78, line 44: print cleanupErr

cmd/kepler/main.go
```diff
@@ -27,46 +29,29 @@ func main() {
...
+	var g run.Group
+	for _, s := range services {
 		g.Add(
 			func() error {
-				logger.Info("HTTP server is running. Press Ctrl+C to stop.")
-				<-ctx.Done()
-				return nil
+				return s.Start(ctx)
 			},
 			func(err error) {
```
vimalk78, line 43: question: is err unused?

internal/device/energy_test.go
```diff
@@ -69,3 +71,58 @@ func TestEnergy_String(t *testing.T) {
...
+
+func TestPower_Watts(t *testing.T) {
+	tests := []struct {
+		name  string
+		power Power
+		want  float64
+	}{
+		{"Zero", 0, 0.0},
+		{"One", 1_000_000, 1.0},
+		{"1.5 Joule", 1_500_000, 1.5},
```
vimalk78, line 83: Watts

internal/device/energy_test.go
```diff
@@ -69,3 +71,58 @@ func TestEnergy_String(t *testing.T) {
...
+}
+
+func TestPower_MicroWatts(t *testing.T) {
+	tests := []struct {
+		name  string
+		power Power
+		want  float64
+	}{
+		{"Zero", 0, 0.0},
+		{"2 Million MicroJoules", 2_000_000, 2_000_000},
```
vimalk78, line 101: MicroWatts

internal/device/energy.go
```diff
@@ -23,3 +25,20 @@ func (e Energy) MicroJoules() uint64 {
 func (e Energy) String() string {
 	return fmt.Sprintf("%fJ", e.Joules())
 }
+
+// Power represents power usage as an float64 MicroWatts.
+// Use functions Watts and MicroWatts to get the energy value as
```
vimalk78, line 30: get the power value

cmd/kepler/main.go
sthaha: +1, we should log both, run and cleanup err.

internal/service/service.go
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright 2025.
+
```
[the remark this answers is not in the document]
sthaha: @vprashar2929 it would be great if we can get a validator (I was using `reuse`) for this in CI.

cmd/kepler/main.go
sthaha: Also we should log both errors separately.

internal/device/energy.go
```diff
@@ -23,3 +25,30 @@ func (e Energy) MicroJoules() uint64 {
...
+// Power represents power usage as an float64 MicroWatts.
+// Use functions Watts and MicroWatts to get the power value as
+// Watts or MicroWatts
+type Power float64
+
+const (
+	MicroWatt Power = 1.0
+	MilliWatt       = 1000 * MicroWatt
+	Watt            = 1000 * MilliWatt
+)
```
sthaha, line 38: We should do something similar to this for Joules as well in a separate commit. Inspired by time.Duration

vimalk78 (approved):

PR #1974, opened 2025-04-03
chore: improve test coverage of config

sthaha:
This commit adds missing tests for config.String and config.Load functions.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -64,6 +64,20 @@ func TestLoadEmptyFromYAML(t *testing.T) {
 	assert.Equal(t, defaultCfg.Log.Format, cfg.Log.Format)
 }
 
+func TestLoadInvalidConfigFromYAML(t *testing.T) {
+	// Test loading an empty configuration
+	yamlData := `
+log:
+  level: FATAL
+  format: json
+`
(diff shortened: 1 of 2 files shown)


vprashar2929 (approved):

internal/config/config_test.go
```diff
@@ -210,8 +224,54 @@ func TestInvalidConfigurationValues(t *testing.T) {
...
 
+	// test manual string builder approach
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Get string representation
+			str := tc.config.manualString()
+
+			// Verify it's valid YAML and contains the expected values
+			assert.Contains(t, str, "log.level: "+tc.config.Log.Level)
+			assert.Contains(t, str, "log.format: "+tc.config.Log.Format)
```
vimalk78, line 274: question: is it a deliberate choice to have "level:" in case of yaml marshalling and "log.level" in case of manual ?

vimalk78: lgtm otherwise

sthaha: yes, that was intentional. It tells us the config failed to Marshall properly (for whatever reason).

PR #1977, opened 2025-04-03
chore(makefile): omit main from test coverage

vprashar2929:
This commit omits the `cmd/` folder from the test coverage.


sthaha (approved):
We need to move the existing tests from cmd to internal, with this change.

vprashar2929:
> We need to move the existing tests from cmd to internal, with this change.

yeah will move them

PR #1985, opened 2025-04-10
feat(config): make config module public

KaiyiLiu1234:
Moved config module from internal to pkg as it should be a reusable public module. Immediate usecase is making use of config module in the kepler-operator.


SamYuan1990:
emmmm？should we merge it to main branch?

KaiyiLiu1234:
@SamYuan1990 no. this is for the reboot branch. Specifically, the config module is made public so that the operator can reuse it. Its separate from the main branch.

cmd/kepler/main.go
```diff
@@ -12,11 +12,11 @@ import (
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/oklog/run"
-	"github.com/sustainable-computing-io/kepler/internal/config"
 	"github.com/sustainable-computing-io/kepler/internal/logger"
 	"github.com/sustainable-computing-io/kepler/internal/monitor"
 	"github.com/sustainable-computing-io/kepler/internal/service"
 	"github.com/sustainable-computing-io/kepler/internal/version"
+	"github.com/sustainable-computing-io/kepler/pkg/config"
```
sthaha, line 19: @KaiyiLiu1234 lets put this under `kepler` itself and not `pkg`

vprashar2929:
@KaiyiLiu1234 I cannot see this commit in the [reboot](https://github.com/sustainable-computing-io/kepler/commits/reboot/) branch? Looks like you pushed to [public-config](https://github.com/sustainable-computing-io/kepler/tree/public-config)

PR #1986, opened 2025-04-10
feat: add Prometheus exporter and API server to Kepler

sthaha:
This commit introduces a  HTTP API server and stub for Prometheus exporter. Changes include:

- Prometheus Exporter: package that implements a Prometheus exporter (`Exporter`) to export power data. It integrates with the existing power monitor and registers metrics endpoints via the API server. Supports debug collectors (e.g., Go runtime metrics) configurable via options.

- API Server: HTTP `APIServer` implementation, providing a configurable HTTP server for registering endpoints e.g., `/metrics` from Prometheus Exporter. Supports multiple listen addresses and graceful shutdown.
- Updated `main.go` to instantiate and run the new Prometheus exporter and API server

** New Dependencies **
- Added Prometheus-related dependencies


internal/exporter/prometheus/prometheus.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		e.logger.Info("Enabling debug collector", "collector", c)
+		e.registry.MustRegister(collector)
+	}
+
+	err := e.server.Register("/metrics", "Metrics", "Prometheus metrics",
+		promhttp.HandlerFor(
+			e.registry,
+			promhttp.HandlerOpts{
+				EnableOpenMetrics: true,
```
vimalk78, line 120: hard coded value. should this be controlled via options?

internal/exporter/prometheus/prometheus.go
```diff
@@ -0,0 +1,145 @@
...
+		server:          s,
+		logger:          opts.logger.With("service", "prometheus"),
+		debugCollectors: opts.debugCollectors,
+		registry:        prom.NewRegistry(),
+	}
+
+	return monitor
+}
+
+func collectorForName(name string) (prom.Collector, error) {
```
vimalk78, line 91: add `collectors.NewBuildInfoCollector` also?

internal/server/server.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (s *APIServer) Name() string {
+	return "api-server"
+}
+
+func (s *APIServer) Start(ctx context.Context) error {
+	s.logger.Info("Starting HTTP server", "listening-on", s.listenAddrs)
+	if len(s.listenAddrs) == 0 {
+		return fmt.Errorf("no listening address provided")
+	}
```
vimalk78, line 96: check this before above log?

internal/server/server.go
```diff
@@ -0,0 +1,135 @@
...
+// APIServer implements APIServer
+type APIServer struct {
+	// input
+	logger      *slog.Logger
+	listenAddrs []string
+
+	// http
+	server              *http.Server
+	mux                 *http.ServeMux
+	endpointDescription string
```
vimalk78, line 32: question: is this only used for testing?

internal/exporter/prometheus/prometheus.go
```diff
@@ -0,0 +1,145 @@
...
+	err := e.server.Register("/metrics", "Metrics", "Prometheus metrics",
+		promhttp.HandlerFor(
+			e.registry,
+			promhttp.HandlerOpts{
+				EnableOpenMetrics: true,
+				Registry:          e.registry,
+			},
+		))
+	if err != nil {
+		return err
```
vimalk78, line 125: log the error

internal/server/server.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	errCh := make(chan error)
+	go func() {
+		webCfg := &web.FlagConfig{
+			WebListenAddresses: &s.listenAddrs,
+			WebConfigFile:      pointer(""), // TODO: support passing config file
+		}
+		errCh <- web.ListenAndServe(s.server, webCfg, s.logger)
+	}()
```
vimalk78, line 105: suggestion: service `Start()` already runs in a separate go-routine, here we are creating another. if possible please simplify.

internal/exporter/prometheus/prometheus.go
sthaha: yes, but I wanted that in separate patch.

internal/server/server.go
sthaha: no, actually, I removed the part that builds the landing page. I am adding that now.

internal/server/server.go
sthaha: I feel it help to log in all cases.

internal/server/server.go
sthaha: there are 2 things
1. watch for context (supplied from main) to be cancelled and quit
2. or end when ListenAndServe() exits

both are blocking and thus we need a go routine for the server

vimalk78 (dismissed):

internal/exporter/prometheus/prometheus.go
sthaha: All startup errors are automatically logged in main with logger.Error. I think a good rule of thumb (as in prometheus) is that errors get returned **or** those that do not are logged.

vimalk78 (approved):

PR #1988, opened 2025-04-11
ci: label PR based on commit messages

vprashar2929:
This commit adds a labeler job to PR workflow to label PRs based on commit messages. Following labels are added:
- feat
- fix
- chore
- docs
- ci
- refactor
- perf
- test
- revert

Partially addresses #1989 

Ref: https://github.com/grafana/pr-labeler-action?tab=readme-ov-file#default-config
Relevant labels are already added to project: https://github.com/sustainable-computing-io/kepler/labels

### DIFF
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,0 +1,12 @@
+include-title: false
+include-commits: true
+label-mapping:
+  fix: [fix, bugfix]
+  feat: [feat, feature]
+  chore: [chore]
+  docs: [docs, documentation]
+  test: [test]
+  refactor: [refactor]
+  ci: [ci]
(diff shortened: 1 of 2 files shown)


vprashar2929:
Not enabling label-pr job as it requires config to be present on the `reboot` branch in order to run. Ref: https://github.com/sustainable-computing-io/kepler/actions/runs/14396700125/job/40373570981?pr=1988

After this PR merge I will uncomment the job

sthaha:
Can we make a 0.0.2 release after this?

vprashar2929:
I think we have to add the same to the `main` branch as well since grafana/pr-labeler-action expects the config to be available in the default branch

PR #1995, opened 2025-04-13
feat(config): publicize configuration object

KaiyiLiu1234:
For external projects like kepler-operator, the config object should be made into a public project so it can be reused.


sthaha (changes requested):
@KaiyiLiu1234 , could you please move config to root of the project instead of under `pkg/` ?

sthaha:
@KaiyiLiu1234 I am closing this since #1997 fixes this.

PR #2000, opened 2025-04-14
feat(cpu_info): Added cpu info collector

vimalk78:
Added a collector for cpu info. It depends on outout generated by lscpu command.
  lscpu is an architecture aware utility which gather cpu information
  from sysfs and /proc/cpuinfo

### DIFF
--- a/internal/exporter/prometheus/collectors/cpu_info.go
+++ b/internal/exporter/prometheus/collectors/cpu_info.go
@@ -1,0 +1,193 @@
+// SPDX-FileCopyrightText: 2025 The Kepler Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package collectors
+
+import (
+	"bufio"
+	"os/exec"
+	"strings"
+	"sync"
+
+	prom "github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	architecture     = "architecture"
+	model_name       = "model_name"
+	cpus             = "cpus"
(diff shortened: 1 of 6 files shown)


internal/exporter/prometheus/collectors/cpu_info.go
```diff
@@ -0,0 +1,193 @@
...
+		ci.cache[model_name],
+		ci.cache[cpus],
+		ci.cache[cores_per_socket],
+		ci.cache[sockets],
+		ci.cache[vendor_id],
+	)
+}
+
+func (ci *cpuInfo) getCPUInfo() (map[string]string, error) {
+	kv, err := ci.runLscpu()
```
sthaha, line 107: suggestion: wondering if https://github.com/prometheus/procfs/blob/master/cpuinfo.go is simpler?

vimalk78: this implementation is also not inventing anything new, uses a utility which is quite quite standard and architecture independent.

vimalk78: let me know if this is not acceptable. i will close the PR and create a new based on your suggestion.

sthaha: happy to merge this but I do think we should consider not having a dependency (also consider security implications) especially because a package we depend on already provides this without any additional dependency.

vimalk78:
closing

PR #2011, opened 2025-04-16
chore(device): add fake CPU power meter for testing

sthaha:
Adds a new `FakeCPUMeter` implementation in the `device` package to simulate CPU power metrics for testing purposes. The fake meter is not intended for production use and is for development / testing use only

Changes include:

- New with `fakeEnergyZone` and `fakeRaplMeter` types  that implement the `EnergyZone` and `CPUPowerMeter` interfaces.
- Configuration support for enabling the fake meter via `dev.fake-cpu-meter.enabled` in `config.yaml` ONLY and not through flags
- Refactored `main.go` to use a new `createCPUMeter` function that conditionally creates either a real or fake CPU power meter based on the configuration.
- Updated `config.go` to include `Dev` struct for development settings
- Added default configuration in `hack/config.yaml` with `fake-cpu-meter` disabled.

### DIFF
--- a/internal/device/fake_cpu_power_meter.go
+++ b/internal/device/fake_cpu_power_meter.go
@@ -1,0 +1,174 @@
+// SPDX-FileCopyrightText: 2025 The Kepler Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package device
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"math/rand"
+	"path/filepath"
+	"sync"
+)
+
+// NOTE: This fake meter is not intended to be used in production and is for testing only
+
+type Zone = string
+
+const (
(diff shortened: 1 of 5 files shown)


internal/device/fake_cpu_power_meter.go
```diff
@@ -0,0 +1,174 @@
...
+	return func(m *fakeRaplMeter) {
+		for _, z := range m.zones {
+			if fz, ok := z.(*fakeEnergyZone); ok {
+				fz.maxEnergy = e
+			}
+		}
+	}
+}
+
+// WithFakeMaxEnergy sets the maximum energy value before wrap-around
```
vprashar2929, line 112: minor: update comment

vprashar2929 (approved):

PR #2016, opened 2025-04-17
feat(monitor): improve power monitor with dynamic collection based on interval

sthaha:
This refactor enhances the monitor as follows:

- Replace mutex-based synchronization with atomic operations for thread safety
- Implement dynamic collection scheduling with configurable intervals
- Add option to disable automatic collection (interval=0) for on-demand only mode
- Pass previous snapshot explicitly to calculateNodePower
- Simplify zone name handling by removing unnecessary synchronization
- Use a context for clean cancellation of collection goroutines
- Ensure freshness checks always return recent data

### DIFF
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -39,6 +39,7 @@ type PowerMonitor struct {
 	logger *slog.Logger
 	cpu    device.CPUPowerMeter
 
+	interval     time.Duration
 	clock        clock.WithTicker
 	maxStaleness time.Duration
 
@@ -49,6 +50,10 @@ type PowerMonitor struct {
 	snapshot     atomic.Pointer[Snapshot]
 
 	zonesNames []string // cache of all zones
+
+	// For managing the collection loop
+	collectionCtx    context.Context
+	collectionCancel context.CancelFunc
 }
 
 var _ Service = (*PowerMonitor)(nil)
@@ -60,12 +65,17 @@ func NewPowerMonitor(meter device.CPUPowerMeter, applyOpts ...OptionFn) *PowerMo
 		apply(&opts)
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	monitor := &PowerMonitor{
-		logger:       opts.logger.With("service", "monitor"),
-		cpu:          meter,
-		clock:        opts.clock,
-		dataCh:       make(chan struct{}, 1),
-		maxStaleness: opts.maxStaleness,
+		logger:           opts.logger.With("service", "monitor"),
+		cpu:              meter,
+		clock:            opts.clock,
+		interval:         opts.interval,
+		dataCh:           make(chan struct{}, 1),
+		maxStaleness:     opts.maxStaleness,
+		collectionCtx:    ctx,
+		collectionCancel: cancel,
 	}
 
 	return monitor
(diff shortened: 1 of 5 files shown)


internal/monitor/monitor.go
```diff
@@ -144,6 +156,36 @@ func (pm *PowerMonitor) initZones() error {
...
+func (pm *PowerMonitor) scheduleNextCollection() {
+	timer := pm.clock.After(pm.interval)
+
+	go func() {
+		select {
+		case <-timer:
+			if err := pm.collectData(); err != nil {
+				pm.logger.Error("Failed to collect power data", "error", err)
+			}
+			pm.scheduleNextCollection()
```
vimalk78, line 180: call `collectionLoop()` here?

internal/monitor/monitor.go
vimalk78, line 1: no tests for `collectionLoop` and `scheduleNextCollection`

internal/monitor/monitor.go
sthaha: I find this "easier" to reason ... 
* `collectionLoop` starts the collection and then `schedules Next` if there is an interval otherwise it exits 
 
*  scheduleNext() schedules the next collection (recursive) and aborts if canceled

vimalk78:
please rebase to `reboot`

internal/monitor/monitor.go
```diff
@@ -88,17 +98,27 @@ func (pm *PowerMonitor) Init() error {
 func (pm *PowerMonitor) signalNewData() {
 	select {
 	case pm.dataCh <- struct{}{}: // send signal to any waiting goroutinel
+		pm.logger.Debug("Data channel updated")
 	default:
+		pm.logger.Debug("Data channel is full, dropping signal")
```
vimalk78, line 103: nit: no need to say `dropping signal` , seems something negative has happened.

internal/monitor/monitor.go
```diff
@@ -88,17 +98,27 @@ func (pm *PowerMonitor) Init() error {
...
 	}
 }
 
 func (pm *PowerMonitor) Run(ctx context.Context) error {
 	pm.logger.Info("Monitor is running...")
+	go pm.collectionLoop() // NOTE: runs in background
```
vimalk78, line 109: there is no need to run this in separate go routine, its lifetime is short and is not a loop.

internal/monitor/monitor.go
```diff
@@ -135,18 +155,48 @@ func (pm *PowerMonitor) initZones() error {
...
+		pm.scheduleNextCollection()
+	}
+}
+
+// scheduleNextCollection schedules the next data collection
+func (pm *PowerMonitor) scheduleNextCollection() {
+	timer := pm.clock.After(pm.interval)
+	go func() {
+		select {
+		case <-timer:
```
vimalk78, line 174: we can get the current time from this channel, can avoid calling `Now()` in refresh data

internal/monitor/monitor.go
sthaha: its lifetime will be long as soon as we add the process and containers and pods - right?

internal/monitor/monitor.go
sthaha: Now() should refer to run and not when the collection started, so caching/ reusing the current time isn't IMHO the right approach

internal/monitor/monitor.go
vimalk78: this function does first collection, schedules timer, and returns.

vimalk78 (approved):

PR #2018, opened 2025-04-17
fix(collector): prevent race conditions in PowerCollector with ready flag and tests

sthaha:
This PR checks thread-safety of Power Collector and thereby also the Monitor indirectly using `fakeCPUMeter`. The PowerCollector was prone to race conditions when Describe and Collect were called concurrently (without any waits in between) before metric descriptors were initialized, potentially causing panics or incomplete metrics.
   
Additionally, the PowerMonitor's ZoneNames method used an unnecessary mutex, impacting performance.

This commit also introduces a `ready` flag to ensure Describe and Collect only execute after descriptor initialization, protected by a mutex. 

The PowerMonitor's zones mutex is removed, as zones are immutable post-initialization.

The test suite (`power_collector_concurrency_test.go`) validates thread safety under concurrent Describe, Collect, registration, and rapid calls.

### DIFF
--- a/internal/exporter/prometheus/collector/power_collector.go
+++ b/internal/exporter/prometheus/collector/power_collector.go
@@ -27,6 +27,7 @@ type PowerCollector struct {
 	mutex sync.RWMutex
 
 	// Node power metrics
+	ready                 bool
 	nodeJoulesDescriptors map[string]*prometheus.Desc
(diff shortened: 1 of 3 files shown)


KaiyiLiu1234 (approved):
I do not see any issues with power_collector.go. Tests are thorough enough as well.

PR #2021, opened 2025-04-19
refactor: service lifecycle into optional Init Run Shutdown

sthaha:
This commit modifies the start up behavior. It refactors the service lifecycle management, replacing the Start and Stop methods with
* `Init`: used to initiliaze a service and can assume it isn't called from go routines and must be called before `Run`. 
* `Run`: Runs the service and blocks until context is cancelled or an error is encountered
* `Shutdown`:  method is the cleanup and indicates that you can't call `Run` after `Shutdown`.

This change aims to separate initialization which is run serially and allows services to initialize without worrying about thread-safety. I.E. Its okay for an Init() to call another Init().

Thus, this change separates initialization, runtime, and cleanup phases for better clarity and control. All Methods are optional and main calls the methods if they are defined. 

Tests are updated keep up with the change.

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -28,41 +28,24 @@ func main() {
 	if err != nil {
 		os.Exit(1)
 	}
+
 	logger := logger.New(cfg.Log.Level, cfg.Log.Format)
 	logVersionInfo(logger)
 	printConfigInfo(logger, cfg)
 
-	// create & register all services with run group
 	services, err := createServices(logger, cfg)
 	if err != nil {
 		logger.Error("failed to create services", "error", err)
 		os.Exit(1)
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	var g run.Group
-	for _, s := range services {
-		g.Add(
-			func() error {
-				return s.Start(ctx)
-			},
-			func(err error) {
-				if err != nil {
-					logger.Warn("service terminated with error", "service", s.Name(), "error", err)
-				}
-
-				if cleanupErr := s.Stop(); cleanupErr != nil {
-					logger.Warn("service cleanup failed with error", "service", s.Name(), "error", cleanupErr)
-				}
-				cancel()
-			},
-		)
+	if err = initServices(logger, services); err != nil {
+		logger.Error("failed to initialize services", "error", err)
+		os.Exit(1)
 	}
-	g.Add(waitForInterrupt(ctx, logger, os.Interrupt))
 
-	// run all groups
 	logger.Info("Starting Kepler")
-	if err := g.Run(); err != nil {
+	if err := runServices(logger, services); err != nil {
 		logger.Error("Kepler terminated with an error", "error", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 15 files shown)


cmd/kepler/main.go
```diff
@@ -82,8 +65,47 @@ func logVersionInfo(logger *slog.Logger) {
...
+				}
+
+				logger.Warn("shutting down", "service", s.Name())
+				if shutdownErr := srv.Shutdown(); shutdownErr != nil {
+					logger.Warn("service shutdown failed with error", "service", s.Name(), "error", shutdownErr)
+				}
+			},
+		)
+	}
+	g.Add(waitForInterrupt(ctx, cancel, logger, os.Interrupt))
```
vimalk78, line 103: add handling for `SIGTERM` . this is used in kubernetes to gracefully shutdown a pod.

internal/device/fake_cpu_power_meter_test.go
```diff
@@ -43,9 +44,18 @@ func TestFakeRaplMeter_Name(t *testing.T) {
...
+	go func() {
+		errCh <- meter.Run(ctx)
+	}()
+
+	err := <-errCh
+	duration := time.Since(start)
+	assert.NoError(t, err, "Run() should not return an error")
+	assert.GreaterOrEqual(t, duration, 50*time.Millisecond, "Run() should run until the context is cancelled")
 
 	err = meter.Stop()
```
vimalk78, line 60: `Stop`  can be removed since fake do need to implement service.Shutdown

internal/device/power_meter.go
```diff
@@ -11,8 +11,12 @@ type powerMeter interface {
...
-	Start(ctx context.Context) error
+	// Init() initializes the power meter and makes it ready for use. This method
+	// is not required to be thread-safe
+	Init(ctx context.Context) error
+
+	// Run() power meter for reading energy or power
+	Run(ctx context.Context) error
 
 	// Stop() stops the power meter and releases any resources held
 	Stop() error
```
vimalk78, line 22: change to `Shutdown`?

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -67,6 +67,11 @@ func (r *raplPowerMeter) Start(ctx context.Context) error {
 	return err
 }
 
+func (r *raplPowerMeter) Run(ctx context.Context) error {
+	<-ctx.Done()
+	return nil
+}
+
```
vimalk78, line 74: since real and fake power meter do not actually `Run` and `Stop/Shutdown` , no need to implement `Runner` and `Shutdown. the `runServices` checks for interface implementation anyway.

internal/exporter/prometheus/prometheus.go
vimalk78, line 1: prometheus service does not implement `Shutdown`. it can unregister the prometheus handler from server

vimalk78: perometheus service does not implement `Runner` 👍🏽

internal/service/service.go
```diff
@@ -9,8 +9,24 @@ import "context"
...
+
+// Runner is the interface that all services must implement that needs to run in background
+type Runner interface {
+	Service
+	// Run runs the service and is expected to block and be thread safe
+	Run(ctx context.Context) error
+}
+
+// Shutdown is the interface that all services must implement that are to be shutdown / cleaned up
+type Shutdown interface {
```
vimalk78, line 28: change name to `Shutdowner`

cmd/kepler/main.go
```diff
@@ -82,8 +65,47 @@ func logVersionInfo(logger *slog.Logger) {
 	)
 }
 
-func waitForInterrupt(ctx context.Context, logger *slog.Logger, signals ...os.Signal) (func() error, func(error)) {
-	ctxInternal, cancel := context.WithCancel(ctx)
+func runServices(logger *slog.Logger, services []service.Service) error {
+	logger.Info("Running all services")
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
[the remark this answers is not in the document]
sthaha: cancel gets called in `waitForInterrupt` and is also deferred.

internal/device/fake_cpu_power_meter_test.go
sthaha: Fake implements CPUPowerMeter and isn't a service so to satisfy that interface, we need a Stop method.

internal/device/power_meter.go
sthaha: The idea behind shutdown was to indicate that you can't `start` after `shutdown` where as for devices, I am not sure (at this point in time) if those should be treated as a Service. Any thoughts?

internal/device/rapl_sysfs_power_meter.go
sthaha: PowerMeters aren't services. It implements PowerMeter so these are specific to that interface.

internal/exporter/prometheus/prometheus.go
sthaha: +1 to unregister idea but with shutdown being called, there is no point in unregister since all services are being shutdown -right?

cmd/kepler/main.go
```diff
@@ -145,6 +165,43 @@ Configuration
 `, cfg)
 }
 
+func initServices(logger *slog.Logger, services []service.Service) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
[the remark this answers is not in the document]
sthaha: I agree, we can add context to `Init` if there is a need. 
Can I address that in a separate patch so that the rest of the PRs can be merged?

internal/device/power_meter.go
```diff
@@ -11,8 +11,12 @@ type powerMeter interface {
...
 	Name() string
 
-	// Start() initialuzes and starts the power meter for reading energy or power
-	Start(ctx context.Context) error
+	// Init() initializes the power meter and makes it ready for use. This method
+	// is not required to be thread-safe
+	Init(ctx context.Context) error
+
+	// Run() power meter for reading energy or power
+	Run(ctx context.Context) error
```
sthaha, line 19: Lets make use of service.Runner and service.Initializer to replace the Init in CPUPowerMeter. 
Power Meter will only return Zones()

vimalk78 (approved):
/lgtm
with few changes to be done in a followup PR.

PR #2022, opened 2025-04-19
refactor(monitor): replace RWMutex with atomic.Pointer for snapshot management

sthaha:
This commit replaces sync.RWMutex with atomic.Pointer for thread-safe snapshot
access in PowerMonitor.

* Move zone initialization to initZones method and defer snapshot creation.
* Update calculateNodePower to accept previous node data, decoupling it from snapshot locking.
* update tests to reflect new logic.

### DIFF
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -7,7 +7,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/sustainable-computing-io/kepler/internal/device"
@@ -39,18 +39,16 @@ type PowerMonitor struct {
 	logger *slog.Logger
 	cpu    device.CPUPowerMeter
 
-	clock clock.WithTicker
+	clock        clock.WithTicker
+	maxStaleness time.Duration
 
 	// signals when a snapshot has been updated
 	dataCh chan struct{}
 
 	computeGroup singleflight.Group
-	maxStaleness time.Duration
+	snapshot     atomic.Pointer[Snapshot]
 
-	zones []string // cache of all zones read
-
-	snapshotMu sync.RWMutex
-	snapshot   *Snapshot
+	zonesNames []string // cache of all zones
 }
 
 var _ Service = (*PowerMonitor)(nil)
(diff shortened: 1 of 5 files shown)


internal/monitor/monitor.go
```diff
@@ -191,21 +194,36 @@ func (pm *PowerMonitor) ensureFreshData() error {
...
 // computePower create a new snapshot of the power consumption of various levels( currently only node)
 func (pm *PowerMonitor) computePower() error {
 	started := time.Now()
 	defer func() { pm.logger.Info("Computed power", "duration", time.Since(started)) }()
 
+	prevSnapshot := pm.snapshot.Load()
+	// ensure snapshot is not nil from here on
+	if prevSnapshot == nil {
+		prevSnapshot = NewSnapshot()
+	}
```
vimalk78, line 216: cant we simplify it to
```
if prevSnapshot == nil {
  	snapshot := NewSnapshot()
      	snapshot.Timestamp = time.Now()
	pm.snapshot.Store(snapshot)
        return err
}
```

otherwise both prevSnapshot and newSnapshot are same

internal/monitor/node.go
```diff
@@ -7,17 +7,14 @@ import (
...
 
-func (pm *PowerMonitor) calculateNodePower(snapshot *Snapshot) error {
-	// Get previous measurements for calculating watts
-	pm.snapshotMu.RLock()
-	prevReadTime := pm.snapshot.Node.Timestamp
-	prevZones := pm.snapshot.Node.Zones
-	pm.snapshotMu.RUnlock()
-
-	if prevReadTime.IsZero() {
+func (pm *PowerMonitor) calculateNodePower(snapshot *Snapshot, prevNode *Node) error {
```
vimalk78, line 10: shouldnt `computeNodePower` have prev and current *Node as parameters?

internal/monitor/node.go
```diff
@@ -7,17 +7,14 @@ import (
...
+	if prevNode == nil || prevNode.Timestamp.IsZero() {
 		// No previous data, nothing to do
 		return pm.firstNodeRead(snapshot)
```
vimalk78, line 13: this can still trigger `pm.signalNewData()`

sthaha: yes, it can. Something wrong with that?

internal/monitor/monitor.go
sthaha: sorry, I am not getting it :( 
When monitor starts, we don't allocate any memory for the snapshot. In the `computePower` we ensure that previousSnapshot which is going to be referred from now on has a blank (0 ed out) snapshot. We specifically don't want to set the timestamp of the previous snapshot and it must be 0. This information is useful to compute the first - collection.

vimalk78: ref: https://github.com/sustainable-computing-io/kepler/issues/2025

vimalk78 (approved):

PR #2027, opened 2025-04-23
feat(energy_zone): Added Energy Zone Collector

vimalk78:
- Added Energy Zone Collector
- Added Tests

```
❯ curl localhost:28282/metrics | grep _rapl_zone
# HELP kepler_node_rapl_zone Rapl Zones from sysfs
# TYPE kepler_node_rapl_zone gauge
kepler_node_rapl_zone{index="0",name="core",path="/sys/class/powercap/intel-rapl:0:0"} 1
kepler_node_rapl_zone{index="0",name="package",path="/sys/class/powercap/intel-rapl-mmio:0"} 1
kepler_node_rapl_zone{index="0",name="package",path="/sys/class/powercap/intel-rapl:0"} 1
kepler_node_rapl_zone{index="0",name="psys",path="/sys/class/powercap/intel-rapl:1"} 1
kepler_node_rapl_zone{index="0",name="uncore",path="/sys/class/powercap/intel-rapl:0:1"} 1
```

### DIFF
--- a/internal/exporter/prometheus/collector/cpuinfo.go
+++ b/internal/exporter/prometheus/collector/cpuinfo.go
@@ -8,29 +8,7 @@ import (
 	"sync"
 
 	prom "github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/procfs"
 )
-
-// procFS is an interface for CPUInfo.
-type procFS interface {
-	CPUInfo() ([]procfs.CPUInfo, error)
-}
-
-type realProcFS struct {
-	fs procfs.FS
-}
-
-func (r *realProcFS) CPUInfo() ([]procfs.CPUInfo, error) {
-	return r.fs.CPUInfo()
-}
-
-func newProcFS(mountPoint string) (procFS, error) {
-	fs, err := procfs.NewFS(mountPoint)
-	if err != nil {
-		return nil, err
-	}
-	return &realProcFS{fs: fs}, nil
-}
 
 // cpuInfoCollector collects CPU info metrics from procfs.
 type cpuInfoCollector struct {
--- a/internal/exporter/prometheus/collector/procfs.go
+++ b/internal/exporter/prometheus/collector/procfs.go
@@ -1,0 +1,26 @@
+package collector
+
+import (
+	"github.com/prometheus/procfs"
+)
+
+// procFS is an interface to prometheus/procfs
+type procFS interface {
+	CPUInfo() ([]procfs.CPUInfo, error)
+}
+
+type realProcFS struct {
+	fs procfs.FS
+}
+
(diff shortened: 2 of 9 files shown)


internal/exporter/prometheus/collector/procfs.go
```diff
@@ -0,0 +1,26 @@
+package collector
+
+import (
+	"github.com/prometheus/procfs"
+)
+
+// procFS is an interface to prometheus/procfs
+type procFS interface {
```
sthaha, line 8: How about cpuInfoProvider or something more specific to the cpuInfo?

internal/exporter/prometheus/collector/sysfs.go
```diff
@@ -0,0 +1,24 @@
...
+// sysFS is an interface to prometheus/procfs/sysfs
+type sysFS interface {
+	Zones() ([]sysfs.RaplZone, error)
+}
+
+type realSysFS struct {
+	sysfs sysfs.FS
+}
+
+func (s *realSysFS) Zones() ([]sysfs.RaplZone, error) {
```
sthaha, line 14: we should reuse cpuPowerMeter here since the raplReader removes non-standard rapl zones ( a functionality that we want to preserve here as well). Also when we implement filtering of zones, we should only show the zones that are filtered. 

We could instead have an interface that raplPowerMeter satisfies which returns only the EnergyZones() which provides all information.

internal/exporter/prometheus/collector/cpuinfo.go
```diff
@@ -8,30 +8,8 @@ import (
 	"sync"
 
 	prom "github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/procfs"
 )
 
-// procFS is an interface for CPUInfo.
```
sthaha, line 14: Can this refactor /reorg be a separate commit?

sthaha (changes requested):
Requesting change to reuse existing power-meter that provides the zone information. We can add an `Init` which computes the zones information and cache it.

```
kepler_node_rapl_zone{index="0",name="package",path="/sys/class/powercap/intel-rapl-mmio:0"} 1
```
we shouldn't expose this.

internal/exporter/prometheus/collector/sysfs.go
vimalk78: then we can have this metric as part of power collecor itself.

internal/exporter/prometheus/collector/cpuinfo.go
vimalk78: perhaps if this refactor was done, comment would ask for a refactor

internal/exporter/prometheus/collector/sysfs.go
sthaha: Indeed, didn't think of that :)

PR #2029, opened 2025-04-24
feat(pprof): Added pprof

vimalk78:
Added pprof

image

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -33,18 +33,22 @@ type (
 	}
 
 	Config struct {
-		Log  Log  `yaml:"log"`
-		Host Host `yaml:"host"`
-		Dev  Dev  `yaml:"dev"` // WARN: do not expose dev settings as flags
+		Log         Log  `yaml:"log"`
+		Host        Host `yaml:"host"`
+		Dev         Dev  `yaml:"dev"` // WARN: do not expose dev settings as flags
+		EnablePprof bool `yaml:"enable-pprof"`
 	}
 )
 
 const (
 	// Flags
-	LogLevelFlag   = "log.level"
-	LogFormatFlag  = "log.format"
+	LogLevelFlag  = "log.level"
+	LogFormatFlag = "log.format"
+
 	HostSysFSFlag  = "host.sysfs"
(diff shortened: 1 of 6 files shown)


internal/server/server.go
```diff
@@ -159,6 +159,7 @@ func (s *APIServer) Shutdown() error {
 }
 
 func (s *APIServer) Register(endpoint, summary, description string, handler http.Handler) error {
+	s.logger.Debug("Endpoint Registered", "endppoint", endpoint)
```
sthaha, line 162: nit: typo 
```suggestion
	s.logger.Debug("Endpoint Registered", "endpoint", endpoint)
```

internal/server/pprof.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/sustainable-computing-io/kepler/internal/service"
+)
+
+type pp struct {
+	api APIService
+}
+
+var _ service.Service = &pp{}
+var _ service.Initializer = &pp{}
```
sthaha, line 19: minor: we don't need an object to be created for this assertion 

```
var _ service.Service = (*pp)(nil)
```

config/config.go
```diff
@@ -33,19 +33,23 @@ type (
 	}
 
 	Config struct {
-		Log  Log  `yaml:"log"`
-		Host Host `yaml:"host"`
-		Dev  Dev  `yaml:"dev"` // WARN: do not expose dev settings as flags
+		Log         Log  `yaml:"log"`
+		Host        Host `yaml:"host"`
+		Dev         Dev  `yaml:"dev"` // WARN: do not expose dev settings as flags
+		EnablePprof bool `yaml:"enable-pprof"`
```
sthaha, line 39: question: What do you think about
```
debug:
  enable-pprof: true #
  collectors:  # so that we group all debugging under debug
   - go
   - process
```

sthaha (approved):
LGTM, need a followup for minor fixes.

vimalk78:
> LGTM, need a followup for minor fixes.

pushed 2 changes, for typpo and for interface assertion

sthaha (approved):

PR #2031, opened 2025-04-28
feat: add RAPL zone filtering support through configuration

sthaha:
Changes include:
* Add RAPL zone filtering option in config (rapl/zones).
* Add zone filtering in raplPowerMeter and log excluded and included zones
* Update createCPUMeter to use RAPL config and filter zones.
* Test for RAPL config and zone filtering.

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -163,5 +163,14 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 	if fake := cfg.Dev.FakeCpuMeter; fake.Enabled {
 		return device.NewFakeCPUMeter(fake.Zones, device.WithFakeLogger(logger))
 	}
-	return device.NewCPUPowerMeter(cfg.Host.SysFS)
+
+	if len(cfg.Rapl.Zones) > 0 {
+		logger.Info("rapl zones are filtered", "zones-enabled", cfg.Rapl.Zones)
+	}
+
+	return device.NewCPUPowerMeter(
+		cfg.Host.SysFS,
+		device.WithRaplLogger(logger),
+		device.WithZoneFilter(cfg.Rapl.Zones),
+	)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,11 @@ type (
 		ProcFS string `yaml:"procfs"`
 	}
 
+	// Rapl configuration
+	Rapl struct {
+		Zones []string `yaml:"zones"`
+	}
+
 	// Development mode settings; disabled by default
 	Dev struct {
 		FakeCpuMeter struct {
@@ -37,6 +42,7 @@ type (
 		Host        Host `yaml:"host"`
 		Dev         Dev  `yaml:"dev"` // WARN: do not expose dev settings as flags
 		EnablePprof bool `yaml:"enable-pprof"`
+		Rapl        Rapl `yaml:"rapl"`
 	}
 )
 
@@ -63,6 +69,9 @@ func DefaultConfig() *Config {
 		Host: Host{
 			SysFS:  "/sys",
 			ProcFS: "/proc",
+		},
+		Rapl: Rapl{
+			Zones: []string{},
 		},
 	}
 
@@ -128,6 +137,7 @@ func RegisterFlags(app *kingpin.Application) ConfigUpdaterFn {
 	hostSysFS := app.Flag(HostSysFSFlag, "Host sysfs path").Default("/sys").ExistingDir()
 	hostProcFS := app.Flag(HostProcFSFlag, "Host procfs path").Default("/proc").ExistingDir()
 	enablePprof := app.Flag(EnablePprofFlag, "Enable pprof").Default("false").Bool()
+
 	return func(cfg *Config) error {
 		// Logging settings
 		if flagsSet[LogLevelFlag] {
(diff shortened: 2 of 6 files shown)


config/config_rapl_test.go
vprashar2929, line 1: Shouldn't we keep all config related tests inside `config.go` ?

cmd/kepler/main.go
```diff
@@ -256,5 +256,18 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 	if fake := cfg.Dev.FakeCpuMeter; fake.Enabled {
 		return device.NewFakeCPUMeter(fake.Zones, device.WithFakeLogger(logger))
 	}
-	return device.NewCPUPowerMeter(cfg.Host.SysFS)
+
+	if !cfg.Rapl.Enabled {
```
vprashar2929, line 260: What should be the behaviour when rapl is disabled?

config/config_rapl_test.go
sthaha: no, there is no hard requirement like that. It needs to be under the same package. Thats all

cmd/kepler/main.go
sthaha: Kepler should continue to monitor power of other meters. E.g. say there is platform power and gpu ...

vprashar2929: When I disable rapl I still can see Kepler producing node metrics

vprashar2929: > Kepler should continue to monitor power of other meters. E.g. say there is platform power and GPU ...

For now it should only produce metrics like CPU info and build info right?

sthaha: @vprashar2929, on second thought, I am removing `enable.rapl` flag for now since it requires other changes to monitor (more than the scope of this work). Lets revisit this when we have platform power as well.

config/config.go
```diff
@@ -243,6 +253,7 @@ func (c *Config) manualString() string {
 		{LogFormatFlag, c.Log.Format},
 		{HostSysFSFlag, c.Host.SysFS},
 		{HostProcFSFlag, c.Host.ProcFS},
+		{"rapl.zones", strings.Join(c.Rapl.Zones, ", ")},
```
vprashar2929, line 256: We should also sanatize the rapl zone input. For eg:
```
rapl:
  zones: [" core"] # zones to be enabled, empty enables all default zones

```
is a valid config and its should use `core`

sthaha: Tests are added now to address this

hack/config.yaml
```diff
@@ -6,6 +6,10 @@ host:
   sysfs: /sys # Path to sysfs filesystem (default: /sys)
   procfs: /proc # Path to procfs filesystem (default: /proc)
 
+rapl:
+  enabled: false # Enable RAPL (default: true)
```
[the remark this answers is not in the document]
sthaha: done

hack/config.yaml
```diff
@@ -6,8 +6,12 @@ host:
...
 
+rapl:
+  enabled: true # Enable RAPL (default: true)
+  zones: [] # zones to be enabled, empty enables all default zones
+
 # WARN DO NOT ENABLE THIS IN PRODUCTION - for development / testing only
 dev:
   fake-cpu-meter:
-    enabled: false
+    enabled: true
```
vprashar2929, line 16: This should stay false

cmd/kepler/main.go
```diff
@@ -163,5 +163,14 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 	if fake := cfg.Dev.FakeCpuMeter; fake.Enabled {
 		return device.NewFakeCPUMeter(fake.Zones, device.WithFakeLogger(logger))
 	}
-	return device.NewCPUPowerMeter(cfg.Host.SysFS)
+
+	if len(cfg.Rapl.Zones) > 0 {
+		logger.Info("Filtering RAPL zones", "zones", cfg.Rapl.Zones)
```
vimalk78, line 168: `cfg.Rapl.Zones` are the zones we want to keep, not the ones we want to filter out.
better say it as enabled zones, like mentioned in comment below
```
 # zones to be enabled, empty enables all default zones
```

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -79,6 +128,12 @@ func (r *raplPowerMeter) Zones() ([]EnergyZone, error) {
 		return nil, fmt.Errorf("no RAPL zones found")
 	}
 
+	zones = r.filterZones(zones)
+	if len(zones) == 0 {
+		return nil, fmt.Errorf("no RAPL zones found after filtering")
+	}
+
+	// filter out non-standard zones
```
vimalk78, line 136: what if a non-standard zone is configured.

cmd/kepler/main.go
vimalk78: comment was not just about t he log but about replacing the word `filter` with `enable` or `select`.
instead of `device.WithZoneFilter` we have `device.WithSelectedZones` or `device.WithEnabledZones`
instead of `filterZones` we say `selectZones` or `enabledZones`

but this is a minor comment. we can change it in a follow-up PR if needed.

vimalk78 (approved):

PR #2035, opened 2025-04-28
chore(config): add skip validation logic from operator

vprashar2929:
This commit adds the logic to skip validations for the config from the operator. With this we can skip certain configs that we don't want Kepler to validate.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -40,6 +40,12 @@ type (
 	}
 )
 
+type SkipValidation int
+
+const (
+	SkipHostValidation SkipValidation = 1
+)
+
 const (
 	// Flags
 	LogLevelFlag  = "log.level"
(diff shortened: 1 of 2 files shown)


sthaha:
@vprashar2929 , would you mind adding missing tests that validates skips?

config/config.go
```diff
@@ -163,7 +165,11 @@ func (c *Config) sanitize() {
 }
 
 // Validate checks for configuration errors
-func (c *Config) Validate() error {
+func (c *Config) Validate(skips ...SkipValidation) error {
+	validationSkipped := make(map[SkipValidation]bool, len(skips))
+	for _, v := range skips {
+		validationSkipped[v] = true
```
sthaha, line 171: could you also please bring in skipping the host validation as well?

sthaha (approved):

PR #2038, opened 2025-04-29
chore: expose pprof in config file

vprashar2929:
This commit exposes the `enable-pprof` in the config file.


hack/config.yaml
```diff
@@ -9,6 +9,8 @@ host:
 rapl:
   zones: [] # zones to be enabled, empty enables all default zones
 
+enable-pprof: false # Enable pprof debug endpoints
```
sthaha, line 12: Lets enable pprof

vprashar2929: Done

sthaha:
@vprashar2929 unrelated to this PR .. I think having a job run in CI that builds and runs -- 
```
kepler --config.file=./hack/config.yaml
```
will be quite useful.

vprashar2929:
I was thinking of using compose as we will anyway require it once we add e2e's

Issue #2025, opened 2025-04-23
monitor's first collection unnecessarily allocates a previous snapshot 


sthaha:
```
if prevSnapshot == nil {
  	snapshot := NewSnapshot()
      	snapshot.Timestamp = time.Now()
	pm.snapshot.Store(snapshot)
        return err
}
```

otherwise both prevSnapshot and newSnapshot are same

_Originally posted by @vimalk78 in https://github.com/sustainable-computing-io/kepler/pull/2022#discussion_r2055516916_

Improve the first collection so that 2 snapshots aren't allocated.

PR #2040, opened 2025-04-29
refactor(monitor): split  snapshot refresh into initial and subsequent readings

sthaha:
Split the refreshSnapshot logic to handle initial and subsequent power readings separately. This commit adds firstReading and calculatePower methods to improve clarity and maintainability.

Also
* Update calculateNodePower to remove initial reading logic.
* Adjust tests to reflect the new method signatures and behavior.

fixes: #2025

### DIFF
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -245,27 +245,51 @@ func (pm *PowerMonitor) isFresh() bool {
 	return age <= pm.maxStaleness
 }
 
-// refreshSnapshot create a new snapshot of the power consumption of various levels( currently only node)
+// refreshSnapshot creates a new snapshot of the power consumption
+// It handles both initial and subsequent collections assuming previous Snapshot
+// is nil only on first call.
 func (pm *PowerMonitor) refreshSnapshot() error {
 	started := pm.clock.Now()
-	defer func() { pm.logger.Info("Computed power", "duration", pm.clock.Since(started)) }()
-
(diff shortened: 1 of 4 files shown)


internal/monitor/monitor.go
```diff
@@ -245,27 +245,51 @@ func (pm *PowerMonitor) isFresh() bool {
...
+	}
+	return nil
+}
+
+func (pm *PowerMonitor) calculatePower(prev, newSnapshot *Snapshot) error {
+	if err := pm.calculateNodePower(prev.Node, newSnapshot.Node); err != nil {
+		return fmt.Errorf(nodePowerError, err)
+	}
+	return nil
+}
```
vimalk78, line 295: nit: these methods are called nowhere else, not even in test code, so why not inline these methods at the place these are called?

 not a blocker.

vimalk78 (approved):

internal/monitor/monitor.go
sthaha: This will expand in future to ... 

```
func firstReading(snapshot) {
  firstNodeRead(snapshot.Node)
  firstProcessRead(snapshot.Process)
  firstContainerRead(snapshot.Container)
  firstVMRead(snapshot.VM)
}

func computePower(prev, new) {
  computeNodePower(prev.Node, new.Node)
  computeProcessRead(prev.Processes, new.Processes)
  // .. container
  // .. VM
  // .. Pods 
}

```

PR #2046, opened 2025-05-04
ci: add job for labeling PR

vprashar2929:
This commits adds a job which will label PRs based on conventional commit message standards.


vprashar2929:
For this workflow to work, it has to be merged as the action is triggered on `pull_request_target`, meaning it will trigger once any pull request is created against the `reboot` branch.

vprashar2929:
Once the PR is merged, we have to set the GH workflow permissions to Read and write under repo settings
Screenshot 2025-05-04 at 12 23 18 PM

sthaha (approved):
lgtm but I think we should be consistent/enforce upper/lower-case for `type` in `type(scope): msg` (being lowercase).

vprashar2929:
I think pre-commit hook enforces that 🤔

sthaha:
> I think pre-commit hook enforces that 🤔

So why does this honor `FEAT`, `feat` and `Feat` 🤔 ?

vprashar2929:
updated the nouns in config

sthaha (approved):

PR #2057, opened 2025-05-08
Introduce PRODUCTION flag to enable/disable stripping binaries after build

egypcio:
by merging this PR we would:

* be able to set `PRODUCTION` to strip the exporter and validator binaries;
* avoid adding [DWARF](https://dwarfstd.org/doc/DWARF5.pdf) debug information to newer releases;
* reduce the size/footprint of the built binaries;
* still be able to find/follow debug symbols using [Delve](https://github.com/go-delve/delve), should necessary;
* promote the usage of less energy/power consumption.

example using `PRODUCTION` to enable/disable stripped build for the exporter locally:

```bash
$ PRODUCTION=1 make _build_local
$ file _output/bin/linux_amd64/kepler 
_output/bin/linux_amd64/kepler: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=45a457a614ec12d4f18e983f7f28c9bfd3dfc9f3, for GNU/Linux 3.2.0, stripped
$ ls -laish _output/bin/linux_amd64/kepler
27662536 37M -rwxrwxr-x 1 egypcio egypcio 37M May  8 11:42 _output/bin/linux_amd64/kepler
```
```bash
$ make _build_local
$ file _output/bin/linux_amd64/kepler 
_output/bin/linux_amd64/kepler: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=ee248b3f1e691fa128d33baa5f91eaa82286a12b, for GNU/Linux 3.2.0, with debug_info, not stripped
$ ls -laish _output/bin/linux_amd64/kepler
27662536 53M -rwxrwxr-x 1 egypcio egypcio 53M May  8 11:46 _output/bin/linux_amd64/kepler
```

impact on the binary created after building the validator:

```
8.4M -rwxrwxr-x 1 egypcio egypcio 8.4M May  8 11:49 _output/bin/validator-debug
5.7M -rwxrwxr-x 1 egypcio egypcio 5.7M May  8 11:49 _output/bin/validator
```

> Host: Linux version 6.11.0-25-generic (buildd@lcy02-amd64-027) (x86_64-linux-gnu-gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, GNU ld (GNU Binutils for Ubuntu) 2.42), a.k.a. Ubuntu 24.04.2 LTS (Noble Numbat)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,19 @@ ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
     GPU_TAGS := habana
 endif
 
+# Use PRODUCTION to enable/disable stripping binaries for the kepler exporter
+# 	and validator. That reduces the size of the binaries and removes debug symbols.
+# Expected values for this flag are:
+#	0 - Unstripped binaries including debug symbols (default);
+#	1 - Stripped binaries.
+# More about stripped binaries and debug symbols can be found at the RedHat's developer portal here:
+#	* https://developers.redhat.com/articles/2024/04/03/how-add-debug-support-go-stripped-binaries
+#
+PRODUCTION ?= 0
+ifeq ($(PRODUCTION),1)
+  LDFLAGS+= -s -w
+endif
+
 # set GOENV
 GOOS := $(shell go env GOOS)
 GOARCH := $(shell go env GOARCH)


sthaha:
@egypcio Thank you for the patch but TBH I don't feel strongly about potential saving .. 

As mentioned in the article you linked ... 

> Symbol information takes up a lot of storage space, so when there are space constraints, it makes sense to work with stripped binaries.

Where exactly are you hitting space constraint ? 

Also if we are to enable this, we would want `make build PRODUCTION=1` so that the default build contains debug information. 


Could you also please squash the commits into a single one?

egypcio:
> @egypcio Thank you for the patch but TBH I don't feel strongly about potential saving ..
> 
> As mentioned in the article you linked ...
> 
> > Symbol information takes up a lot of storage space, so when there are space constraints, it makes sense to work with stripped binaries.
> 
> Where exactly are you hitting space constraint ?
> 
> Also if we are to enable this, we would want `make build PRODUCTION=1` so that the default build contains debug information.
> 
> Could you also please squash the commits into a single one?

olá @sthaha,

many thanks for the input/feedback. much appreciated!

I've applied the suggested changes and squashed the commits into a single one.

now, on the space constraint: I wouldn't take that specific cut from that blog post so literal, as per serving a smaller binary without debug symbols would reduce I/O and memory consumption (now: consider a large scale deployment, or multiple pipelines running to perform CI/CD; consider multiple deployments world wide).

one of the motivations for the proposed change in the PR comes from this article:
* https://www.cncf.io/blog/2023/10/11/exploring-keplers-potentials-unveiling-cloud-application-power-consumption/

Makefile
```diff
@@ -47,6 +47,21 @@ ifeq ($(shell ldconfig -p | grep -q libhlml.so && echo exists),exists)
...
+#	1 - Stripped binaries.
+# More about stripped binaries and debug symbols can be found at the RedHat's developer portal here:
+#	* https://developers.redhat.com/articles/2024/04/03/how-add-debug-support-go-stripped-binaries
+#
+PRODUCTION ?= 0
+ifeq ($(PRODUCTION),1)
+  go_ldflags_strip_args=-s -w
+else
+  go_ldflags_strip_args=
+endif
```
sthaha, line 63:
```suggestion
ifeq ($(PRODUCTION),1)
  LDFLAGS+= -s -w
endif
```

Makefile
```diff
@@ -177,7 +192,7 @@ _build_local: generate ##  Build Kepler binary locally.
 	@mkdir -p "$(CROSS_BUILD_BINDIR)/$(GOOS)_$(GOARCH)"
 	+@$(GOENV) go build \
 		-v -tags ${GO_BUILD_TAGS} \
-		-ldflags "$(LDFLAGS)" \
+		-ldflags "$(LDFLAGS) $(go_ldflags_strip_args)" \
```
sthaha, line 195:
```suggestion
		-ldflags "$(LDFLAGS)" \
```

Makefile
```diff
@@ -430,7 +445,7 @@ build-validator: tidy-vendor format ## Build Validator.
 	+@$(GOENV) go build \
 		-v -tags ${GO_BUILD_TAGS} \
 		-o $(CROSS_BUILD_BINDIR)/$(GOOS)_$(GOARCH)/validator \
-		-ldflags "$(LDFLAGS)" \
+		-ldflags "$(LDFLAGS) $(go_ldflags_strip_args)" \
```
sthaha, line 448:
```suggestion
		-ldflags "$(LDFLAGS)" \
```

sthaha:
@egypcio , could you please elaborate what exactly in the blog is the motivation that you are referring to ..

> one of the motivations for the proposed change in the PR comes from this article:
> https://www.cncf.io/blog/2023/10/11/exploring-keplers-potentials-unveiling-cloud-application-power-consumption/

egypcio:
`ping` 😃 

any particular blocker here?

sthaha:
@egypcio , no the patch looks good to me but as you can see the CI is failing, which needs to pass. Can you please take a look?

sthaha (approved):

sthaha:
@egypcio , thank you for the patch  🙏

PR #2058, opened 2025-05-10
chore(build): add production build support and enhance debug builds

sthaha:
* Introduce PRODUCTION variable to strip debug symbols for production builds.
* Refactor LDFLAGS to separate version flags and support production optimizations.
* Add BUILD_ARGS for customizable build arguments (such as -cover) in both build and build-debug targets.
* Output debug builds to a distinct -debug binary with race detection.


vprashar2929 (approved):

vprashar2929:
Maybe we have to update the Dockerfile as well with `make build PRODUCTION=1`

PR #2061, opened 2025-05-12
feat: add process and container power monitoring

sthaha:
This commit extends power monitoring to include process and container-level metrics alongside node-level metrics.

Changes include

* A new resource informer to track processes and containers, calculating power consumption based on CPU time ratios.
* Implement container detection for Docker, Containerd, CRI-O, Podman, and Kubernetes runtimes. Include tests for container detection and resource tracking.
* Add Prometheus metrics for energy (joules) and power (watts) per RAPL zone, with labels for process and container metadata.


internal/monitor/container.go
```diff
@@ -0,0 +1,122 @@
...
+	containers := pm.resources.Containers()
+
+	// Skip if no containers
+	if len(containers.Running) == 0 {
+		pm.logger.Debug("No running containers found, skipping container power calculation")
+		return nil
+	}
+
+	pm.logger.Debug("Calculating container power",
+		"node-cputime", containers.NodeCPUTimeDelta,
```
vprashar2929, line 62: Is it in seconds?

config/config_test.go
```diff
@@ -423,3 +428,90 @@ func TestValidateWithSkip(t *testing.T) {
...
+		name     string
+		args     []string
+		expected expect
+	}{{
+		name:     "default",
+		args:     []string{},
+		expected: expect{interval: 5 * time.Second, staleness: 500 * time.Millisecond, parseError: nil},
+	}, {
+		name:     "invalid-interval flag",
+		args:     []string{"--monitor.interval=-10Fs"},
```
vprashar2929, line 485: typo: `10s`

config/config.go
```diff
@@ -140,8 +153,13 @@ func RegisterFlags(app *kingpin.Application) ConfigUpdaterFn {
 	// Logging
 	logLevel := app.Flag(LogLevelFlag, "Logging level: debug, info, warn, error").Default("info").Enum("debug", "info", "warn", "error")
 	logFormat := app.Flag(LogFormatFlag, "Logging format: text or json").Default("text").Enum("text", "json")
+	// host
 	hostSysFS := app.Flag(HostSysFSFlag, "Host sysfs path").Default("/sys").ExistingDir()
 	hostProcFS := app.Flag(HostProcFSFlag, "Host procfs path").Default("/proc").ExistingDir()
+
+	// monitor
+	monitorInterval := app.Flag(MonitorIntervalFlag, "Rate at which kepler should monitor processes; 0 to disable").Default("5s").Duration()
```
vprashar2929, line 161: Isn't this irrespective of the process?

sthaha:
> node-container diff doesn't look correct 🤔
> 
What does `node - container` diff do? And how is it incorrect 🤔 ?

internal/resource/informer.go
```diff
@@ -0,0 +1,271 @@
...
+
+	cached, exists := ri.containerCache[c.ID]
+	if !exists {
+		cached = c.Clone()
+		ri.containerCache[c.ID] = cached
+	}
+
+	if !visited {
+		cached.CPUTimeDelta = 0
+	}
```
vimalk78, line 206: `visited` means that the container is already added to `containersRunning` during the same monitor interval, and we want to reset the CPUTimeDelta for the cached Container if we are seeing this container for the first time in the interval.

it this understanding correct? can we reset the cputime explicitly at the end of interval?

internal/resource/procfs_reader.go
```diff
@@ -0,0 +1,126 @@
...
+	return p.proc.Environ()
+}
+
+func (p *procWrapper) CmdLine() ([]string, error) {
+	return p.proc.CmdLine()
+}
+
+// userHZ is the number of clock ticks per second
+// hardcoded just like in procfs
+const userHZ = 100
```
vimalk78, line 73:
```
/*
#include <unistd.h>
*/
import "C"

userHZ = int64(C.sysconf(C._SC_CLK_TCK))

```

internal/resource/informer.go
```diff
@@ -0,0 +1,271 @@
...
+	ri.processes.Running = procsRunning
+	ri.processes.Terminated = procsTerminated
+
+	ri.containers.NodeCPUTimeDelta = nodeCPUDelta
+	ri.containers.Running = containersRunning
+	ri.containers.Terminated = containersTerminated
+
+	now := ri.clock.Now()
+	ri.lastScanTime = now
+	duration := ri.clock.Since(started)
```
vimalk78, line 180:
```suggestion
	duration := now.Sub(started)
```

internal/resource/procfs_informer_test.go
```diff
@@ -0,0 +1,498 @@
...
+		assert.ErrorContains(t, err, "failed to get processes")
+
+		mockInformer.AssertExpectations(t)
+	})
+}
+
+// Test for the procfs fixture to ensure the test fixture directory is available
+// and to test the integration with procfs package
+func TestProcWrapper(t *testing.T) {
+	fs, err := procfs.NewFS("./testdata/good/proc")
```
vimalk78, line 408: folder does not exist

config/config.go
```diff
@@ -269,6 +300,8 @@ func (c *Config) manualString() string {
 		{LogFormatFlag, c.Log.Format},
 		{HostSysFSFlag, c.Host.SysFS},
 		{HostProcFSFlag, c.Host.ProcFS},
+		{MonitorIntervalFlag, c.Monitor.Interval.String()},
+		{"monitor.staleness", c.Monitor.Staleness.String()},
```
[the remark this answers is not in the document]
sthaha: "monitor.staleness" isn't a flag, we can define a const but so far the consts are only defined for command line options.

internal/resource/informer.go
sthaha: Do you mean have timer that resets it ? If thats what you meant, then we will have to think of concurrency :(

internal/resource/procfs_reader.go
sthaha: Just like in node-exporter, lets only hardcode it and not have CGO unless really needed

internal/resource/container.go
```diff
@@ -0,0 +1,132 @@
...
+func containerNameFromCmdLine(cmdline []string) string {
+	if len(cmdline) <= 1 {
+		return ""
+	}
+
+	exe := filepath.Base(cmdline[0])
+
+	// look for --name flag
+	for i, arg := range cmdline {
+		if i > 0 && strings.HasPrefix(arg, "--name=") {
```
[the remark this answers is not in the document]
sthaha: fixed and added tests

internal/resource/informer.go
vimalk78: no, at the end of monitor interval loop.

vimalk78:
can we reduce the number of files in testdata? do we need all the files there?
```
autogroup
auxv
cgroup
cmdline
comm
coredump_filter
cpu_resctrl_groups
environ
gid_map
ksm_merging_pages
ksm_stat
latency
mem
mountinfo
mounts
numa_maps
oom_adj
oom_score_adj
patch_state
personality
projid_map
sched
schedstat
setgroups
stack
stat
timens_offsets
timerslack_ns
uid_map
```

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -77,13 +96,60 @@ func (c *PowerCollector) updateDescriptors() {
...
+				nil,
+			)
+		}
+
+		// container metric descriptors
+		if _, exists := c.containerJoulesDescriptors[zoneName]; !exists {
+			c.containerJoulesDescriptors[zoneName] = prometheus.NewDesc(
+				prometheus.BuildFQName(namespace, containerRAPL, zoneName+"_joules_total"),
+				"Energy consumption in joules for RAPL zone "+zoneName+" by container",
+				[]string{"id", "name", "runtime", "namespace", "pod"},
```
vimalk78, line 124: can we update the Help string to mention what each label means?

```suggestion
				"Energy consumption in joules for RAPL zone "+zoneName+" by container.  Labels: id = container id, name: container name, runtime: container runtime, namespace: kubernetes namespace (if available), pod: kubernetes pod name (if available)",
				[]string{"id", "name", "runtime", "namespace", "pod"},
```
same for other descriptors.

"namespace" is kubernetes namespace, it should be clear from metric description.

vimalk78 (approved):

PR #2065, opened 2025-05-14
chore(dashboard): add support for process and container-level metrics

vprashar2929:
This commit updates the existing dashboard to support process-level and container-level metrics.


vprashar2929:
Changes:

Grafana-05-15-2025_04_45_PM

sthaha (approved):

PR #2066, opened 2025-05-14
feat(server) : Support TLS with web config

vimalk78:
- Updated flags/config to support passing web config
- api server uses web config to support TLS
- added tests for config
- added tests for tls support

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -36,6 +36,9 @@ type (
 			Zones   []string `yaml:"zones"`
 		} `yaml:"fake-cpu-meter"`
 	}
+	Web struct {
+		Config string `yaml:"configFile"`
+	}
 
 	Config struct {
 		Log         Log  `yaml:"log"`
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -36,13 +36,17 @@ type (
 			Zones   []string `yaml:"zones"`
 		} `yaml:"fake-cpu-meter"`
 	}
+	Web struct {
+		Config string `yaml:"config"`
```
sthaha, line 40:
```suggestion
		ConfigFile string `yaml:"configFile"`
```

sthaha (approved):
Looks good to me. Please consider calling `web.config` `web.config-file` and 

```yaml
web:
  configFile: 
```

vimalk78:
> Looks good to me. Please consider calling `web.config` `web.config-file` and
> 
> ```yaml
> web:
>   configFile: 
> ```

updated

sthaha (approved):

PR #2067, opened 2025-05-15
chore: add TLS config to sample config

vprashar2929:
This commit adds the TLS related config to the sample configuration file

### DIFF
--- a/hack/config.yaml
+++ b/hack/config.yaml
@@ -11,6 +11,9 @@ rapl:
 
 enable-pprof: true # Enable pprof debug endpoints
 
+web:
+  configFile: "" # Path to TLS server config file
+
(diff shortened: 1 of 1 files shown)


hack/config.yaml
```diff
@@ -11,6 +11,9 @@ rapl:
 
 enable-pprof: true # Enable pprof debug endpoints
 
+web:
+  configFile: "" # Path to TLS server config file
```
vimalk78, line 15: the default value is already empty string.

vimalk78 (approved):

hack/config.yaml
vimalk78: we could add a separate sample which has a defined web config also, and refers to that file in this config.yaml

sthaha:
@vprashar2929 please wait for #2061  to merge 🙏

PR #2069, opened 2025-05-16
workflow(xcrypto-check): add workflow to track x/crypto deps

KaiyiLiu1234:
Workflow tracks golang.org/x/crypto deps introduced prior to PR, during PR, and any direct imports of golang.org/x/crypto. Any usage of x/crypto will result in a warning output in the logs but will not enforce a failure if x/crypto is discovered.

Note, x/crypto contains cryptographic modules that are unacceptable for fips compliance. It is important to keep track of x/crypto usage to aid in manual audits.


KaiyiLiu1234:
Sample CI Run when x/crypto dependencies get introduced: https://github.com/KaiyiLiu1234/kepler/actions/runs/15060128577/job/42333613680?pr=5.

KaiyiLiu1234:
Workflow added to Operator here: https://github.com/sustainable-computing-io/kepler-operator/pull/512

.github/workflows/x-crypto-check.yaml
vprashar2929, line 1: @KaiyiLiu1234 Can you move this inside `.github/crypto-check` ?

.github/workflows/x-crypto-check.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Get PR branch dependencies
+        run: |
+          git checkout ${{ inputs.head_sha }}
+          go mod tidy
+          go mod graph > deps-pr.txt
+
+      - name: Analyze crypto dependencies
+        shell: bash
+        run: |
+          echo "=== Dependency Report ==="
```
vprashar2929, line 42: Just wondering if we can shift this into a script which can be inside `hack/` for easy usage and maintainability

vprashar2929:
@KaiyiLiu1234 Can you check if you have configured pre-commit hooks for the repo? Ref: https://github.com/sustainable-computing-io/kepler/blob/reboot/docs/developer/pre-commit.md `workflow(xcrypto-check)` is not a valid conventional commit message

KaiyiLiu1234:
@vprashar2929 I have configured pre-commit and it passes all the tests. I will amend it to follow the precommit documentation on commit messages.

.github/workflows/x-crypto-check.yaml
KaiyiLiu1234: will do!

.github/workflows/x-crypto-check.yaml
KaiyiLiu1234: sure!

.github/workflows/x-crypto-check.yaml
KaiyiLiu1234: crypto check has been updated to be an action

.github/workflows/x-crypto-check.yaml
KaiyiLiu1234: script has been moved to hack/reports

KaiyiLiu1234:
@vprashar2929 @sthaha I have updated the script to now be runnable on local machine. All that is required is to pass it the --base-ref (ex. reboot), --head-sha, and optionally the --report-path. I also made it so that the report is written to the PR as a comment as shown here: https://github.com/KaiyiLiu1234/kepler/pull/6#issuecomment-2902802802. The reason the comment does not appear here is because of this error: Error: Resource not accessible by integration. This indicates the workflow does not have the needed permissions. Does anyone know the proper way to provide this workflow with write permissions?

hack/reports/print-xcrypto-report.sh
```diff
@@ -0,0 +1,150 @@
...
+    log_and_save "ℹ️ Changes introduced by this PR:"
+    if ! grep -q " golang.org/x/crypto@" /tmp/deps-pr.txt; then
+        log_and_save "✅ PR does not introduce any new dependencies on golang.org/x/crypto"
+    else
+        new_deps=$(comm -13 <(sort /tmp/deps-base.txt) <(sort /tmp/deps-pr.txt) | grep " golang.org/x/crypto@" || true)
+        if [ -z "$new_deps" ]; then
+            log_and_save "❗PR doesn't add new x/crypto dependencies (note it is possible this PR depends on existing x/crypto dependencies)"
+        else
+            log_and_save "⚠️ PR introduces new dependencies on golang.org/x/crypto:"
+            pr_deps=$(echo "$new_deps" | sed 's/^/ 🔹 /')
```
vprashar2929, line 67:
```suggestion
pr_deps=" 🔹 ${new_deps//$'\n'/$'\n' 🔹 }"
```

hack/reports/print-xcrypto-report.sh
```diff
@@ -0,0 +1,150 @@
...
+    log_and_save ""
+
+    # Check for direct imports of x/crypto
+    log_and_save "ℹ️ Locate any direct dependencies on golang.org/x/crypto:"
+    occurrences=$(< /tmp/deps-direct.txt)
+    if [ -z "$occurrences" ]; then
+        log_and_save "✅ No direct imports of golang.org/x/crypto found"
+    else
+        log_and_save "⚠️ Discovered direct imports of golang.org/x/crypto:"
+        direct_deps=$(echo "$occurrences" | sed 's/^/ 🔹 /')
```
vprashar2929, line 82:
```suggestion
         direct_deps=" 🔹 ${occurrences//$'\n'/$'\n' 🔹 }"

```

.github/workflows/pr-checks.yaml
```diff
@@ -39,6 +39,35 @@ jobs:
         shell: bash
         run: make fmt && git diff --exit-code
 
+  check-x-crypto-deps:
```
vprashar2929, line 42: you can try re-running the worklfow by adding this here: https://github.com/thollander/actions-comment-pull-request?tab=readme-ov-file#permissions

KaiyiLiu1234: Added but the same permissions error appears.

hack/reports/print-xcrypto-report.sh
KaiyiLiu1234: Added change

hack/reports/print-xcrypto-report.sh
KaiyiLiu1234: Added change

.github/workflows/pr-checks.yaml
vprashar2929: Can you create a separate workflow for this like assign-labels, instead of changing the event in PR checks? Post that we can merge and create a dummy PR to validate if it works. I believe it should work as the assign label workflow works fine.

KaiyiLiu1234: Sounds good. I made the check x crypto into a separate workflow. Please let me know if it was done correctly. We can merge it and I will make a dummy PR to see if the comment will work.

PR #2074, opened 2025-05-20
chore(logger): prefer buffered stdout over stderr

sthaha:
This commit modified logger to use stdout (buffered out) unless the stdout exporter is enabled. On enabling stdout exporter, the logger is configured to write to stderr.

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -30,7 +30,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	logger := logger.New(cfg.Log.Level, cfg.Log.Format)
+	// Configure logger - use stderr if stdout exporter is enabled to prevent output interleaving
+	logOut := os.Stdout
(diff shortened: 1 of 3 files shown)


cmd/kepler/main.go
```diff
@@ -78,7 +84,7 @@ func parseArgsAndConfig() (*config.Config, error) {
 	updateConfig := config.RegisterFlags(app)
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
-	logger := logger.New("info", "text")
+	logger := logger.New("info", "text", os.Stdout)
```
vimalk78, line 87: this function is using separate logger?

vimalk78 (approved):

cmd/kepler/main.go
sthaha: yes, because parseArgs needs to log before the "actual / application" logger is created.

PR #2087, opened 2025-05-20
ci: capture diff after running make deps

vprashar2929:
This commit ensures that the diff is taken after running make deps and fails if the diff is not empty.

It also fixes the issue where the go.mod file contains the diff due to which VERSION being computed has `-dirty` suffix.


vprashar2929:
Ref: https://quay.io/repository/sustainable_computing_io/kepler-reboot?tab=tags&tag=latest

sthaha (approved):

PR #2089, opened 2025-05-21
chore(dashboard): test dashboard for node power metrics

KaiyiLiu1234:
Implemented dashboard which illustrates node level power metrics for energy zones and instances. The dashboard shows historical power consumption of node zones, total power consumption of zones in cluster, average power consumption, and current power consumption. All power consumption metrics are represented in Watts.


KaiyiLiu1234:
Below is what the dashboard looks like @sthaha @vprashar2929 
Screenshot from 2025-05-20 16-40-00
Screenshot from 2025-05-20 16-39-47

sthaha (approved):

PR #2102, opened 2025-05-26
chore: add merge config logic from operator

vprashar2929:
This commit adds the logic to merge the config with additional configs from the operator.


sthaha (dismissed):
@vprashar2929 could we please implement the builder pattern here for config?

PR #2103, opened 2025-05-27
chore: implement builder pattern for config merging

vprashar2929:
This commit replaces the current MergeAdditionalConfigs implementation with a builder pattern. It also fixes boolean merging issue by switching from `mergo.WithOverride` to `mergo.WithOverwriteWithEmptyValue`, ensuring additional Config object mirror default content with modifications during YAML unmarshalling. This is a workaround for the issue https://github.com/darccio/mergo/issues/249.


config/config.go
```diff
@@ -427,31 +433,34 @@ func (c *Config) manualString() string {
 	return sb.String()
 }
 
-func (c *Config) merge(additional *Config) error {
-	return mergo.Merge(c, additional, mergo.WithOverride)
+// UseDefault sets the default configuration
+func (b *Builder) UseDefault() *Builder {
+	b.Config = DefaultConfig()
+	return b
 }
```
sthaha, line 440: we don't need this method since the `Build` starts with a default. I think we should replace it with `WithConfig` which set the base config.

config/config.go
```diff
@@ -427,31 +440,53 @@ func (c *Config) manualString() string {
...
+		if err := yaml.Unmarshal([]byte(y), additional); err != nil {
+			return nil, fmt.Errorf("failed to parse YAML config: %w", err)
+		}
 
-// MergeAdditionalConfigs merges additional YAML configurations into a default config
-func MergeAdditionalConfigs(defaultConfig *Config, additionalConfigs ...string) error {
-	for _, cfg := range additionalConfigs {
-		if err := defaultConfig.mergeYAML(cfg); err != nil {
-			return fmt.Errorf("failed to merge additional config: %w", err)
+		if err := mergo.Merge(b.Config, additional, mergo.WithOverride, mergo.WithTransformers(PtrBoolTransformer{})); err != nil {
```
sthaha, line 483: @vprashar2929 , lets make sure this is covered by the tests.

sthaha (approved):

PR #2112, opened 2025-05-29
chore(compose): update dashboard to support recent zone changes

vprashar2929:
With the recent zone changes to include zone+index in the name the current dashboard fails to render the queries that don't follow the new zone name format. This commit updates the dashboard to include a `zone_clean` variable that filters out the zone index and then use that variable where zone+index is not supported.

### DIFF
--- a/compose/dev/grafana/dashboards/dev/dashboard.json
+++ b/compose/dev/grafana/dashboards/dev/dashboard.json
@@ -547,7 +547,7 @@
             "uid": "PDE6745920139CE56"
           },
           "editorMode": "code",
-          "expr": "sum(rate(node_rapl_${zone}_joules_total[20s]))",
+          "expr": "sum(rate(node_rapl_${zone_clean}_joules_total[20s]))",
           "instant": false,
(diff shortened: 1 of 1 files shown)


vimalk78:
@vprashar2929 if you have, pls add a picture

vprashar2929:
Grafana-05-29-2025_11_55_PM

compose/dev/grafana/dashboards/dev/dashboard.json
```diff
@@ -547,7 +547,7 @@
             "uid": "PDE6745920139CE56"
           },
           "editorMode": "code",
-          "expr": "sum(rate(node_rapl_${zone}_joules_total[20s]))",
+          "expr": "sum(rate(node_rapl_${zone_clean}_joules_total[20s]))",
```
sthaha, line 550: Why are we calling it zone_clean ?

vprashar2929: sounds good

sthaha (approved):

PR #2114, opened 2025-05-30
ci: use git archive for source version checkout

vprashar2929:
This commit updates the x-crypto detection script to use git archive for checking out two different versions of source instead of using worktrees
It also fixes the issue where on CI the script fails when creating worktree for base branch as it already exists.

### DIFF
--- a/.github/check-x-crypto/action.yaml
+++ b/.github/check-x-crypto/action.yaml
@@ -13,8 +13,6 @@ runs:
   steps:
     - name: Checkout code
       uses: actions/checkout@v4
-      with:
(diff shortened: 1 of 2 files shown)


vprashar2929:
The current CI run still refers to the script committed in the reboot branch. Hence, it will fail

KaiyiLiu1234 (approved):
looks great! Thanks so much for the help!

KaiyiLiu1234:
Merge when you're ready!

vprashar2929:
Thanks 🙇

PR #2116, opened 2025-05-30
ci: update checkout action to check out head commit from forked repo

KaiyiLiu1234:
Added fetch-depth, ref, and repository parameters to the checkout action. This should be safe because the workflow will not run any code from the forked repository making the pull request, and the workflow uses git archive. Specifically, the workflow will only search for the dependency graph (with go mod graph) and direct imports of "golang.org/x/crypto".

### DIFF
--- a/.github/check-x-crypto/action.yaml
+++ b/.github/check-x-crypto/action.yaml
@@ -11,14 +11,6 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Checkout code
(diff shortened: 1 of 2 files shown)


KaiyiLiu1234:
Let me know if this PR is safe to commit given the fact I am making it so that checkout action checks out a forked repository @vprashar2929. I also i believe we need to merge this for the changes to take effect?

vprashar2929 (approved):

PR #2119, opened 2025-06-01
fix(container): Improve container ID detection using deepest cgroup path match for nested containerIDs in cgroup path

vimalk78:
This PR improves container ID detection using deepest cgroup path match for nested containerIDs in cgroup path, typical in kind cluster.

If multiple container IDs are found within a single cgroup path, the changed function prioritizes the one appearing furthest to the right or "deepest" in the path string.

e.g. for a pod created in kind cluster, the cgroup path for the process running in pod is typically like below:
```
0::/system.slice/docker-fd9d0ea06257a9780827cbc7fd92e3812a54fca26d63e191b73610d5d48b9cbd.scope/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-besteffort.slice/kubelet-kubepods-besteffort-podeab5a334_93fe_48a8_b139_9e8079c1f163.slice/cri-containerd-99f3a16ea25b7724cb56a4f0c0df1113ad9474fbf5545bead97fd5c7f61c13f4.scope
```
 the first container (docker) refers to the kind node
the second container (cri-containerd) refers to the container running as part of the pod

this commit will return the second container id

### DIFF
--- a/internal/resource/container.go
+++ b/internal/resource/container.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


sthaha:
Thanks a lot @vimalk78 ! lgtm but CI is failing. 

Question. if we want the deepest scope, how about we separate each segment by `/` and regex match each segment in reverse order?

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sustainable-computing-io/kepler
 
-go 1.22
+go 1.23
+
+toolchain go1.23.3
```
sthaha: Nice! wondering if you ran go mod tidy?

vimalk78: yes

PR #2120, opened 2025-06-02
ci: make the x-crypto workflow temporarily disable

vprashar2929:
This commit disables the x-crypto workflow temporarily to run on pull requests until the workflow and script are fixed


vimalk78:
The `x/crypto` check is still running for PR

vprashar2929:
yeah It will not once this is merged since we used `pull_request_target` event in xcrypto and it looks for the workflows that are in reboot branch

PR #2122, opened 2025-06-02
ci: remove dependency on codecov for publish

vprashar2929:
This commit removes the publish job dependency on the removed codecov job.


vprashar2929:
Previous CI failure for ref: https://github.com/sustainable-computing-io/kepler/actions/runs/15381405419

sthaha (approved):

PR #2125, opened 2025-06-03
fix(power): correct power attribution by separating used vs idle power

sthaha:
This PR fixes a critical power attribution bug where workloads (processes/containers/VMs)
were being attributed the full node power consumption instead of only the power
corresponding to their actual CPU utilization.

Key Changes:
  - Introduced used vs idle power split: Node power is now divided between "used" (active workload) and "idle" (system overhead) portions based on actual CPU usage ratio from /proc/stat
  - Fixed attribution algorithm: Processes/containers now receive power attribution only from the "used" power portion rather than total node power
  - Enhanced metrics granularity: Added new `kepler_node_cpu_[active|idle]_[watts|joules_total]`  `kepler_node_cpu_usage_ratio` metric

Impact: Provides significantly more accurate power attribution that correctly reflects the relationship between CPU utilization and energy consumption, eliminating over-attribution to workloads.


internal/resource/procfs_reader.go
```diff
@@ -90,15 +90,68 @@ func WrapProc(proc procfs.Proc) procInfo {
 type allProcReader interface {
 	// AllProcs returns a list of all running processes
 	AllProcs() ([]procInfo, error)
+
+	// CPUUsageRatio returns the CPU usage ratio
+	CPUUsageRatio() (float64, error)
```
vimalk78, line 95: can we instead call it `ActiveCPUUsageRatio`  or just `ActiveCPUUsage`?  just saying cpu usage is open to interpretation.

and lets provide the definition in the comments
```
activeCPUUsage := (User + System) / (User + System + Idle)
```

vimalk78: though it is better to define it as
```
totalCPUUsage := User + Nice + System + Idle + Iowait + IRQ + SoftIRQ + Steal + Guest + GuestNice
activeCPUUsage := totalCPUUsage - (Idle + Iowait)
activeCPUUsageRatio =  (activeUsage / totalUsage)

```

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -77,8 +78,12 @@ func NewPowerCollector(monitor PowerDataProvider, logger *slog.Logger) *PowerCol
 	c := &PowerCollector{
 		pm:                      monitor,
 		logger:                  logger.With("collector", "power"),
-		nodeCPUJoulesDescriptor: joulesDesc("node", "cpu", []string{zone, "path"}),
-		nodeCPUWattsDescriptor:  wattsDesc("node", "cpu", []string{zone, "path"}),
+		nodeCPUJoulesDescriptor: joulesDesc("node", "cpu", []string{zone, "path", "mode"}),
+		nodeCPUWattsDescriptor:  wattsDesc("node", "cpu", []string{zone, "path", "mode"}),
+		nodeCPUUsageRatioDescriptor: prometheus.NewDesc(
+			prometheus.BuildFQName(keplerNS, "node", "cpu_usage_ratio"),
```
vimalk78, line 114: same way as in `procfs_reader.go` , name the metric as `active_cpu_usage`

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -77,8 +78,12 @@ func NewPowerCollector(monitor PowerDataProvider, logger *slog.Logger) *PowerCol
...
+			"CPU usage ratio of a node",
```
vimalk78: mention that this ratio is bettween `0.0` and `1.0`  , not in %age terms
also better to restrict the decimal precision

internal/resource/procfs_reader.go
```diff
@@ -90,15 +90,68 @@ func WrapProc(proc procfs.Proc) procInfo {
...
+// user + system over (user + system  + idle )
+func (r *procFSReader) CPUUsageRatio() (float64, error) {
+	current, err := r.fs.Stat()
+	if err != nil {
+		return 0, err
+	}
+
+	pU, pS, pI := r.prevStat.User, r.prevStat.System, r.prevStat.Idle
+	r.prevStat = current.CPUTotal
+	if pU+pS+pI == 0 {
```
vimalk78: no overflow, right? :)

internal/resource/procfs_reader.go
```diff
@@ -90,15 +90,68 @@ func WrapProc(proc procfs.Proc) procInfo {
...
+//
+// 	// find delta
+// 	dU := cU - pU
+// 	dS := cS - pS
+// 	dI := cI - pI
+//
+// 	active := dU + dS
+// 	total := active + dI
+// 	return active / total
+// }
```
vimalk78: remove commented code

internal/resource/procfs_reader.go
```diff
@@ -90,15 +90,68 @@ func WrapProc(proc procfs.Proc) procInfo {
...
+
 // AllProcs returns a list of all running processes
 func (r *procFSReader) AllProcs() ([]procInfo, error) {
+	r.fs.Stat()
```
vimalk78: why call this?

internal/monitor/types.go
```diff
@@ -30,17 +46,22 @@ type Usage struct {
 }
 
 // ZoneUsageMap maps zones to energy data
-type ZoneUsageMap map[EnergyZone]*Usage
+type (
+	ZoneUsageMap     map[EnergyZone]*Usage
```
vimalk78: please add comment about the type, to highlight difference with NodeZoneUsageMap

internal/resource/procfs_reader.go
sthaha: IMHO, CPUUsage implies that is active i.e, its used. Active IMO is superfluous. Just like how we say "what's cpu usage?". 

for the ratio calculation, I will get back to you on that.

internal/resource/procfs_reader.go
sthaha: Indeed there is a chance .. I can compare it to zero value struct.

internal/resource/procfs_reader.go
sthaha: no need to call it .. :D .. what trying something here :)

internal/resource/procfs_reader.go
sthaha: @vimalk78 isn't this better 

```
// Get aggregate CPU stats (sum of all CPUs)
	cpu1 := stat1.CPUTotal
	cpu2 := stat2.CPUTotal

	// Calculate differences for each field
	user := cpu2.User - cpu1.User
	nice := cpu2.Nice - cpu1.Nice
	system := cpu2.System - cpu1.System
	idle := cpu2.Idle - cpu1.Idle
	iowait := cpu2.IOWait - cpu1.IOWait
	irq := cpu2.IRQ - cpu1.IRQ
	softirq := cpu2.SoftIRQ - cpu1.SoftIRQ
	steal := cpu2.Steal - cpu1.Steal

	// Total jiffies is the sum of all fields
	total := user + nice + system + idle + iowait + irq + softirq + steal

	// Avoid division by zero
	if total == 0 {
		return CPUUsage{UsagePercent: 0.0, IdlePercent: 100.0}, nil
	}

	// Calculate percentages
	usagePercent := ((total - idle) / total) * 100  // 👈 note only idle is removed
	idlePercent := (idle / total) * 100
```

internal/exporter/prometheus/collector/power_collector.go
sthaha: IMHO, cpu usage implies what's being used; i.e. active. since there is no other kind, lets call it cpu_usage_ratio. But I will add in description that its a value between 0.0 to 1.0

internal/exporter/prometheus/collector/power_collector.go
sthaha: not sure if we should round lets wait for a reason to round it in a prometheus metrics. Is there a precedence to this? 

Also now the the metric description clarifies it's "between 0.0 and 1.0", and users can apply rounding in queries if their specific use case requires it. This follows the Prometheus principle of providing raw, accurate data and letting users transform it as needed.

internal/resource/procfs_reader.go
vimalk78: deduct iowait also from usage?

internal/exporter/prometheus/collector/power_collector.go
vimalk78: `fmt`  package can restrict  `0.3333333333333333` to 6 decimal places.

sthaha: I agree doing this on stdout exporter but not for prometheus since it is supposed be a raw value.

internal/resource/procfs_reader.go
```diff
@@ -90,11 +90,54 @@ func WrapProc(proc procfs.Proc) procInfo {
...
 // procFSReader is the default implementation of ProcReader using procfs
 type procFSReader struct {
-	fs procfs.FS
+	fs       procfs.FS
+	prevStat procfs.CPUStat
+}
+
+// CPUUsageRatio returns the CPU usage ratio as
+// active over total, where active = total - idle
+// and total = user + nice + system + idle + iowait + irq + softirq + steal
```
vimalk78, line 106: update comment to say `active = total - (idle + iowait)`

internal/resource/informer.go
```diff
@@ -14,32 +14,37 @@ import (
 	"k8s.io/utils/clock"
 )
 
+type Node struct {
+	CPUTimeDelta  float64
```
vimalk78, line 18: nit: this field contains `sum(proc.CPUTimeDelta)` , not related to node

vimalk78 (approved):

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -160,23 +202,57 @@ func (c *PowerCollector) collectNodeMetrics(ch chan<- prometheus.Metric, node *m
...
+		)
+
+		ch <- prometheus.MustNewConstMetric(
+			c.nodeCPUIdleJoulesDesc,
+			prometheus.CounterValue,
+			energy.IdleEnergy.Joules(),
+			zoneName, path,
+		)
+
+		// watts
```
vimalk78, line 236: the node active/idle joules goes up and down, it is more like a gauge than a counter
image

sthaha: what are you plotting? rate ? 

You are right ... I am not incrementing (prev + current) the counter and only calculating the irate.

vimalk78: no, not with rate

sthaha:
@vimalk78 I am merging this so that pod can be rebased on this. I will continue to fix the incorrect implementation of ActiveEnergy counter as a follow up commit.

vprashar2929:
also:
Screenshot 2025-06-07 at 5 05 44 PM

PR #2127, opened 2025-06-03
feat(pod): added pod power

vimalk78:
feat(pod): added pod power

   - podInformer for fetching and caching pods
   - pod energy/power calculation
   - power collector shows pod energy/power
   - tests


config/config.go
```diff
@@ -114,6 +122,11 @@ const (
 	// NOTE: not a flag
 	ExporterPrometheusDebugCollectors = "exporter.prometheus.debug-collectors"
 
+	// kubernetes flags
+	KubernetesFlag = "kube.enable"
+	KubeConfigFlag = "kube.config"
+	NodeNameFlag   = "kube.nodeName"
```
sthaha:
```suggestion
	NodeNameFlag   = "kube.node-name"
```
Lower case for all flags

config/config.go
```diff
@@ -71,6 +71,12 @@ type (
 		Pprof PprofDebug `yaml:"pprof"`
 	}
 
+	Kube struct {
+		Enabled bool   `yaml:"enabled"`
```
sthaha:
```suggestion
		Enabled *bool   `yaml:"enabled"`
```

sthaha: We need all bool to be *bool for config merging to work

config/config.go
```diff
@@ -351,6 +384,14 @@ func (c *Config) Validate(skips ...SkipValidation) error {
 			errs = append(errs, fmt.Sprintf("invalid monitor staleness: %s can't be negative", c.Monitor.Staleness))
 		}
 	}
+	{ // Kubenetes
+		if c.Kube.Config != "" && !c.Kube.Enabled {
+			errs = append(errs, fmt.Sprintf("kube.config supplied but kube.enables set to false"))
```
sthaha: kube.enable .. May be use the Flag const

config/config.go
```diff
@@ -351,6 +384,14 @@ func (c *Config) Validate(skips ...SkipValidation) error {
 			errs = append(errs, fmt.Sprintf("invalid monitor staleness: %s can't be negative", c.Monitor.Staleness))
 		}
 	}
+	{ // Kubenetes
+		if c.Kube.Config != "" && !c.Kube.Enabled {
```
sthaha: We also need to validate if the kube.config file is readable and fail if it isn't. 
We also should allow this check to be skipped (like other host specific tests)

config/config.go
```diff
@@ -351,6 +384,14 @@ func (c *Config) Validate(skips ...SkipValidation) error {
...
+			errs = append(errs, fmt.Sprintf("kube.config supplied but kube.enables set to false"))
+		}
+		if c.Kube.Node != "" && !c.Kube.Enabled {
+			errs = append(errs, fmt.Sprintf("kube.nodeName supplied but kube.enables set to false"))
```
sthaha: `kube.node-name` and  `kube.enable` lets use the flag consts

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("controller-runtime could not create manager: %w", err)
+	}
+	pi.logger.Debug("setupManager with cache")
+	// Add an index on containerID
+	err = mgr.GetCache().IndexField(
+		context.Background(),
+		&corev1.Pod{},    // cache object
+		IndexContainerID, // index Name
+		func(obj client.Object) []string { // keys
```
sthaha: could we refactor this to a different function?

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,280 @@
...
+			pod := obj.(*corev1.Pod)
```
sthaha: Is it guaranteed that the conversion will always succeed? E.g should we 

```
 pod, ok := obj.(*corev1.Pod)
```

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,287 @@
...
+	ctrl.RegisterFlags(ctrlConfigFlagSet)
+	ctrlConfigFlagSet.Set(KubeconfigFlagName, kubeconfig)
+	return nil
+}
+
+func (i *podInformer) Name() string {
+	return "podInformer"
+}
+
+func (i *podInformer) Shutdown() error {
```
sthaha: We can remove this function.

internal/k8s/pod/pod_test.go
```diff
@@ -0,0 +1,22 @@
+package pod
```
sthaha, line 4: license :(

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,287 @@
...
+		&pods,
+		client.MatchingFields{IndexContainerID: containerID},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving pod info from cache: %w", err)
+	}
+
+	if len(pods.Items) == 0 {
+		// pi.logger.Debug("no pod found for container", "container", containerID)
+		return nil, ErrNotFound
```
sthaha: should this be an error? there can we container without pods associated .

internal/resource/informer.go
```diff
@@ -185,7 +210,33 @@ func (ri *resourceInformer) Refresh() error {
...
 
 	}
-	// Node
+
+	containersNoPod := []string{}
+	if ri.podInformer != nil {
+		for _, container := range containersRunning {
+			podInfo, err := ri.podInformer.PodInfo(container.ID)
+			if err != nil {
+				if err == pod.ErrNotFound {
```
sthaha: errors.Is(err, pod.ErrNotFound)

internal/resource/informer.go
```diff
@@ -248,6 +318,9 @@ func (ri *resourceInformer) Refresh() error {
 		"container.terminated", len(containersTerminated),
 		"vm.running", len(vmsRunning),
 		"vm.terminated", len(vmsTerminated),
+		"pods.running", len(podsRunning),
+		"pods.terminated", len(podsTerminated),
+		"containers.nopod", len(containersNoPod),
```
sthaha:
```suggestion
		"pod.running", len(podsRunning),
		"pod.terminated", len(podsTerminated),
		"container.no-pod", len(containersNoPod),
```

sthaha: to match the existing convention (or please feel free change all to plural)

internal/resource/informer.go
```diff
@@ -237,6 +290,23 @@ func (ri *resourceInformer) Refresh() error {
...
+			continue
+		}
+		podsTerminated[id] = pod
+		delete(ri.podCache, id)
+	}
+
+	ri.pods.NodeCPUTimeDelta = nodeCPUDelta
+	ri.pods.Running = podsRunning
+	ri.pods.Terminated = podsTerminated
+	ri.pods.ContainersNoPod = containersNoPod
```
sthaha, line 306: We shouldn't have to provide this information. since `contianer.pod_id == ""` already exists to find containers without a pod associated.

sthaha (changes requested):
Most of the implementation looks good to me.
Few comments though

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+	}
+
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme: pi.scheme,
+		Cache:  cacheOp,
+		// TODO: fix this
+		Metrics: server.Options{
+			BindAddress: "0.0.0.0:8081",
+		},
```
vimalk78: @sthaha any comments here? what should we have here? or just remove this part?

vimalk78: removed it. works fine.

config/config.go
vimalk78: updated

config/config.go
vimalk78: updated

config/config.go
vimalk78: updated

config/config.go
vimalk78: the definition of `kube.config` flag takes care of file existing and readability.

config/config.go
vimalk78: updated

internal/k8s/pod/pod.go
vimalk78: updated

internal/k8s/pod/pod.go
vimalk78: it should be guaranteed, but updated

internal/k8s/pod/pod.go
vimalk78: have added code to cancel the context which is used to start the manager. this clears up resources on api server

internal/k8s/pod/pod_test.go
vimalk78: updated

internal/k8s/pod/pod.go
vimalk78: there can be pods with no containers, but this is used to signal pod not found condition.
the case of pod with no containers, we will not be able to query it anyway.

if still have more comments, can fix in followup PR

internal/resource/informer.go
vimalk78: updated

internal/resource/informer.go
vimalk78: updated

internal/resource/informer.go
vimalk78: this is for quick view in logs.

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,303 @@
...
+		ID        string
+		Name      string
+		Namespace string
+	}
+)
+
+type podInformer struct {
+	logger *slog.Logger
+
+	kubeEnabled    bool
```
sthaha: We shouldn't need this in this struct since the decision to initialize podInformer or not should be handled in main - right 🤔 ?

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,303 @@
...
+func extractContainerID(str string) string {
+	parts := strings.SplitN(str, "://", 2)
+	return parts[len(parts)-1]
+}
+
+func (pi *podInformer) Run(ctx context.Context) error {
+	pi.logger.Info("Starting pod informer")
+	mgrCtx, cancel := context.WithCancel(context.Background())
+	pi.mgrCtx = mgrCtx
+	pi.mgrCancelFunc = cancel
```
sthaha: we shouldn't need this. You should be able to pass the ctx to Start. The service shutdown mechanism already calls cancel.

internal/k8s/pod/pod.go
sthaha: Sorry, should have asked properly. 

What I meant was that it isn't an error for pod to not have a container with a given id. E.g. if there are 2 pods and   - p1 running c1 and p2 running c2, and a static container X1 without any pods associated, then PodInfo(X1) should just return nil, without an error - right?  since it isn't an error condition.

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func (i *podInformer) Name() string {
+	return "podInformer"
+}
+
+func (pi *podInformer) Shutdown() error {
+	pi.logger.Info("stopping pod informer manager")
+	pi.mgrCancelFunc()
+	return nil
```
sthaha: we shouldn't require this since cancel will be called by main.

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to add an index on containerID: %w", err)
+	}
+	return mgr, nil
+}
+
+func (pi *podInformer) indexerFunc(obj client.Object) []string {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return []string{"invalidContainerID"}
```
sthaha: shouldn't we return "" instead?

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,303 @@
...
+	return reconcile.Result{}, nil
+}
+
+func createController(mgr manager.Manager) error {
+	d := &dummyReconciler{}
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithEventFilter(predicate.ResourceVersionChangedPredicate{}).
+		Build(d)
+	return err
```
sthaha: we shouldn't need the reconciler since manager's cache is all that we need - right?

sthaha (changes requested):

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -95,6 +99,7 @@ func NewPowerCollector(monitor PowerDataProvider, logger *slog.Logger) *PowerCol
 		zone   = "zone"
 		cntrID = "container_id"
 		vmID   = "vm_id"
+		podID  = "pod_id"
```
sthaha, line 102: Just noticed that we need `pod_id` in `kepler_container_.*` as well

internal/k8s/pod/pod.go
vimalk78: so user has to do a nil check, instead of `ErrNotFound` which has string "no pod found for container"
ok.

sthaha: That works .. but imho, we can return a bool (PodInfo, found bool, err)

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -195,6 +207,7 @@ func (c *PowerCollector) Collect(ch chan<- prometheus.Metric) {
 	c.collectProcessMetrics(ch, snapshot.Processes)
 	c.collectContainerMetrics(ch, snapshot.Containers)
```
sthaha, line 208: Could you please add pod_id to the containers so that we can "join" containers and pods

internal/k8s/pod/pod.go
```diff
@@ -0,0 +1,254 @@
...
+	pi.cfg = cfg
+
+	mgr, err := pi.setupManager(scheme)
+	if err != nil {
+		return fmt.Errorf("controller-runtime could not create manager: %w", err)
+	}
+	pi.manager = mgr
+
+	opts := zap.Options{
+		Development: true, // enables DebugLevel by default
```
sthaha, line 135: nit: can we set this to true if current pi.logger. log-level == Debug?

internal/monitor/monitor.go
```diff
@@ -311,6 +318,11 @@ func (pm *PowerMonitor) firstReading(newSnapshot *Snapshot) error {
 		return fmt.Errorf(vmPowerError, err)
 	}
 
+	// First read for pods
+	if err := pm.firstPodRead(newSnapshot); err != nil {
+		return fmt.Errorf(containerPowerError, err)
```
sthaha, line 323:
```suggestion
		return fmt.Errorf(podPowerError, err)
```

internal/monitor/monitor_collection_test.go
```diff
@@ -30,6 +30,7 @@ func TestCollectionLoop(t *testing.T) {
 	resourceInformer := &MockResourceInformer{}
 	resourceInformer.SetExpectations(t, tr)
 	resourceInformer.On("Refresh").Return(nil)
+	// resourceInformer.On("Pods").Return(nil)
```
sthaha, line 33: remove?

internal/resource/informer.go
```diff
@@ -185,7 +209,33 @@ func (ri *resourceInformer) Refresh() error {
 		}
 
 	}
-	// Node
+
+	containersNoPod := []string{}
+	if ri.podInformer != nil {
+		for _, container := range containersRunning {
```
sthaha, line 215: missing test? In a follow up PR,  could you please add test to test this functionality?

internal/resource/informer.go
```diff
@@ -185,7 +209,33 @@ func (ri *resourceInformer) Refresh() error {
...
+				Name:      podInfo.Name,
+				Namespace: podInfo.Namespace,
+			}
+			container.Pod = pod
+			_, seen := podsRunning[pod.ID]
+			// reset CPU Time of the pod if it is getting added to the running list for the first time
+			// in the subsequent iteration, the CPUTimeDelta should be incremented by container's CPUTimeDelta
+			resetCPUTime := !seen
+			podsRunning[pod.ID] = ri.updatePodCache(container, resetCPUTime)
+		}
```
sthaha, line 237: missing tests. Could you please add tests in a follow up PR?

internal/resource/informer.go
```diff
@@ -237,6 +289,22 @@ func (ri *resourceInformer) Refresh() error {
 	ri.vms.Running = vmsRunning
 	ri.vms.Terminated = vmsTerminated
 
+	// Find terminated pods
+	totalPodsDelta := float64(0)
```
sthaha, line 293: we don't need to find totalPodsDelta, do we?

internal/resource/types.go
```diff
@@ -100,3 +102,24 @@ func (vm *VirtualMachine) Clone() *VirtualMachine {
...
 }
+
+type Pod struct {
+	ID        string
+	Name      string
+	Namespace string
+
+	// Resource usage tracking
+	CPUTotalTime float64 // total cpu time used by the VM so far
+	CPUTimeDelta float64 // cpu time used by the VM since last refresh
```
sthaha, line 113:
```suggestion
	CPUTimeDelta float64 // cpu time used by the Pod since last refresh
```

internal/resource/types.go
```diff
@@ -100,3 +102,24 @@ func (vm *VirtualMachine) Clone() *VirtualMachine {
...
 	}
 }
+
+type Pod struct {
+	ID        string
+	Name      string
+	Namespace string
+
+	// Resource usage tracking
+	CPUTotalTime float64 // total cpu time used by the VM so far
```
sthaha, line 112:
```suggestion
	CPUTotalTime float64 // total cpu time used by the Pod so far
```

PR #2128, opened 2025-06-04
ci: add workflow to run kepler on k8s

vprashar2929:
This commit adds a new workflow which builds and deploys Kepler on Kind cluster.
This workflow mainly does the following:
1. Installs required dependencies like Docker, Kind, Kubectl on to the runner.
2. Builds Kepler and pushes to local registry.
3. Deploys Kepler on Kind cluster.
4. Validate the metrics. (Will be removing this once we add proper validations for Kepler.)

### DIFF
--- a/.github/workflows/k8s.yaml
+++ b/.github/workflows/k8s.yaml
@@ -1,0 +1,145 @@
+name: Build and Deploy on K8s
+
+on: #yamllint disable-line rule:truthy
+  pull_request:
+    branches: [reboot]
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Docker
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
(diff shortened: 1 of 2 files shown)


KaiyiLiu1234 (approved):
lgtm

.github/workflows/k8s.yaml
```diff
@@ -0,0 +1,145 @@
...
+        id: validate
+        shell: bash
+        run: |
+          kubectl port-forward service/kepler 28282:28282 -n kepler &
+          sleep 20 # sleep for 20 seconds to give the service time to start
+
+          HTTP_STATUS=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:28282/metrics)
+          [[ $HTTP_STATUS -ne 200 ]] && echo "HTTP status code is not 200" && exit 1
+
+          curl -s -o /tmp/metrics.txt http://localhost:28282/metrics
```
sthaha, line 113: Could you also please check that there are no `An error has occurred` ?

sthaha (approved):

.github/workflows/k8s.yaml
vprashar2929: Sure will add this as well

PR #2133, opened 2025-06-08
docs: add tool for generating docs for metrics

vprashar2929:
This commit adds a tool for generating docs for metrics.

The tool extracts the metrics information from the various prometheus collectors and generates a markdown file with the metrics information.

This is useful for generating the metrics documentation for the kepler


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: all build build-debug clean test coverage lint vet fmt deps image push run
+.PHONY: all build build-debug clean test coverage lint vet fmt deps image push run deploy undeploy cluster-restart cluster-down cluster-up gen-metrics-docs
```
sthaha, line 1: Instead of this can we please have multiple .PHONY close to the target so that when you are deleting a target, the .PHONY gets removed as well ?  -- In a separate PR

hack/gen-metric-docs/main.go
```diff
@@ -0,0 +1,295 @@
+// SPDX-FileCopyrightText: 2025 The Kepler Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
```
sthaha, line 4: Very nice! 
Could we write tests as well (in a separate PR) :) so that when we make a change, we don't break doc generation by mistake?

sthaha (approved):
@vprashar2929 Good idea!! 

Can we please add a CI test to ensure that the doc is up to date?

Makefile
vprashar2929: Agree

hack/gen-metric-docs/main.go
vprashar2929: yes will do that

PR #2135, opened 2025-06-09
chore: add .PHONY declarations near to targets

vprashar2929:
This commit adds .PHONY declarations near to targets in the Makefile for better readability and clarity.


Makefile
```diff
@@ -59,13 +57,15 @@ COVER_HTML=coverage.html
 all: clean fmt lint vet build test
```
sthaha, line 57: .PHONY: all

sthaha (approved):

PR #2136, opened 2025-06-09
ci: add missing check for file changes in equinix workflow

vprashar2929:
This commit adds a missing check for file change in one of the jobs in the equinix workflow.

### DIFF
--- a/.github/workflows/k8s-equinix.yaml
+++ b/.github/workflows/k8s-equinix.yaml
@@ -50,7 +50,8 @@ jobs:
           os_name: ubuntu_22_04
 
   build-and-deploy:
-    needs: create-runner
+    needs: [check-changes, create-runner]
+    if: needs.check-changes.outputs.changes == 'true'
     runs-on: self-hosted
     env:
       GOCACHE: /home/ghrunner/.cache/go-build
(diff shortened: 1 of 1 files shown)


.github/workflows/k8s-equinix.yaml
```diff
@@ -197,9 +198,9 @@ jobs:
           echo "::endgroup::"
 
   cleanup-runner:
-    needs: build-and-deploy
+    needs: [check-changes, build-and-deploy]
     runs-on: ubuntu-latest
-    if: always() # Run even if the previous job fails
+    if: always() && needs.check-changes.outputs.changes == 'true' # Run even if the previous job fails
```
sthaha, line 203: what does it mean to run `always` and changes are detected? 
Shouldn't it be just `when changes are detected`?

vprashar2929: run always ensure that if build-and-deploy job fails for any reason cleanup job will be executed ensuring the equinix machine gets destroyed.
We also need check changes as only specifying always() will always trigger this job anyhow due to its nature. For eg: https://github.com/sustainable-computing-io/kepler/actions/runs/15529582815/job/43770258237?pr=2136

So checking both condition ensures that we don't run the job all the time but it also does cleanup if something goes wrong in any of the previous job runs

sthaha: isn't if:  needs.check-changes.outputs.changes == 'true'  sufficient?

sthaha: Also see failure: https://github.com/sustainable-computing-io/kepler/actions/runs/15529582815/job/43770258237?pr=2136#step:3:150

vprashar2929: yes it will fail, as for the change to be in effect, this needs to be merged to the reboot branch 😅

vprashar2929: `pull_request_target` event picks up the workflow yaml from the target branch i.e reboot branch and execute it

PR #2141, opened 2025-06-11
fix(metric): add missing pod_id to container metrics

sthaha:


vimalk78:
this is Work In Progress. will send a separate PR

Issue #2144, opened 2025-06-11
Kepler reboot TestCollectionCancellation fails on CI


vprashar2929:
Ref: https://github.com/sustainable-computing-io/kepler/actions/runs/15579633102/job/43871625255#step:4:736

vprashar2929:
closing as #2146 addresses this

PR #2146, opened 2025-06-12
fix(monitor): race condition between timer and context cancelation

sthaha:
This commit fixes a race condition when timer fires ⚡️ and context gets canceled 💥 at the same time. This race condition was detected the the TestCollectionCancellation and happens as follows ...

```
monitor_collection_test.go:201:
Error Trace:    .../monitor_collection_test.go:201
Error:          Not equal:
                expected: time.Date(2025, time.June, 11, 8, 14, 45, 720156604, time.Local)
                actual  : time.Date(2025, time.June, 11, 8, 14, 45, 732389967, time.Local)
Test:           TestCollectionCancellation
Messages:       No collections should happen after cancellation
```

Race Sequence
--------------

1. **Test Context**: `TestCollectionCancellation` runs with a 100ms timeout context
2. **Timer Setup**: Collection goroutines are scheduled with 10ms intervals
3. **Cancellation**: After ~50ms, test calls `cancel()` which triggers `pm.collectionCancel()`
4. **Run() Exit**: The `Run()` method exits and test captures `snapshotAfterRun.Timestamp`
5. **Race Window**: 💥 There's a brief window where:
   - Context is canceled (`pm.collectionCtx.Done()` becomes ready)
   - Timer also fires (`<-timer` becomes ready)
   - Go's `select` **randomly** chooses between the two ready channels

6. **Failure Case**: When `select` chooses `<-timer`:
   - `synchronizedPowerRefresh()` executes and updates snapshot with new timestamp
   - Test assertion fails because timestamps don't match

Fixes https://github.com/sustainable-computing-io/kepler/issues/2144

### DIFF
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -175,13 +175,20 @@ func (pm *PowerMonitor) scheduleNextCollection() {
 	go func() {
 		select {
 		case <-timer:
+			// Check if context is cancelled before doing any work to avoid a race condition
+			// where the context is cancelled after the timer has expired
+			if err := pm.collectionCtx.Err(); err != nil {
+				pm.logger.Info("Collection loop terminated; context canceled", "reason", err)
+				return
+			}
+
 			if err := pm.synchronizedPowerRefresh(); err != nil {
 				pm.logger.Error("Failed to collect power data", "error", err)
 			}
 			pm.scheduleNextCollection()
 
 		case <-pm.collectionCtx.Done():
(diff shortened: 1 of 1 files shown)


vprashar2929 (approved):

vimalk78 (approved):

vimalk78:
suggestion:
print `pm.collectionCtx.Err()`  in log
```
		case <-pm.collectionCtx.Done():
			pm.logger.Info("Collection loop terminated")
			return
		}
```

PR #2150, opened 2025-06-12
fix(exporter): move WithNodeName option to collector

vimalk78:
Moved WithNodeName() option from prometheus exporter to collector
 added  tests

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -197,6 +197,7 @@ func createPrometheusExporter(logger *slog.Logger, cfg *config.Config, apiServer
 		pm,
 		prometheus.WithLogger(logger),
 		prometheus.WithProcFSPath(cfg.Host.ProcFS),
+		prometheus.WithNodeName(cfg.Kube.Node),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create Prometheus collectors: %w", err)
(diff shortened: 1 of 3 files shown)


internal/exporter/prometheus/collector/power_collector_test.go
```diff
@@ -309,6 +326,86 @@ func TestPowerCollector(t *testing.T) {
...
+					nodeName := valueOfLabel(m, "node_name")
+					assert.Equal(t, "test-node", nodeName, "Process metrics should have node_name label")
+				}
+			}
+			if metric.GetName() == "kepler_process_cpu_watts" {
+				for _, m := range metric.GetMetric() {
+					nodeName := valueOfLabel(m, "node_name")
+					assert.Equal(t, "test-node", nodeName, "Process metrics should have node_name label")
+				}
+			}
```
sthaha, line 345: suggestion: refactor this into a helper function that assert that `assertMetricLabelValue(t, name, label, value)` we have something similar called `asserrtMainMetricValue` which I think can be replaced as well.

cmd/kepler/main.go
```diff
@@ -197,6 +197,7 @@ func createPrometheusExporter(logger *slog.Logger, cfg *config.Config, apiServer
 		pm,
 		prometheus.WithLogger(logger),
 		prometheus.WithProcFSPath(cfg.Host.ProcFS),
+		prometheus.WithNodeName(cfg.Kube.Node),
```
sthaha, line 200: nice!

sthaha (commented):
lgtm! but ci :( .. some suggestions though

sthaha (approved):

PR #2152, opened 2025-06-13
ci: trigger equinix workflow only on source code changes

vprashar2929:
This commit ensures that the check changes job in the equinix workflow only look for the changes in the actual source code. i.e excluding any go code present in `hack/` folder


vprashar2929:
docs job will failure is addressed in #2151

sthaha (approved):

PR #2153, opened 2025-06-13
ci: ensure pod level metrics are exported

vprashar2929:
This commit ensures that pod level metrics are exported when Kepler is deployed on Kind cluster.


sthaha:
@vprashar2929 lets ensure that the workflow is run when a workflow is modified.

vprashar2929:
> lets ensure that the workflow is run when a workflow is modified.

You mean equinix workflow?

sthaha (approved):

PR #2156, opened 2025-06-16
feat(monitor): track terminated processes

sthaha:
Previously, between each call to `Snapshot()` (usually from Prometheus exporter),
processes that terminate go unaccounted for in power consumption metrics.
This creates a data loss problem where workloads that consumed significant power
but are no longer running will not be returned to Prometheus.

**Scenario Example:**
   - Prometheus scrapes every 30 seconds
   - Kepler monitor computes power every 5 seconds (default interval)
   - Between first and second Prometheus scrape, monitor computes 5-6 times
   - Each computation overwrites the snapshot with only currently running processes
   - Terminated processes that consumed power during those 25-30 seconds are lost

Changes include:
   - Add terminated process tracking to power monitor
   - Extend power collector to export terminated process metrics
   - Include comprehensive test coverage for terminated process scenarios
   - Enhance process lifecycle management for accurate power attribution

### DIFF
--- a/internal/exporter/prometheus/collector/power_collector.go
+++ b/internal/exporter/prometheus/collector/power_collector.go
@@ -122,8 +122,8 @@ func NewPowerCollector(monitor PowerDataProvider, nodeName string, logger *slog.
 			"CPU usage ratio of a node (value between 0.0 and 1.0)",
 			nil, prometheus.Labels{nodeNameLabel: nodeName}),
 
-		processCPUJoulesDescriptor: joulesDesc("process", "cpu", nodeName, []string{"pid", "comm", "exe", "type", cntrID, vmID, zone}),
-		processCPUWattsDescriptor:  wattsDesc("process", "cpu", nodeName, []string{"pid", "comm", "exe", "type", cntrID, vmID, zone}),
+		processCPUJoulesDescriptor: joulesDesc("process", "cpu", nodeName, []string{"pid", "comm", "exe", "type", "state", cntrID, vmID, zone}),
+		processCPUWattsDescriptor:  wattsDesc("process", "cpu", nodeName, []string{"pid", "comm", "exe", "type", "state", cntrID, vmID, zone}),
 		processCPUTimeDescriptor:   timeDesc("process", "cpu", nodeName, []string{"pid", "comm", "exe", "type", cntrID, vmID}),
 
 		containerCPUJoulesDescriptor: joulesDesc("container", "cpu", nodeName, []string{cntrID, "container_name", "runtime", zone, podID}),
@@ -206,7 +206,9 @@ func (c *PowerCollector) Collect(ch chan<- prometheus.Metric) {
 	}
 
 	c.collectNodeMetrics(ch, snapshot.Node)
-	c.collectProcessMetrics(ch, snapshot.Processes)
+	c.collectProcessMetrics(ch, "running", snapshot.Processes)
+	c.collectProcessMetrics(ch, "terminated", snapshot.TerminatedProcesses)
+
 	c.collectContainerMetrics(ch, snapshot.Containers)
 	c.collectVMMetrics(ch, snapshot.VirtualMachines)
 	c.collectPodMetrics(ch, snapshot.Pods)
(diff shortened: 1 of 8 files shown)


internal/exporter/prometheus/collector/power_collector_test.go
```diff
@@ -64,20 +64,74 @@ func (m *MockPowerMonitor) TriggerUpdate() {
 	}
 }
 
-func assertMetricLabelValues(t *testing.T, registry *prometheus.Registry, metricName string, expectedLabels map[string]string) {
+// assertMetricLabelValues finds a specific metric by matching all provided labels and checks its value
+func assertMetricLabelValues(t *testing.T, registry *prometheus.Registry, metricName string, matchLabels map[string]string, expectedValue float64) {
```
sthaha, line 70: @vimalk78

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -272,7 +274,7 @@ func (c *PowerCollector) collectNodeMetrics(ch chan<- prometheus.Metric, node *m
 }
 
 // collectProcessMetrics collects process-level power metrics
-func (c *PowerCollector) collectProcessMetrics(ch chan<- prometheus.Metric, processes monitor.Processes) {
+func (c *PowerCollector) collectProcessMetrics(ch chan<- prometheus.Metric, state string, processes monitor.Processes) {
 	if len(processes) == 0 {
 		c.logger.Debug("No processes to export metrics for")
```
vimalk78: change log to print state

vimalk78 (dismissed):

internal/monitor/monitor.go
```diff
@@ -278,6 +292,9 @@ func (pm *PowerMonitor) refreshSnapshot() error {
 		}
 	}
 
+	// Reset exported to keep track of terminated processes until Snapshot is exported
+	pm.exported.Store(false)
```
vimalk78, line 296: nit: set the atomic variable to false only if it was true previousely
```
pm.exported.CompareAndSwap(true, false)
```

also add a debug statement when it is reset (set to false)

internal/monitor/process.go
```diff
@@ -111,9 +135,36 @@ func (pm *PowerMonitor) calculateProcessPower(prev, newSnapshot *Snapshot) error
...
-	// Update the snapshot
+	// Update the snapshot of running processes
 	newSnapshot.Processes = processMap
-	pm.logger.Debug("snapshot updated for process", "process", len(newSnapshot.Processes))
+
+	// Copy existing terminated processes from previous snapshot if not exported
+	if !pm.exported.Load() {
+		// NOTE: no need to deep clone since already terminated processes won't be updated
+		maps.Copy(newSnapshot.TerminatedProcesses, prev.TerminatedProcesses)
+	}
```
vimalk78, line 145: if Snapshot() is not called, the processes could start accumulating here. can we keep a maximum limit?

sthaha: Yes, I thought of the same but with maxRetention, I think it is better to implement a PQ with a fixed size. (In a different PR) ?

internal/monitor/monitor.go
sthaha: Good find!

sthaha: Just tried it out ... 

cpu: AMD Ryzen 7 PRO 7840HS w/ Radeon 780M Graphics
BenchmarkStore-16               371794159                3.227 ns/op
BenchmarkCompareAndSwap-16      345375631                3.473 ns/op

Store is better in all runs.

sthaha:
```go
package main

import (
        "sync/atomic"
        "testing"
)

func BenchmarkStore(b *testing.B) {
        var flag atomic.Bool
        flag.Store(true) // Ensure initial state
        b.ResetTimer()
        for b.Loop() {
                flag.Store(false)
                flag.Store(true)
        }
}

func BenchmarkCompareAndSwap(b *testing.B) {
        var flag atomic.Bool
        flag.Store(true) // Ensure initial state
        b.ResetTimer()
        for b.Loop() {
                flag.CompareAndSwap(true, false)
                flag.CompareAndSwap(false, true)
        }
}

```

vimalk78 (approved):

PR #2161, opened 2025-06-17
chore: Enable TLS in server creation process

SamYuan1990:
I just checked the code in config/config_test.go and internal/server/server_tls_test.go I found that for kepler itself, we enabled TLS config in config structure but not add the invoke for supporting enable TLS in server creation. With this PR, patch this missing part of code.


SamYuan1990:
I hope to enable TLS and test in this PR, as build for test case and apply https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md in next PR.

sthaha (approved):

PR #2165, opened 2025-06-18
[feat] using exporter-toolkit/web for web config

SamYuan1990:
Let's use this PR to discuss how we integrate with exporter-toolkit/web.
From exporter-toolkit/web SDK, we need `*web.FlagConfig` structure ready before we start the server.
but the question here is how we make it in our configuration and code structure.
we now have config in yaml file and those flag reg in same pattern, if we fully moved to exporter-toolkit/web as one example impl in this PR shows.
We still need all config in one place, so that it will not impact main.go, but somehow breaks config.go in yaml and flag reg process.
The benefits here IMO, is that we can leave web config to upstream repo for feature supports and it seems for an exporter we don't need to support listen on multi socket, so... just a default port and default TLS config file will be good enough.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,8 @@ import (
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
+	"github.com/prometheus/exporter-toolkit/web"
+	"github.com/prometheus/exporter-toolkit/web/kingpinflag"
 	"gopkg.in/yaml.v3"
 	"k8s.io/utils/ptr"
 )
@@ -39,7 +41,7 @@ type (
 		} `yaml:"fake-cpu-meter"`
 	}
 	Web struct {
-		Config string `yaml:"configFile"`
+		Config *web.FlagConfig
 	}
 
 	Monitor struct {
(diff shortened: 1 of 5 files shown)


SamYuan1990:
@sthaha and @vimalk78 may I have your feedback on this PR? I know there some test case need to be fix, but maybe we need to vote for using exporter-toolkit/web for web config or not and make a design for how integration it.

sthaha:
> @sthaha and @vimalk78 may I have your feedback on this PR? I know there some test case need to be fix, but maybe we need to vote for using exporter-toolkit/web for web config or not and make a design for how integration it.

I like the idea of embedding webconfig into the config. It is one less configuration to worry about. Let me play around it a bit and get back to you.

---

Just a nit: ... This PR will be a `feat` and not a `chore` https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13#types

SamYuan1990:
> > @sthaha and @vimalk78 may I have your feedback on this PR? I know there some test case need to be fix, but maybe we need to vote for using exporter-toolkit/web for web config or not and make a design for how integration it.
> 
> I like the idea of embedding webconfig into the config. It is one less configuration to worry about. Let me play around it a bit and get back to you.
> 
> Just a nit: ... This PR will be a `feat` and not a `chore` https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13#types

I just make some update in code and fix with test cases.

PR #2170, opened 2025-06-19
refactor(resource): split Refresh into smaller functions with parallel execution

sthaha:
This change breaks down the Refresh method into smaller functions following Single Responsibility Principle. Changes are ..

- refreshProcesses: categorizes processes during initial scan
- refreshContainers/refreshVMs: build workload-specific collections
- refreshPods: handles container-to-pod mapping (depends on containers)
- refreshNode: calculates node-level aggregations

Note that this change introduces parallel execution using goroutines for independent workloads (containers+pods, VMs, and node calculations) while maintaining proper dependency ordering.

Also adds test coverage for concurrency testing, missing coverage and error handling scenarios.

### DIFF
--- a/internal/resource/informer.go
+++ b/internal/resource/informer.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/sustainable-computing-io/kepler/internal/k8s/pod"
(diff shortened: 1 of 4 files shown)


internal/resource/informer.go
```diff
@@ -194,133 +195,215 @@ func (ri *resourceInformer) Refresh() error {
...
+// Refresh updates the internal state by scanning processes, containers, VMs, and pods.
+// This method is NOT thread-safe and should not be called concurrently.
+func (ri *resourceInformer) Refresh() error {
+	started := ri.clock.Now()
+
+	// Refresh workloads in dependency order:
+	// processes -> {
+	//   -> containers -> pod
+	//   -> VMs
+	//   -> node
```
vimalk78, line 356: nit: isn't node the first ? it should not depend on processes

vimalk78 (approved):

internal/resource/informer.go
sthaha: node.NodeCPUDeltaTotal = Σ process[i].CpuDelta

PR #2172, opened 2025-06-19
chore: switch to main

vprashar2929:
This commit updates the GH workflows, Makefile and docs to now point to the main branch instead of reboot branch


kaiyiliu-strala:
lgtm but what is the purpose of this?

SamYuan1990:
> lgtm but what is the purpose of this?

honestly, I suppose what happened( or will happened) is there no going to have a PR from reboot to main but rename the branch from reboot to main.

SamYuan1990:
well as what was happened already, I will just let it go. 
but when we are going to back to main, I hope we should archive today's CI and double check on what are today's CI features and what we are going to keep and maybe re-impl after switch.
cc: @sthaha , @KaiyiLiu1234 , @sunya-ch

vprashar2929:
@SamYuan1990 We will just switch reboot branch to main hence all the existing workflows (for the reboot branch) will now work as it is for the default branch i.e `main`

SamYuan1990:
> @SamYuan1990 We will just switch reboot branch to main hence all the existing workflows (for the reboot branch) will now work as it is for the default branch i.e `main`

@vprashar2929 , IMO, after make today's reboot becomes main, we still have works on CI. For example:
- [ ] today we follow openSSF as a part of CNCF on boarding requirement. we have sbom and score card actions been invoked.
- [ ] daily build.
- [ ] Depbot.
- [ ] RPM release.

I don't want this switch breaks our commitment once we make it during CNCF on boarding.

sthaha:
@vprashar2929 , I agree with @SamYuan1990  on 
- [ ] SBOM generation / scorecard
- [ ] Daily builds 
- [ ] Dependabot (once a fortnight)

I am not too convinced on `RPM` mostly because of maintenance, but open to that suggestion if users find it useful.

vprashar2929:
💯  Agree, creating issue to track this

sthaha (approved):

PR #2173, opened 2025-06-20
nit: remove double validation on folder when register flag

SamYuan1990:
I suppose we double validate `/sys` or `/proc` folder if we follow the default settings, or if we set up another folder other than `/sys` or `/proc` we still validate those folder during flag reg process and we have another validation function to check if the config is validated. So in this PR, just remove validation in flag reg process.

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -105,6 +105,12 @@ func parseArgsAndConfig() (*config.Config, error) {
 		return nil, err
 	}
 
+	// Validate config before run
+	if err := cfg.Validate(); err != nil {
+		logger.Error("Error validate command line flags", "error", err.Error())
+		return nil, err
+	}
+
 	return cfg, nil
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -186,10 +186,6 @@ func Load(r io.Reader) (*Config, error) {
 	}
 	cfg.sanitize()
 
-	if err := cfg.Validate(); err != nil {
-		return nil, err
-	}
-
 	return cfg, nil
 }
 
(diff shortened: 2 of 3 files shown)


SamYuan1990:
@sthaha , I am confused with the code logic in config pkg, as there many times validation been invoked. I hope we can make it clean as a Validation function in public or a validation function as hook. To follow single response, or KISS.

cmd/kepler/main.go
```diff
@@ -105,6 +105,12 @@ func parseArgsAndConfig() (*config.Config, error) {
 		return nil, err
 	}
 
+	// Validate config before run
+	if err := cfg.Validate(); err != nil {
+		logger.Error("Error validate command line flags", "error", err.Error())
+		return nil, err
+	}
+
```
sthaha, line 113: Lets not make it the responsibility of the caller to validate the config.

config/config.go
```diff
@@ -186,10 +186,6 @@ func Load(r io.Reader) (*Config, error) {
 	}
 	cfg.sanitize()
 
-	if err := cfg.Validate(); err != nil {
-		return nil, err
-	}
-
```
sthaha, line 192: Load should validate

config/config.go
```diff
@@ -237,8 +233,8 @@ func RegisterFlags(app *kingpin.Application) ConfigUpdaterFn {
 	logLevel := app.Flag(LogLevelFlag, "Logging level: debug, info, warn, error").Default("info").Enum("debug", "info", "warn", "error")
 	logFormat := app.Flag(LogFormatFlag, "Logging format: text or json").Default("text").Enum("text", "json")
 	// host
-	hostSysFS := app.Flag(HostSysFSFlag, "Host sysfs path").Default("/sys").ExistingDir()
-	hostProcFS := app.Flag(HostProcFSFlag, "Host procfs path").Default("/proc").ExistingDir()
```
sthaha, line 241: Lets not make it String since we must ensure the that the directory exists

sthaha (changes requested):
Double validation is fine since config can be loaded through yaml which can be the overriden by flags.

SamYuan1990:
> Double validation is fine since config can be loaded through yaml which can be the overriden by flags.

but wait, if `/sys` and `/proc` have validation as those folder must exist, are those configurable? 
or say, do we need those configurable?

For instance, on Mac OS, `/sys` not exists at init configuration phase, `ExistingDir ` been executed and error at here blocked the configuration.
Otherwise, on Linux, `/sys` is a default path and always there, so user don't need to overwrite this config?

sthaha:
> > Double validation is fine since config can be loaded through yaml which can be the overriden by flags.
> 
> but wait, if `/sys` and `/proc` have validation as those folder must exist, are those configurable? or say, do we need those configurable?
> 

I agree that if `--host.sys=/foo/bar` is set, then `/sys` must not be validated. If I understand correctly, it should be the case now but I will check and get back to you.

config/config.go
kaiyiliu-strala: +1

SamYuan1990:
close as POC was done.

PR #2174, opened 2025-06-20
feat(metrics): add configurable metrics level filtering for Prometheus exporter

vimalk78:
- Add configurable metrics level filtering to control which metric types are exported by the Prometheus exporter
  - Implement metrics level configuration via command line flags and YAML config
  - Add support for granular control over node, process, container, VM, and pod metrics
  - Added tests

  Supported metric levels: `node, process, container, vm, pod, all, none`

Usage: 
```
sudo ./bin/kepler  --metrics=node --metrics=pod --kube.enable --kube.node-name="kind-worker2" --kube.config=/home/vimalkum/.kube/config
```

### DIFF
--- a/compose/dev/kepler-dev/etc/kepler/config.yaml
+++ b/compose/dev/kepler-dev/etc/kepler/config.yaml
@@ -36,6 +36,10 @@ exporter:
     debugCollectors:
       - go
       - process
+    metricsLevel:
+      - node
+      - container
+      - process
 
 debug: # debug related config
   pprof: # pprof related config
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
+	"github.com/sustainable-computing-io/kepler/internal/exporter/prometheus/metrics"
 	"gopkg.in/yaml.v3"
 	"k8s.io/utils/ptr"
 )
@@ -53,8 +54,9 @@ type (
 	}
 
 	PrometheusExporter struct {
-		Enabled         *bool    `yaml:"enabled"`
-		DebugCollectors []string `yaml:"debugCollectors"`
+		Enabled         *bool         `yaml:"enabled"`
+		DebugCollectors []string      `yaml:"debugCollectors"`
+		MetricsLevel    metrics.Level `yaml:"metricsLevel"`
 	}
 
 	Exporter struct {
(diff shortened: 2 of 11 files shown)


internal/exporter/prometheus/metrics/level.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// Individual metric levels using bit patterns
+	MetricsLevelNode      Level = 1 << (iota - 1) // 1
+	MetricsLevelProcess                           // 2
+	MetricsLevelContainer                         // 4
+	MetricsLevelVM                                // 8
+	MetricsLevelPod                               // 16
+
+	// MetricsLevelAll represents all metrics enabled
+	MetricsLevelAll Level = MetricsLevelNode | MetricsLevelProcess | MetricsLevelContainer | MetricsLevelVM | MetricsLevelPod
```
sthaha: Lets avoid the none and the all until there is a specific usecase. 
```
metricsLevels: [] | null
```
indicates all 


Is there a specific usecase of none ?

compose/dev/kepler-dev/etc/kepler/config.yaml
```diff
@@ -36,6 +36,11 @@ exporter:
     debugCollectors:
       - go
       - process
+    # metricsLevel: all
+    metricsLevel:
```
sthaha, line 39:
```suggestion
    metricsLevels:
```
to indicate multiple

config/config.go
```diff
@@ -91,6 +93,55 @@ type (
 	}
 )
 
+// MetricsLevelValue is a custom kingpin.Value that parses metrics levels directly into metrics.Level
+type MetricsLevelValue struct {
+	level *metrics.Level
```
sthaha, line 98: should this be pointer? won't value suffice?

config/config.go
```diff
@@ -91,6 +93,55 @@ type (
...
+
+	// Handle special cases
+	if value == "all" {
+		*m.level = metrics.MetricsLevelAll
+		return nil
+	}
+	if value == "none" {
+		*m.level = metrics.MetricsLevelNone
+		return nil
+	}
```
sthaha: Lets avoid the explicit "none" and  and "all" cases until there is a specific use case.

internal/exporter/prometheus/metrics/level.go
sthaha: Second thought, yes, we need none but not `all`

config/config.go
vimalk78: i don't see the need to remove these, the meanings are quite clear.

sthaha: Its not about the meaning but about the purpose ... when would user want to set kepler to not export any metric for instance?

sthaha: the argument against all is similar. By default all metrics are exported, so 
```
metricLevels: [] | null 
``` 
exports all metrics as well as ... 

```
metricLevels: 
 - all
```

the `all` example looks odd IMHO.

vimalk78:
The `MetricsLevelAll` and `MetricsLevelNone` are removed as asked

PR #2176, opened 2025-06-22
 feat(monitor): add terminated workload tracking and reporting

sthaha:
This commit extends the functionality to track terminated process (5e91e0fe) to now track all workloads - Containers, Pods, VMs. Changes include:

- Add terminated process/container/VM/pod tracking to Snapshot struct
- Extend power collector to expose terminated workload metrics
- Update monitor components to track terminated workload energy accumulation
- Add comprehensive test coverage for terminated workload scenarios

### DIFF
--- a/internal/monitor/container.go
+++ b/internal/monitor/container.go
@@ -3,98 +3,128 @@
 
 package monitor
 
+import (
+	"maps"
+
+	"github.com/sustainable-computing-io/kepler/internal/resource"
+)
+
 // firstContainerRead initializes container power data for the first time
 func (pm *PowerMonitor) firstContainerRead(snapshot *Snapshot) error {
-	// Get the available zones to initialize each container with the same zones
-	zones, err := pm.cpu.Zones()
-	if err != nil {
-		return err
-	}
-
-	// Get the current running
 	running := pm.resources.Containers().Running
 	containers := make(Containers, len(running))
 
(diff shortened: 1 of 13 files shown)


internal/monitor/container.go
```diff
@@ -3,98 +3,128 @@
...
+		if !exists {
+			continue
+		}
+
+		// Only include terminated containers that have consumed energy
+		if prevContainer.Zones.HasZeroEnergy() {
+			pm.logger.Debug("Filtering out terminated container with zero energy", "id", id)
+			continue
+		}
+		pm.logger.Debug("Including terminated container with non-zero energy", "id", id)
```
vimalk78, line 95: can we have this log line after the loop, with one log line showing included terminated containers

internal/monitor/container.go
```diff
@@ -3,98 +3,128 @@
...
+
+		terminatedContainer := prevContainer.Clone()
+		newSnapshot.TerminatedContainers[id] = terminatedContainer
+	}
 
 	// Skip if no containers
-	if len(containers.Running) == 0 {
+	if len(cntrs.Running) == 0 {
```
vimalk78: should this check also check for length of `cntrs.Terminated`

vimalk78 (dismissed):

vimalk78 (approved):

PR #2179, opened 2025-06-23
fix(zones): fix metric label for multiple sockets

vimalk78:
This PR fixes the issue of zone label in metrics having a suffix  even when there is a single cpu socket.
  This commit makes sysfsRaplReader make use of number of cpu sockets to   decide a ZoneLabel which is then used in metrics.  if there is a single cpu socket, the index suffix is not used; but if   there are multiple cpu sockets, the zone label is set to use name and  index both.


vimalk78:
please close this if not needed.

sthaha:
closing this since #2183 implements this

PR #2183, opened 2025-06-23
feat(device/cpu): aggregate multi-socket zones into single zone

sthaha:
This commit implements `AggregatedZone` to consolidate multiple  EnergyZones with same name (e.g., package zones in multi-socket systems) into single zone. The aggregation also handles  counter wrapping and is transparent to the caller. This enables power attribution across multi-socket systems while maintaining compatibility with single-socket deployments.

Additionally it also solves the zone label which used to have the index suffix as a hack.

Key changes:
- New `AggregatedZone` type that sums energy from multiple zones of the same type
- Proper handling of counter wrapping at MaxEnergy
- Thread-safe energy aggregation with overflow protection

image

### DIFF
--- a/internal/device/energy_zone.go
+++ b/internal/device/energy_zone.go
@@ -1,0 +1,154 @@
+// SPDX-FileCopyrightText: 2025 The Kepler Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package device
+
+import (
+	"fmt"
+	"math"
+	"sync"
+)
+
+type Zone = string
+
+const (
+	ZonePackage Zone = "package"
+	ZoneCore    Zone = "core"
+	ZoneDRAM    Zone = "dram"
+	ZoneUncore  Zone = "uncore"
+	ZonePSys    Zone = "psys"
+	ZonePP0     Zone = "pp0" // Power Plane 0 - processor cores
+	ZonePP1     Zone = "pp1" // Power Plane 1 - uncore (e.g., integrated GPU)
+)
+
+// zoneKey uniquely identifies a zone by name and index
+type zoneKey struct {
+	name  string
+	index int
+}
+
+// AggregatedZone implements EnergyZone interface by aggregating multiple zones
+// of the same type (e.g., multiple package zones in multi-socket systems).
+// It handles energy counter wrapping for each individual zone and provides
+// a single consolidated energy reading.
+type AggregatedZone struct {
+	name          string
+	index         int
+	zones         []EnergyZone
+	lastReadings  map[zoneKey]Energy
+	currentEnergy Energy // Aggregated energy counter
+	maxEnergy     Energy // Cached sum of all zone MaxEnergy values
+	mu            sync.RWMutex
+}
+
+// NewAggregatedZone creates a new AggregatedZone for zones of the same type
+// The name is taken from the first zone
+// Panics if zones is empty or nil
+func NewAggregatedZone(zones []EnergyZone) *AggregatedZone {
+	// Panic on invalid inputs
+	if len(zones) == 0 {
+		panic("NewAggregatedZone: zones cannot be empty")
+	}
+
+	// Use the first zone's name as the aggregated zone name
+	name := zones[0].Name()
+	// Calculate and cache the combined MaxEnergy during construction
+	// Check for overflow when summing MaxEnergy values
+	var totalMax Energy
+	for _, zone := range zones {
+		zoneMax := zone.MaxEnergy()
+		// Check for overflow before adding
+		if totalMax > 0 && zoneMax > math.MaxUint64-totalMax {
+			// Overflow would occur, use MaxUint64 as safe maximum
+			totalMax = Energy(math.MaxUint64)
+			break
+		}
+		totalMax += zoneMax
+	}
+
+	return &AggregatedZone{
+		name:          name,
+		index:         -1, // Indicates this is an aggregated zone
+		zones:         zones,
+		lastReadings:  make(map[zoneKey]Energy),
+		currentEnergy: 0,
+		maxEnergy:     totalMax, // Cache the combined MaxEnergy
+	}
+}
+
+// Name returns the zone name
+func (az *AggregatedZone) Name() string {
+	return az.name
+}
+
+// Index returns the zone index (-1 for aggregated zones)
+func (az *AggregatedZone) Index() int {
+	return az.index
+}
+
+// Path returns path for the aggregated zone
+func (az *AggregatedZone) Path() string {
+	// TODO: decide if all the paths should be returned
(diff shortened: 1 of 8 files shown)


vprashar2929:
Tested on multi-socket machine. 

Available RAPL domains on machine:

```
root@aa4a13c3:/home/ghrunner/kepler/compose/dev# for d in /sys/devices/virtual/powercap/intel-rapl/intel-rapl:*; do   [ -f "$d/name" ] && echo "${d#/sys/devices/virtual/powercap/intel-rapl/}: $(cat "$d/name")";   for s in "$d"/intel-rapl:*; do     [ -f "$s/name" ] && echo "${s#/sys/devices/virtual/powercap/intel-rapl/}: $(cat "$s/name")";   done; done
intel-rapl:0: package-0
intel-rapl:0/intel-rapl:0:0: dram
intel-rapl:1: package-1
intel-rapl:1/intel-rapl:1:0: dram
```

Kepler creating aggregated zone:

```
kepler-dev-1  | time=2025-07-01T12:47:56.812Z level=DEBUG source=internal/device/rapl_sysfs_power_meter.go:177 msg="Created aggregated zone" service=rapl name=package zone_count=2 zones="[package-0 package-1]"
kepler-dev-1  | time=2025-07-01T12:47:56.812Z level=DEBUG source=internal/device/rapl_sysfs_power_meter.go:177 msg="Created aggregated zone" service=rapl name=dram zone_count=2 zones="[dram-0 dram-1]"
```

Comparing against node exporter

Prometheus-Time-Series-Collection-and-Processing-Server-07-01-2025_06_24_PM

vprashar2929:
@sthaha Right now, if I want to filter only the package-0 zone, it's not possible?

sthaha:
> @sthaha Right now, if I want to filter only the package-0 zone, it's not possible?

Thats right ... but if you run a version before this you should be able to see the values.

internal/device/energy_zone.go
```diff
@@ -0,0 +1,154 @@
...
+// Index returns the zone index (-1 for aggregated zones)
+func (az *AggregatedZone) Index() int {
+	return az.index
+}
+
+// Path returns path for the aggregated zone
+func (az *AggregatedZone) Path() string {
+	// TODO: decide if all the paths should be returned
+	return fmt.Sprintf("aggregated-%s", az.name)
+}
```
vimalk78, line 93: `Path()` returns name?

internal/device/energy_zone.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// Update aggregated energy counter
+	az.currentEnergy += totalDelta
+
+	// Wrap at maxEnergy boundary to match hardware counter behavior
+	// This is required for the power attribution algorithm's calculateEnergyDelta()
+	if az.maxEnergy > 0 {
+		az.currentEnergy %= az.maxEnergy
+	}
```
vimalk78, line 145: what is  the need for this wraparound when totalMax already takes care of overflow

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -146,13 +146,52 @@ func (r *raplPowerMeter) Zones() ([]EnergyZone, error) {
...
+	var result []EnergyZone
+	for name, zones := range zoneGroups {
+		if len(zones) == 1 {
+			// Single zone - use as-is
+			result = append(result, zones[0])
+			continue
+
+		}
+
+		// Multiple zones with same name - create AggregatedZone
```
vimalk78, line 174: is there an option to not aggregate in some rare case?

internal/device/energy_zone.go
```diff
@@ -0,0 +1,154 @@
...
+// a single consolidated energy reading.
+type AggregatedZone struct {
+	name          string
+	index         int
+	zones         []EnergyZone
+	lastReadings  map[zoneKey]Energy
+	currentEnergy Energy // Aggregated energy counter
+	maxEnergy     Energy // Cached sum of all zone MaxEnergy values
+	mu            sync.RWMutex
+}
```
vimalk78, line 42: Can we add a method `Label` in interface, for which `AggregatedZone` will return just name, and sysFsRaplZone will return "{Name}-{Index}" ?
And we use `Label` in power collector?

internal/device/energy_zone.go
```diff
@@ -0,0 +1,154 @@
...
+func (az *AggregatedZone) Path() string {
+	// TODO: decide if all the paths should be returned
+	return fmt.Sprintf("aggregated-%s", az.name)
+}
+
+// Energy returns the total energy consumption across all aggregated zones,
+// handling wrap-around for each individual zone
+func (az *AggregatedZone) Energy() (Energy, error) {
+	az.mu.Lock()
+	defer az.mu.Unlock()
```
vimalk78, line 99: why need mutex?  `sysfsRaplZone` didn't need this

internal/device/energy_zone.go
sthaha: Since we are replacing multiple zones with a single zone, and won't be exposing the sub zones I don't think having a label helps in this case.

internal/device/energy_zone.go
sthaha: Yes, since we can't point to single path, I thought an easy way to identify is to set Path as aggregated. We definitely can improve this by finding the common part of the parts and replacing the difference with a regex. I will add an issue for us to fix this later.

sthaha: Or do you think appending the parts into a single string is sufficient?

internal/device/energy_zone.go
sthaha: I will mark this as resolved, please re-open if I didn't understand this correctly.

internal/device/energy_zone.go
sthaha: Added #2195

internal/device/rapl_sysfs_power_meter.go
sthaha: Good question and I think we will need that when/if kepler supports measuring individual socket consumption and is able to attribute that to process running on those sockets. 

Or are you talking about some other requirement that we may have at the moment?

internal/device/energy_zone.go
sthaha: I think it may not be necessary but let me check that again .. for now this shouldn't change how the aggregated zone should behave. 
Adding an issue

internal/device/rapl_sysfs_power_meter.go
vimalk78: some reason we cant think of now. the default can be to aggregate, but if needed a config change (unexposed perhaps) will stop the aggregation.

vimalk78 (approved):

PR #2202, opened 2025-07-07
feat: implement energy-based terminated workload tracking

sthaha:
Replace unbounded terminated workload maps with bounded, energy-prioritized
tracking system. Automatically retain the N highest energy consuming
terminated resources (processes, containers, VMs, pods) based on optimal
RAPL energy zone selection.

Key changes:
- Add TopEnergyZone() method to CPUPowerMeter interface
- Implement generic TerminatedResourceTracker with min-heap priority queue
- Add Resource interface to Process, Container, VM, and Pod types
- Replace map-based terminated fields with bounded slices in Snapshot
- Add maxTerminated configuration option (default: 500)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -49,6 +49,10 @@ type (
 		Interval  time.Duration `yaml:"interval"`  // Interval for monitoring resources
 		Staleness time.Duration `yaml:"staleness"` // Time after which calculated values are considered stale
 
+		// MaxTerminated controls terminated workload tracking behavior:
+		// <0: Any negative value indicates to track unlimited terminated workloads (no capacity limit)
+		// =0: Disable terminated workload tracking completely
+		// >0: Track top N terminated workloads by energy consumption
 		MaxTerminated int `yaml:"maxTerminated"`
 	}
 
@@ -297,8 +301,8 @@ func RegisterFlags(app *kingpin.Application) ConfigUpdaterFn {
 	// monitor
 	monitorInterval := app.Flag(MonitorIntervalFlag,
 		"Interval for monitoring resources (processes, container, vm, etc...); 0 to disable").Default("5s").Duration()
-	maxTerminated := app.Flag(MonitorMaxTerminatedFlag,
-		"Maximum number of terminated workloads to keep in memory until exported; 0 for unlimited").Default("500").Int()
+	monitorMaxTerminated := app.Flag(MonitorMaxTerminatedFlag,
+		"Maximum number of terminated workloads to track; 0 to disable, -1 for unlimited").Default("500").Int()
 
 	enablePprof := app.Flag(pprofEnabledFlag, "Enable pprof debug endpoints").Default("false").Bool()
 	webConfig := app.Flag(WebConfigFlag, "Web config file path").Default("").String()
@@ -338,9 +342,8 @@ func RegisterFlags(app *kingpin.Application) ConfigUpdaterFn {
 		if flagsSet[MonitorIntervalFlag] {
 			cfg.Monitor.Interval = *monitorInterval
 		}
-
 		if flagsSet[MonitorMaxTerminatedFlag] {
-			cfg.Monitor.MaxTerminated = *maxTerminated
+			cfg.Monitor.MaxTerminated = *monitorMaxTerminated
 		}
 
 		if flagsSet[pprofEnabledFlag] {
--- a/internal/device/cpu_power_meter.go
+++ b/internal/device/cpu_power_meter.go
@@ -32,4 +32,9 @@ type CPUPowerMeter interface {
 
 	// Zones() returns a slice of the energy measurement zones
 	Zones() ([]EnergyZone, error)
+
+	// PrimaryEnergyZone() returns the zone with the highest energy coverage/priority
+	// This zone represents the most comprehensive energy measurement available
+	// E.g. Psys > Package > Core > DRAM > Uncore
+	PrimaryEnergyZone() (EnergyZone, error)
 }
--- a/internal/device/rapl_sysfs_power_meter.go
+++ b/internal/device/rapl_sysfs_power_meter.go
@@ -17,6 +17,7 @@ type raplPowerMeter struct {
 	cachedZones []EnergyZone
 	logger      *slog.Logger
 	zoneFilter  []string
+	topZone     EnergyZone
 }
 
 type OptionFn func(*raplPowerMeter)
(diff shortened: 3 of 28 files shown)


config/config.go
```diff
@@ -46,6 +46,10 @@ type (
 		Interval  time.Duration `yaml:"interval"`  // Interval for monitoring resources
 		Staleness time.Duration `yaml:"staleness"` // Time after which calculated values are considered stale
 
+		// MaxTerminated controls terminated workload tracking behavior:
+		// -1: Track unlimited terminated workloads (no capacity limit)
```
sthaha: Any negative number is unlimited tracking ..

internal/device/cpu_power_meter.go
```diff
@@ -32,4 +32,9 @@ type CPUPowerMeter interface {
 
 	// Zones() returns a slice of the energy measurement zones
 	Zones() ([]EnergyZone, error)
+
+	// TopEnergyZone() returns the zone with the highest energy coverage/priority
+	// This zone represents the most comprehensive energy measurement available
+	// E.g. Psys > Package > Core > DRAM > Uncore
```
vimalk78, line 38: `Dram` and `uncore` do not belong in the same ordering

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -133,42 +134,59 @@ func (r *raplPowerMeter) Zones() ([]EnergyZone, error) {
...
 
 		// ignore non-standard zones if a standard zone already exists
 		if existingZone, exists := stdZoneMap[key]; exists && isStandardRaplPath(existingZone.Path()) {
 			continue
 		}
+
+		// If this is a new key, add it to the ordered list
+		if _, exists := stdZoneMap[key]; !exists {
+			orderedKeys = append(orderedKeys, key)
+		}
```
vimalk78: previous `if` also checks for presence of key, can we combine the two checks for clarity? 
the `isStandardRaplPath` for existing entry in map is hard to understand already.

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -133,42 +134,59 @@ func (r *raplPowerMeter) Zones() ([]EnergyZone, error) {
...
 		stdZoneMap[key] = zone
 	}
 
+	// Create ordered zones based on the key order
+	var orderedZones []EnergyZone
+	for _, key := range orderedKeys {
+		orderedZones = append(orderedZones, stdZoneMap[key])
```
vimalk78: does the ordering mean the order in which zones are returned by `sysfsReader` ?

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -133,42 +134,59 @@ func (r *raplPowerMeter) Zones() ([]EnergyZone, error) {
...
-	r.cachedZones = r.groupZonesByName(stdZoneMap)
+	r.cachedZones = r.groupZonesByName(orderedZones)
 	return r.cachedZones, nil
 }
 
-// groupZonesByName groups zones by their base name and creates AggregatedZone
-// instances when multiple zones share the same name (multi-socket systems)
-func (r *raplPowerMeter) groupZonesByName(stdZoneMap map[zoneKey]EnergyZone) []EnergyZone {
+// groupZonesByName groups zones by their base name while preserving order
+func (r *raplPowerMeter) groupZonesByName(zones []EnergyZone) []EnergyZone {
```
vimalk78: in case of multiple sockets, the final ordering is as per ordering in the first socket?

internal/device/rapl_sysfs_power_meter.go
```diff
@@ -192,6 +210,43 @@ func (r *raplPowerMeter) zoneNames(zones []EnergyZone) []string {
...
+		return nil, fmt.Errorf("no energy zones available")
+	}
+
+	zoneMap := map[string]EnergyZone{}
+	for _, zone := range zones {
+		zoneMap[strings.ToLower(zone.Name())] = zone
+	}
+
+	// Priority hierarchy for RAPL zones (highest to lowest priority)
+	priorityOrder := []string{"psys", "package", "core", "dram", "uncore"}
```
vimalk78, line 218: `Dram` and `uncore` do not belong in the same hierarchy. these are at same level as `core`

vimalk78, line 218: for flexibility, can we not hard code this ordering? have the ability for user to define it at runtime?

internal/device/cpu_power_meter.go
```diff
@@ -32,4 +32,9 @@ type CPUPowerMeter interface {
...
+	TopEnergyZone() (EnergyZone, error)
```
vimalk78: can we use something like `PrimaryEnergyZone` or `HighestEnegyZone`
`Top` means ordered based on some runtime criteria

internal/monitor/terminated_resource_tracker.go
vimalk78, line 1: if terminated resources are cleared after each Snapshot export, do we really need this tracking? accumulation will occur only if export is somehow disabled, right?

vimalk78 (commented):
if terminated processes, containers, vm, pod will be accumulated for at most one scrape interval, do we really need to track it in a heap based on estimated value of energy used after last snapshot?

internal/device/rapl_sysfs_power_meter.go
sthaha: ordering represents the order of the input. E.g if the input is `package, core, dram, core, package, dram'` the returned zones will preserve the order `package, core, dram`

internal/device/cpu_power_meter.go
sthaha: This is just the preferred order so that in the absence of Psys and Package, Core will be preferred over dram if both are present.

internal/device/cpu_power_meter.go
sthaha: Lets pick `PrimaryEnergyZone()`

internal/device/rapl_sysfs_power_meter.go
sthaha: Will combine both.

internal/device/rapl_sysfs_power_meter.go
sthaha: imho, the PrimaryEnergyZone shouldn't be left to user to decide. If there are special cases, this implementation should handle it.

internal/device/rapl_sysfs_power_meter.go
sthaha: Lets remove the ordering of zones with the caveat that PrimaryEnergyZone can return a random zone from the list of zone if there are no "known" zones.

internal/device/rapl_sysfs_power_meter.go
sthaha: resolved in earlier comment

internal/monitor/terminated_resource_tracker.go
sthaha: I think it is important to keep the items tracked within a limit. Here we are picking the top consumers to keep in memory until it is exported. IMHO, it will be useful in case of edge compute where the exporter may not be scrapped or guaranteed to be scrapped at regular intervals.

PR #2206, opened 2025-07-08
feat(config): add configurable web server listen addresses

sthaha:
Add support for configurable listen addresses for the web server through both CLI flags and YAML configuration, instead of the hardcoded :28282 default. This change is needed to support a reverse proxy (kube-rbac-proxy) setup where the listen-address will use 127.0.0.1:28282 instead of 0.0.0.0:28282

Changes include:
- Add Web.ListenAddresses []string field to config structure
- Add --web.listen-address CLI flag (repeatable for multiple addresses)

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -158,6 +158,7 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 
 	apiServer := server.NewAPIServer(
 		server.WithLogger(logger),
+		server.WithListenAddress(cfg.Web.ListenAddresses),
 		server.WithWebConfig(cfg.Web.Config),
 	)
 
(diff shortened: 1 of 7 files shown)


SamYuan1990 (dismissed):
wait a min.... listen address is a part of webconfig, so https://github.com/sustainable-computing-io/kepler/pull/2165 works or not?

sthaha:
@SamYuan1990 

> wait a min.... listen address is a part of webconfig, so https://github.com/sustainable-computing-io/kepler/pull/2165 works or not?

#2165  PR isn't complete and makes lots of changes unrelated to adding --web.config.file and --web.listen-address. We can revisit that PR once this merges. This pr is required for a critical feature related to securing kepler and make minimal change need to add `--web.listen-address`. This shouldn't block #2165 from progressing.

PR #2214, opened 2025-07-10
[ci]: add depbot, and scorecord back to CI

SamYuan1990:
as title, add depend bot and score cord back in CI.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      day: monday
+      interval: weekly
+    groups:
+      go-dependencies:
+        patterns:
+          - "*"
+    ignore:
+      - dependency-name: k8s.io/*
+        update-types: [version-update:semver-major, version-update:semver-minor]
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      day: monday
+      interval: weekly
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,64 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on: # yamllint disable-line rule:truthy
+  workflow_call:
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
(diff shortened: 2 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,30 @@
...
+    schedule:
+      day: monday
+      interval: weekly
+    groups:
+      go-dependencies:
+        patterns:
+          - "*"
+    ignore:
+      # Ignore gohlml updates as this will require manual testing
+      - dependency-name: github.com/HabanaAI/gohlml
```
vprashar2929, line 17: Not required

.github/dependabot.yml
```diff
@@ -0,0 +1,30 @@
...
+      # Ignore go-nvml updates as this will require manual testing
+      - dependency-name: github.com/NVIDIA/go-nvml
```
vprashar2929: same here

.github/dependabot.yml
vprashar2929: not required

SamYuan1990:
@sthaha , may I know your opinion on https://github.com/sustainable-computing-io/kepler/blob/archived/.github/workflows/image.yml#L96-L131 about SBOM and cosign? as current image build is through make file... which seems not integrated well with SBOM and cosign actions.

.github/dependabot.yml
SamYuan1990: k

.github/dependabot.yml
SamYuan1990: k

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,64 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@cbb722410c2e876e24abbe8de2cc27693e501dcb # v4.1.1
```
vprashar2929: @SamYuan1990 Why are we using SHA instead of version?

SamYuan1990: I just get those config and settings from archived repo.

sthaha: @SamYuan1990 I agree with @vprashar2929 , if the `version` works, lets use that 🙏 

https://github.com/actions/checkout example says `- uses: actions/checkout@v4`

sthaha:
> @sthaha , may I know your opinion on https://github.com/sustainable-computing-io/kepler/blob/archived/.github/workflows/image.yml#L96-L131 about SBOM and cosign? as current image build is through make file... which seems not integrated well with SBOM and cosign actions.

I think we should modify that so that we can sign the artifacts. Since I know @vprashar2929 is working on a Dockerfile change to move requirement of `.git` outside of `Dockerfile` perhaps he could fix this part as well. what say @vprashar2929  ?

.github/workflows/scorecard.yml
SamYuan1990: done

sthaha (approved):

PR #2223, opened 2025-07-11
[ci]: add depbot, and scorecord back to CI

sthaha:
Reverts sustainable-computing-io/kepler#2214 because of failure 
@SamYuan1990 https://github.com/sustainable-computing-io/kepler/actions/runs/16211044819/job/45771026658


vprashar2929:
We can close this as #2227 addresses this

PR #2226, opened 2025-07-11
ci: remove redundant expression for config change comments

vprashar2929:
This commit fixes the issue where the config-change workflow was commenting on every PR regardless of whether the config change was actually detected.

The workflow also includes config detection for helm values.yaml as well

### DIFF
--- a/.github/workflows/config-change.yaml
+++ b/.github/workflows/config-change.yaml
@@ -2,7 +2,8 @@ name: Check Config Changes
 
 on: #yamllint disable-line rule:truthy
   pull_request_target:
-
+  # Using `pull_request_target` event type to allow the workflow to comment on the PR.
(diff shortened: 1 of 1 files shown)


vprashar2929:
Sample PR where no config changes were made: https://github.com/vprashar2929/kepler/pull/23
Sample PR where config changes were made: https://github.com/vprashar2929/kepler/pull/24

vprashar2929:
Please ignore https://github.com/sustainable-computing-io/kepler/pull/2226#issuecomment-3062526333 as the workflow event is `pull_request_target` and requires the changes to be actually added to the default branch i.e `main` in order for those changes to be in effect. Refer: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

sthaha (approved):

PR #2231, opened 2025-07-12
chore: enable check-x-crypto workflow

KaiyiLiu1234:
Switched workflow_dispatch with pull_request. Removed the comment functionality due to permission errors. The checks will be visible in the workflow logs.


kaiyiliu-strala:
@vprashar2929 The logs look very strange. Could it have something to do with the migration of reboot to main? Edit: never mind it was because the base branch was referencing my forked repo and not the upstream.

SamYuan1990:
@KaiyiLiu1234 , may I know why we are checking x-crypto?

SamYuan1990:
and btw, @KaiyiLiu1234 , as if the historical commits are free from the package, then we just need to check the increase part as PR specific content having the package or not.
https://github.com/sustainable-computing-io/kepler/blob/main/.github/check-x-crypto/action.yaml#L16

so which means, we have space to improve here, as remove --head-sha or --base-ref ... by just scan current code base?(which is head for now.)

kaiyiliu-strala:
@SamYuan1990 We are checking x/crypto to help us with FIPS compliance. For downstream builds, we need to make sure our use of x/crypto is not violating FIPS.

kaiyiliu-strala:
@SamYuan1990 the purpose of this is for local tests so that we can run this script locally if we want.

kaiyiliu-strala:
@vprashar2929 can you review whenever you can? Thanks!

SamYuan1990:
I am not sure, for example a downstream may replace their crypto lib and build their own version. In that case, if they back to us with a crypto issue...
> for example, when they enabled TLS, after replaced crypto lib and a TLS issue happened...

in this case, we can only provide limit help. As either we have FIPS crypto lib or the way they make the replacement is not back to us, or....

Should we just wait for golang support FIPS 140-3?

If I am the personal who need take task to support FIPS by replace go dependency, tools like GoViz, GMChart... etc are specific for this kind of task.


| |
Sam Yuan
|
|
***@***.***
Thanks and Regards
|


---- Replied Message ----
| From | ***@***.***> |
| Date | 07/17/2025 21:25 |
| To | ***@***.***> |
| Cc | Sam ***@***.***> ,
***@***.***> |
| Subject | Re: [sustainable-computing-io/kepler] chore: enable check-x-crypto workflow (PR #2231) |
KaiyiLiu1234 left a comment (sustainable-computing-io/kepler#2231)

@SamYuan1990 We are checking x/crypto to help us with FIPS compliance. For downstream builds, we need to make sure our use of x/crypto is not violating FIPS.

kaiyiliu-strala:
@SamYuan1990 You need not worry about supporting FIPS for upstream kepler at the moment. FIPS validation is just a requirement for downstream products and not a concern for upstream. This tools is just to help with FIPS validation by seeing where x/crypto (a very sensitive package) is being imported. Of course, if we can try to track and minimize use of x/crypto in our upstream kepler, that will be helpful for the downstream builds.

SamYuan1990:
> @SamYuan1990 You need not worry about supporting FIPS for upstream kepler at the moment. FIPS validation is just a requirement for downstream products and not a concern for upstream. This tools is just to help with FIPS validation by seeing where x/crypto (a very sensitive package) is being imported. Of course, if we can try to track and minimize use of x/crypto in our upstream kepler, that will be helpful for the downstream builds.

in this case, @KaiyiLiu1234, I would like to suggest as we make SBOM back. 
I don't get the point as reboot use https://github.com/sustainable-computing-io/kepler/tree/main/.github/publish-image
instead of https://github.com/sustainable-computing-io/kepler/blob/archived/.github/workflows/image.yml#L124

The SBOM structure give downstream a view for any library how and where they been invoked, I hope once we have #2235 back, we remove those checking script specific for x/crypto, wdyt?

PR #2240, opened 2025-07-16
ci: publish Kepler binary on releases

vprashar2929:
This commit updates the release workflow to publish the Kepler binary when release is created.
Currently we only publish the binary for Linux-amd64. Once we have multi-arch support, we can publish the binary for other platforms as well and update the workflow accordingly.


vprashar2929:
Sample release for ref: https://github.com/vprashar2929/kepler/releases/tag/v0.11.0

PR #2243, opened 2025-07-18
ci: specify version when running make targets

vprashar2929:
This commit fixes the issue where the release workflow failed due to referencing a non-existent version. Now when building and pushing images we specify the version when running make targets.


SamYuan1990 (approved):
Let's have it as hot fix.

PR #2245, opened 2025-07-18
doc: enhancement proposal for supporing Redfish

sthaha:


aharivel:
Hi @sthaha , 
I read your draft and it looks good. But the main problem I see is how will you handle the different endpoint where we can find the power consumed ? Here an example on 2 machines: 
/redfish/v1/Chassis/Tray_1/Power#/0/PowerConsumedWatts
/redfish/v1/Chassis/System.Embedded.1/Power/PowerConsumedWatts

But there is also other one in the Redfish specification. Please check this: https://www.dmtf.org/sites/default/files/standards/documents/DSP0268_2024.4.html#environmentmetrics-132
And look for "PowerWatts".
Sometimes it's available, sometimes not..

sthaha:
> But the main problem I see is how will you handle the different endpoint where we can find the power consumed ? Here an example on 2 machines:
/redfish/v1/Chassis/Tray_1/Power#/0/PowerConsumedWatts
/redfish/v1/Chassis/System.Embedded.1/Power/PowerConsumedWatts

Hi @aharivel! I believe gofish library solves this problem that you pointed out but we should test this of course ...  see kepler implementation in progress https://github.com/sustainable-computing-io/kepler/pull/2263  

E.g ... https://github.com/stmcginnis/gofish/blob/v0.15.0/redfish/chassis.go#L443-L478

I think for each vendor, gofish creates and endpoint specific to the vendor 
- **Dell iDRAC** returns: `"Power": {"@odata.id": "/redfish/v1/Chassis/System.Embedded.1/Power"}`
- **HPE iLO** returns: `"Power": {"@odata.id": "/redfish/v1/Chassis/1/Power"}`  
- **Lenovo XCC** returns: `"Power": {"@odata.id": "/redfish/v1/Chassis/Tray1/Power"}`


I believe kepler code  should uniformly work across all vendors supported by gofish.
Our implementation should include vendor testing with realistic responses. Also thinking of adding a tool to capture and update testdata. 

Also .. any fix to redfish support should be contributed to gofish library.

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Metrics
+
+```prometheus
+# New platform metrics
+kepler_node_platform_watts{source="redfish",node_name="worker-1"} 450.5
+kepler_node_platform_joules_total{source="redfish",node_name="worker-1"} 123456.789
+
+# Existing CPU metrics unchanged
+kepler_node_cpu_watts{zone="package",node_name="worker-1"} 125.2
```
vimalk78, line 211: should we add a `source` label in existing cpu_watts metrics? set the label according to the cpu power source.  currently `source=rapl` will be there.

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,223 @@
...
+    Rel(monitor, rapl, "Uses")
+    Rel(platform, redfish, "Uses")
+    Rel(exporter, monitor, "Collects")
+    Rel(exporter, platform, "Collects")
+    Rel(user, exporter, "Scrapes")
+```
+
+## Node Identification
+
+Nodes identified via `--platform.node-id` flag or `platform.nodeID` config,
```
vimalk78:
```suggestion
Nodes identified via `--platform.node-id` flag or `nodes.nodeID` config,
```

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,223 @@
...
+--platform.redfish.enabled=true
+--platform.redfish.config=/etc/kepler/redfish.yaml
+```
+
+**BMC Configuration (`/etc/kepler/redfish.yaml`):**
+
+```yaml
+nodes:
+  worker-1: bmc-1
+  worker-2: bmc-2
```
vimalk78: how are the nodes identified here? with kubernetes node name? or whatever is passed to the flag `platform.node-id` by the operator?

vimalk78: can the bmc IP be auto discovered? the Redfish/BMC appears as a virtual nic. and BMC may appear in /etc/hosts

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,223 @@
...
+
+bmcs:
+  bmc-1:
+    endpoint: "https://192.168.1.100"
+    username: "admin"
+    password: "secret"
+    insecure: true  # TLS verification
```
vimalk78, line 172: api calls for redfish require a `ChassisId` , do we need config item for that?

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,223 @@
...
+- Connection failures: Log and continue with RAPL-only
+- Authentication errors: Retry once, then disable for node
+- Timeouts: 30-second context timeout for BMC requests
+- Graceful degradation when BMCs unavailable
+
+## Security
+
+- Credentials in Kubernetes secrets or secure files (mode 0600)
+- No credential logging
+- Require explicit opt-in via configuration
```
vimalk78, line 225: pls add a small section for kepler config changes too

docs/developer/proposal/EP_001-redfish-support.md
sthaha: I am not sure about this at all. Do you have any reference to this? 
Also, since this is going to be implemented as an experimental feature. I think we should be able to adapt as we get more information from users. The old implementation in kepler didn't support any autodiscovery. 
May be we should address that as a separate enhancement what say 

@aharivel any inputs on this?

vimalk78: +1 lets do it in separate enhancement

sthaha: I will update the doc. So the idea is to maintain a list of BMC and node to BMC mapping. 

the node will be identified either by `--platform.redfish.node-id=foobar` flag or by 


```yaml
platform:
  redfish
    nodeID : foobar
```

docs/developer/proposal/EP_001-redfish-support.md
sthaha: +1. I have added this as something to resolve in `# Open Questions` section. I am inclined to not add ChassisID since gofish reports all available Chassis. Perhaps we can add a chassis-id or UUID  https://pkg.go.dev/github.com/stmcginnis/gofish@v0.20.0/redfish#Chassis to the data exported.

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,241 @@
...
+the underlying bare metal server (via BMC), regardless of whether Kepler runs on bare
+metal or within a VM. This separation enables:
+
+- Multiple VMs on the same bare metal to report the same platform power
+- Clear distinction between attributed workload power and total platform power
+- Aggregation by BMC ID to get actual bare metal consumption: `max by(bmc) (kepler_platform_watts)`
+
+```prometheus
+# New platform metrics (bare metal power consumption)
+kepler_platform_watts{source="redfish",node_name="worker-1",bmc="bmc-1",chassis_id="System.Embedded.1"} 450.5
```
sthaha, line 208: @vimalk78 @aharivel Any thoughts on calling this just `kepler_platform_watts` since the power reported could be at a level higher than `node` in cases where kepler is running on a VM (later) and the power reported is that of the underlying baremetal (which can be hosting other VMs)

docs/developer/proposal/EP_001-redfish-support.md
sthaha: definitely but not as part of this proposal.

docs/developer/proposal/EP_001-redfish-support.md
sthaha: done!

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,259 @@
...
+kepler_platform_joules_total{source="redfish",node_name="worker-1",bmc="bmc-1",chassis_id="System.Embedded.1"} 123456.789
```
sthaha: Lets remove this since RedFish only provides instantaneous power.

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -0,0 +1,259 @@
...
+
+# Existing node metrics unchanged (workload attribution)
+kepler_node_cpu_watts{zone="package",node_name="worker-1"} 125.2
+```
+
+## Error Handling
+
+- Connection failures: Log and continue with RAPL-only
```
sthaha:
```suggestion
- Connection failures: Log and continue without redfish power
```

docs/developer/proposal/EP_001-redfish-support.md
vimalk78: 👍🏽

Issue #2225, opened 2025-07-11
Use precommit hook for helm lint


sthaha:
Not using precommit hook here?

_Originally posted by @vprashar2929 in https://github.com/sustainable-computing-io/kepler/pull/2220#discussion_r2200215038_

ExplorerRay:
Does this issue wants to use others' hooks to replace local one? 
Because I saw a local pre-commit helm lint check [here](https://github.com/sustainable-computing-io/kepler/blob/main/.pre-commit-config.yaml#L81).

I search some existing pre commit hook for helm lint, I only see [this one](https://github.com/gruntwork-io/pre-commit/blob/master/hooks/helmlint.sh), and I also try it with following config, seems to work well.

```
  - repo: https://github.com/gruntwork-io/pre-commit
    rev: v0.1.30
    hooks:
      - id: helmlint
        name: helm-lint
        files: ^manifests/helm/[^/]+/(Chart\.yaml|values\.yaml|templates/.*)$
```

ExplorerRay:
I just found [this docs](https://github.com/ExplorerRay/kepler/blob/main/docs/developer/pre-commit.md). And I think it needs to be updated, too.

vprashar2929:
closing as CI job and docs are updated

PR #2248, opened 2025-07-20
Use precommit hook for helm lint

ExplorerRay:
Use https://github.com/gruntwork-io/pre-commit/blob/master/hooks/helmlint.sh to pre commit hook for helm lint so that developers don't need to install helm in their hosts for pre-commit.

https://github.com/sustainable-computing-io/kepler/issues/2225

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,12 +78,9 @@ repos:
       - id: shellcheck
         args: [-x]
 
-  - repo: local
+  - repo: https://github.com/gruntwork-io/pre-commit
+    rev: v0.1.30
     hooks:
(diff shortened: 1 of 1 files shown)


ExplorerRay:
Oh, I found that it still needs `helm` command for that hook by testing in a VM without `helm`... Otherwise, it will report `command not found`.

sthaha:
@ExplorerRay Thank you for the patch but in order to  merge, we need DCO to pass as well. Could  you please follow the instructions here https://github.com/sustainable-computing-io/kepler/pull/2248/checks?check_run_id=46339146025 to amend the commit and re-submit?

PR #2249, opened 2025-07-21
doc: add enhancement proposal template

sthaha:
Add structured template for enhancement proposals to improve documentation consistency and provide clear guidelines for contributors proposing new features or significant changes.


docs/developer/proposal/EP_TEMPLATE.md
```diff
@@ -0,0 +1,235 @@
...
+
+### API/Interface Changes
+
+[Describe any changes to public APIs, service interfaces, or data structures. Show code snippets for new or modified interfaces.]
+
+## Configuration
+
+[Describe how the new feature will be configured. Include details on new configuration files, CLI flags, and environment variables.]
+
+### Main Configuration Changes
```
SamYuan1990, line 96: Major？

docs/developer/proposal/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+This directory contains Enhancement Proposals (EPs) for major features and changes to Kepler.
+
+## Active Proposals
+
+| ID | Title | Status | Author | Created |
+|----|-------|--------|--------|---------|
+| [EP-000](EP_TEMPLATE.md) | Enhancement Proposal Template | Accepted |Sunil Thaha | 2025-01-18 |
+
+## Proposal Status
```
SamYuan1990, line 11: I suppose we don't need this, for example a proposal in a PR, the PR merged means Accepted?

docs/developer/proposal/EP_TEMPLATE.md
```diff
@@ -0,0 +1,235 @@
...
+
+- **Description**: ...
+- **Reason for Rejection**: ...
+
+### Alternative 2: [Name of Alternative]
+
+- **Description**: ...
+- **Reason for Rejection**: ...
+
+## Success Metrics
```
SamYuan1990, line 222: do we need this field? maybe just if accpetable?

SamYuan1990 (commented):
I suppose it likes RFC right?

docs/developer/proposal/index.md
sthaha: That isn't right. We may have proposal that are rejected but merged. The idea is to keep track of proposals for history so that anyone proposing a change only need to check in proposal. E.g. check - https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/0000-kep-process/README.md

docs/developer/proposal/EP_TEMPLATE.md
sthaha: yes, and no. We can use it where it makes sense. I will submit few proposals for reference.

docs/developer/index.md
```diff
@@ -6,6 +6,10 @@ This section contains documentation for developers working on Kepler.
 
 - [Power Attribution Guide](power-attribution-guide.md) - Comprehensive guide on how Kepler measures and attributes power consumption to processes, containers, VMs, and pods
 
+## Enhancement Proposals
+
+- [Enhancement Proposals](proposal/) - Technical proposals for major features and changes to Kepler
```
sunya-ch, line 11: May give an example what should be considered as major and what shouldn't.
Ref: https://github.com/kubernetes/enhancements/blob/master/README.md#is-my-thing-an-enhancement

docs/developer/proposal/EP_TEMPLATE.md
sthaha: Sorry, could you please expand what you meant by `major?`

sthaha: Here is an example proposal - https://github.com/sustainable-computing-io/kepler/pull/2245/files#diff-93a1c5139682e88e3b21e45f121c0dfd2fabf9ff6a68628e1f46f28be91b4e2a

docs/developer/index.md
sthaha: In the context of kepler, I think we should consider any pr that requires some amount of discussion to be major. E.g.
* Redfish, ACPI support
* Add helm repository support 
* GPU (to or not to)
* GeoPM integration (to or not to)

etc .. 
My idea is to 
* try this out and then adapt based on the feedback from community. 
* Allow anyone can write a proposal without having to implement it.
* Provide a pathway/ guidance for future contributors to work on something they are interested in
* keep track of all ideas presented (accepted or rejected). Reading a discussion thread and summarizing the outcome isn't quite easy :)

docs/developer/proposal/EP_TEMPLATE.md
SamYuan1990: > Here is an example proposal - https://github.com/sustainable-computing-io/kepler/pull/2245/files#diff-93a1c5139682e88e3b21e45f121c0dfd2fabf9ff6a68628e1f46f28be91b4e2a

don't worry, I just think "main configuration change" or "major configuration change."

PR #2251, opened 2025-07-22
docs: add pre-commit available hooks

ExplorerRay:
When searching for developer docs, I just found that [this](https://github.com/sustainable-computing-io/kepler/blob/main/docs/developer/pre-commit.md) hasn't been updated. Therefore, I add some existing hooks which in pre-commit but not in docs to it.

https://github.com/sustainable-computing-io/kepler/issues/2225


vprashar2929 (approved):

vprashar2929:
@ExplorerRay Thanks for updating the doc 🙇

PR #2274, opened 2025-08-12
docs: update start command

lampajr:
As also highlighted by the `--help` command, the `--config` option does not exist (anymore?) and `--config.file` should be use instead.


vprashar2929 (approved):

vprashar2929:
@lampajr Thanks for the update! 🙇

lampajr:
> @lampajr Thanks for the update! 🙇

Glad to be helpful even with such small changes

PR #2275, opened 2025-08-12
ci: add workflow to profile Kepler versions

vprashar2929:
This commit adds a workflow and script to profile `kepler-dev` and `kepler-latest` using Docker compose

The script requires both services to be active, captures CPU/Memory profiles, compare then and generate top cumulative diff. The workflow uses this to comment on the PR.

Steps include:
1. Deploy kepler-dev and kepler-latest via Docker compose
2. Use fake CPU power meter (runs on VM; BM machine post-merge)
3. Run profiling script to collect profiling data
4. Compare CPU/Memory profiles and generate PR comment

### DIFF
--- a/.github/workflows/profiling.yaml
+++ b/.github/workflows/profiling.yaml
@@ -1,0 +1,157 @@
+name: Profiling Report
+on:
+  pull_request_target:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  profiling:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Graphviz
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y graphviz
+
+      - name: Install Docker
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list
+          sudo apt-get update
+          sudo apt-get install -y docker-ce docker-ce-cli containerd.io
+          sudo usermod -aG docker $USER
+
+      - name: Verify Docker installation
+        shell: bash
+        run: |
+          docker ps
+          docker --version
+
+      - name: Checkout source
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@v5.5.0
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: Install pprof
+        run: go install github.com/google/pprof@latest
+
+      - name: Enable fake cpu meter
+        shell: bash
+        run: |
+          sed -i '/fake-cpu-meter:/{n;s/enabled: false/enabled: true/}' \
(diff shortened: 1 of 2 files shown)


sthaha:
@vprashar2929 could you please make a change to a go file so that we can get [Build and Deploy K8s on Equinix / build-and-deploy (pull_request_target)](https://github.com/sustainable-computing-io/kepler/actions/runs/16903999541/job/47889391002?pr=2275) run as well ?

vprashar2929:
@sthaha This runs on VM as of now. Once Oracle cloud runners are added will update the workflow accordingly

vprashar2929:
@sthaha can you help check if the diff artifacts is something we are interested in?

.github/workflows/profiling.yaml
```diff
@@ -0,0 +1,168 @@
...
+          # Build and start containers for services defined in compose.yaml in detach mode
+          docker compose up kepler-dev --build -d --wait --wait-timeout 300
+
+      - name: Sleep for 30s
+        run: sleep 30
+
+      - name: Capture CPU and Memory profiling for kepler-dev service
+        run: |
+          # NOTE: KEPLER_PORT is set to 28283 as kepler-dev service is running on port 28283
+          KEPLER_PORT=28283 DURATION=60 ./hack/reports/profiling.sh capture
```
sthaha: How about ... 

```suggestion
          ./hack/reports/profiling.sh capture --base-port 28284 --pr-port 28284 --duration 60 
```

sthaha: How about we capture it simultaneously?

vprashar2929: It doesn't make a difference in numbers when run simultaneously vs separately

vprashar2929:
With faulty code

hack/reports/profiling.sh
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+PROJECT_ROOT="$(git rev-parse --show-toplevel)"
+
+# config
+declare DURATION=${DURATION:-30}
```
sthaha, line 8: duration .. 45s ?

hack/reports/profiling.sh
```diff
@@ -0,0 +1,208 @@
...
+		"http://localhost:$KEPLER_LATEST_PORT/debug/pprof/profile" &
+
+	(
+		sleep 5
+		for i in {1..5}; do
+			info "Querying metric endpoints [$i/5]"
+			curl -s "http://localhost:$KEPLER_DEV_PORT/metrics" >/dev/null &
+			curl -s "http://localhost:$KEPLER_LATEST_PORT/metrics" >/dev/null &
+			wait
+			sleep 5
```
sthaha, line 134: lets hit .. 

- latest first 
- dev next
wait 4 seconds 

- dev first
- latest next
wait 4 secs 

So that the distribution is equal between both containers as much as possible

.github/workflows/profiling.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+      - name: Download profiling artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: profile-artifacts-${{ github.event.pull_request.number }}
+          path: ./tmp
+
+      - name: Generate comment messages
+        id: generate_message
+        run: |
```
sthaha, line 136: may be better in a script than in a yaml ?

hack/reports/profiling.sh
vprashar2929: It's set to 60s https://github.com/sustainable-computing-io/kepler/pull/2275/files#diff-1d89a551d727b3d4c01288c4e2f917867850b6de28c1660d7bc997c8a39658eeR99

.github/workflows/profiling.yaml
vprashar2929: I don't want to maintain github specific stuff like pull request number org etc inside script. Moved the basic messaging inside the script under `profiling.sh output` command

PR #2286, opened 2025-08-20
fix: skip pre commit check for depend bot

SamYuan1990:
I suppose we don't need pre commit hook check for depend bot


.github/workflows/pr-checks.yaml
```diff
@@ -102,6 +102,7 @@ jobs:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   pre-commit:
+    if: ${{ github.actor != 'dependabot[bot]' && github.actor != 'dependabot' }}
```
vprashar2929: Shouldn't this be or as in `|| instead of &&` ?

SamYuan1990: Good founding.

PR #2288, opened 2025-08-24
ci: switch from Equinix to self hosted runners

vprashar2929:


vprashar2929:
→ Existing equinix jobs will fail and will be removed once this PR gets merged

PR #2294, opened 2025-08-26
chore: add compose manifests for kepler latest

vprashar2929:
This commit adds the compose manifests for deploying kepler latest.


vprashar2929:
Grafana dashboard:
Grafana-08-26-2025_04_15_PM

compose/compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - type: bind
+        source: ./default/kepler/etc/kepler/
+        target: /etc/kepler
+        read_only: true
+
+      # NOTE: place kubeconfig here
+      # e.g. cp $KUBECONFIG ./shared/kube/kubeconfig
+      # for kind cluster, rename host in kubeconfig to kind-control-plane:6443
+      - type: bind
+        source: ./shared/kube
```
sthaha, line 33: we don't need `shared/kube` since there is only on kepler using this kubeconfig. 

In dev/compose.yaml, its different since 2 keplers are running and need to share data such as this.

compose/compose.yaml
```diff
@@ -0,0 +1,47 @@
+name: kepler
```
sthaha, line 1: Can dev-compose reuse (or include ) this so reduce the amount of repeatition we have? Perhaps just like the `monitoring` stack?

sthaha (approved):

sthaha:
@vprashar2929 can you please test if the dev grafana dashboard works?

compose/dev/compose.yaml
```diff
@@ -1,7 +1,7 @@
 name: dev
 include:
   - path:
-      - ../monitoring/compose.yaml
```
sthaha, line 4: This change breaks grafana :( , no dashboards are to be found

sthaha (changes requested):
dev compose needs fixing

vprashar2929:
Issues addressed @sthaha Can you give a run locally?

sthaha:
@vprashar2929 .. could you please rebase 🙏

PR #2295, opened 2025-08-26
ci: ensure profiling job only runs on source code changes

vprashar2929:
This commit ensures that the profiling job only runs on source code changes only.

### DIFF
--- a/.github/workflows/profiling.yaml
+++ b/.github/workflows/profiling.yaml
@@ -3,7 +3,26 @@ on:
   pull_request:
 
 jobs:
+  check-changes:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/profiling.yaml
```diff
@@ -3,7 +3,26 @@ on:
...
+
+      - name: filter changes
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            changes:
+              - '**/*.go'
+              - 'go.mod'
+              - 'go.sum'
```
sthaha, line 22: Also when the profiling script or this file changes 🙏

vprashar2929: oh yes will add those in another PR ❤️

PR #2297, opened 2025-08-26
ci: extract pr-number from the artifact

vprashar2929:
This commit fixes the issue where the pr-number required by the PR comment workflow was not being passed due to GitHub restricting the `github.event.workflow_run.pull_requests` array for security reasons, which causes it to be empty when PRs are created from forks


vprashar2929:
Ref: https://github.com/sustainable-computing-io/kepler/actions/runs/17236481587

sthaha (approved):

PR #2301, opened 2025-08-27
ci: update check for generating comment message

vprashar2929:
This commit fixes the condition for generating comment message in the workflow file.

### DIFF
--- a/.github/workflows/profiling.yaml
+++ b/.github/workflows/profiling.yaml
@@ -129,7 +129,7 @@ jobs:
 
   generate-comment-message:
     needs: profiling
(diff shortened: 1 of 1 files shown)


.github/workflows/profiling.yaml
```diff
@@ -129,7 +129,7 @@ jobs:
 
   generate-comment-message:
     needs: profiling
-    if: needs.profiling.outcome == 'success'
+    if: needs.profiling.result == 'success'
```
sthaha, line 132: Testing in production :) ?

sthaha (approved):

Issue #2079, opened 2025-05-19
[reboot]: Support Platform Metrics / ACPI


sthaha:

ExplorerRay:
If I want to take this issue, should I write a proposal like [this](https://github.com/sustainable-computing-io/kepler/pull/2245/files)?

sthaha:
@ExplorerRay thank you for helping out. Yes, lets start with the proposal.

ExplorerRay:
@sthaha When writing the proposal, I am confused about success metrics which follows [here](https://github.com/sustainable-computing-io/kepler/pull/2263/files#diff-93a1c5139682e88e3b21e45f121c0dfd2fabf9ff6a68628e1f46f28be91b4e2a). 

- In functional metrics, it says `Platform power metrics available on 95%+ of nodes with BMCs`, but how to determine the implemented Kepler is supported by so many nodes?
- For performance metrics, is there an existing way to test Kepler's performance? With it, all developers will have the same standard for how to judge the overhead.

sthaha:
@ExplorerRay 

> In functional metrics, it says Platform power metrics available on 95%+ of nodes with BMCs, but how to determine the implemented Kepler is supported by so many nodes?

we can skip that in the proposal. In Redfish case, I was relying on gofish as a way to support reading from redfish. 

> For performance metrics, is there an existing way to test Kepler's performance? With it, all developers will have the same standard for how to judge the overhead. 

Ideally CI should give us the performance difference (cc: @vprashar2929 ) when a PR is added to kepler but comparing `main` vs `pr` after running both versions. But again, in ACPI's case unless we have the hardware, it may not even be possible.

ExplorerRay:
When considering the implementation of this feature, I am wondering why the Redfish PR uses new `internal/platform` directory instead of `internal/device`. I think it's because that the Redfish client is different from reading files in sysfs. Therefore, it's better to use a split directory which may make `internal/device` look cleaner. But for ACPI, it also read files in sysfs while it collects power data for platform. I am not sure which directory is better for ACPI power meter code.
For my opinion, I want to follow the Redfish PR because I think the ACPI power meter code won't reuse RAPL ones although it is actually a device. However, I don't know whether it's good or not to follow a PR that hasn't been merged. Will that change a lot before it merged?

[...]

Issue #1774, opened 2024-09-10
Kepler reports abnormally high platform power on Equinix


sthaha:
See: https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/docs/train-validate-e2e/2024-09-09_22-41-38/AbsPower/BPFOnly/ExponentialRegressionTrainer/report-v0.7.11-212-gac5ee8b8.md#platform---idle 

Which reports  platform power use as `160,000 W` (as if there is a divide / 1000 error) measured through ACPI

<img width="1511" alt="image" src="https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/docs/train-validate-e2e/2024-09-09_22-41-38/AbsPower/BPFOnly/ExponentialRegressionTrainer/images/metal-vs-vm-platform_idle.png?raw=true">

PR #2304, opened 2025-08-28
docs(proposal): add EP_004 ACPI support for Kepler

ExplorerRay:
This proposal aims to broaden the support of platform power and energy metrics with ACPI for Kepler.
https://github.com/sustainable-computing-io/kepler/issues/2079

### DIFF
--- a/docs/developer/proposal/EP_004-ACPI-support.md
+++ b/docs/developer/proposal/EP_004-ACPI-support.md
@@ -3,7 +3,7 @@
 **Status**: Draft
 **Author**: Ray Huang
 **Created**: 2025-08-28
-**Last Updated**: 2025-08-28
+**Last Updated**: 2025-08-29
 
 ## Summary
 
@@ -58,30 +58,26 @@ Not all nodes have Redfish for getting platform power metrics, so this proposal 
 Add ACPI option which is an alternative of Redfish to Kepler.
 ACPI can collect platform power metrics from reading sysfs. The platform collector is able to reuse the one that Redfish uses if merged.
 
-```text
-┌─────────────────┐    ┌───────────────────┐
-│   CPU Power     │    │  Platform Power   │
-│   (RAPL)        │    │ (Redfish or ACPI) │
-└─────────────────┘    └───────────────────┘
-         │                        │
-         ▼                        ▼
-┌─────────────────┐    ┌────────────────────┐
-│ Power Monitor   │    │ Power Monitor      │
-│ (Attribution)   │    │ (For ACPI,         │
-│                 │    │ no attribution)    │
-└─────────────────┘    └────────────────────┘
-         │                        │
-         ▼                        ▼
-┌─────────────────┐    ┌────────────────────┐
-│ Power Collector │    │ Platform Collector │
-│                 │    │                    │
-└─────────────────┘    └────────────────────┘
-         └──────────┬─────────────┘
-                    ▼
-           ┌──────────────────┐
-           │ Prometheus       │
-           │ Exporter         │
-           └──────────────────┘
+```mermaid
+graph TB
+  CPUPowerMeter[CPU Power Meter]
+  PowerMonitorNode[Power Monitor **With** attribution]
+  PlatformPowerMeter[Platform Power Meter with ACPI]
+  PowerMonitorPlatform[Power Monitor **Without** attribution]
+  PowerCollector[Power Collector]
+  PlatformCollector[Platform Collector]
+  PromExporter[Prometheus Exporter]
+
+  CPUPowerMeter --> PowerMonitorNode --> PowerCollector --> PromExporter
+  PlatformPowerMeter --> PowerMonitorPlatform --> PlatformCollector --> PromExporter
+
+  style CPUPowerMeter fill:#1e88e5,color:#fff
+  style PlatformPowerMeter fill:#1e88ff,color:#fff
+  style PowerMonitorNode fill:#43a047,color:#fff
+  style PowerMonitorPlatform fill:#43a067,color:#fff
+  style PowerCollector fill:#fb8c00,color:#fff
+  style PlatformCollector fill:#f57c00,color:#fff
+  style PromExporter fill:#8e24aa,color:#fff
 ```
 
 ### Aggregate power metrics
@@ -114,8 +110,6 @@ internal/
 
 ### API/Interface Changes
 
-[Describe any changes to public APIs, service interfaces, or data structures. Show code snippets for new or modified interfaces.]
-
 ```go
 // acpi_power_meter.go - Implements powerMeter interface
 type acpiPowerMeter struct {
@@ -147,9 +141,9 @@ type ACPI struct {
 }
 ```
 
-**CLI flags**
+**CLI flags:**
 
-```
+```bash
 --platform.acpi.enabled=true  # Enable ACPI monitoring
 ```
 
@@ -223,6 +217,7 @@ spec:
 ## Migration and Compatibility
 
 ### Backward Compatibility
+
 No conflict with existing RAPL support.
 
 ### Migration Path
@@ -237,16 +232,16 @@ Set `ACPI.Enabled` to false and restart the Kepler service.
 
 ## Metrics Output
 
-ACPI provides native power **average** metrics, so we can expose `kepler_platform_joules_total` by multiplying `power[1-*]_average` with `power[1-*]_average_interval`.
+Because ACPI does not support `energy[1-*]_input`, which accumulates energy consumption over time, and since the result of multiplying `power[1-*]_average` by `power[1-*]_average_interval` cannot be validated, only `kepler_platform_watts` is provided.
+
+We will be able to implement `kepler_platform_joules_total` after ACPI supports `energy[1-*]_input`.
 
 ```prometheus
 # Platform power metrics with new source
 kepler_platform_watts{source="acpi",node_name="worker-1"} 450.5
-kepler_platform_joules_total{source="acpi",node_name="worker-1"} 123456.789
 
 # Existing CPU power metrics (unchanged)
 kepler_node_cpu_watts{zone="package",node_name="worker-1"} 125.2
-kepler_node_cpu_joules_total{zone="package",node_name="worker-1"} 89234.567
 ```
 
 ## Implementation Plan


docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,292 @@
...
+- **Testability**: Mock sysfs for unit/integration tests.
+
+## Proposed Solution
+
+### High-Level Architecture
+
+Add ACPI option which is an alternative of Redfish to Kepler.
+ACPI can collect platform power metrics from reading sysfs. The platform collector is able to reuse the one that Redfish uses if merged.
+
+```text
```
sthaha: nit: mermaid (like in MSR proposal) is a better way to draw this.

docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,292 @@
...
+
+type ACPI struct {
+    Enabled    *bool  `yaml:"enabled"`
+}
+```
+
+**CLI flags**
+
+```
+--platform.acpi.enabled=true  # Enable ACPI monitoring
```
sthaha, line 147: Please note that for the actual feature implementation, lets introduce this as an `experimental` feature, get feedback and then promote it as a supported feature.

docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,292 @@
...
+Set `ACPI.Enabled` to false and restart the Kepler service.
+
+## Metrics Output
+
+ACPI provides native power **average** metrics, so we can expose `kepler_platform_joules_total` by multiplying `power[1-*]_average` with `power[1-*]_average_interval`.
+
+```prometheus
+# Platform power metrics with new source
+kepler_platform_watts{source="acpi",node_name="worker-1"} 450.5
+kepler_platform_joules_total{source="acpi",node_name="worker-1"} 123456.789
```
sthaha:
```suggestion
```


As discussed, we can't support energy joules total just like redfish from instantaneous power. In this case the 'average_power' is only with in the average_interval (which seems like a rolling window). I don't think we should read the power over the average_interval and integrate it. 

Hence, lets not support energy computation from power since it is incorrect (based on my reading of the documentation: https://www.kernel.org/doc/html/latest/hwmon/acpi_power_meter.html 


Incidentally I found https://github.com/torvalds/linux/blob/master/Documentation/hwmon/sysfs-interface.rst#energy but ACPI doesn't seem to implement it. Lets support this feature if and when kernel support it as well.

docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,292 @@
...
+- Kubernetes deployment testing
+- ACPI-related testing
+- User documentation and deployment guides
+
+## Risks and Mitigations
+
+### Technical risks
+
+- **Several measurements under one power meter**: Aggregate `power*_average` to the power meter reading value.
+- **Several ACPI power meters detected**: Implement aggregation logic for to combine readings from multiple ACPI power meters (`ACPI000D:xx`).
```
sthaha, line 270: I think we need to take another look at this. We need to ensure that there is no hierarchy/topology in these directories otherwise we run the risk of double counting .. E.g. 

E.g. what if .. 

- Multiple ACPI000D:XX devices detected
- If meter A measures total platform power
- And meter B measures a subset (CPU power)
- Then A + B would be the wrong information to provide - right?

We need ...
1. Real world data /evidence of Multiple Meters .. and the structure

2. Do you know if we can make use of https://github.com/torvalds/linux/blob/master/drivers/hwmon/acpi_power_meter.c#L254 ?

docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,292 @@
...
+  - Ensure metrics are correctly labeled with `source="acpi"` and exposed via the Prometheus exporter.
+- **E2E Tests**:
+  - Deploy Kepler in a Kubernetes cluster and verify metrics are available in Prometheus.
+  - Test a node without ACPI support and ensure Kepler falls back to other available collectors (e.g., RAPL).
+
+### Test Infrastructure
+
+- Testdata Structure: Mimic real sysfs structure under `internal/device/platform/acpi/testdata/`
+  - Fake multiple power meters (`ACPI000D:xx`) or multiple measurements (`power*_average`) for testing
+  - e.g., `testdata/sys/bus/acpi/drivers/power_meter/ACPI000D:00/power1_average`, with sample values for power readings (e.g., 450500000 microwatts for 450.5 watts).
```
sthaha, line 215: Can there be a different unit as well? Eg. milliwatts instead of micro?  We have had cases in the past (with old kepler) where the power reported was abnormally high .. 
See: https://github.com/sustainable-computing-io/kepler/issues/1774 -> https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/docs/train-validate-e2e/2024-09-09_22-41-38/AbsPower/BPFOnly/ExponentialRegressionTrainer/report-v0.7.11-212-gac5ee8b8.md#platform---idle

sthaha: https://github.com/torvalds/linux/blob/master/Documentation/hwmon/sysfs-interface.rst#power does say `microWatt`

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: So, what should I do for an experimental feature? Change `--platform.acpi.enabled=true` into `--experimental.platform.acpi.enabled=true`?

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: What I am going to implement is reading the `power[1-*]_average` every `power[1-*]_average_interval` so that every reading will occur in the rolling window. I don't think the result will be incorrect. But the overhead might be higher depends on different intervals.

sthaha: How are we going to validate that the result is valid? IMHO, we should avoid providing a feature that isn't supported by the driver and can't be validated.

docs/developer/proposal/EP_004-ACPI-support.md
sthaha: Thats right .. May be for proposals, we should add along with `status: Implemented`, a field for  `maturity: Experimental|Stable` .

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: OK, I got it.

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: There doesn’t seem to be any document explaining how to handle experimental features. For example, how to configure them or when to promote them to stable. I think a guide is needed. Otherwise, it may be hard for people to decide which feature should be an experimental one.

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: Thanks for pointing this out. 

If there are multiple ACPI000D devices, we can use `measures/` directory to check which device is measured by this power meter. [ref](https://www.kernel.org/doc/Documentation/hwmon/acpi_power_meter)

About multiple measurements `power[1-*]_xxx` part, I haven't found a reference that telling how to determine how `power1`, `power2` are generated and what they measure.

The host I can access currently only has `ACPI000D:00` and `power1_xxx` in it, so I can only refer to docs or code to know how multiple power meters or measurements work.

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: I fix this, but I think I need to open another PR after all things are discussed and checked. Otherwise, the merge commit for resolving conflict will also be included.

docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,287 @@
...
+- Replace Redfish support. (Just complementary)
+- Support all ACPI features such as setting power limits. (focus on reading power_meter)
+- Integration with virtualized environments (e.g., VMs without ACPI pass-through).
+- Backward compatibility with legacy ACPI versions (<4.0) or non-Linux systems.
+
+## Requirements
+
+### Functional Requirements
+
+- Use [golang sysfs package](https://github.com/prometheus/procfs) which is used by RAPL to read ACPI power_meter so that no additional dependencies.
```
vimalk78, line 41: what path(s) will be read for ACPI power by procfs? is reading `/sys/class/hwmon` supported in procfs?

vimalk78, line 41: what permissions are needed to read acpi power?

docs/developer/proposal/EP_004-ACPI-support.md
```diff
@@ -0,0 +1,287 @@
...
+  style PowerMonitorPlatform fill:#43a067,color:#fff
+  style PowerCollector fill:#fb8c00,color:#fff
+  style PlatformCollector fill:#f57c00,color:#fff
+  style PromExporter fill:#8e24aa,color:#fff
+```
+
+### Aggregate power metrics
+
+According to [the ACPI spec for power meter](https://docs.kernel.org/hwmon/acpi_power_meter.html#special-features), there may be more than one power meter available per platform (`/sys/bus/acpi/drivers/power_meter/ACPI000D:XX/power[1-*]_average`).
+To acquire power metrics for whole platform, I will aggregate metrics from all available power meters.
```
vimalk78, line 86: aggregate in kepler? with `sum` or `average` ?

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: In the Ubuntu 24.04 host I accessed before, the `power[1-*]_average` are readable for all users. So we only need to mount into container, and this has already done because of RAPL (mount host sysfs). There are no additional capabilities needed for reading.

ExplorerRay: I originally thought reading `/sys/bus/acpi/drivers/power_meter/ACPI000D:XX/power[1-*]_average`. 
But I am not sure if hwmon path and this path which is much common in different OS, this path should be the one that most OS support so that less issues will occur when OS changes. I am not sure which one is better.

I found that golang sysfs module doesn't support ACPI reading both the path I mentioned and hwmon path. I will change using sysfs module into reading file directly.

 Thanks.

docs/developer/proposal/EP_004-ACPI-support.md
ExplorerRay: My original idea is using `sum` to add them together, but I found that there are no docs mentioning that multiple power meters [here](https://github.com/sustainable-computing-io/kepler/pull/2304#discussion_r2309133930). I think this requires more discussion before implementing.

PR #2311, opened 2025-09-04
refactor(redfish): enhance power reading with PowerSubsystem API support

sthaha:
This refactoring improves reliability by supporting both new and legacy 
Redfish implementations. This commit adds support for PowerSubsystem API
and uses deprecated/legagy Power API as fallback for old BMC

See: https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2025.2.html#power-173-deprecated

Key Changes:
- Add PowerSubsystem API support as primary power reading method
- Fallback to deprecated Power API when needed
- Move client lifecycle management from Service to PowerReader
- Strategy pattern for determining optimal power reading approach
- redfish Init() fails if it can't create a gofish client that connects
  to BMC

Other changes:
- Reorganize test utilities into dedicated testutil package
- Add fallback scenario test
- JSON fixtures for PowerSubsystem and PowerSupplies responses
- Mock server handles configurable response scenarios
- Add validation helpers for power data verification


internal/exporter/prometheus/collector/platform_collector.go
```diff
@@ -53,8 +53,8 @@ func NewRedfishCollector(redfish RedfishDataProvider, logger *slog.Logger) *Plat
 		bmcID:    redfish.BMCID(),
 		wattsDesc: prometheus.NewDesc(
 			prometheus.BuildFQName(keplerNS, platformSubsystem, "watts"),
-			"Current platform power consumption in watts from BMC PowerControl entries",
-			[]string{"source", "node_name", "bmc_id", "chassis_id", "power_control_id", "power_control_name"},
+			"Current platform power in watts from BMC (PowerSubsystem or deprecated Power API)",
```
vimalk78, line 56: perhaps we can drop the `(PowerSubsystem or deprecated Power API)` part as it may confuse someone

docs/developer/proposal/EP_001-redfish-support.md
vimalk78, line 1: can we mention the Redfish data model spec we support https://www.dmtf.org/standards/redfish

internal/platform/redfish/power_reader.go
```diff
@@ -4,48 +4,176 @@
...
+		return fmt.Errorf("BMC service is not available")
+	}
+
+	// Get chassis collection to test with
+	chassis, err := service.Chassis()
+	if err != nil {
+		return fmt.Errorf("failed to get chassis collection: %w", err)
+	}
+
+	if len(chassis) == 0 {
```
vimalk78, line 97: is this an error

internal/platform/redfish/power_reader.go
```diff
@@ -4,48 +4,176 @@
...
 
-	service := pr.client.Service
-	if service == nil {
-		return nil, fmt.Errorf("BMC service is not available")
+	pr.client.Logout()
+	pr.client = nil
+}
+
+// determineStrategy tests chassis until it finds one with a supported API that has data
+func (pr *PowerReader) determineStrategy(chassis []*redfish.Chassis) (PowerAPIStrategy, error) {
```
vimalk78, line 126: is there a case where some chassis support one API and some support old deprecated API? 
if a chassis supports old api, we assume all support old api

vimalk78: will it be better to check all chassis for new api first before fallback check for all chassis using old api

internal/platform/redfish/power_reader.go
```diff
@@ -126,9 +223,115 @@ func (pr *PowerReader) ReadAll() ([]Chassis, error) {
 		return nil, fmt.Errorf("no chassis with valid power readings found")
 	}
 
-	pr.logger.Info("Successfully collected PowerControl readings",
-		"endpoint", pr.endpoint, "chassis_count", len(chassisList),
-		"total_readings", totalReadings)
+	pr.logger.Info("Successfully collected power readings",
+		"endpoint", pr.endpoint, "strategy", pr.strategy,
+		"chassis_count", len(chassisList), "total_readings", totalReadings)
```
vimalk78, line 228: print `len(chassis)` also to show if all chassis were read.

internal/platform/redfish/power_reader.go
```diff
@@ -57,64 +185,33 @@ func (pr *PowerReader) ReadAll() ([]Chassis, error) {
...
+		default:
+			return nil, fmt.Errorf("unknown power reading strategy: %s", pr.strategy)
 		}
 
-		power, err := ch.Power()
 		if err != nil {
-			pr.logger.Warn("Failed to get power information from chassis",
-				"chassis_id", chassisID, "error", err)
+			pr.logger.Warn("Failed to read power data from chassis",
+				"chassis_id", ch.ID, "strategy", pr.strategy, "error", err)
```
vimalk78, line 208: print `pr.endpoint` in warning message

docs/developer/proposal/EP_001-redfish-support.md
```diff
@@ -223,17 +227,17 @@ The Redfish service implements a **on-demand collection with caching**:
...
-### Multiple Chassis and PowerControl Support
+### Multiple Chassis and PowerSupply Support
 
-- `Power()` method returns `*PowerReading` (single reading containing multiple chassis with detailed PowerControl data)
-- `PowerReading` struct contains `[]Chassis` slice, each with `ID` and `[]Reading` for individual PowerControl entries
-- Iterates through all available chassis on the BMC and their PowerControl arrays
-- Filters and returns only PowerControl entries with valid power readings
-- Each reading includes `ControlID`, `Name`, and `Power` for granular power domain monitoring
-- Exposes individual PowerControl entries as separate metrics (e.g., Server Power Control, CPU Sub-system Power, Memory Power)
+- `Power()` method returns `*PowerReading` (single reading containing multiple chassis with detailed power supply data via PowerSubsystem)
```
vimalk78, line 235: this is for old API right? we can mention it so.

internal/exporter/prometheus/collector/platform_collector.go
sthaha: I feel this is relevant information especially when reporting issues to us.

internal/platform/redfish/power_reader.go
sthaha: Yes, IMHO ..I think the point of configuring redfish to monitor power would be to expect a reading. We can change this behavior based on feedback but for now, my opinion is to error if there no power could be read.

internal/platform/redfish/power_reader.go
sthaha: > is there a case where some chassis support one API and some support old deprecated API? if a chassis supports old api, we assume all support old api

I am not sure if we can make this assumption either. Lets wait for realworld examples. In the test that @vprashar2929 and I conducted, the new version of the API was present but did not support power reading.

docs/developer/proposal/EP_001-redfish-support.md
sthaha: Isn't this the case even now? It returns a single object 🤔

internal/exporter/prometheus/collector/platform_collector.go
sthaha: BTW, happy to remove it .. we can add it back on when there is a need.

PR #2313, opened 2025-09-04
ci: add simple install test for the new helm chart

jklare:
* to allow a minimal verification if the helm chart can be installed, this commit adds a corresponding GitHub workflow
* the workflow
  * is triggered on pull_requests to ensure that those do not break the deployment of Kepler via the helm chart
  * installs a simple KinD cluster, utilizing the GitHub action from https://github.com/marketplace/actions/kind-cluster
  * installs helm using the GitHub action from https://github.com/Azure/setup-helm
  * installs the helm chart directly from the current git checkout to the KinD cluster
  * waits for the rollout of the kepler daemonset to finish
  * shows all deployed kubernetes resources in the kepler namespace
* ideally this minimal test can be expanded in the future by exposing the service and checking if the /metrics endpoint is reachable as intended

### DIFF
--- a/.github/workflows/helm-deploy-test.yaml
+++ b/.github/workflows/helm-deploy-test.yaml
@@ -1,0 +1,27 @@
+name: Deploy Kepler with Helm on a k8s Kind cluster
+
+on:
+  pull_request:
+
+jobs:
+  install-kepler-on-kind:
+    runs-on: ubuntu-latest
+    steps:
+      # https://github.com/actions/checkout
+      - name: Checkout
+        uses: actions/checkout@v3
+      # https://github.com/marketplace/actions/kind-cluster
+      - name: Create k8s Kind Cluster
+        uses: helm/kind-action@v1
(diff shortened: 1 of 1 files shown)


jklare:
In preparation for releasing the new helm chart to a chart repository, I thought it might be useful to have a simple test if it can be deployed to a k8s (kind) cluster. Please let me know what you think, @sthaha. :)

sthaha:
@jklare Thank you for help us with the helm charts. For CI related changes it is best to work in @vprashar2929. But I think we already have workflows that test the helm charts.

vprashar2929:
@jklare We are already testing helm manifests deployment in CI: https://github.com/sustainable-computing-io/kepler/blob/3626252d4678a7ccf672cb85496a3b4195c3ce4c/.github/k8s/action.yaml#L167

jklare:
Oh, I did not see that :) thanks for the quick feedback!

PR #2314, opened 2025-09-05
chore(compose): add support for deploying redfish using sushy tool

vprashar2929:
This commit adds support for deploying sushy-tool using dev compose. Sushy-tool provides a static redfish responder that uses mockups enabling redfish API for testing and development.

The kepler-dev service now connects to redfish API endpoint to provide platform metrics.

*NOTE: The mockups contain static Power data and are not real data. This is for testing purposes only and not recommended for production.*

### DIFF
--- a/compose/dev/compose.yaml
+++ b/compose/dev/compose.yaml
@@ -44,8 +44,12 @@ services:
     command:
       - --config.file=/etc/kepler/config.yaml
 
+    depends_on:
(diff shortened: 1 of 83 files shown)


vimalk78:
do we need services other than Chassis?

vprashar2929:
not really but only if we need complete redfish like implementation

PR #2317, opened 2025-09-09
feat: add ultra-lightweight health check endpoints for Kubernetes probes

### LINKED ISSUE #2282: API endpoints for health(z)
### Target Version

### Feature Description

### Problem Statement

### Proposed Solution

Both endpoints should:

robinlem:
Implements dedicated /probe/livez and /probe/readyz endpoints optimized for Kubernetes liveness and readiness probes, addressing issue #2282.

Key features:
- /probe/readyz: Returns ready when first sample collected
- /probe/livez: Returns alive if sampling loop active (data within 30s)
- Ultra-lightweight: uses cached timestamp, no expensive operations
- Backward compatible: fallback to legacy /metrics endpoint via Helm config
- Configurable probe settings in Helm chart

Also updates all test mocks to implement the new LastCollectionTime() method required by the PowerDataProvider interface.

Closes #2282

Robin

### DIFF
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -21,6 +21,10 @@ type PowerDataProvider interface {
 	// Snapshot returns the current power data
 	Snapshot() (*Snapshot, error)
 
+	// LastCollectionTime returns the timestamp of the last successful collection.
+	// Returns zero time if no collection has occurred yet. This is ultra-lightweight.
+	LastCollectionTime() time.Time
+
 	// DataChannel returns a channel that signals when new data is available
 	DataChannel() <-chan struct{}
 
@@ -197,6 +201,16 @@ func (pm *PowerMonitor) Snapshot() (*Snapshot, error) {
 	pm.exported.Store(true)
 
 	return snapshot.Clone(), nil
+}
+
+// LastCollectionTime returns the timestamp of the last successful collection.
+// Returns zero time if no collection has occurred yet. This is ultra-lightweight.
+func (pm *PowerMonitor) LastCollectionTime() time.Time {
+	snapshot := pm.snapshot.Load()
+	if snapshot == nil {
+		return time.Time{}
+	}
+	return snapshot.Timestamp
 }
 
 func (pm *PowerMonitor) initZones() error {
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -21,10 +21,6 @@ type PowerDataProvider interface {
 	// Snapshot returns the current power data
 	Snapshot() (*Snapshot, error)
 
-	// LastCollectionTime returns the timestamp of the last successful collection.
-	// Returns zero time if no collection has occurred yet. This is ultra-lightweight.
-	LastCollectionTime() time.Time
-
 	// DataChannel returns a channel that signals when new data is available
 	DataChannel() <-chan struct{}
 
@@ -203,15 +199,6 @@ func (pm *PowerMonitor) Snapshot() (*Snapshot, error) {
 	return snapshot.Clone(), nil
 }
 
-// LastCollectionTime returns the timestamp of the last successful collection.
-// Returns zero time if no collection has occurred yet. This is ultra-lightweight.
-func (pm *PowerMonitor) LastCollectionTime() time.Time {
-	snapshot := pm.snapshot.Load()
-	if snapshot == nil {
-		return time.Time{}
-	}
-	return snapshot.Timestamp
-}
 
 func (pm *PowerMonitor) initZones() error {
 	// zone names need to be collected only once and can be cached
--- a/internal/server/probe.go
+++ b/internal/server/probe.go
@@ -6,7 +6,6 @@ package server
 import (
 	"encoding/json"
 	"net/http"
-	"time"
 
 	"github.com/sustainable-computing-io/kepler/internal/monitor"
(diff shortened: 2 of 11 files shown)


internal/monitor/monitor.go
```diff
@@ -21,6 +21,10 @@ type PowerDataProvider interface {
 	// Snapshot returns the current power data
 	Snapshot() (*Snapshot, error)
 
+	// LastCollectionTime returns the timestamp of the last successful collection.
+	// Returns zero time if no collection has occurred yet. This is ultra-lightweight.
+	LastCollectionTime() time.Time
+
```
sthaha: Lets not return LastCollectionTime from Monitor. Snapshot() already has the collection time. For ready we should return ok when **all** services are `Running` so that even if the monitor has not collected anything (interval: 0). Kepler is still ready.

internal/server/probe.go
```diff
@@ -0,0 +1,124 @@
...
+		maxStaleTime: 30 * time.Second, // Consider stale if no sample in 30s
+	}
+}
+
+func (p *probe) Name() string {
+	return "probe"
+}
+
+func (p *probe) Init() error {
+	return p.api.Register("/probe/", "probe", "Health check endpoints", p.handlers())
```
sthaha, line 37: can't we register both endpoints here?

internal/server/probe.go
```diff
@@ -0,0 +1,124 @@
...
+	if r.Method != http.MethodGet {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Check if we have at least one successful sample (ultra-lightweight)
+	lastCollection := p.powerMonitor.LastCollectionTime()
+	if lastCollection.IsZero() {
+		p.respondWithError(w, "not ready", "no successful sample yet")
+		return
```
sthaha: I think we should mark `readyz`  -> `ok` when all services are in `running` state.

internal/server/probe.go
```diff
@@ -0,0 +1,124 @@
...
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Get the last collection time (ultra-lightweight)
+	lastCollection := p.powerMonitor.LastCollectionTime()
+	if lastCollection.IsZero() {
+		p.respondWithError(w, "not alive", "no samples available")
+		return
+	}
```
sthaha: This will mark kepler as not ready if monitor.interval = 0, won't it?
IMHO, we should mark the system as live if all services are running just like readyz. Detecting stalled state isn't straight forward.

internal/monitor/monitor.go
robinlem: Thank you so much @sthaha for your detailed feedback ! This is my first PR on this project and I really apreciate you taking time to guide me.

 You're absolutely right about using Snapshot() instead of LastCollectionTime(). I initially chose LastCollectionTime() because I wanted something very lightweight and low-overhead for the health checks, but I now understand that Snapshot() is much more appropriate for a complete health test.

 Your point about interval=0 is particularly insightful - the readiness check should indeed return "ok" when all services are running, even if the monitor hasn't collected anything automatically.

  I've implemented your suggestions :
  - Replaced LastCollectionTime() with Snapshot() calls
  - Updated both readiness and liveness checks to work properly with interval=0

internal/server/probe.go
robinlem: I used the structured approach for readability, but I'm happy to simplify and register both endpoints directly in Init() if you prefer, Just let me know

internal/server/probe.go
robinlem: if I'm right, Snapshot() already validates the entire monitoring pipeline (hardware access, resource reading, calculations) which ensures all critical services  are operational. Do you think we need additional service state checking beyond this comprehensive test?

internal/server/probe.go
robinlem: switched both endpoints to use Snapshot() instead of time-based checks. Now both readyz and livez verify that services are ok (including with interval=0) rather than checking stale timestamps. Both endpoints now follow the same logic you suggested

internal/monitor/monitor.go
sthaha: @robinlem , like I mentioned, liveness should just return `ok` if after all services have run just like readiness. We shouldn't rely on `Snapshot()` since it cause monitor to calculate (unncessarily).

internal/server/probe.go
sthaha: Calling Snapshot is an expensive operation so and may not be even operational if the user has disabled `rapl` based power reporting (and in future only have platform power).

internal/monitor/monitor.go
robinlem: I really appreciate your guidance on this.I want to make sure I implement exactly what you have in mind.

I understand the concern about Snapshot() being expensive, though I've noticed that in practice it's often cached when monitoring interval < probe frequency 

However, I completely understand your preference for a conceptual approach focused on "services have started" rather than "monitoring pipeline works."

My question: In the current architecture, I don't see a direct way to check that "all services are running" without calling their methods. Could you give me a concrete example of how you'd implement this check?

I want to implement this the right way according to your vision - just need a bit more guidance on the "how"

robinlem:
Closing this PR. I’m going to completely rethink the implementation and take a different approach. I’ll open a new PR once the new design is ready. Thanks for the feedback!

PR #2322, opened 2025-09-12
chore: add experimental config validation skip option

vprashar2929:
Enable skipping validation of experimental configuration in controller manager. When experimental config is enabled and Kepler is deployed via operator, the controller manager previously attempted to validate the experimental config, which could cause issues. This change allows the controller manager to skip experimental config validation, deferring validation to the Kepler pods where the actual configuration will be properly validated.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -189,8 +189,8 @@ func (m *MetricsLevelValue) IsCumulative() bool {
 type SkipValidation int
 
 const (
-	SkipHostValidation SkipValidation = 1
-	SkipKubeValidation SkipValidation = 2
+	SkipHostValidation         SkipValidation = 1
+	SkipExperimentalValidation SkipValidation = 2
 )
 
 const (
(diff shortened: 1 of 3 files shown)


sthaha (approved):
Thanks @vprashar2929 ! good find!

config/config.go
```diff
@@ -714,19 +714,23 @@ func (c *Config) Validate(skips ...SkipValidation) error {
...
 	var errs []string
 
-	if c.IsFeatureEnabled(ExperimentalRedfishFeature) {
-		if c.Experimental.Platform.Redfish.ConfigFile == "" {
-			errs = append(errs, fmt.Sprintf("%s not supplied but %s set to true", ExperimentalPlatformRedfishConfigFlag, ExperimentalPlatformRedfishEnabledFlag))
-		} else {
-			if err := canReadFile(c.Experimental.Platform.Redfish.ConfigFile); err != nil {
-				errs = append(errs, fmt.Sprintf("unreadable Redfish config file: %s: %s", c.Experimental.Platform.Redfish.ConfigFile, err.Error()))
+	{ // Validate experimental settings
+		if _, skip := validationSkipped[SkipExperimentalValidation]; !skip {
```
sthaha:
```suggestion
func (c *Config) validateExperimentalConfig(validationSkipped map[SkipValidation]bool) []string {
if !c.experimentalFeatureEnabled() || validationSkipped[SkipExperimentalValidation];  {
return nil
}

var errs []string

	{ // Validate experimental settings
```

sthaha (commented):
Consider exiting early to reduce the nesting.

sthaha (approved):

Issue #2302, opened 2025-08-27
Publish helm repo to github oci section


jcpunk:
### Target Version

Current version (0.10.0+) - New Architecture

### Feature Description

Modern versions of helm are able to pull charts from an OCI repo.  The github container registry is already used by a number of projects for this purpose.

### Problem Statement

Currently the helm install instructions point to an assembled tarball that doesn't work with all helm abstraction layers.

### Proposed Solution

https://github.com/marketplace/actions/helm-publish

### Alternatives Considered

Setting up my own hacky repo mirror

### Additional Context

https://helm.sh/docs/topics/registries/
https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#install-helm-chart

ExplorerRay:
Hi, after seeing the provided context and referring to [how Prometheus does](https://github.com/prometheus-community/helm-charts/blob/main/.github/workflows/release.yaml#L56), I think it's okay to add lines in `release.yaml` for push helm chart to quay.io so that we can use `helm install [RELEASE_NAME] oci://quay.io/sustainable_computing_io/charts/kepler`.

```
     - name: Push Helm Chart to OCI registry
        shell: bash
        run: |
          VERSION=${{ steps.version.outputs.version }}
          # Remove 'v' prefix from version
          CHART_VERSION=${VERSION#v}
          helm push helm-releases/kepler-helm-${CHART_VERSION}.tgz oci://${{ env.IMG_BASE }}/charts
```

I am not sure whether it's okay to occupy `charts` repo under `sustainable_computing_io` organization and if this fits the need of this issue, so I wrote these here before making a PR.

sunya-ch:
@ExplorerRay shall this issue closed by https://github.com/sustainable-computing-io/kepler/pull/2324?

ExplorerRay:
The current state of `oci://quay.io/sustainable_computing_io/charts/kepler` is still unavailable because it requires a release to trigger that CI.

I think whether this issue can be closed depends on @jcpunk. Maybe we need another CI for pushing helm chart to registry when merging to main in addition to only release. Or some lines of docs to tell users that we can use this way to install kepler helm chart.

jcpunk:
I'd be hesitant to overwrite the chat at every merge into `main` - I mostly use helm as a tool for predicable deployments.  Provided there is some sort of "regular release" to get the chart published, I'll be fine.

[...]

PR #2324, opened 2025-09-16
feat: push helm chart to OCI registry

ExplorerRay:
I refer to [how Prometheus does with their helm chart release to OCI registry](https://github.com/prometheus-community/helm-charts/blob/main/.github/workflows/release.yaml#L56) and provide one more step to push packaged helm chart to OCI registry.

After pushing, we can run the following command for installing kepler helm chart `helm install [RELEASE_NAME] oci://quay.io/sustainable_computing_io/charts/kepler-helm`

https://github.com/sustainable-computing-io/kepler/issues/2302

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -108,3 +108,11 @@ jobs:
         shell: bash
         run: |
           make push VERSION=${{ steps.version.outputs.version }}
+
+      - name: Push Helm Chart to OCI registry
+        shell: bash
+        run: |
+          VERSION=${{ steps.version.outputs.version }}
+          # Remove 'v' prefix from version
+          CHART_VERSION=${VERSION#v}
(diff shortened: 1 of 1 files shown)


sthaha:
@ExplorerRay thank you for helping out. Do you have an example for this workflow run (may be on your fork) ?

ExplorerRay:
Thanks for your comment. I just use another branch for testing this workflow by changing OCI repo to my ghcr.io one and everything looks as expected.

https://github.com/ExplorerRay/kepler/actions/runs/17783227945/job/50546120501

vprashar2929:
@ExplorerRay! Thanks for this contribution ❤️ 

The commit message uses `feat:` prefix, but according to [The Conventional Commits
specification](https://www.conventionalcommits.org/en/v1.0.0/), this change should use `ci:` instead since it modifies the CI/CD pipeline rather than adding a feature to the source code.

Could you please update the commit message to:
`ci: push helm chart to OCI registry`

vprashar2929 (approved):

PR #2327, opened 2025-09-17
feat: add Kubernetes health probe endpoints

### LINKED ISSUE #2282: API endpoints for health(z)

robinlem:
Add dedicated health check endpoints for improved Kubernetes integration:
- /probe/livez: Liveness probe with heartbeat-based monitoring
- /probe/readyz: Readiness probe checking data availability

Key features:
* Lightweight checks (<10µs response time) using atomic operations
* Support for both passive (interval=0) and active collection modes
* JSON response format with status, timestamp, and duration
* Configurable tolerance (2x collection interval) for liveness detection
* Thread-safe implementation with comprehensive test coverage

Implementation:
* Add LiveChecker and ReadyChecker interfaces to service package
* Implement health checks in PowerMonitor with heartbeat tracking
* Create HealthProbeService for HTTP endpoint handling
* Update Helm chart to use new endpoints by default

Breaking change: Helm chart now uses /probe/* endpoints instead of /metrics for health probes, providing more accurate health status detection.

Closes [2282](https://github.com/sustainable-computing-io/kepler/issues/2282)

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -167,11 +167,15 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		server.WithWebConfig(cfg.Web.Config),
 	)
 
+	// Create health probe service
+	healthProbeService := server.NewHealthProbeService(apiServer, pm, pm, logger)
+
 	services = append(services,
 		resourceInformer,
 		cpuPowerMeter,
 		apiServer,
 		pm,
+		healthProbeService,
 	)
 
 	// Add Redfish service if enabled
--- a/internal/monitor/monitor.go
+++ b/internal/monitor/monitor.go
@@ -70,6 +70,9 @@ type PowerMonitor struct {
 
 	// lastCollectUnixNano tracks the last collection timestamp for liveness checks
 	lastCollectUnixNano int64
+	
+	// healthCheckTolerance is the multiplier for interval tolerance in liveness checks
+	healthCheckTolerance float64
 
 	zonesNames []string // cache of all zones
 
@@ -107,6 +110,8 @@ func NewPowerMonitor(meter device.CPUPowerMeter, applyOpts ...OptionFn) *PowerMo
 
 		maxTerminated:                opts.maxTerminated,
 		minTerminatedEnergyThreshold: opts.minTerminatedEnergyThreshold,
+		
+		healthCheckTolerance: opts.healthCheckTolerance,
 
 		collectionCtx:    ctx,
 		collectionCancel: cancel,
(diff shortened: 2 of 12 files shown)


internal/monitor/monitor.go
```diff
@@ -429,3 +443,47 @@ func (pm *PowerMonitor) calculatePower(prev, newSnapshot *Snapshot) error {
...
+	
+	// Passive mode: ready even without periodic collection
+	if pm.interval == 0 {
+		return true, nil
+	}
+	
+	// Active collection: require at least one snapshot
+	if pm.snapshot.Load() == nil {
+		return false, fmt.Errorf("no data yet")
+	}
```
sthaha, line 485: This doesn't make it **not** ready. No snapshot only means that no scrape has been made and does not mean monitor is not in ready state ...

sthaha: Monitor is Ready once `Run` is called.

cmd/kepler/main.go
```diff
@@ -167,11 +167,15 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		server.WithWebConfig(cfg.Web.Config),
 	)
 
+	// Create health probe service
+	healthProbeService := server.NewHealthProbeService(apiServer, pm, pm, logger)
```
sthaha, line 171: why do we need 2 pm 🤔 ? Shouldn't the health-probe have access to **all** services, filter those that have Liveness and Readyness checks?

 Also keep in mind that when a service's Init() is done, and all services's Run  (see internal/service.Runner) is blocked, kepler should be in Ready state. (We may have to rethink the readiness probe, there is chance to simplify it).

robinlem: The pm (PowerMonitor) is passed twice because it implements both LiveChecker and ReadyChecker interfaces, serving as both the liveness and readiness probe checker. I wanted a clear split for both, but we can change.

internal/monitor/monitor.go
robinlem: I think snapshot-based readiness check is actually better.

 Why: Run() can be called successfully, but if _firstReading()_ or _calculatePower()_ fail in _refreshSnapshot()_ (called with _run()_ ), we get:
  - running = true (service started)
  - snapshot = nil (no data due to collection error)

 Don't you think for a monitoring service, "ready" should mean "can provide data", not just "is running".
 I think your thoughs make more sense for a startup probe but not a readyness probe.
  
 I have read this documentation : https://kubernetes.io/docs/reference/using-api/health-checks/ 
 It's not crystal clear but they say “The kubelet uses readiness probes to know when a container is ready to start accepting traffic.”

nikimanoledaki:
Hi @robinlem, are you still working on this PR?

Could you rebase please and let us know if you are willing to continue working on this PR, and we'll help with reviews. Thank you!

PR #2331, opened 2025-10-06
feat: add health check probes for k8s liveness and readiness

vimalk78:
Implement health probe endpoints following k8s best practices:
  - Add LiveChecker and ReadyChecker interfaces to service framework
  - Implement IsLive() and IsReady() methods in PowerMonitor using atomic operations
  - Create HealthProbe service with /probe/livez and /probe/readyz endpoints
  - Update K8s daemonset to use new health endpoints instead of /metrics
  - unit tests for health check

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -206,6 +206,10 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		services = append(services, stdoutExporter)
 	}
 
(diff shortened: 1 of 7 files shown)


laurall974:
@vimalk78 would you have to take this on and rebase? We should be able to close  one of this 2 PR: https://github.com/sustainable-computing-io/kepler/pull/2327 as it seems to be a duplicate.

PR #2338, opened 2025-10-15
ci(dep): update dependabot configuration

vprashar2929:
This commit updates the existing dependabot configuration:

Changes include:
* Explicitly set PR limits to 5
* Add labels - All PRs will be tagged with:
  * `dependencies` (common label)
  * Ecosystem label (go, github-actions)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,5 @@
 # Please see the documentation for all configuration options:
 # https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
-
 version: 2
 updates:
   - package-ecosystem: gomod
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -15,6 +14,12 @@ updates:
     ignore:
       - dependency-name: k8s.io/*
         update-types: [version-update:semver-major, version-update:semver-minor]
+    commit-message:
+      include: scope
+    open-pull-requests-limit: 5
```
SamYuan1990, line 19: do we need limit PR number to 5?
in an edge case, we meet this limitation and a new cve happens in one of our dependency.
a new patch is ready on upstream, can we benefits from https://github.com/sustainable-computing-io/kepler/security/dependabot to open a new PR as #2247 for now, over the limitation, having the 6 PRs and patched?

sthaha: Lets start with 5 (since the PR is already mergable). If in future, we have to raise the number, we can deal with it then.

PR #2345, opened 2025-10-25
ci: harden github actions permissions for scorecard compliance

vprashar2929:
Implement OpenSSF scorecard security recommendations to
improve workflow security

- Set default  `read-all` permissions at the workflow level
  across all workflows
- Define minimal job-level permissions following least privilege
  principle
- Refactor release workflow to separate build and release steps
- Add scorecard job to PR checks workflow
- Reduce artifact retention to 1 day for scorecard results

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -3,6 +3,9 @@ name: PR Checks
 on: # yamllint disable-line rule:truthy
   pull_request:
 
+# default permissions as read only
+permissions: read-all
+
 jobs:
   check-changes:
     runs-on: ubuntu-latest
@@ -97,9 +100,18 @@ jobs:
   test-and-codecov:
     needs: check-changes
     if: needs.check-changes.outputs.changes == 'true'
+    permissions:
+      contents: read
     uses: ./.github/workflows/test-and-codecov.yaml
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  scorecard:
+    permissions:
+      contents: read
+      security-events: write
+      id-token: write
+    uses: ./.github/workflows/scorecard.yml
 
   pre-commit:
     runs-on: ubuntu-latest
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -4,14 +4,22 @@ on: # yamllint disable-line rule:truthy
   push:
     branches: [main]
 
+# default permissions as read only
+permissions: read-all
+
 jobs:
   test-and-codecov:
+    permissions:
+      contents: read
     uses: ./.github/workflows/test-and-codecov.yaml
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-  # for each PR merge, openSSF scan
   scorecard:
+    permissions:
+      contents: read
+      security-events: write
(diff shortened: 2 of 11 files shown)


vprashar2929:
Severity on `main` branch: https://github.com/sustainable-computing-io/kepler/security/code-scanning?query=is%3Aopen+branch%3Amain+severity%3Ahigh
Severity on `chore-scorecard` branch(With this PR): https://github.com/sustainable-computing-io/kepler/security/code-scanning?query=is%3Aopen+branch%3Achore-scorecard

.github/workflows/pr-checks.yaml
```diff
@@ -97,10 +100,19 @@ jobs:
   test-and-codecov:
     needs: check-changes
     if: needs.check-changes.outputs.changes == 'true'
+    permissions:
```
SamYuan1990, line 103: a question here for this.
if https://github.com/sustainable-computing-io/kepler/pull/2345/files#diff-7a6a4578fe04c4fc86b28e2dc9672e48d1a9f721315e540b90bac0ef12263053R6-R8 set global permission for all jobs, why do we need a permission setting at here?

.github/workflows/push.yaml
```diff
@@ -4,14 +4,22 @@ on: # yamllint disable-line rule:truthy
   push:
     branches: [main]
 
+# default permissions as read only
+permissions: read-all
+
 jobs:
   test-and-codecov:
+    permissions:
+      contents: read
```
SamYuan1990, line 13: will codecov write anything? as this job will run after push, can we double check with document or anyway to test it?

.github/workflows/release.yaml
```diff
@@ -116,3 +110,38 @@ jobs:
...
+        id: version
+        run: |
+          TAG_NAME=${{ github.ref_name }}
+          echo "version=$TAG_NAME" >> "$GITHUB_OUTPUT"
+
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ github.ref_name }}
+          name: release-${{ steps.version.outputs.version }}
```
SamYuan1990, line 139: release-${{ github.ref_name }} ?
if in this new design we just have bash here for ... value injection?

.github/workflows/pr-checks.yaml
vprashar2929: When using reusable workflows (jobs with the uses: keyword), permissions are not inherited from the
  global permissions setting at the workflow level. This is a security feature in GitHub Actions.

  The global permissions: read-all in the workflow only applies to regular jobs that run steps directly.
  For jobs that call reusable workflows, we must explicitly declare what permissions should be passed
   to the called workflow.

  Without the explicit permissions, the test-and-codecov.yaml workflow would receive
  no permissions (or minimal default permissions), which could cause it to fail.

.github/workflows/push.yaml
vprashar2929: No, Codecov does not write anything to the repository. `contents: read` is correct and
  sufficient.

.github/workflows/release.yaml
vprashar2929: yes we can use `release-${{ github.ref_name }}` How about I update it in this https://github.com/sustainable-computing-io/kepler/pull/2347 ?

PR #2347, opened 2025-11-03
ci: restrict workflows to run only on main repo

vprashar2929:
Prevent workflows from running on forks by adding repository owner checks to workflows that publish artifacts or use self-hosted runners.

- Add check to `push.yaml` to prevent image publishing on forks
- Add check to `release.yaml` to prevent release creation on forks
- Add check to `k8s-bm.yaml` to prevent self-hosted runner usage on forks
- Add check to `profiling.yaml` to prevent self-hosted runner usage on forks
- Removed `Extract version` step from `release.yaml` since it is
      redundant and just copying `github.ref_name` to a step variable
      output without any transformation or processing

This ensures fork contributors don't need to manually disable workflows and prevents unnecessary runs on forks.

### DIFF
--- a/.github/workflows/k8s-bm.yaml
+++ b/.github/workflows/k8s-bm.yaml
@@ -30,7 +30,7 @@ jobs:
 
   build-and-deploy:
     needs: [check-changes]
-    if: needs.check-changes.outputs.changes == 'true'
+    if: needs.check-changes.outputs.changes == 'true' && github.repository_owner == 'sustainable-computing-io'
     runs-on: self-hosted
     steps:
       # NOTE: In case of self-hosted runners, tools like docker, make, kubectl etc. are installed at the time of setting up the runner.
--- a/.github/workflows/profiling.yaml
+++ b/.github/workflows/profiling.yaml
@@ -28,7 +28,7 @@ jobs:
 
   profiling:
     needs: check-changes
-    if: needs.check-changes.outputs.changes == 'true'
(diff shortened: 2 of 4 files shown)


.github/workflows/k8s-bm.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
 
   build-and-deploy:
     needs: [check-changes]
-    if: needs.check-changes.outputs.changes == 'true'
+    if: needs.check-changes.outputs.changes == 'true' && github.repository_owner == 'sustainable-computing-io'
```
SamYuan1990, line 33: For pipeline developer, if they run PR on their fork, it may fail as there no self-hosted runner.
Which is fine, if he/she just adjust for example sbom or helm chart test, or say a CI developer should have the knowledge to ignore error in this case.

For some one need test kepler on their own fork with their own self hosted runner, they can just reuse this workflow without changes, which means they also help tested integration on other platform for us.

I suppose we can think twice.

vprashar2929: >For some one need test kepler on their own fork with their own self hosted runner, they can just reuse this workflow without changes, which means they also help tested integration on other platform for us.

Thats the trade-off. Developers with self-hosted runners need to make a one-line change. Its a reasonable ask given that the security benefits of the restriction, improved UX for the majority of contributors without self-hosted runners. Plus anyways they have to update the workflows if they are using different name/label for their self-hosted runner. For eg: https://github.com/sustainable-computing-io/kepler/blob/main/.github/workflows/k8s-bm.yaml#L34

Issue #2354, opened 2025-11-13
Guaranteed QoS pods are not detected in 0.10.x


laurall974:
### Kepler Version

0.10.0 or later (Current/Supported)

### Bug Description

In Kepler 0.10.x, pods with **Guaranteed QoS** are not detected because the regex in `kubepodsPattern` assumes that cgroup paths always include a QoS class directory (`/kubepods/<qos_class>/pod<uid>/<container_id>`).

However, Guaranteed pods may have paths like:

```
/kubepods/pod<uid>/<container_id>
```

without a QoS class segment. This causes the regex to fail and the pod is skipped.

Guaranteed pods correspond to the first form and are **currently not detected** by Kepler 0.10.x. More on Guaranteed QoS in Kubernetes: [Kubernetes Pod QoS](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/).

### Proposed fix

Make the middle QoS class directory optional in the regex:

```go
kubepodsPattern = regexp.MustCompile(`/kubepods/(?:[^/]+/)?pod[0-9a-f\-]+/([0-9a-f]{64})`)
```

### Steps to Reproduce

1. Create a pod with QOS Guaranteed (cf below).
2. Inspect /proc/< pid >/cgroup for the pod.
3. Observe `/kubepods/pod<uid>/<container_id>` — no QoS class directory.
4. Run Kepler 0.10.x → pod not detected.

For a Pod to be given a QoS class of Guaranteed:

- Every Container in the Pod must have a memory limit and a memory request.
- For every Container in the Pod, the memory limit must equal the memory request.
- Every Container in the Pod must have a CPU limit and a CPU request.
- For every Container in the Pod, the CPU limit must equal the CPU request.

### Expected Behavior

All pod power consumption should be reported.

### Environment

- Kepler version: 0.10.x (issue not present in 0.8.x)
- Kubernetes version:  v1.33.1
- CRI runtime: CRI-O
- Hardware: Linux Bare Metal Node
- Deployment Method: Helm chart (daemonset)

### Logs and Error Messages

```shell

```

vimalk78:
@laurall974 can you please add steps to create QOS guaranteed pod? is the container detected? i guess not. wonder why this issue is not present in 0.8

laurall974:
@vimalk78 this is the steps to have a QOS class pods: 

For a Pod to be given a QoS class of Guaranteed:

- Every Container in the Pod must have a memory limit and a memory request.
[...]

PR #2355, opened 2025-11-13
fix(kubernetes): make kubepods pattern compatible with qos guaranteed pods

laurall974:
Fixes https://github.com/sustainable-computing-io/kepler/issues/2354


vimalk78:
@laurall974 can you please check the failing tests

PR #2357, opened 2025-11-18
fix(kubernetes): make kubepods pattern compatible with qos guaranteed pods

laurall974:
Fixes regex pattern to detect Kubernetes Guaranteed QoS pods by making the QoS class directory optional in cgroup paths (/kubepods/pod<uid>/ vs /kubepods/<qos>/pod<uid>/). Adds test coverage for both Guaranteed and Burstable QoS pod detection scenarios.

Fixes https://github.com/sustainable-computing-io/kepler/issues/2354


vimalk78:
hello @laurall974 
Can you please add tests in  internal/resource/container_test.go

vimalk78 (approved):

PR #2364, opened 2025-12-03
docs: add CLAUDE.md symlink

vprashar2929:
Creates a symbolic link CLAUDE.md pointing to the AGENT.md to enable CLAUDE.md references in Claude code.

Claude code specifically looks for CLAUDE.md file, and using a symlin avoids content duplication whole keeping AGENTS.md as the source of truth.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


vprashar2929:
as per https://github.com/anthropics/claude-code/issues/6235

PR #2365, opened 2025-12-04
feat: update energy interface to expose a power method

KaiyiLiu1234:
Alternative power meters (non-RAPL) such as Hwmon actually expose their metrics in microwatts/milliwatts instead of microjoules. Thus, the interface needs to handle acquring power metrics with a Power() method. This PR updates the interface and Kepler to make use of Power() method.

Note: for now node.go will expect microwatts as the input but this may need to be updated to handle different multiples in future PRs.

### DIFF
--- a/internal/device/cpu_power_meter.go
+++ b/internal/device/cpu_power_meter.go
@@ -27,7 +27,7 @@ type EnergyZone interface {
 
 	// Power() returns the current power consumption by the zone.
 	// This method is used for zones that provide instantaneous power readings.
-	Power() (float64, error)
+	Power() (Power, error)
 }
 
 // CPUPowerMeter implements powerMeter
--- a/internal/device/energy_zone.go
+++ b/internal/device/energy_zone.go
@@ -153,12 +153,12 @@ func (az *AggregatedZone) MaxEnergy() Energy {
 	return az.maxEnergy
 }
 
-// Power returns the total power consumption across all aggregated zones
-func (az *AggregatedZone) Power() (float64, error) {
+// Power returns the total power consumption across all aggregated zones in microwatts
+func (az *AggregatedZone) Power() (Power, error) {
 	az.mu.RLock()
 	defer az.mu.RUnlock()
 
-	var totalPower float64
+	var totalPower Power
 	for _, zone := range az.zones {
 		power, err := zone.Power()
 		if err != nil {
--- a/internal/device/energy_zone_test.go
+++ b/internal/device/energy_zone_test.go
@@ -20,7 +20,7 @@ type mockEnergyZone struct {
 	path      string
 	energy    Energy
 	maxEnergy Energy
-	power     float64
+	power     Power
 	err       error
 	powerErr  error
 	mu        sync.RWMutex
@@ -35,7 +35,7 @@ func (m *mockEnergyZone) Energy() (Energy, error) {
 	return m.energy, m.err
 }
 func (m *mockEnergyZone) MaxEnergy() Energy { return m.maxEnergy }
-func (m *mockEnergyZone) Power() (float64, error) {
+func (m *mockEnergyZone) Power() (Power, error) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 	return m.power, m.powerErr
@@ -49,7 +49,7 @@ func (m *mockEnergyZone) SetEnergy(energy Energy) {
 }
 
 // SetPower safely updates the power value for testing
-func (m *mockEnergyZone) SetPower(power float64) {
+func (m *mockEnergyZone) SetPower(power Power) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.power = power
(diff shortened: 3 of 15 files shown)


kaiyiliu-strala:
@vprashar2929 @vimalk78 I think something strange is happening with the profiling report. When i attempt to run it locally on the upstream, it also produces a container for an outdated kepler which errors immediately. Do you know what seems to be happening?

kaiyiliu-strala:
@vimalk78 @vprashar2929 The TestCollectionCancellation keeps failing. It works fine when I run make test locally however. It might have to do with the unit tests I have added which takes more time but it could also be a race condition that is not related to my changes. Any ideas on why this is occurring? 

```
2025/12/08` 20:00:59 INFO Monitor has terminated. service=monitor
    monitor_collection_test.go:190: Run exited as expected
2025/12/08 20:00:59 INFO Computed power service=monitor duration=2.021051ms
2025/12/08 20:00:59 INFO Collection loop terminated service=monitor reason="context canceled"
    monitor_collection_test.go:204: 
        	Error Trace:	/home/runner/work/kepler/kepler/internal/monitor/monitor_collection_test.go:204
        	Error:      	Not equal: 
        	            	expected: time.Date(2025, time.December, 8, 20, 0, 59, 749119773, time.Local)
        	            	actual  : time.Date(2025, time.December, 8, 20, 0, 59, 761478141, time.Local)
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(time.Time) 2025-12-08 20:00:59.749119773 +0000 UTC m=+0.196659835
        	            	+(time.Time) 2025-12-08 20:00:59.761478141 +0000 UTC m=+0.209018203
        	            	 
        	Test:       	TestCollectionCancellation
        	Messages:   	No collections should happen after cancellation
--- FAIL: TestCollectionCancellation (0.09s)
```

kaiyiliu-strala:
codecov patch is lower because I added more test code to the mock_cpu_power_meter.go file which is meant to be a test file but codecov does not view it that way. I added all the relevant unit tests.

kaiyiliu-strala:
> @vimalk78 @vprashar2929 The TestCollectionCancellation keeps failing. It works fine when I run make test locally however. It might have to do with the unit tests I have added which takes more time but it could also be a race condition that is not related to my changes. Any ideas on why this is occurring?
> 
> ```
> 2025/12/08` 20:00:59 INFO Monitor has terminated. service=monitor
>     monitor_collection_test.go:190: Run exited as expected
> 2025/12/08 20:00:59 INFO Computed power service=monitor duration=2.021051ms
> 2025/12/08 20:00:59 INFO Collection loop terminated service=monitor reason="context canceled"
>     monitor_collection_test.go:204: 
>         	Error Trace:	/home/runner/work/kepler/kepler/internal/monitor/monitor_collection_test.go:204
>         	Error:      	Not equal: 
>         	            	expected: time.Date(2025, time.December, 8, 20, 0, 59, 749119773, time.Local)
>         	            	actual  : time.Date(2025, time.December, 8, 20, 0, 59, 761478141, time.Local)
>         	            	
>         	            	Diff:
>         	            	--- Expected
>         	            	+++ Actual
>         	            	@@ -1,2 +1,2 @@
>         	            	-(time.Time) 2025-12-08 20:00:59.749119773 +0000 UTC m=+0.196659835
>         	            	+(time.Time) 2025-12-08 20:00:59.761478141 +0000 UTC m=+0.209018203
>         	            	 
>         	Test:       	TestCollectionCancellation
>         	Messages:   	No collections should happen after cancellation
> --- FAIL: TestCollectionCancellation (0.09s)
> ```

I was able to fix this issue by adding a wait in TestCollectionCancellation (time.Sleep(100 * time.Millisecond)) before it checks if a collection has happened. Essentially, I gave it more time for collection goroutines to complete to get the most recent timestamp. I think this is a race condition however that existed during the initial implementation.

internal/device/cpu_power_meter.go
```diff
@@ -24,6 +24,10 @@ type EnergyZone interface {
 	// When energy usage reaches this value, the energy value returned by Energy()
 	// will wrap around and start again from zero.
 	MaxEnergy() Energy
+
+	// Power() returns the current power consumption by the zone.
+	// This method is used for zones that provide instantaneous power readings.
+	Power() (float64, error)
```
vimalk78: would be beneficial to add a method to return the type/kind? i.e. a method which returns the units or type of unit (Energy/ Power) . because the name of the go type still remains `EnergyZone`

internal/device/cpu_power_meter.go
vimalk78, line 1: the `EnergyZone` though an interface, is biased towards RAPL. now we have HWMON. we could improve abstractions. 


```
  // SensorType indicates what kind of value the sensor provides
  type SensorType int

  const (
      SensorEnergy SensorType = iota  // Cumulative (needs delta calculation)
      SensorPower                       // Instantaneous (direct value)
  )

  // Sensor represents a power/energy measurement source
  type Sensor interface {
      Name() string
      Index() int
      Path() string

      // Type returns whether this sensor provides cumulative energy or instantaneous power
      Type() SensorType

      // Read returns the current value in base units (microjoules or microwatts)
      Read() (uint64, error)

      // MaxValue returns the wraparound point for cumulative sensors (0 if N/A)
      MaxValue() uint64
  }


  // RAPL
  type raplSensor struct { ... }
  func (r raplSensor) Type() SensorType { return SensorEnergy }
  func (r raplSensor) Read() (uint64, error) { /* read energy_uj */ }
  func (r raplSensor) MaxValue() uint64 { return r.maxMicrojoules }

  // hwmon
  type hwmonSensor struct { ... }
  func (h hwmonSensor) Type() SensorType { return SensorPower }
  func (h hwmonSensor) Read() (uint64, error) { /* read power1_input */ }
  func (h hwmonSensor) MaxValue() uint64 { return 0 }

```

kaiyiliu-strala: This makes sense. I can make this change. It will be a much larger interface change, but I agree.

vprashar2929:
>@vprashar2929 @vimalk78 I think something strange is happening with the profiling report. When i attempt to run it locally on the upstream, it also produces a container for an outdated kepler which errors immediately. Do you know what seems to be happening?

@KaiyiLiu1234 fixed the issue on the runner that was causing profiling report to fail

internal/device/cpu_power_meter.go
kaiyiliu-strala: @vimalk78 So I agree with improving the interface, but I think this may have to be a separate PR with also an enhancement proposal which I will make. The changes not only changes every file, but there needs to be changes to the data types we use (ex. Energy, Power). I think it would be best for me to make a proposal of what changes would need to be done and then make separate PR if that is ok with you.

kaiyiliu-strala:
Updating Power methods to use the Power type instead of float64.

internal/device/cpu_power_meter.go
vimalk78: deferred to another PR after enhancement proposal

internal/device/energy_zone.go
```diff
@@ -152,3 +152,20 @@ func (az *AggregatedZone) Energy() (Energy, error) {
 func (az *AggregatedZone) MaxEnergy() Energy {
 	return az.maxEnergy
 }
+
+// Power returns the total power consumption across all aggregated zones
+func (az *AggregatedZone) Power() (float64, error) {
```
vimalk78: is there an aggregation in hwmon? 
```
// AggregatedZone implements EnergyZone interface by aggregating multiple zones
// of the same type (e.g., multiple package zones in multi-socket systems).
```

internal/device/cpu_power_meter.go
kaiyiliu-strala: Yep, I am adding the Power unit.

kaiyiliu-strala: Added

internal/device/energy_zone.go
vimalk78: this is for multi-socket case.

kaiyiliu-strala: Will conduct test for multi-sockets. We will need access to that or someone from community to help with that

PR #2369, opened 2025-12-12
feat(k8s): replace API server watch with kubelet /pods polling

vimalk78:
Reduces API server load on large clusters by polling local kubelet `/pods` endpoint instead of maintaining persistent watch connections.

  - Add kubeletPodInformer that polls kubelet at NODE_IP:10250/pods
  - Use downward API (status.hostIP) to get node IP
  - Add nodes/proxy RBAC for kubelet webhook authorization
  - Keep apiserver mode as fallback via kube.podInformer.mode config

### DIFF
--- a/.github/workflows/config-change.yaml
+++ b/.github/workflows/config-change.yaml
@@ -25,9 +25,11 @@ jobs:
         uses: dorny/paths-filter@v3
         id: filter
         with:
+          predicate-quantifier: every
           filters: |
             config:
               - 'config/**/*.go'
+              - '!config/**/*_test.go'
             doc:
               - 'docs/user/configuration.md'
             compose_dev:
--- a/.github/workflows/profiling.yaml
+++ b/.github/workflows/profiling.yaml
@@ -1,34 +1,20 @@
 name: Profiling Report
 on: # yamllint disable-line rule:truthy
   pull_request:
+    paths:
+      - "**/*.go"
+      - "!**/*_test.go"
+      - go.mod
+      - go.sum
+      - .github/workflows/profiling.yaml
+      - hack/reports/profiling.sh
 
 # default permissions as read only
 permissions: read-all
 
 jobs:
-  check-changes:
-    runs-on: ubuntu-latest
-    outputs:
-      changes: ${{ steps.filter.outputs.changes }}
-    steps:
-      - name: checkout source
-        uses: actions/checkout@v6
-
-      - name: filter changes
-        uses: dorny/paths-filter@v3
-        id: filter
-        with:
-          filters: |
-            changes:
-              - '**/*.go'
-              - 'go.mod'
-              - 'go.sum'
-              - '.github/workflows/profiling.yaml'
-              - 'hack/reports/profiling.sh'
-
   profiling:
-    needs: check-changes
-    if: needs.check-changes.outputs.changes == 'true' && github.repository_owner == 'sustainable-computing-io'
+    if: github.repository_owner == 'sustainable-computing-io'
     runs-on: self-hosted
     steps:
       # NOTE: In case of self-hosted runners, tools like docker, make, kubectl etc. are installed at the time of setting up the runner.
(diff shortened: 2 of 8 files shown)


vprashar2929:
@vimalk78 Can we address the CI failures?

vprashar2929:
@vimalk78 Can we also address config changes: https://github.com/sustainable-computing-io/kepler/pull/2369#issuecomment-3810468507

manifests/k8s/configmap.yaml
```diff
@@ -42,7 +42,11 @@ data:
     dev:
       fake-cpu-meter:
         enabled: false
-        zones: []
+        # enabled: true
```
vprashar2929: This should be reverted right?

vimalk78: reverted

config/config.go
```diff
@@ -117,10 +117,16 @@ type (
 		Pprof PprofDebug `yaml:"pprof"`
 	}
 
+	PodInformer struct {
+		Mode         string        `yaml:"mode"`         // "kubelet" (default) or "apiserver"
+		PollInterval time.Duration `yaml:"pollInterval"` // Poll interval for kubelet mode (default: 15s)
```
vprashar2929, line 122: Doesn't this introduces staleness in data compare to previous implementation?

internal/k8s/pod/kubelet.go
```diff
@@ -0,0 +1,288 @@
...
+	if port == 0 {
+		port = defaultKubeletPort
+	}
+	i.kubeletPort = port
+
+	// Extract InternalIP from node addresses
+	host := ""
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == corev1.NodeInternalIP {
+			host = addr.Address
```
vprashar2929, line 151: Do we have test for this in case of ipv6 address?

config/config.go
vimalk78: Yes, polling introduces some staleness compared to the watch-based informer, api-server is more sophisticated. `pollInterval` is configurable and can be tuned based on your needs.

internal/k8s/pod/kubelet.go
vimalk78: added

config/config.go
vimalk78: pls resolve if acceptable

vimalk78:
@vprashar2929 @KaiyiLiu1234 please take a look

config/config.go
vprashar2929: I think it's good that we document this someplace for reference

vprashar2929:
@vimalk78 Should we reconsider making apiserver as default (safe and always work) rather than kubelet considering it may fail in hardened environments if set to default?

vimalk78:
> fail in hardened environments

which environment? 

no issues with keeping backward compatibility and have api-server as default. what about large clusters?

vprashar2929:
> > fail in hardened environments
> 
> which environment?

clusters having strict NetworkPolicies, or custom clusters without InternalIP

> no issues with keeping backward compatibility and have api-server as default. what about large clusters?

It's configurable right and users can configure it as per their cluster needs

vimalk78:
> It's configurable right and users can configure it as per their cluster needs

kubelet failure is loud, fails fast and users can switch to use api-server
whereas api-server load increases gradually with cluster scaling, hard to notice to switch to kubelet

manifests/k8s/configmap.yaml
```diff
@@ -39,10 +39,16 @@ data:
...
       config: ""
       nodeName: ""
+      podInformer:
+        mode: kubelet          # "kubelet" (default) or "apiserver"
+        pollInterval: 15s      # Poll interval for kubelet mode (default: 15s)
     dev:
       fake-cpu-meter:
         enabled: false
-        zones: []
+        # zones:
```
vprashar2929: This should be reverted right?

vimalk78: reverted

PR #2370, opened 2025-12-15
feat(hack): add GPU time-slicing scripts for nvidia gpu

vimalk78:
- apply-gpu-timeslicing.sh: enables 4x GPU time-slicing with prerequisite checks (GPU Operator, ClusterPolicy auto-detection)
  - verify-gpu-timeslicing.sh: comprehensive verification of config
  - enable-gpu-timeslicing.yaml: ConfigMap with 4 replicas


vprashar2929:
@vimalk78 Should be chore right instead of feat?

kaiyiliu-strala:
Would you like this to be reviewed? Since it is just in hack directory, I see no issue with merging this.

PR #2372, opened 2025-12-18
feat(hwmon): implement device reader for architectures with hwmon sensors

KaiyiLiu1234:
Implemented hwmon device reader to allow Kepler to acquire power metrics (watts) from hwmon subsystem. Hwmon enablement added as an experimental feature in config.

Precise changes:
- hwmon subsystem reader added to devices called hwmon_power_meter.go
- Kepler Config updated to include experimental field called Hwmon with Enable boolean


kaiyiliu-strala:
TODO: Update documentation to include experimental hwmon enable feature.

kaiyiliu-strala:
TODO: Add unit tests for config.go

config/config.go
```diff
@@ -51,6 +54,13 @@ type (
 		Zones []string `yaml:"zones"`
 	}
 
+	// Hwmon configuration (Set in Experimental)
+	Hwmon struct {
+		Enabled *bool    `yaml:"enabled"` // Development mode settings (sensor detection should be set based on architecture in future)
+		Zones   []string `yaml:"zones"`
+		Chips   []string `yaml:"chips"`
```
vimalk78: add comments briefly explaining what are `zones` and `chips` in hwmon.
are these hwmon concepts or arm concepts?

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+	n, err := unix.Read(int(f.Fd()), b)
+	if err != nil {
+		return nil, err
+	}
+	if n < 0 {
+		return nil, fmt.Errorf("failed to read file: %q, read returned negative bytes value: %d", file, n)
+	}
+
+	return b[:n], nil
+}
```
vimalk78, line 465: can we do simple `ioutil.ReadFile` 
in rapl support, procfs also uses `filepath` and `os.ReadFile` 
https://github.com/prometheus/procfs/blob/master/sysfs/class_powercap.go#L90-L94

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+func (z *hwmonPowerZone) Energy() (Energy, error) {
+	// hwmon provides power, not energy
+	// Return 0 for interface compatibility
+	return 0, fmt.Errorf("hwmon zones do not provide energy readings")
+}
+
+func (z *hwmonPowerZone) MaxEnergy() Energy {
+	// No maximum for power sensors
+	return 0
+}
```
vimalk78, line 498: is this function used anywhere? what is the implication of returning 0 as the MaxPower

internal/device/testdata/sys/class/hwmon/hwmon0/power1_label
vimalk78, line 1: is lebel being read in hwmon_power_meter?

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+	}
+
+	return zones, nil
+}
+
+// findSensorsByType finds all sensors of a given type (e.g., "power")
+// Returns a map of sensor number -> sensor files
+func (r *sysfsHwmonReader) findSensorsByType(files []os.DirEntry, sensorType string) map[int]map[string]string {
+	sensors := make(map[int]map[string]string)
+	pattern := regexp.MustCompile(fmt.Sprintf(`^%s(\d+)_(.+)$`, sensorType))
```
vimalk78, line 299: does this get compiled only at startup?

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+// sysfsHwmonReader implements hwmonReader by reading directly from sysfs
+type sysfsHwmonReader struct {
+	basePath string // /sys/class/hwmon
+}
+
+var (
+	hwmonInvalidMetricChars = regexp.MustCompile("[^a-z0-9:_]")
+)
+
+func (r *sysfsHwmonReader) Zones() ([]EnergyZone, error) {
```
vimalk78, line 233: add a function comment

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("hwmon not available: %w", err)
+		}
+		return nil, fmt.Errorf("failed to read hwmon directory: %w", err)
+	}
+
+	var zones []EnergyZone
+	for _, entry := range hwmonDirs {
+		// check for valid hwmon devices
+		if !entry.IsDir() && !isSymlink(filepath.Join(r.basePath, entry.Name())) {
```
vimalk78, line 245: it needs to be a symlink?

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+		return nil, err
+	}
+
+	// Get human-readable chip name
+	humanName, _ := r.getHumanReadableChipName(hwmonPath)
+
+	// Scan for power sensors
+	files, err := os.ReadDir(hwmonPath)
+	if err != nil {
+		return nil, err
```
vimalk78: need better error message

internal/device/hwmon_power_meter.go
vimalk78: better move it to package level

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+	if len(zones) == 0 {
+		return nil, fmt.Errorf("no hwmon power zones found")
+	}
+
+	return zones, nil
+}
+
+func (r *sysfsHwmonReader) discoverZones(hwmonPath string) ([]EnergyZone, error) {
+	// Get chip name
+	chipName, err := r.getChipName(hwmonPath)
```
vimalk78, line 267: why not filter the chips here itself?

internal/device/hwmon_power_meter.go
```diff
@@ -0,0 +1,542 @@
...
+// WithHwmonLogger sets the logger for hwmonPowerMeter
+func WithHwmonLogger(logger *slog.Logger) HwmonOptionFn {
+	return func(pm *hwmonPowerMeter) {
+		pm.logger = logger.With("service", "hwmon")
+	}
+}
+
+// WithChipFilter sets chip names to include for monitoring
+// If empty, all chips are included
+func WithChipFilter(chips []string) HwmonOptionFn {
```
vimalk78: how will the users decide which chip names are to be included here?

internal/device/hwmon_power_meter.go
vimalk78, line 1: lets clearly document what we mean by a "zone". does hwmon_power_meter use sensor as zone?

```
hwmon0                                                     << chip
├── power1_input                                   << sensor
└── power2_input                                   << sensor
hwmon1
├── power1_input
└── power2_input
```

internal/device/hwmon_power_meter.go
kaiyiliu-strala: So this is quite interesting. Counterintuitively, you actually should not use os.ReadFile. I initially did this, but according to node exporter themselves, for reading hwmon drivers, you should use system call directly:

```
// On some machines, hwmon drivers are broken and return EAGAIN.  This causes
// Go's os.ReadFile implementation to poll forever.
//
// Since we either want to read data or bail immediately, do the simplest
// possible read using system call directly.
```

internal/device/hwmon_power_meter.go
kaiyiliu-strala: This is not MaxPower it is MaxEnergy. This is for satisfying the EnergyZone interface. Essentially, it means that because this zone only exposes Power, Energy and MaxEnergy will return 0.

internal/device/testdata/sys/class/hwmon/hwmon0/power1_label
kaiyiliu-strala: Yes absolutely. We need the label to determine what kind of "zone" this sensor satisfies at the node level (ex. cpu_power, io_power)

internal/device/hwmon_power_meter.go
kaiyiliu-strala: With the config:
```
experimental:
    hwmon:
      enabled: true
      zones: ["package", "core"]  # Filter by zone names
      chips: ["k10temp", "amdgpu"]  # Filter by chip names
```
Empty means everything will be included. Chip names are determined based on how node rapl determines them:
1. Try to construct name from device path
2. Fall back to /sys/class/hwmon/hwmonX/name file
3. Fall back to hwmon directory name (hwmon0, hwmon1, etc.)

internal/device/hwmon_power_meter.go
kaiyiliu-strala: It can be put there, but I followed rapl's setup which instead placed it in Zones() to be consistent. Please refer to rapl_sysfs_power_meter.go

internal/device/hwmon_power_meter.go
kaiyiliu-strala: I will update the documentation

internal/device/hwmon_power_meter.go
vimalk78: how will the user know about the chip names itself

internal/device/hwmon_power_meter.go
kaiyiliu-strala: yes it needs to be. All hwmon class devices in hwmon are symlinks. We need to filter it based on that to avoid folders that are not relevant.

internal/device/hwmon_power_meter.go
kaiyiliu-strala: As in remove it as a method (remove r *sysfsHwmonReader)? This can be done.

internal/device/hwmon_power_meter.go
kaiyiliu-strala: Fixed

internal/device/hwmon_power_meter.go
kaiyiliu-strala: Chip filtering was removed. The use of chip name to represent hardware monitor will remain but users won't be able to filter the chips.

internal/device/hwmon_power_meter.go
kaiyiliu-strala: So turns out this is not needed. Following the format in rapl_sysfs_power_meter.go for consistency, caching is used in the Zone() function. Thus, this function will not be run over and over.

config/config.go
kaiyiliu-strala: This will be added to the documentation and once done I will mark as resolved.

kaiyiliu-strala:
@vimalk78 I have tested locally and on the arm machine one more time. It works. Please let me know if this can be approved. All comments have been addressed.

vimalk78 (approved):

vimalk78:
need to rebase the branch on latest

kaiyiliu-strala:
Note that multi-socket testing has to be done still. Aggregated zones follow the rapl format but it is uncertain if this is always going to be consistent. Expect variations between arm machines.

kaiyiliu-strala:
Issue with Multi socket scenario:

  1. Multi-socket CPUs (legitimate aggregation case)
  2. Different power rails on the same device (should be separate)
  3. Different devices with coincidentally same labels (should be separate)

We should consider modifying the aggregation case for hwmon. @vimalk78, recording this so we know this major limitation in the future when we test this and fix this with help of community.

internal/device/hwmon_power_meter.go
kaiyiliu-strala: Also for clarification, the labels are the zones. the sensor/chip is the hwmon0. Note that I have recorded possible issues where we have different sensors/chips but the same labels.

kaiyiliu-strala:
@vimalk78 Unit tests added and codecov now passing.

PR #2373, opened 2025-12-22
feat(platform): set hostUsers explicitly

jcpunk:
The daemons require host privileges. Setting this explicitly both documents the incompatibility with user namespaces and ensures, if the default changes, the daemonset will continue to function as expected.


vprashar2929:
@jcpunk Thanks for this patch! Shouldn't this be a `chore` instead of `feat` https://github.com/sustainable-computing-io/kepler/blob/main/docs/developer/pre-commit.md#commit-message-format

jcpunk:
Changed name to `chore` in commit message.  Anything else I'm missing?

vprashar2929:
No LGTM! Thanks 🙇

PR #2374, opened 2025-12-26
feat(helm): add optional networkPolicy

jcpunk:
This adds an optional network policy for the kepler daemonset.


vprashar2929:
@jcpunk Can you use chore instead of feat?

jcpunk:
Switched to `chore`.

PR #2375, opened 2025-12-26
feat(helm): allow for startupProbe if wanted

jcpunk:
If folks want to set a startupProbe, it is now easy to load.


vprashar2929:
@jcpunk Can you use chore instead of feat?

jcpunk:
Switched to `chore`.

vprashar2929 (approved):

PR #2384, opened 2026-01-21
feat(gpu): add NVIDIA GPU power monitoring via NVML

vimalk78:
Implement GPU power monitoring using NVIDIA NVML library with support
  for time-slicing and exclusive GPU sharing modes.

  Key changes:
  - Add NVIDIA collector with process-level power attribution
  - Integrate GPU meters into monitor via WithGPUPowerMeters() option
  - Add GPU Prometheus metrics:
    - kepler_node_gpu_info (device metadata with labels)
    - kepler_node_gpu_watts (total GPU power)
    - kepler_node_gpu_idle_watts (minimum observed power)
    - kepler_node_gpu_active_watts (power above idle baseline)
    - kepler_process_gpu_watts (per-process GPU power attribution)
  - Add GPUPower field to Process and GPUStats to Snapshot
  - Enable CGO for NVML library linking
  - Add experimental.gpu.enabled config flag (default: false)

  Power attribution uses SM utilization ratio for time-slicing mode:
`P_process = P_active × (SmUtil_process / ΣSmUtil_all)`

  Idle power is tracked as minimum observed power per device. For accurate
  idle calibration, Kepler should be started before GPU workloads begin.
  
  
  
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ GOFMT=$(GOCMD) fmt
 GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 
-CGO_ENABLED ?= 0
+CGO_ENABLED ?= 1
 
 # Project parameters
 BINARY_NAME=kepler
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/sustainable-computing-io/kepler/config"
 	"github.com/sustainable-computing-io/kepler/internal/device"
 	"github.com/sustainable-computing-io/kepler/internal/device/gpu"
+	_ "github.com/sustainable-computing-io/kepler/internal/device/gpu/nvidia" // Register NVIDIA backend
 	"github.com/sustainable-computing-io/kepler/internal/exporter/prometheus"
 	"github.com/sustainable-computing-io/kepler/internal/exporter/stdout"
 	"github.com/sustainable-computing-io/kepler/internal/k8s/pod"
@@ -25,9 +26,6 @@ import (
 	"github.com/sustainable-computing-io/kepler/internal/server"
 	"github.com/sustainable-computing-io/kepler/internal/service"
 	"github.com/sustainable-computing-io/kepler/internal/version"
-
-	// Register GPU backends via init()
-	_ "github.com/sustainable-computing-io/kepler/internal/device/gpu/nvidia"
 )
 
 func main() {
@@ -133,6 +131,10 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 	if err != nil {
 		return nil, fmt.Errorf("failed to create CPU power meter: %w", err)
 	}
+
+	// GPU meters are optional - returns empty slice if not available
+	gpuMeters := createGPUMeters(logger, cfg)
+
 	var services []service.Service
 
 	var podInformer pod.Informer
@@ -153,15 +155,20 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		return nil, fmt.Errorf("failed to create resource informer: %w", err)
 	}
 
-	pm := monitor.NewPowerMonitor(
-		cpuPowerMeter,
+	// Build monitor options
+	pmOpts := []monitor.OptionFn{
 		monitor.WithLogger(logger),
 		monitor.WithResourceInformer(resourceInformer),
 		monitor.WithInterval(cfg.Monitor.Interval),
 		monitor.WithMaxStaleness(cfg.Monitor.Staleness),
 		monitor.WithMaxTerminated(cfg.Monitor.MaxTerminated),
-		monitor.WithMinTerminatedEnergyThreshold(monitor.Energy(cfg.Monitor.MinTerminatedEnergyThreshold)*monitor.Joule),
-	)
+		monitor.WithMinTerminatedEnergyThreshold(monitor.Energy(cfg.Monitor.MinTerminatedEnergyThreshold) * monitor.Joule),
+	}
+	if len(gpuMeters) > 0 {
+		pmOpts = append(pmOpts, monitor.WithGPUPowerMeters(gpuMeters))
+	}
+
+	pm := monitor.NewPowerMonitor(cpuPowerMeter, pmOpts...)
 
 	// Create Redfish service if enabled (experimental feature)
 
@@ -178,6 +185,11 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		pm,
 	)
 
+	// Add GPU meters to services for lifecycle management (Init/Shutdown)
+	for _, m := range gpuMeters {
+		services = append(services, m)
+	}
+
 	// Add Redfish service if enabled
 	var redfishService *redfish.Service
 	if cfg.IsFeatureEnabled(config.ExperimentalRedfishFeature) {
@@ -188,22 +200,6 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		} else {
 			services = append(services, rs)
 			redfishService = rs
-		}
-	}
-
-	// Discover GPU devices if enabled (experimental feature)
-	if cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
-		gpuMeters := gpu.DiscoverAll(logger)
-		if len(gpuMeters) > 0 {
-			logger.Info("GPU power monitoring enabled",
-				"vendors", len(gpuMeters),
-				"registered_vendors", gpu.RegisteredVendors())
-			// GPU meters are initialized by DiscoverAll, store for future use
-			// TODO: integrate with monitor for per-process power attribution
-			_ = gpuMeters
-		} else {
-			logger.Warn("GPU power monitoring enabled but no GPUs discovered",
-				"registered_vendors", gpu.RegisteredVendors())
 		}
 	}
 
@@ -305,4 +301,30 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 		device.WithRaplLogger(logger),
 		device.WithZoneFilter(cfg.Rapl.Zones),
 	)
+}
+
+// createGPUMeters discovers and initializes GPU power meters for all vendors.
+// Uses the registry pattern to support multiple GPU vendors (NVIDIA, AMD, Intel).
+// Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
+func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
+	if !cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
+		logger.Info("GPU feature disabled")
+		return nil
+	}
+
+	// DiscoverAll probes all registered GPU backends and returns initialized meters
+	meters := gpu.DiscoverAll(logger)
+	if len(meters) == 0 {
+		logger.Info("no GPUs discovered on this node")
+		return nil
+	}
+
+	// Log all discovered GPUs
+	for _, m := range meters {
+		logger.Info("GPU discovered",
+			"vendor", m.Vendor(),
+			"devices", len(m.Devices()))
+	}
+
+	return meters
 }
--- a/docs/user/metrics.md
+++ b/docs/user/metrics.md
@@ -95,6 +95,42 @@ These metrics provide energy and power information at the node level.
 - **Constant Labels**:
   - `node_name`
 
+#### kepler_node_gpu_active_watts
+
+- **Type**: GAUGE
+- **Description**: GPU active power (total - idle) in watts
+- **Labels**:
+  - `gpu`
+  - `gpu_uuid`
+  - `gpu_name`
+  - `vendor`
+- **Constant Labels**:
+  - `node_name`
+
+#### kepler_node_gpu_idle_watts
+
+- **Type**: GAUGE
+- **Description**: GPU idle power (auto-detected minimum) in watts
+- **Labels**:
+  - `gpu`
+  - `gpu_uuid`
+  - `gpu_name`
+  - `vendor`
+- **Constant Labels**:
+  - `node_name`
+
+#### kepler_node_gpu_watts
+
+- **Type**: GAUGE
+- **Description**: Total GPU power consumption in watts
+- **Labels**:
+  - `gpu`
+  - `gpu_uuid`
+  - `gpu_name`
+  - `vendor`
+- **Constant Labels**:
+  - `node_name`
+
 ### Container Metrics
 
 These metrics provide energy and power information for containers.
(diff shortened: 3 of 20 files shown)


kaiyiliu-strala:
Is this GPU setup separate from the interfaces we use to divide power metrics among processes, containers, etc that is used by CPU setup? I am wondering if it is possible to instead reuse what we have setup for RAPL/HWMON or is it too different?

kaiyiliu-strala:
Not a requirement for this PR but for the minimum observed power to represent idle, we could make use of a linear model by saving two data points (highest known and lowest known) and the y intercept can be kept for the idle power. If we ever encounter a scenario where we have 0 GPU utilization, the simple linear model will pick up on that and if we don't, the model should still help us provide a more accurate prediction. If we were to persist this data, we would persist the both datapoints instead of just the minimum power.

kaiyiliu-strala:
Is there anywhere we can document in detail how the NVIDIA GPU Power Monitoring works (for developers) and all of its limitations?

docs/user/metrics.md
```diff
@@ -95,6 +95,39 @@ These metrics provide energy and power information at the node level.
...
+  - `gpu`
+  - `gpu_uuid`
+  - `gpu_name`
+- **Constant Labels**:
+  - `node_name`
+
+#### kepler_node_gpu_idle_watts
+
+- **Type**: GAUGE
+- **Description**: GPU idle power (auto-detected minimum) in watts
```
kaiyiliu-strala, line 113: Two Questions regarding this:
 - How long before idle power stabilizes? (At the beginning if gpu utilization is high, its possible we get very inaccurate active power and idle power - some degree of time should pass to allow stabilization)
 - Should idle power reset after GPU mode changes?

kaiyiliu-strala:
Is there metrics for GPU memory used by a process or GPU memory usage in general? Does that need to be factored in?

kaiyiliu-strala:
Other than some minor comments and the idle power calculation, this LGTM.

Makefile
```diff
@@ -12,7 +12,7 @@ GOFMT=$(GOCMD) fmt
 GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 
-CGO_ENABLED ?= 0
```
vprashar2929, line 15: Dockerfile already handles it right? Shouldn't we keep the default as 0?

docs/user/metrics.md
vimalk78: yes this is a known issue

cmd/kepler/main.go
```diff
@@ -306,3 +302,29 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 		device.WithZoneFilter(cfg.Rapl.Zones),
 	)
 }
+
+// createGPUMeters discovers and initializes GPU power meters for all vendors.
+// Uses the registry pattern to support multiple GPU vendors (NVIDIA, AMD, Intel).
+// Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
+func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
+	if !cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
+		logger.Info("GPU feature disabled")
```
vprashar2929, line 311: If experimental is not enabled should we even log this?

vimalk78:
> Is there metrics for GPU memory used by a process or GPU memory usage in general? Does that need to be factored in?

As of now, per process gpu power is based only on SM utilization

internal/monitor/process.go
```diff
@@ -83,6 +112,47 @@ func (pm *PowerMonitor) calculateProcessPower(prev, newSnapshot *Snapshot) error
...
+					TotalPower:  stats.TotalPower,
+					IdlePower:   stats.IdlePower,
+					ActivePower: stats.ActivePower,
+				})
+			}
+		}
+		newSnapshot.GPUStats = gpuStats
+		pm.logger.Debug("GPU process power", "gpu_processes", len(gpuPowerByPID), "power_map", gpuPowerByPID)
+	} else {
+		pm.logger.Debug("No GPU meters available")
```
vprashar2929: Same for this as well

internal/monitor/process.go
```diff
@@ -83,6 +112,47 @@ func (pm *PowerMonitor) calculateProcessPower(prev, newSnapshot *Snapshot) error
...
+			}
+			// Aggregate power across all meters
+			for pid, watts := range power {
+				gpuPowerByPID[pid] += watts
+			}
+
+			// Collect GPU device stats for debugging/monitoring
+			devices := meter.Devices()
+			for _, dev := range devices {
+				stats, err := meter.GetDevicePowerStats(dev.Index)
```
vprashar2929, line 136: Aren't we calculating twice once in firstProcessRead and here?

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -191,6 +213,11 @@ func (c *PowerCollector) Describe(ch chan<- *prometheus.Desc) {
 		ch <- c.podCPUJoulesDescriptor
 		ch <- c.podCPUWattsDescriptor
 	}
+
+	// GPU device power metrics (always exposed if available)
```
vprashar2929: Shouldn't these be controlled in same way as others are through metricsLevel config?

internal/device/gpu/nvidia/collector.go
```diff
@@ -268,48 +289,74 @@ func (c *GPUPowerCollector) attributeTimeSlicing(deviceIndex int, result map[uin
...
 		return nil
 	}
 
-	// Update lastSeen timestamp
-	var maxTimestamp uint64
-	for _, u := range utils {
-		if u.Timestamp > maxTimestamp {
-			maxTimestamp = u.Timestamp
+	// Step 3: Build utilization map by PID
+	utilMap := make(map[uint32]uint32) // PID -> ComputeUtil
```
vprashar2929, line 317: Minor: A new map will be allocated each time on fuc call Can't we reuse it somehow?

internal/exporter/prometheus/collector/gpuinfo.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"sync"
+
+	prom "github.com/prometheus/client_golang/prometheus"
+	"github.com/sustainable-computing-io/kepler/internal/monitor"
+)
+
+// gpuInfoCollector collects GPU device info metrics.
+type gpuInfoCollector struct {
+	sync.Mutex
```
vprashar2929, line 16: why not sync.RWMutex?

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -137,6 +144,20 @@ func NewPowerCollector(monitor PowerDataProvider, nodeName string, logger *slog.
 
 		podCPUJoulesDescriptor: joulesDesc("pod", "cpu", nodeName, []string{podID, "pod_name", "pod_namespace", "state", zone}),
 		podCPUWattsDescriptor:  wattsDesc("pod", "cpu", nodeName, []string{podID, "pod_name", "pod_namespace", "state", zone}),
+
+		// GPU device power metrics (node-level)
+		gpuTotalWattsDescriptor: prometheus.NewDesc(
+			prometheus.BuildFQName(keplerNS, "node", "gpu_watts"),
+			"Total GPU power consumption in watts",
+			[]string{"gpu", "gpu_uuid", "gpu_name"}, prometheus.Labels{nodeNameLabel: nodeName}),
```
vprashar2929: Shouldn't we also include vendor here so that user's can query based upon different vendors they have in their cluster?

internal/monitor/monitor.go
```diff
@@ -346,6 +358,7 @@ func (pm *PowerMonitor) refreshSnapshot() error {
 		"containers", len(newSnapshot.Containers),
 		"vms", len(newSnapshot.VirtualMachines),
 		"pods", len(newSnapshot.Pods),
+		"gpu_stats", len(newSnapshot.GPUStats),
```
vprashar2929: better name than gpu_stats?

vimalk78:
> Is this GPU setup separate from the interfaces we use to divide power metrics among processes, containers, etc that is used by CPU setup? I am wondering if it is possible to instead reuse what we have setup for RAPL/HWMON or is it too different?

there are few differences
1. Energy vs Power: RAPL returns cumulative energy, so kepler must calculate the diff periodically to get instantaneous power. whereas for NVIDIA GPUs we get instantaneous power directly
2. % resource usage: in case of CPUs we get cumulative values of cpu time for process and node from /proc, and so must calculate the %cpu usage by doing periodic diff. whereas for NVIDIA GPU we get it directly from NVML as SMUtil from NVML
so there is lot more logic to be done in case of RAPL, and lot less for NVML as the driver does most of the work

internal/monitor/process.go
```diff
@@ -83,6 +112,47 @@ func (pm *PowerMonitor) calculateProcessPower(prev, newSnapshot *Snapshot) error
...
+					Name:        dev.Name,
+					Vendor:      string(dev.Vendor),
+					TotalPower:  stats.TotalPower,
+					IdlePower:   stats.IdlePower,
+					ActivePower: stats.ActivePower,
+				})
+			}
+		}
+		newSnapshot.GPUStats = gpuStats
+		pm.logger.Debug("GPU process power", "gpu_processes", len(gpuPowerByPID), "power_map", gpuPowerByPID)
```
vprashar2929: Can't we just keep count and not logging map?

internal/exporter/prometheus/collector/power_collector.go
```diff
@@ -340,6 +370,17 @@ func (c *PowerCollector) collectProcessMetrics(ch chan<- prometheus.Metric, stat
 				zoneName,
 			)
 		}
+
+		// GPU power metric (only for processes actively using GPU)
+		if proc.GPUPower > 0 {
+			ch <- prometheus.MustNewConstMetric(
```
vprashar2929, line 380: Should we also include the attribution method as metric label? like `attribution_method=equal`

Makefile
vimalk78: local builds will fail.

cmd/kepler/main.go
vimalk78: yes. better be clear.

internal/monitor/process.go
vimalk78: does this show if feature is disabled?

internal/exporter/prometheus/collector/gpuinfo.go
vimalk78: what is the advantage?

cmd/kepler/main.go
vprashar2929: If I haven't enabled any experimental feature then why to log this? Even we don't log experimental config unless we enable any?

internal/monitor/process.go
vprashar2929: yes

internal/exporter/prometheus/collector/gpuinfo.go
vprashar2929: because we don't modify gpu info in every scrape. It is done once at the time of running

cmd/kepler/main.go
vimalk78: this is subjective opinion. GPU power is a big feature, logging or not logging this line does not impact much.
i can remove this in the interest of resolving the comment

vprashar2929: My view is that we should keep it consistent. We didn't apply this in case of Redfish or hwmon so doing it would be inconsistent

internal/monitor/process.go
vimalk78: No. firstProcessRead runs once on startup, calculateProcessPower runs on subsequent intervals. Mutually exclusive paths.

internal/device/gpu/nvidia/collector.go
vimalk78: Map is small (few PIDs), short-lived, GC handles it.

internal/monitor/monitor.go
vimalk78: removed.

internal/monitor/process.go
vimalk78: the information could be useful in debugging, or not? only the processes using the gpu are logged

internal/exporter/prometheus/collector/gpuinfo.go
vimalk78: this follows same pattern as cpu_info collector

internal/exporter/prometheus/collector/power_collector.go
vimalk78: this is only a fallback mechanism. adding this label will produce different series
```
  kepler_process_gpu_watts{pid="123", method="utilization"} 50
  kepler_process_gpu_watts{pid="123", method="equal"} 45
```
lets decide based on real usage

internal/monitor/process.go
vimalk78: removed

internal/exporter/prometheus/collector/gpuinfo.go
vprashar2929: yeah I saw. We should update there as well

internal/exporter/prometheus/collector/power_collector.go
vprashar2929: or maybe in gpu_info metric? 🤔

Makefile
vprashar2929: `CGO_ENABLED=1 make build` should work fine

vimalk78:
> Is there anywhere we can document in detail how the NVIDIA GPU Power Monitoring works (for developers) and all of its limitations?

Added a doc

Makefile
vimalk78: of course. now that we have a dependency on nvml, we need CGO enabled

vimalk78: what is the purpose to keep it CGO=0

kaiyiliu-strala: With CGO set to 0, it will produce static builds (no dependencies and cross compiling is easier). But if it is necessary to have for the GPU setup then it must be done. To clarify we aren't making two separate builds right (one for GPU and one for no gpu)?

kaiyiliu-strala:
@vimalk78 I have no other concerns regarding this. I can approve but do fully address @vprashar2929 before merging.

Makefile
vprashar2929: I don't think so. My initial understanding was that since it was an experimental feature should we make this change now or wait until it is matured? @vimalk78 Final thoughts?

vprashar2929:
@vimalk78 Can you rebase?

vimalk78:
> @vimalk78 Can you rebase?

rebased

vimalk78:
> @vimalk78 I have no other concerns regarding this. I can approve but do fully address @vprashar2929 before merging.

what is pending to be addressed?

Makefile
vprashar2929: We can take this up separately. Resolving this for now

PR #2386, opened 2026-01-28
test: resolve race condition in power collector tests

vprashar2929:
This commit resolves the race condition in power collector tests. The test spawned gorutine that could still be running when the test ended causing a data race condition
It adds synchronization to ensure monitor goroutine completes before test cleanup runs eliminating the race condition.


vimalk78:
> This commit resolves the race condition in power collector tests. It adds synchronization to ensure monitor goroutine completes before test cleanup runs

can you please add comments about the issue was, and how it is fixed with this change

vprashar2929:
Updated the comment

kaiyiliu-strala:
lgtm now

PR #2387, opened 2026-01-30
fix(helm): use image.tag="" otherwise it will always use the latest tag

yellowhat:
Currently the rendered helm chart will always use the `latest` image tag, ignoring the `appVersion` value.

The template is already checking if `image.tag` is empty, otherwise use `.Chart.AppVersion`:

```
image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```


vimalk78:
@yellowhat Thanks for the fix.
the test fails because you have created PR from your fork's main branch. can you change the branch to be other than `main`?

yellowhat:
Sure

yellowhat:
Superseeded by https://github.com/sustainable-computing-io/kepler/pull/2389

PR #2388, opened 2026-01-30
feat(gpu): add idle power detection and configurable override

vimalk78:
Track minimum observed GPU power only when no compute processes are
  running to establish true idle baseline. Add configurable idle power
  override (experimental.gpu.idle-power) for cases where true idle
  cannot be observed.

image

  Test: **GPU Idle Power Detection**

  Tested on a Tesla T4 node with five scenarios:

  1. **Kepler starts idle** (19:30–19:42): No workloads running. Kepler observes true idle power at ~17W.
  2. **Workload starts** (19:42–20:03): Idle baseline holds at 17W. Process power correctly shows ~15-55W (total minus idle).
  3. **Kepler restarted under load** (~20:05): Idle drops to 0W (never observed true idle). Process power jumps to ~60-70W since all GPU power is attributed as active — conservative but correct behavior.
  4. **Workload stops** (~20:10): Kepler detects true idle, brief spike as GPU settles, then locks in ~17W baseline.
  5. **Workload restarts** (20:12+): Idle baseline persists at 17W. Process power returns to correct ~15-20W attribution.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -140,6 +140,12 @@ type (
 	ExperimentalGPU struct {
 		// Enabled controls whether GPU power monitoring is enabled
 		Enabled *bool `yaml:"enabled"`
+
+		// IdlePower is the GPU idle power in Watts. When set (> 0), this value
+		// is used instead of auto-detected idle power. Useful when Kepler cannot
+		// observe true idle (e.g. GPUs always under load).
+		// 0 means auto-detect (track minimum power when no compute processes are running).
+		IdlePower float64 `yaml:"idlePower"`
 	}
 
 	// Experimental contains experimental features (no stability guarantees)
@@ -256,7 +262,8 @@ const (
 	ExperimentalHwmonZonesFlag   = "experimental.hwmon.zones"
 
 	// Experimental GPU flags
-	ExperimentalGPUEnabledFlag = "experimental.gpu.enabled"
+	ExperimentalGPUEnabledFlag   = "experimental.gpu.enabled"
+	ExperimentalGPUIdlePowerFlag = "experimental.gpu.idle-power"
 
 // WARN:  dev settings shouldn't be exposed as flags as flags are intended for end users
 )
@@ -414,6 +421,7 @@ func RegisterFlags(app *kingpin.Application) ConfigUpdaterFn {
 
 	// experimental GPU
 	gpuEnabled := app.Flag(ExperimentalGPUEnabledFlag, "Enable experimental GPU power monitoring").Default("false").Bool()
+	gpuIdlePower := app.Flag(ExperimentalGPUIdlePowerFlag, "GPU idle power in Watts (0 = auto-detect from idle observations)").Default("0").Float64()
 
 	return func(cfg *Config) error {
 		// Logging settings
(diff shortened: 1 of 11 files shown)


config/config.go
```diff
@@ -628,11 +636,16 @@ func applyGPUConfig(cfg *Config, flagsSet map[string]bool, enabled *bool) {
...
+
+	if flagsSet[ExperimentalGPUIdlePowerFlag] {
+		cfg.Experimental.GPU.IdlePower = *idlePower
+	}
 }
 
 // hasGPUFlags returns true if any GPU experimental flags are set
 func hasGPUFlags(flagsSet map[string]bool) bool {
-	return flagsSet[ExperimentalGPUEnabledFlag]
+	return flagsSet[ExperimentalGPUEnabledFlag] ||
```
sunya-ch: This logic will make setting idle power affect enable flag (if idle is set, it considers gpu enabled even though the enable flag is false).
Instead, I think we should not apply idle flag if it is not enabled. 

Meaning that we should leave flag check to only enable flag.

sunya-ch (commented):
Nice work! The logic behind replacing min with checking no proc makes sense to me. Just leave a minor comment.

config/config.go
vimalk78: removed the function, and added check directly in applyGPUConfig

manifests/helm/kepler/values.yaml
```diff
@@ -114,6 +114,7 @@ config:
       zones: [] # List of zones to enable (default enable all)
     gpu:
       enabled: false # Enable experimental GPU power monitoring
+      idlePower: 0 # GPU idle power in Watts (0 = auto-detect)
```
sunya-ch, line 117: Can this field omit especially when enabled is false?

sunya-ch (approved):
LGTM. Leave a question.

manifests/helm/kepler/values.yaml
vimalk78: yes, it can be omitted.

config/config.go
```diff
@@ -628,11 +636,11 @@ func applyGPUConfig(cfg *Config, flagsSet map[string]bool, enabled *bool) {
 	if flagsSet[ExperimentalGPUEnabledFlag] {
 		cfg.Experimental.GPU.Enabled = enabled
 	}
-}
 
-// hasGPUFlags returns true if any GPU experimental flags are set
-func hasGPUFlags(flagsSet map[string]bool) bool {
-	return flagsSet[ExperimentalGPUEnabledFlag]
+	// Only apply idle power if GPU is enabled
+	if cfg.IsFeatureEnabled(ExperimentalGPUFeature) && flagsSet[ExperimentalGPUIdlePowerFlag] {
```
vprashar2929, line 641: can we have UT for this?

vimalk78: we can add in subsequent PRs

internal/device/gpu/nvidia/collector.go
```diff
@@ -189,6 +211,15 @@ func (c *GPUPowerCollector) getDevicePowerStatsLocked(deviceIndex int) (gpu.GPUP
 	}, nil
 }
 
+// SetIdlePower sets the configured idle power in Watts.
+// When set (> 0), this value always takes precedence over observed idle power.
+// Set to 0 to use auto-detected idle power.
+func (c *GPUPowerCollector) SetIdlePower(watts float64) {
```
kaiyiliu-strala, line 217: very minor but you could add basic validation to setting the idle power with this public method (ensure watts is not 0 or or negative)

kaiyiliu-strala:
lgtm

internal/device/gpu/nvidia/collector.go
vimalk78: updated

KaiyiLiu1234 (approved):
lgtm

PR #2389, opened 2026-01-30
fix(helm): use image.tag="" otherwise it will always use the latest tag

yellowhat:
Currently the rendered helm chart will always use the `latest` image tag, ignoring the `appVersion` value.

The template is already checking if `image.tag` is empty, otherwise use `.Chart.AppVersion`:

```
image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```


vimalk78 (approved):
LGTM

PR #2392, opened 2026-02-02
chore(helm): trigger DaemonSet rollout on ConfigMap changes

yellowhat:
Currently a change to the `ConfigMap` does not trigger an update of the `DaemonSet`, you need to manually restart/delete pods.

This PR adds the `checksum/config` annotation to `DaemonSet` pod template that includes a SHA256 hash of the `ConfigMap` content.

This ensures pods are automatically restarted when configuration changes are deployed via `helm upgrade`.

### DIFF
--- a/manifests/helm/kepler/templates/daemonset.yaml
+++ b/manifests/helm/kepler/templates/daemonset.yaml
@@ -20,10 +20,11 @@ spec:
         {{- with .Values.labels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.annotations }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- with .Values.annotations }}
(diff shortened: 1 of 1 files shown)


vprashar2929:
@yellowhat Isn't using operator address this concern: https://github.com/sustainable-computing-io/kepler-operator/blob/main/docs/user/installation/kubernetes.md, https://github.com/sustainable-computing-io/kepler-operator/blob/main/docs/user/reference/custom-configmaps.md

yellowhat:
I am not familiar with the `kepler-operator`, but if you just use the helm chart you have this problem.

vprashar2929:
@yellowhat Can you update the commit message from `fix` to `chore` according to: https://github.com/sustainable-computing-io/kepler/blob/main/docs/developer/pre-commit.md#commit-message-format

yellowhat:
Done

vprashar2929 (approved):

PR #2393, opened 2026-02-02
test(monitor): resolve race condition in mock cleanup

vprashar2929:
This commit removes t.Cleanup() from SetExpectations() that raced with singleflight go routines and adds explicit ClearExpectations() helper for test isolation.


vprashar2929:
closing this in lieu of https://github.com/sustainable-computing-io/kepler/pull/2394

PR #2394, opened 2026-02-02
fix(monitor): resolve race condition in collection goroutine lifecycle

vimalk78:
fix(monitor): resolve race conditions in collection tests

  Add sync.WaitGroup to track goroutines spawned by scheduleNextCollection
  so Run() and Shutdown() wait for in-flight collections to complete.

  Test fixes:
  - TestConcurrentRegistration: Add assert.Eventually to wait for first
    snapshot before gathering (waitForData() goroutine scheduling race)
  - TestCollectionCancellation: Add t.Cleanup to ensure Run() goroutine
    completes before test teardown, increase timeout from 5ms to 100ms
  - Remove unsafe t.Cleanup that cleared mock expectations while background
    goroutines were still reading them
  - Add explicit ClearExpectations() for subtest isolation

### Race Condition Fix Summary

  The race conditions in these tests were caused by insufficient synchronization
  between test assertions and background goroutines:

  1. **TestConcurrentRegistration**: The `waitForData()` goroutine in `PowerCollector`
     wasn't guaranteed to be scheduled before `Gather()` ran, causing collectors
     to return empty metrics on slow CI runners.

  2. **TestCollectionCancellation**: When `assertDataUpdated` timed out (5ms) and
     called `t.Fatalf`, test teardown began while the `Run()` goroutine was still
     mid-collection, racing with mock method calls.

  ### Fix

  - Add `assert.Eventually` to wait for the first snapshot before gathering
  - Add `t.Cleanup` to ensure `Run()` goroutine completes before test teardown
  - Increase `assertDataUpdated` timeout from 5ms to 100ms for slow CI
  - The `collectionWg sync.WaitGroup` (added earlier) ensures clean goroutine
    shutdown in `Run()` and `Shutdown()`

### DIFF
--- a/internal/exporter/prometheus/collector/power_collector_concurrency_test.go
+++ b/internal/exporter/prometheus/collector/power_collector_concurrency_test.go
@@ -358,6 +358,12 @@ func TestConcurrentRegistration(t *testing.T) {
 		monitorWg.Wait()
 	})
 
+	// Wait for monitor to produce first snapshot
+	assert.Eventually(t, func() bool {
+		_, err := fakeMonitor.Snapshot()
+		return err == nil
+	}, 2*time.Second, 10*time.Millisecond)
+
 	// Create registries
 	registries := make([]*prometheus.Registry, numRegistries)
 	for i := range numRegistries {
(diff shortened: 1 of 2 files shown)


vprashar2929:
@vimalk78 Isn't this same https://github.com/sustainable-computing-io/kepler/pull/2393 😅

kaiyiliu-strala:
I believe Claude also recommends this as the solution

PR #2402, opened 2026-02-04
chore: add podman-compose fix and readme

sunya-ch:
The current dev compose cannot be up with podman-compose command.

The main problem with podman-compose compatibility is the include directive, which is a Docker Compose v2.20+ feature that podman-compose doesn't fully support.

This PR adds that fix by explicitly merge including compose and add README with instruction to install and maintain for podman-compose compatibility. 

Also, must note that the Scaphandre service is not available for ARM architecture (not available image for this architecture).

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,10 @@ Choose your preferred method:
 make build && sudo ./bin/kepler
 
 # ✨ Docker Compose (with Prometheus & Grafana)
-cd compose/dev && docker-compose up -d
+cd compose/dev && docker compose up -d
+
+# 🐋 Podman Compose
+cd compose/dev && podman-compose -f compose-podman.yaml up -d
 
 # 🐳 Kubernetes with Kustomize
 kubectl kustomize manifests/k8s | \
--- a/compose/dev/README-podman.md
+++ b/compose/dev/README-podman.md
@@ -1,0 +1,177 @@
+<!-- SPDX-FileCopyrightText: 2025 The Kepler Authors -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# Podman Compose Setup for Kepler Development
+
+This directory contains a podman-compose compatible version of the development environment.
+
+## Why a Separate File?
+
+The standard `compose.yaml` uses Docker Compose v2.20+ features like the `include` directive, which podman-compose doesn't fully support. The `compose-podman.yaml` file merges all services from:
+
+- `../compose.yaml` (base Kepler services)
+- `../monitoring/compose.yaml` (Prometheus & Grafana)
+- `./override.yaml` (dev-specific overrides)
+
+into a single file that works with podman-compose.
+
+## Usage
+
+### With Podman Compose
+
+```bash
+cd compose/dev
+
+# Start all services
+podman-compose -f compose-podman.yaml up -d
+
+# View logs
+podman-compose -f compose-podman.yaml logs -f kepler-dev
+
+# Stop and clean up
+podman-compose -f compose-podman.yaml down --volumes
+```
+
+### With Docker Compose (Standard)
+
+For Docker Compose users, continue using the standard file:
+
+```bash
+cd compose/dev
+
+# Start all services
+docker compose up -d
+
+# View logs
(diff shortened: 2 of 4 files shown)


vimalk78:
I tested this PR.



**Testing Environment:** Fedora 42, Podman 5.7.1, `podman-compose` 1.5.0, `podman compose` (docker-compose v5.0.1 backend)

---

### What works

- `docker compose up -d` works out of the box with the standard `compose.yaml` (no changes needed)
- `sudo podman compose up -d` can start all 6 services using `compose-podman.yaml`
- All services come up, kepler reads RAPL, scaphandre and node-exporter collect metrics

### Issues found

#### 1. Grafana cannot reach Prometheus (network isolation)

Podman compose creates all networks with `"isolate": "true"` by default, which blocks inter-container communication. Docker compose does not do this.

```
Post "http://prometheus:9090/api/v1/query": dial tcp: lookup prometheus: i/o timeout
```

- Tried `driver_opts: { isolate: "false" }` in compose file → ignored ([known bug](https://github.com/containers/podman/issues/27349))
- Tried creating networks manually (`podman network create`) and using `external: true` → DNS stops working entirely (`connection refused` on port 53)
- Docker compose on the same machine creates networks without isolation and DNS works fine

#### 2. RAPL requires sudo

Docker daemon runs as root, so `docker compose` accesses RAPL without sudo. Podman is rootless by default, so `sudo podman compose` is required. This also means images are stored in root's storage, separate from the user's.

#### 3. `podman-compose` (Python) vs `podman compose` (plugin)

- `podman-compose`: `-d` flag doesn't properly detach from terminal
- `podman compose`: `-d` works correctly (uses docker-compose backend)
- `podman-compose`: `build --network=host` not supported (builds fail when container can't reach internet)

#### 4. `include` directive works with `podman compose`

Tested: `podman compose -f compose.yaml config` parses the `include` directive correctly. However, `podman compose` has a bug where **dependencies from included files are not honored**:

```bash
# Docker: starts prometheus (grafana's dependency from included file)
docker compose up grafana kepler-dev -d  → starts grafana, prometheus, kepler-dev, sushy-static ✅

# Podman: skips prometheus
sudo podman compose up grafana kepler-dev -d  → starts grafana, kepler-dev, sushy-static ❌

# Podman with flattened file: starts prometheus correctly
sudo podman compose -f compose-podman.yaml up grafana kepler-dev -d  → starts all 4 ✅
```

This validates the need for `compose-podman.yaml` to fix the dependency resolution.

#### 5. SELinux (Fedora/RHEL)

When running rootless podman, bind mounts fail with `permission denied` due to SELinux. This is moot since sudo is required anyway for RAPL (sudo bypasses SELinux).

### What we couldn't resolve

The network isolation issue (`isolate: true`) appears to be a Podman bug with no working workaround from the compose file. This means **Grafana cannot connect to Prometheus** using service names when using podman compose on native Linux.

### Suggestions

1. Could you confirm your test environment? (macOS with `podman machine` may have different networking behavior)
2. Consider documenting in README-podman.md:
   - `sudo` is required for RAPL
   - Recommend `podman compose` over `podman-compose`
   - Known networking limitations on Linux
3. Minor: remove `# Made with Bob` comment (line 204)

vprashar2929:
I think we can simplify this by having a single compose manifest. Adding this will increase the effort to maintain both manifests and duplication

vimalk78:
> I think we can simplify this by having a single compose manifest. Adding this will increase the effort to maintain both manifests and duplication

we cannot have single compose file. `podman compose` does not start dependencies from other include compose files, thats why we need a single composite compose file

sunya-ch:
I totally forgot about this PR. 
Will be back to work on this within this month.
@vimalk78 @vprashar2929 Thank you for the review.

krishpinto:
Hey @sunya-ch , I saw your comment that you're planning to get back to it, but are you still picking it up, or should I go ahead and try to do it? I went through the other review comments too, so happy to start working through those.

sunya-ch:
@krishpinto
Please go ahead taking over this PR. Let me close this PR.

PR #2407, opened 2026-02-07
ci: switch to using large runners

vprashar2929:
This commit switches to using large runners for the CI pipeline.

### DIFF
--- a/internal/resource/vm.go
+++ b/internal/resource/vm.go
@@ -12,7 +12,7 @@ import (
 
 var (
 	// QEMU/KVM patterns - matches both qemu-system-* and qemu-kvm variants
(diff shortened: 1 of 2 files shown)


SamYuan1990 (commented):
May I have more background for this PR?

SamYuan1990:
https://github.com/sustainable-computing-io/kepler/actions/runs/21778739792/job/62839465756
https://github.com/sustainable-computing-io/kepler/actions/runs/21779214986/job/62840593059

I random picked another job, also 7mins ... which means the large runner doesn't save time for the PR job.... so why do we switch to a large runner?

SamYuan1990 (approved):
LGTM

PR #2408, opened 2026-02-07
test(monitor): increase assertion timeout

vprashar2929:
This commit increases the assertion timeout for the monitor test. The 5ms timeoit on assertDataUpdated is too tight for CI The go routine needs to be scheduled, sleep 1ms in assertDataChannelEmpty, then run the collectionLoop all before the main goroutine exits. On CI runners that doesn't happen in 5ms

c164049e fixed the exact same problem in TestCollectionCancellation

### DIFF
--- a/internal/monitor/monitor_collection_test.go
+++ b/internal/monitor/monitor_collection_test.go
@@ -113,7 +113,7 @@ func TestPeriodicCollection(t *testing.T) {
 	}()
 
 	t.Log("Waiting for first collection")
-	assertDataUpdated(t, dataCh, 5*time.Millisecond, "expected first collection as soon as run is invoked")
+	assertDataUpdated(t, dataCh, 100*time.Millisecond, "expected first collection as soon as run is invoked")
 	assert.GreaterOrEqual(t, energyCalls.Load(), int32(1), "Should have made at least one energy call")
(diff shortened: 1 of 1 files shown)


vprashar2929:
Addresses CI failure in #2369

internal/monitor/monitor_collection_test.go
```diff
@@ -113,7 +113,7 @@ func TestPeriodicCollection(t *testing.T) {
 	}()
 
 	t.Log("Waiting for first collection")
-	assertDataUpdated(t, dataCh, 5*time.Millisecond, "expected first collection as soon as run is invoked")
+	assertDataUpdated(t, dataCh, 100*time.Millisecond, "expected first collection as soon as run is invoked")
```
vimalk78, line 116: if we have to increase this timeout to pass the test, then test itself requires rethinking.

vprashar2929: maybe it is to early for runner and not a test issue? 🤔

vimalk78: is this still failing? i didn't observe recently

vprashar2929: me neither. What are your thoughts? Should we close it or have it?

vimalk78: close it.

PR #2411, opened 2026-02-08
ci(codecov): raise patch coverage target from 70% to 90%

vprashar2929:
At 70% well tested files in a PR can mask entirely untested
files because codecov computes patch coverage as a blended
average across all changed lines. 90% limits uncovered lines to
10% preventing whole functions from passing without tests

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,7 +8,7 @@ coverage:
         threshold: 1%
     patch:
       default:
-        target: 70%
+        target: 90%
 
 ignore:
   - internal/device/gpu/nvidia/nvml_lib.go


vimalk78:
is it path coverage or patch coverage?

vimalk78:
can we increase it in small steps? keep it at 80%?

vprashar2929:
> can we increase it in small steps? keep it at 80%?

I feel 80% would still be borderline and still some of the line will remain untested making it ineffective

vimalk78:
> > can we increase it in small steps? keep it at 80%?
> 
> I feel 80% would still be borderline and still some of the line will remain untested making it ineffective

yes anything less than 100% will mean untested lines. but we have to balance the effort against benefit

vimalk78:
what is the need for this?

vprashar2929:
> what is the need for this?

because patches that requires tests can be easily skipped by code patch that is heavily tested > https://github.com/sustainable-computing-io/kepler/pull/2411#issue-3912323382

Issue #2397, opened 2026-02-03
we can add in subsequent PRs


vprashar2929:
we can add in subsequent PRs

_Originally posted by @vimalk78 in https://github.com/sustainable-computing-io/kepler/pull/2388#discussion_r2758327569_

vprashar2929:
closing as #2412 addresses it

PR #2412, opened 2026-02-08
test(config): add UT for GPU idle power

vprashar2929:
Addresses https://github.com/sustainable-computing-io/kepler/issues/2397


vprashar2929:
This https://github.com/sustainable-computing-io/kepler/pull/2412#issuecomment-3866635496 is addressed in #2410

vimalk78 (approved):

PR #2415, opened 2026-02-10
chore: remove GPU logs when feature is disabled

vprashar2929:
This commit removes the GPU related log noise that appears when the experimental GPU feature is disabled

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -362,7 +362,6 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 // Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
 func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
 	if !cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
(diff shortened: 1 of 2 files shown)


cmd/kepler/main.go
```diff
@@ -353,7 +353,6 @@ func createCPUMeter(logger *slog.Logger, cfg *config.Config) (device.CPUPowerMet
 // Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
 func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
 	if !cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
-		logger.Info("GPU feature disabled")
```
vimalk78, line 365: i don't agree that this is noise.

vprashar2929: This creates confusion. If something is not used why logging then?

vimalk78: because kepler has this feature present. just not enabled.

vprashar2929: Not all experimental features, such as hwmon or redfish, follow this approach. This seems to diverge from the standard design principles we typically adhere to. It also brings back similarities to the earlier Kepler design.

PR #2417, opened 2026-02-12
feat(kvm): support kvm vm processes on proxmox VE

brunnert:
I recently tried using Kepler in a Proxmox virtual environment (https://www.proxmox.com/en/). Although Proxmox uses KVM/QEMU for virtualisation, Kepler was unable to detect the corresponding processes as virtual machines because they look like this:

`/usr/bin/kvm -id 109 -name some-vm-name [...]`

This can easily be fixed by adjusting the qemuPattern in the internal/resource/vm.go file, as demonstrated in this pull request. Therefore, it would be great if this fix could be merged.

### DIFF
--- a/internal/resource/vm.go
+++ b/internal/resource/vm.go
@@ -12,7 +12,7 @@ import (
 
 var (
 	// QEMU/KVM patterns - matches both qemu-system-* and qemu-kvm variants
-	qemuPattern = regexp.MustCompile(`(bin/qemu-system-\w+|libexec/qemu-kvm)`)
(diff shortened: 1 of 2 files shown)


vprashar2929:
@brunnert Thanks for this addition. Can you try raising the PR from different branch than `main` as one workflow fails for which I have a fix https://github.com/sustainable-computing-io/kepler/pull/2418 meanwhile?

brunnert:
> @brunnert Thanks for this addition. Can you try raising the PR from different branch than `main` as one workflow fails for which I have a fix #2418 meanwhile?

@vprashar2929 Here is a replacement PR from a different source branch: https://github.com/sustainable-computing-io/kepler/pull/2419 - If this doesn't work, I will wait until your fix is merged and update my branch accordingly.

PR #2420, opened 2026-02-15
test: add fake NVML backend for GPU dev/testing

vprashar2929:
This commit adds a FakeNVMLBackend that implements the NVMLBackend/NVMLDevice interfaces, enabling GPU feature development and e2e testing without real NVIDIA hardware.

- Add FakeNVMLBackend and FakeNVMLDevice in nvidia package
- Add WithNVMLBackend() collector option for backend injection
- Add dev.fake-gpu-meter config (enabled, deviceCount, sharingMode)
- Wire fake backend in createGPUMeters() when config is enabled
- Add bare-metal e2e GPU tests (metrics presence, labels, conservation)
- Add fake-gpu-meter config to all deployment manifests
- Update e2e testing docs with gpu_test.go

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -11,11 +11,12 @@ import (
 	"syscall"
 
 	"github.com/alecthomas/kingpin/v2"
+	"k8s.io/utils/ptr"
 
 	"github.com/sustainable-computing-io/kepler/config"
 	"github.com/sustainable-computing-io/kepler/internal/device"
 	"github.com/sustainable-computing-io/kepler/internal/device/gpu"
-	_ "github.com/sustainable-computing-io/kepler/internal/device/gpu/nvidia" // Register NVIDIA backend
+	"github.com/sustainable-computing-io/kepler/internal/device/gpu/nvidia"
 	"github.com/sustainable-computing-io/kepler/internal/exporter/prometheus"
 	"github.com/sustainable-computing-io/kepler/internal/exporter/stdout"
 	"github.com/sustainable-computing-io/kepler/internal/k8s/pod"
@@ -319,6 +320,10 @@ func createPrometheusExporter(
 // Uses the registry pattern to support multiple GPU vendors (NVIDIA, AMD, Intel).
 // Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
 func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
+	if fake := cfg.Dev.FakeGPUMeter; ptr.Deref(fake.Enabled, false) {
+		return createFakeGPUMeter(logger, fake.DeviceCount, fake.SharingMode)
+	}
+
 	if !cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
 		return nil
 	}
@@ -338,4 +343,57 @@ func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMete
 	}
 
 	return meters
+}
+
+// createFakeGPUMeter creates a GPU meter backed by FakeNVMLBackend for
+// development and testing. Not for production use.
+func createFakeGPUMeter(logger *slog.Logger, deviceCount int, sharingMode string) []gpu.GPUPowerMeter {
+	if deviceCount <= 0 {
+		logger.Error("invalid fake GPU device count", "deviceCount", deviceCount)
+		return nil
+	}
+
+	var mode nvidia.ComputeMode
+	switch sharingMode {
+	case "exclusive":
+		mode = nvidia.ComputeModeExclusiveProcess
+	case "time-slicing":
+		mode = nvidia.ComputeModeDefault
+	default:
+		logger.Error("invalid fake GPU sharing mode", "sharingMode", sharingMode)
+		return nil
+	}
+
+	logger.Warn("using FAKE NVML backend (dev/testing mode)",
+		"deviceCount", deviceCount,
+		"sharingMode", sharingMode)
+
+	devices := make([]*nvidia.FakeNVMLDevice, deviceCount)
+	for i := range devices {
+		devices[i] = &nvidia.FakeNVMLDevice{
+			Idx:         i,
+			DeviceUUID:  fmt.Sprintf("FAKE-GPU-%04d", i),
+			DeviceName:  "Fake NVIDIA GPU",
+			DevicePower: 225 * device.Watt,
+			Mode:        mode,
+		}
+	}
+	// TODO: Add fake GPU processes when e2e workloads can supply stable PIDs.
+
+	backend := nvidia.NewFakeNVMLBackend(devices...)
+	collector, err := nvidia.NewGPUPowerCollector(logger, nvidia.WithNVMLBackend(backend))
+	if err != nil {
+		logger.Error("failed to create GPU collector with fake backend", "error", err)
+		return nil
+	}
+
+	// Init must be called before returning, matching the real path (gpu.Discover
+	// calls Init before returning the meter). This ensures Devices() is populated
+	// when the monitor logs GPU meter info during its own Init.
+	if err := collector.Init(); err != nil {
+		logger.Error("failed to initialize fake GPU collector", "error", err)
+		return nil
+	}
+
+	return []gpu.GPUPowerMeter{collector}
 }
--- a/docs/developer/e2e-testing.md
+++ b/docs/developer/e2e-testing.md
@@ -101,6 +101,7 @@ The e2e tests are organized by concern across multiple files:
 | `metrics_test.go`    | Metric format validation: required labels, non-negative values, multiple scrapes       |
 | `invariants_test.go` | Energy conservation laws: Total = Active + Idle, process power attribution             |
 | `workload_test.go`   | Workload detection: stress-ng detection, power changes under load, terminated tracking |
+| `gpu_test.go`        | GPU metrics: presence, labels, non-negative values, power conservation (uses fake GPU) |
 
 ### Test Configuration
 
@@ -130,6 +131,12 @@ monitor:
 - `kepler_node_cpu_idle_joules_total` / `kepler_node_cpu_idle_watts` - Idle power
 - `kepler_node_cpu_usage_ratio` - CPU utilization (0.0 to 1.0)
 - `kepler_node_cpu_info` - CPU hardware information
+- `kepler_node_gpu_watts` - Total GPU power (fake GPU)
+- `kepler_node_gpu_active_watts` / `kepler_node_gpu_idle_watts` - Active and idle GPU power (currently exercises the idle case)
+- `kepler_node_gpu_info` - GPU device information
+
+GPU tests require GPU metrics to be enabled, as in the default e2e configuration.
+Custom configurations passed through `-kepler.config` must also enable GPU metrics when running these tests.
 
 ### Process-Level Metrics
 
@@ -154,12 +161,13 @@ monitor:
 
 ## What Is NOT Covered (and Why)
 
-| Feature                    | Reason Not Tested in Bare-Metal E2E                                      | Where It's Tested                             |
-|----------------------------|--------------------------------------------------------------------------|-----------------------------------------------|
-| VM metrics (`kepler_vm_*`) | Requires libvirt/hypervisor not available in test environment            | Unit tests in `internal/monitor/vm_test.go`   |
-| Redfish platform metrics   | Requires BMC hardware access not available in CI/dev environments        | Unit tests in `internal/platform/redfish/`    |
-| Metrics level filtering    | Low priority; config parsing is unit tested; e2e uses full metrics level | Unit tests in `config/`                       |
-| pprof debug endpoints      | Debug feature with low e2e value                                         | Unit tests in `internal/server/pprof_test.go` |
+| Feature                    | Reason Not Tested in Bare-Metal E2E                                      | Where It's Tested                                            |
+|----------------------------|--------------------------------------------------------------------------|--------------------------------------------------------------|
+| GPU pod/container metrics  | The fake GPU meter does not expose any fake processes yet                | Unit tests in `internal/monitor/`; Phase 2: fake-gpu K8s e2e |
+| VM metrics (`kepler_vm_*`) | Requires libvirt/hypervisor not available in test environment            | Unit tests in `internal/monitor/vm_test.go`                  |
+| Redfish platform metrics   | Requires BMC hardware access not available in CI/dev environments        | Unit tests in `internal/platform/redfish/`                   |
+| Metrics level filtering    | Low priority; config parsing is unit tested; e2e uses full metrics level | Unit tests in `config/`                                      |
+| pprof debug endpoints      | Debug feature with low e2e value                                         | Unit tests in `internal/server/pprof_test.go`                |
 
 > **Note**: Container and Pod metrics are tested in the [Kubernetes E2E tests](#kubernetes-e2e-tests) which run against a real cluster.
 
--- a/internal/device/gpu/nvidia/fake_backend.go
+++ b/internal/device/gpu/nvidia/fake_backend.go
@@ -1,0 +1,216 @@
+// SPDX-FileCopyrightText: 2025 The Kepler Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package nvidia
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/sustainable-computing-io/kepler/internal/device"
+	"github.com/sustainable-computing-io/kepler/internal/device/gpu"
+)
+
+// NOTE: This fake backend is not intended for production use. It is for
+// development and testing only, similar to fakeRaplMeter for CPU power.
+
+// FakeNVMLBackend implements NVMLBackend for testing without real GPUs.
+type FakeNVMLBackend struct {
+	mu      sync.RWMutex
+	devices []*FakeNVMLDevice
+}
+
+// FakeNVMLDevice implements NVMLDevice with configurable behavior.
+// Set exported fields before concurrent use. Callers must not read or write
+// them directly while device methods are running. Methods synchronize internal
+// updates to cumulative energy.
+type FakeNVMLDevice struct {
+	// Idx is the device index (0-based).
+	Idx int
+
+	// DeviceUUID is the unique identifier for this fake GPU.
+	DeviceUUID string
+
+	// DeviceName is the product name for this fake GPU.
+	DeviceName string
+
+	// DevicePower is the current power reading returned by GetPowerUsage.
+	DevicePower device.Power
+
+	// TotalEnergy is cumulative energy; GetTotalEnergy integrates DevicePower
+	// over the elapsed time since its previous call.
+	TotalEnergy device.Energy
+
+	// Mode is the compute mode (ComputeModeDefault for time-slicing,
+	// ComputeModeExclusiveProcess for exclusive).
+	Mode ComputeMode
+
+	// MIG controls whether IsMIGEnabled returns true.
+	MIG bool
+
+	// Processes is the list of "running" GPU processes returned by
+	// GetComputeRunningProcesses.
+	Processes []gpu.ProcessGPUInfo
+
+	// Utilization is the per-process utilization returned by GetProcessUtilization.
+	Utilization []gpu.ProcessUtilization
+
+	// UtilError, if set, is returned by GetProcessUtilization instead of Utilization.
+	// This allows testing the time-slicing fallback path.
+	UtilError error
+
+	mu sync.RWMutex
+
+	lastEnergyRead time.Time
+	now            func() time.Time
+}
+
+// Verify interface compliance at compile time.
+var (
+	_ NVMLBackend = (*FakeNVMLBackend)(nil)
+	_ NVMLDevice  = (*FakeNVMLDevice)(nil)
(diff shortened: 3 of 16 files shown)


nikimanoledaki:
@vprashar2929 just check in here - do you still have capacity for this PR? I can help with getting reviews if so. thank you!

vprashar2929:
@nikimanoledaki Yes please. That would be great 🙂

cmd/kepler/main.go
```diff
@@ -315,6 +315,10 @@ func createPrometheusExporter(
 // Uses the registry pattern to support multiple GPU vendors (NVIDIA, AMD, Intel).
 // Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
 func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
+	if fake := cfg.Dev.FakeGpuMeter; *fake.Enabled {
```
bitflicker64: This panics at startup when the key is present with an empty value, since yaml sets the pointer to nil:

```yaml
dev:
  fake-gpu-meter:
    enabled:
```

```
panic: runtime error: invalid memory address or nil pointer dereference
main.createGPUMeters(...) cmd/kepler/main.go:318
```

Reproduced with a build of this branch. The fake-cpu-meter path uses `ptr.Deref` for the same reason (config.go:398).

```suggestion
	if fake := cfg.Dev.FakeGpuMeter; ptr.Deref(fake.Enabled, false) {
```

Needs `k8s.io/utils/ptr` added to the imports.

cmd/kepler/main.go
```diff
@@ -335,3 +339,48 @@ func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMete
...
+	logger.Warn("using FAKE NVML backend (dev/testing mode)",
+		"deviceCount", deviceCount,
+		"sharingMode", sharingMode)
+
+	if deviceCount <= 0 {
+		deviceCount = 1
+	}
+
+	mode := nvidia.ComputeModeDefault
+	if sharingMode == "exclusive" {
```
bitflicker64: `deviceCount <= 0` is silently coerced to 1, and any string other than exactly `exclusive` falls back to time-slicing. The startup warning prints the raw values, but nothing says a fallback happened, so a typo like `exclusve` just runs in the wrong mode. `config.Validate()` already checks enums like `cpu.preferredMeters`; validating both fields there would fail fast. If you would rather keep it permissive, a log line on the fallback would make it visible.

cmd/kepler/main.go
```diff
@@ -315,6 +315,10 @@ func createPrometheusExporter(
 // Uses the registry pattern to support multiple GPU vendors (NVIDIA, AMD, Intel).
 // Returns empty slice if GPU is not enabled or no GPUs are available (soft-fail).
 func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMeter {
+	if fake := cfg.Dev.FakeGpuMeter; *fake.Enabled {
+		return createFakeGPUMeter(logger, fake.DeviceCount, fake.SharingMode)
+	}
+
 	if !cfg.IsFeatureEnabled(config.ExperimentalGPUFeature) {
```
bitflicker64, line 327: Since the fake path returns before this check, `dev.fake-gpu-meter.enabled: true` turns on the GPU pipeline even with the experimental flag off. I assume that is intentional (the e2e config relies on it), but a line in configuration.md about the precedence would help, same as the note the fake-cpu-meter deprecation warning gets.

internal/device/gpu/nvidia/fake_backend.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (d *FakeNVMLDevice) GetTotalEnergy() (device.Energy, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	// Auto-increment energy based on power (simulates 5s collection interval).
+	if d.energyIncrement == 0 && d.DevicePower > 0 {
+		// energy = power * time; 5s * watts -> joules
+		d.energyIncrement = device.Energy(d.DevicePower.Watts() * 5 * float64(device.Joule))
```
bitflicker64: This assumes 5 s between calls, but `monitor.interval` is configurable and the e2e config uses 3 s. Running this branch, `kepler_node_gpu_joules_total` grew by 1125 J per 3 s cycle, so the joules counter reads 5/3 of what the watts metric implies. Nothing asserts joules against watts today, so no test breaks, but computing the increment from the time since the previous read would remove the surprise:

```go
now := time.Now()
if !d.lastRead.IsZero() {
	elapsed := now.Sub(d.lastRead).Seconds()
	d.TotalEnergy += device.Energy(d.DevicePower.Watts() * elapsed * float64(device.Joule))
}
d.lastRead = now
```

Needs a `lastRead time.Time` field in place of `energyIncrement`. If the fixed step was deliberate to keep the fake deterministic, a comment saying so would do.

internal/device/gpu/nvidia/fake_backend_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/sustainable-computing-io/kepler/internal/device"
+	"github.com/sustainable-computing-io/kepler/internal/device/gpu"
+)
+
+func TestFakeNVMLBackend_InterfaceCompliance(t *testing.T) {
+	// Compile-time checks are in fake_backend.go; this verifies at runtime.
+	var _ NVMLBackend = (*FakeNVMLBackend)(nil)
+	var _ NVMLDevice = (*FakeNVMLDevice)(nil)
+}
```
bitflicker64: This duplicates the compile-time assertions in fake_backend.go and cannot fail, so it adds a test to the count without adding coverage.

```suggestion
```

test/e2e/gpu_test.go
```diff
@@ -0,0 +1,158 @@
...
+		computed := active + idle
+
+		t.Logf("GPU %s: total=%.4f W, active=%.4f W, idle=%.4f W, computed=%.4f W",
+			gpuLabel, m.Value, active, idle, computed)
+
+		diff := math.Abs(m.Value - computed)
+		tolerance := math.Max(m.Value*powerTolerance, absolutePowerTolerance)
+
+		assert.LessOrEqual(t, diff, tolerance,
+			"GPU %s: total (%.4f) should equal active + idle (%.4f)", gpuLabel, m.Value, computed)
```
bitflicker64: invariants_test.go has `assertWithinTolerance` for exactly this; using it keeps the tolerance logic in one place (the `math` import goes away with it):

```suggestion
		assertWithinTolerance(t, m.Value, computed, powerTolerance, absolutePowerTolerance,
			"GPU %s: total (%.4f) should equal active + idle (%.4f)", gpuLabel, m.Value, computed)
```

docs/developer/e2e-testing.md
```diff
@@ -154,12 +155,13 @@ monitor:
...
-| VM metrics (`kepler_vm_*`) | Requires libvirt/hypervisor not available in test environment            | Unit tests in `internal/monitor/vm_test.go`   |
-| Redfish platform metrics   | Requires BMC hardware access not available in CI/dev environments        | Unit tests in `internal/platform/redfish/`    |
-| Metrics level filtering    | Low priority; config parsing is unit tested; e2e uses full metrics level | Unit tests in `config/`                       |
-| pprof debug endpoints      | Debug feature with low e2e value                                         | Unit tests in `internal/server/pprof_test.go` |
+| Feature                    | Reason Not Tested in Bare-Metal E2E                                      | Where It's Tested                                            |
+|----------------------------|--------------------------------------------------------------------------|--------------------------------------------------------------|
+| GPU pod/container metrics  | Fake GPU uses static PIDs that don't map to real containers              | Unit tests in `internal/monitor/`; Phase 2: fake-gpu K8s e2e |
```
bitflicker64: Small mismatch: this reason describes the phase 2 setup. Today `createFakeGPUMeter` wires no processes at all, so there are no static PIDs; pod/container metrics are untested here simply because no fake GPU processes exist yet. Worth rewording so the table matches current behavior.

bitflicker64 (changes requested):
Built this branch and ran it with the fake meter enabled (deviceCount: 2): node metrics, labels, and total = active + idle all check out, and the unit tests pass. One startup crash needs fixing before merge (nil deref when `enabled` is null); that plus a few smaller items are inline. Also a typo in the PR title: "backed" should be "backend".

Two things read as phase 2 territory: active watts is always 0 because no fake processes are wired in, and the GPU e2e tests inherit the RAPL prerequisite so they skip on hosts where RAPL is not readable. Fine to leave for now, but a TODO would keep them from getting lost.

cmd/kepler/main.go
vprashar2929: Addressed

cmd/kepler/main.go
vprashar2929: Addressed

internal/device/gpu/nvidia/fake_backend.go
vprashar2929: Addressed

docs/developer/e2e-testing.md
vprashar2929: Addressed

bitflicker64 (commented):
All seven points from the last round hold up on this push. Re-checked each against `114aea8d` and at runtime on a bare-metal host at `monitor.interval: 3s`.

| earlier comment | state |
| --- | --- |
| nil deref on an empty `enabled:` key | fixed, `ptr.Deref` at main.go:323. `dev.fake-gpu-meter.enabled:` with no value now starts clean instead of panicking |
| silent fallback on bad `deviceCount` / `sharingMode` | fixed in `Validate()`. I fed it `deviceCount: 0`, `deviceCount: -3` and `sharingMode: exclusve`; each one fails fast with the field name in the error. The new config test covers the first and third |
| precedence over `experimental.gpu.enabled` undocumented | fixed in configuration.md |
| energy integrated over a hardcoded 5 s step | fixed, measured below |
| duplicate compile-time assertions in the test file | removed |
| hand-rolled tolerance check in gpu_test.go | now `assertWithinTolerance` |
| stale row in "What Is NOT Covered" | reworded |

The energy fix at the interval that read 5/3 high before:

```
J1       = 14015.120866
J2       = 20771.033128
dt       = 30.0242 s
implied  = 225.0155 W
reported = 225 W
```

`go build ./...` and `go vet ./...` clean, `go test -race -count=1 ./config/... ./internal/device/gpu/...` all pass, and the full bare-metal e2e suite as root gives 21 pass, 0 fail, 9 skip (the workload tests, no stress-ng on that box). The four new GPU tests pass and nothing existing regressed from turning the fake meter on in `e2e-config.yaml`.

Nothing here is a correctness bug, so this is not blocking. Two things worth a second look inline (the inert `sharingMode` and the conservation test that cannot fail), and three nits (skip-vs-require, the fake's partial locking, the unreachable guards).

Two bits of doc drift the diff walked past, same class as the `e2e-testing.md` row that already got fixed:

- `docs/developer/design/architecture/configuration.md` reproduces the `Dev` struct in Go, and it still shows only `FakeCpuMeter`.
- `docs/developer/e2e-testing.md` gained a `gpu_test.go` row and a GPU row under "What Is NOT Covered", but its metric inventory under "What Is Covered" still lists only `kepler_node_cpu_*` and `kepler_process_cpu_*`. The new node GPU series belong there.

One typo: the commit subject reads "fake NVML backed", should be "backend". The PR title has it right.

cmd/kepler/main.go
```diff
@@ -339,3 +344,56 @@ func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMete
...
+	for i := range devices {
+		devices[i] = &nvidia.FakeNVMLDevice{
+			Idx:         i,
+			DeviceUUID:  fmt.Sprintf("FAKE-GPU-%04d", i),
+			DeviceName:  "Fake NVIDIA GPU",
+			DevicePower: 225 * device.Watt,
+			Mode:        mode,
+		}
+	}
+	// TODO: Add fake GPU processes when e2e workloads can supply stable PIDs.
```
bitflicker64, line 381: While this TODO stands, `sharingMode` cannot change any metric. The mode is stored, logged at Init and read by `hasMIG` and `cacheMIGHierarchy`, but the fake leaves `MIG` false, so the only place it changes behavior is `collectProcessPower`, and with `Processes` empty both `attributeExclusive` and `attributeTimeSlicing` return early on `len(procs) == 0`.

Ran `114aea8d` twice, same config, only the mode differing (per device):

```
time-slicing -> node_gpu_watts 225, active 0, idle 225, 0 kepler_process_gpu_* series
exclusive    -> node_gpu_watts 225, active 0, idle 225, 0 kepler_process_gpu_* series
```

Identical apart from the log line. So the field carries validation, user docs, a config test and five shipped config files plus the e2e fixture for something that cannot yet change a metric. Fine as phase 2 scaffolding, but configuration.md reads as if it does something today. Worth a line there saying it takes effect once fake GPU processes land.

test/e2e/gpu_test.go
```diff
@@ -0,0 +1,156 @@
...
+		gpuLabel := m.Labels["gpu"]
+		active := activeByGPU[gpuLabel]
+		idle := idleByGPU[gpuLabel]
+		computed := active + idle
+
+		t.Logf("GPU %s: total=%.4f W, active=%.4f W, idle=%.4f W, computed=%.4f W",
+			gpuLabel, m.Value, active, idle, computed)
+
+		assertWithinTolerance(t, m.Value, computed, powerTolerance, absolutePowerTolerance,
+			"GPU %s: total (%.4f) should equal active + idle (%.4f)", gpuLabel, m.Value, computed)
```
bitflicker64, line 147: With a constant-power fake this assertion cannot fail. `DevicePower` is a fixed 225 W and `GetComputeRunningProcesses` returns empty, so `getDevicePowerStatsLocked` takes the `len(procs) == 0` branch, pins `minObservedPower` at 225 and reports idle == total, active == 0 every cycle. The check reduces to 225 being within tolerance of 0 + 225. From the e2e run on the committed fixture (one fake GPU):

```
gpu_test.go:146: GPU 0: total=225.0000 W, active=0.0000 W, idle=225.0000 W, computed=225.0000 W
```

`TestGPUNodeMetricsPresent` already covers "GPU metrics exist", which is all this catches today. For the active/idle split to be exercised the fake needs power that moves, so idle settles below total. A deterministic step in `GetPowerUsage`, driven from `createFakeGPUMeter`, would do it and keep the test reproducible.

test/e2e/gpu_test.go
```diff
@@ -0,0 +1,156 @@
...
+		"kepler_node_gpu_info should exist when fake GPU meter is enabled")
+}
+
+// TestGPUNodeMetricsLabels verifies GPU node metrics have required labels.
+func TestGPUNodeMetricsLabels(t *testing.T) {
+	_, scraper := setupKeplerForTest(t)
+
+	if !waitForGPUMetrics(t, scraper, 30*time.Second) {
+		t.Skip("Skipping: no GPU metrics present (fake GPU meter may not be enabled)")
+	}
```
bitflicker64: This skip, and the same two at lines 90-92 and 116-118, guard on the condition `TestGPUNodeMetricsPresent` requires at line 40. `suite_test.go` hard-codes `test/testdata/e2e-config.yaml`, which always sets `enabled: true`, so the branch is unreachable by design and only pays off if GPU metrics break. In that case three of the four tests go green by skipping and one goes red. `require.True` in all four would fail loudly instead of quietly shrinking the suite.

internal/device/gpu/nvidia/fake_backend.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	utils := make([]gpu.ProcessUtilization, len(d.Utilization))
+	copy(utils, d.Utilization)
+	return utils, nil
+}
+
+func (d *FakeNVMLDevice) GetComputeMode() (ComputeMode, error) {
+	return d.Mode, nil
+}
```
bitflicker64, line 200: `Idx` and `DeviceUUID` are read both ways: under `d.mu` inside `GetComputeRunningProcesses`, and unlocked in `Index()` and `UUID()`. `GetComputeMode` and `IsMIGEnabled` here take no lock at all, while `GetPowerUsage`, `GetTotalEnergy`, `GetComputeRunningProcesses` and `GetProcessUtilization` do. The struct comment invites callers to set these fields directly, so the mutex reads as a guarantee it does not give for half the accessors.

`go test -race ./internal/device/gpu/nvidia/...` is clean today because nothing mutates a device after construction. It is a trap for whoever writes the phase 2 tests that flip `Mode` or `Processes` mid-run. Either take the read lock in all the accessors, or drop the mutex and document the fake as configure-then-use.

cmd/kepler/main.go
```diff
@@ -339,3 +344,56 @@ func createGPUMeters(logger *slog.Logger, cfg *config.Config) []gpu.GPUPowerMete
...
 	return meters
 }
+
+// createFakeGPUMeter creates a GPU meter backed by FakeNVMLBackend for
+// development and testing. Not for production use.
+func createFakeGPUMeter(logger *slog.Logger, deviceCount int, sharingMode string) []gpu.GPUPowerMeter {
+	if deviceCount <= 0 {
+		logger.Error("invalid fake GPU device count", "deviceCount", deviceCount)
+		return nil
+	}
```
bitflicker64, line 354: Since `Validate()` now rejects both `deviceCount <= 0` and an unknown `sharingMode` before `createGPUMeters` runs, this guard and the `default:` case below are unreachable. Harmless as defense in depth, so no action needed if that is deliberate.

vprashar2929:
@bitflicker64  Please review the changes again

bitflicker64:
> @bitflicker64 Please review the changes again

Sure

bitflicker64 (approved):
All five inline points are resolved, along with the two doc items and the commit subject.

| point | how it landed |
| --- | --- |
| `sharingMode` inert | `docs/user/configuration.md` states it plainly now: changes the detected mode, no effect on exported metrics because the fake configures no processes |
| conservation test cannot fail | the skip became a `require`, but the conservation assertion itself is unchanged. The limitation is documented instead, as a parenthetical on the active/idle line in the e2e metric list |
| skip vs require | all three converted, no GPU-metric skips left. The RAPL prerequisite skip in `setupKeplerForTest` still applies, per the file's own TODO, and the `-kepler.config` consequence is spelled out in e2e-testing.md |
| partial locking | struct doc replaced with a configure-then-use contract |
| unreachable guards | left in place, as agreed |
| `Dev` struct doc drift | `FakeGPUMeter` added to `docs/developer/design/architecture/configuration.md`; field names, types and yaml tags match config.go |
| e2e-testing.md coverage list | GPU series added |
| commit subject | reads "backend" now |

One correction on my side. I asked for the node GPU series in "What Is Covered" and then said there were seven of them. That section lists what the tests assert, not what the exporter emits, and `gpu_test.go` never touches `kepler_node_gpu_joules_total`, `_active_joules_total` or `_idle_joules_total`. The narrower list you went with is the right one.

On the skip-to-require change: the default `e2e-config.yaml` enables the fake meter and CI passes no `-kepler.config`, so nothing regresses. The new failure mode is narrow, a RAPL-capable host running a custom config without the fake GPU now gets three hard failures instead of skips, which is the intended trade and the doc covers it.

Two leftovers, neither blocking:

- The conservation check is documented now but still cannot fail, so it is coverage on paper. Reasonable to leave for phase 2 alongside the fake processes.
- "Methods synchronize internal updates to cumulative energy" undersells the mutex, which also covers `DevicePower`, `Processes`, `Utilization`, `UtilError`, `Idx` and `DeviceUUID`. `Index`, `UUID`, `GetComputeMode` and `IsMIGEnabled` still read unlocked, which is legitimate under the configure-then-use contract in the sentence before it. That contract is the part that closes the point, so this is cosmetic.

"Build and run e2e tests" passed on `57949114`, which is the workflow covering the only code file in this delta. The job still pending is the self-hosted k8s deploy, which does not exercise `test/e2e/gpu_test.go`.

LGTM.

bitflicker64:
@nikimanoledaki @sunya-ch tagging you both here. re-reviewed and approved on `57949114`, everything from my earlier round landed.

two leftovers i noted, neither blocking: `TestGPUNodePowerConservation` still cannot fail while the fake reports a constant 225 W with no processes (idle ends up equal to total, active 0), and the new `FakeNVMLDevice` doc undersells what the mutex actually covers. both look like phase 2 material to me, alongside the fake GPU processes the TODO mentions.

a second pair of eyes before merge would be good if either of you has time.

vprashar2929:
> @nikimanoledaki @sunya-ch tagging you both here. re-reviewed and approved on `57949114`, everything from my earlier round landed.
> 
> two leftovers i noted, neither blocking: `TestGPUNodePowerConservation` still cannot fail while the fake reports a constant 225 W with no processes (idle ends up equal to total, active 0), and the new `FakeNVMLDevice` doc undersells what the mutex actually covers. both look like phase 2 material to me, alongside the fake GPU processes the TODO mentions.
> 
> a second pair of eyes before merge would be good if either of you has time.

@bitflicker64 Happy to address them in subsequent PR

bitflicker64:
@nikimanoledaki as mentioned on the sept 15 call, this one looks mergeable from my side. approved on `57949114`, and @vprashar2929 is taking the two small leftovers in a follow-up PR.

the 12 open review threads are all mine and all addressed. i can't resolve them without write access, so vibhu or a maintainer would need to. it also needs an update from main before it can merge.

bitflicker64 (approved):

PR #2422, opened 2026-02-16
feat(gpu): add MIG power attribution via dcgm-exporter

vimalk78:
Add support for NVIDIA Multi-Instance GPU (MIG) power attribution by
  querying per-instance activity from dcgm-exporter's Prometheus endpoint.

  - Rewrite MIG enumeration to use GetGpuInstances() API (sees all GPU instances, not just container-visible ones)
  - Add DCGMExporterBackend HTTP client with 2s metrics cache
  - Distribute parent GPU active power by DCGM GR_ENGINE_ACTIVE ratio, then by per-process SM utilization within each MIG instance
  - Fall back to equal distribution when DCGM is unavailable
  - Add dcgmEndpoint config option (auto-discovered if empty)
  - Add NVIDIA_VISIBLE_DEVICES/NVIDIA_MIG_MONITOR_DEVICES env vars to k8s and helm manifests
  
  
  
  
  
  
image

### DIFF
--- a/internal/device/gpu/nvidia/collector.go
+++ b/internal/device/gpu/nvidia/collector.go
@@ -44,8 +44,10 @@ type GPUPowerCollector struct {
 
 	// MIG support
 	dcgm                 DCGMBackend
-	dcgmEndpoint         string // pre-configured endpoint, set before Init()
+	dcgmEndpoint         string // pre-configured endpoint, applied during Init() or SetDCGMEndpoint()
 	migInstancesByDevice map[int][]MIGGPUInstance
+
+	initialized bool
 
 	mu sync.RWMutex
 
@@ -77,10 +79,15 @@ func (c *GPUPowerCollector) Name() string {
 	return "nvidia-gpu-power-collector"
 }
 
-// Init initializes the NVML backend and discovers devices
+// Init initializes the NVML backend and discovers devices.
+// Idempotent: subsequent calls after the first are no-ops.
 func (c *GPUPowerCollector) Init() error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
+
+	if c.initialized {
+		return nil
+	}
 
 	if err := c.nvml.Init(); err != nil {
 		return err
@@ -101,31 +108,15 @@ func (c *GPUPowerCollector) Init() error {
 	c.sharingModes = modes
 
 	// Log detected modes and check for MIG
-	hasMIG := false
 	for idx, mode := range modes {
 		c.logger.Info("GPU sharing mode detected",
 			"device", idx,
 			"mode", mode.String())
-		if mode == gpu.SharingModePartitioned {
-			hasMIG = true
-		}
 	}
 
 	// Initialize DCGM backend if MIG is detected
-	if hasMIG {
-		dcgm := NewDCGMExporterBackend(c.logger)
-
-		// If endpoint was pre-configured via SetDCGMEndpoint, it's already set
-		if c.dcgmEndpoint != "" {
-			dcgm.SetEndpoint(c.dcgmEndpoint)
-		}
-
-		if err := dcgm.Init(context.Background()); err != nil {
-			c.logger.Warn("DCGM backend initialization failed, MIG power attribution will use fallback",
-				"error", err)
-		} else {
-			c.dcgm = dcgm
-		}
+	if c.hasMIG() {
+		c.initDCGM()
 
 		// Cache MIG hierarchy from NVML (static topology)
 		if err := c.cacheMIGHierarchy(); err != nil {
@@ -133,7 +124,40 @@ func (c *GPUPowerCollector) Init() error {
 		}
 	}
 
+	c.initialized = true
 	return nil
+}
+
+// hasMIG returns true if any GPU is in MIG (partitioned) mode.
+// Caller must hold c.mu.
+func (c *GPUPowerCollector) hasMIG() bool {
+	for _, mode := range c.sharingModes {
+		if mode == gpu.SharingModePartitioned {
+			return true
+		}
+	}
+	return false
+}
+
+// initDCGM initializes (or reinitializes) the DCGM exporter backend.
+// Shuts down the existing backend if replacing. Caller must hold c.mu.
+func (c *GPUPowerCollector) initDCGM() {
+	if c.dcgm != nil {
+		_ = c.dcgm.Shutdown()
+		c.dcgm = nil
+	}
+
+	dcgm := NewDCGMExporterBackend(c.logger)
+	if c.dcgmEndpoint != "" {
+		dcgm.SetEndpoint(c.dcgmEndpoint)
+	}
+
+	if err := dcgm.Init(context.Background()); err != nil {
+		c.logger.Warn("DCGM backend initialization failed, MIG power attribution will use fallback",
+			"error", err)
+	} else {
+		c.dcgm = dcgm
+	}
 }
 
 // Shutdown cleans up NVML and DCGM resources
@@ -435,11 +459,18 @@ func (c *GPUPowerCollector) attributeTimeSlicing(deviceIndex int, result map[uin
 }
 
 // SetDCGMEndpoint sets the dcgm-exporter endpoint URL.
-// Must be called before Init() to take effect.
+// Can be called before or after Init(). If called after Init() on a system
+// with MIG GPUs, reinitializes the DCGM backend with the new endpoint.
 func (c *GPUPowerCollector) SetDCGMEndpoint(endpoint string) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	c.dcgmEndpoint = endpoint
+
+	if c.initialized && c.hasMIG() {
+		c.logger.Info("reinitializing DCGM backend with configured endpoint",
+			"endpoint", endpoint)
+		c.initDCGM()
+	}
 }
 
 // cacheMIGHierarchy enumerates all MIG instances from NVML at startup.
@@ -513,9 +544,11 @@ func (c *GPUPowerCollector) attributePartitioned(deviceIndex int, result map[uin
 	// Fallback to equal distribution among running processes when DCGM is
 	// unavailable (not deployed, unreachable, or initialization failed).
 	if c.dcgm == nil || !c.dcgm.IsInitialized() || len(instances) == 0 {
-		c.logger.Debug("DCGM not available or no MIG instances, using fallback",
-			"dcgm_available", c.dcgm != nil && c.dcgm.IsInitialized(),
-			"instances", len(instances))
+		c.logger.Warn("DCGM not available, using fallback attribution",
+			"dcgm_nil", c.dcgm == nil,
+			"dcgm_initialized", c.dcgm != nil && c.dcgm.IsInitialized(),
+			"instances", len(instances),
+			"device", deviceIndex)
 		return c.attributePartitionedFallback(nvmlDev, stats.ActivePower, result)
 	}
 
@@ -530,7 +563,7 @@ func (c *GPUPowerCollector) attributePartitioned(deviceIndex int, result map[uin
 	for _, gi := range instances {
 		activity, err := c.dcgm.GetMIGInstanceActivity(context.Background(), deviceIndex, gi.GPUInstanceID)
 		if err != nil {
-			c.logger.Debug("GetMIGInstanceActivity failed",
+			c.logger.Warn("failed to get MIG instance activity",
 				"device", deviceIndex,
 				"gpuInstanceID", gi.GPUInstanceID,
 				"error", err)
@@ -546,7 +579,7 @@ func (c *GPUPowerCollector) attributePartitioned(deviceIndex int, result map[uin
 
 		migDevice, err := nvmlDev.GetMIGDeviceByInstanceID(gi.GPUInstanceID)
 		if err != nil {
-			c.logger.Debug("GetMIGDeviceByInstanceID failed",
+			c.logger.Warn("failed to get MIG device handle",
 				"device", deviceIndex,
 				"gpuInstanceID", gi.GPUInstanceID,
 				"error", err)
@@ -579,6 +612,11 @@ func (c *GPUPowerCollector) attributePartitioned(deviceIndex int, result map[uin
 	}
 
 	if len(data) == 0 || totalActivity == 0 {
+		c.logger.Warn("no active MIG instances found, skipping power attribution",
+			"device", deviceIndex,
+			"instances_checked", len(instances),
+			"data_collected", len(data),
+			"total_activity", totalActivity)
 		return nil
 	}
 
--- a/internal/device/gpu/nvidia/dcgm_exporter.go
+++ b/internal/device/gpu/nvidia/dcgm_exporter.go
@@ -22,9 +22,18 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-// metricsCacheTTL is how long cached metrics are valid before refetching.
-// This prevents HTTP request storms when querying multiple MIG instances.
-const metricsCacheTTL = 2 * time.Second
+const (
+	// metricsCacheTTL is how long cached metrics are valid before refetching.
+	// This prevents HTTP request storms when querying multiple MIG instances.
+	metricsCacheTTL = 2 * time.Second
+
+	// maxConsecutiveFailures is how many HTTP failures before the circuit breaker trips.
+	maxConsecutiveFailures = 3
+
+	// reInitInterval is how long to wait before re-attempting initialization
+	// after the circuit breaker has tripped.
+	reInitInterval = 30 * time.Second
+)
 
 // DCGMExporterBackend provides MIG metrics by querying dcgm-exporter's Prometheus endpoint.
 // This is an alternative to the go-dcgm library that doesn't require libdcgm.so.
@@ -43,6 +52,11 @@ type DCGMExporterBackend struct {
 
 	// Cached metrics with TTL to avoid HTTP request storms
 	cachedMetrics *dcgmMetrics
+
+	// Circuit breaker: tracks consecutive failures and disables the backend
+	// after maxConsecutiveFailures. Re-initialization is attempted after reInitInterval.
+	consecutiveFailures int
+	circuitOpenTime     time.Time
 
 	// discoverEndpoint discovers the local dcgm-exporter endpoint.
 	// Defaults to discoverLocalDCGMExporter. Overridable for testing.
@@ -126,12 +140,18 @@ func (d *DCGMExporterBackend) Init(ctx context.Context) error {
 		return nil
 	}
 
+	return d.initLocked(ctx)
+}
+
+// initLocked performs endpoint discovery and initialization.
+// Caller must hold d.mu.
+func (d *DCGMExporterBackend) initLocked(ctx context.Context) error {
 	// If endpoint is already set, use it
 	if d.endpoint != "" {
 		if err := d.testEndpoint(ctx, d.endpoint); err != nil {
 			return fmt.Errorf("configured endpoint %s not reachable: %w", d.endpoint, err)
 		}
-		d.initialized = true
+		d.markInitialized()
 		d.logger.Info("DCGM exporter backend initialized", "endpoint", d.endpoint)
 		return nil
 	}
@@ -146,11 +166,11 @@ func (d *DCGMExporterBackend) Init(ctx context.Context) error {
 	if localEndpoint := d.discoverEndpoint(); localEndpoint != "" {
 		if err := d.testEndpoint(ctx, localEndpoint); err == nil {
 			d.endpoint = localEndpoint
-			d.initialized = true
+			d.markInitialized()
 			d.logger.Info("DCGM exporter backend initialized", "endpoint", localEndpoint, "discovery", "local-pod")
 			return nil
 		}
-		d.logger.Debug("local dcgm-exporter not reachable", "endpoint", localEndpoint)
+		d.logger.Warn("local dcgm-exporter not reachable", "endpoint", localEndpoint)
 	}
 
 	// Fallback to static endpoints. This covers cases where K8s API discovery
@@ -159,7 +179,7 @@ func (d *DCGMExporterBackend) Init(ctx context.Context) error {
 	for _, ep := range d.fallbackEndpoints {
 		if err := d.testEndpoint(ctx, ep); err == nil {
 			d.endpoint = ep
-			d.initialized = true
+			d.markInitialized()
 			d.logger.Info("DCGM exporter backend initialized", "endpoint", ep, "discovery", "fallback")
 			return nil
 		}
@@ -167,6 +187,14 @@ func (d *DCGMExporterBackend) Init(ctx context.Context) error {
 	}
 
 	return fmt.Errorf("no reachable dcgm-exporter endpoint found")
+}
+
+// markInitialized resets circuit breaker state and marks the backend as ready.
+// Caller must hold d.mu.
+func (d *DCGMExporterBackend) markInitialized() {
+	d.initialized = true
+	d.consecutiveFailures = 0
+	d.circuitOpenTime = time.Time{}
 }
 
 // discoverLocalDCGMExporter finds the dcgm-exporter pod IP on the same node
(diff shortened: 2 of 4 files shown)


internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,386 @@
...
+	if err != nil {
+		d.logger.Debug("failed to create kubernetes client", "error", err)
+		return ""
+	}
+
+	// Find dcgm-exporter pod on the same node
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	pods, err := clientset.CoreV1().Pods("nvidia-gpu-operator").List(ctx, metav1.ListOptions{
```
vprashar2929: What happens in case of different deployment methods (e.g., standalone dcgm-exporter, custom namespaces) Shouldn't we consider making the namespace and
label selector configurable?

manifests/k8s/daemonset.yaml
```diff
@@ -71,6 +71,11 @@ spec:
                   fieldPath: spec.nodeName
             - name: LD_LIBRARY_PATH
               value: /run/nvidia/driver/usr/lib64
+            # Required for NVML to see all MIG devices when running in a container
+            - name: NVIDIA_VISIBLE_DEVICES
+              value: all
```
vprashar2929, line 76:
```suggestion
              value: "all"
```

manifests/k8s/daemonset.yaml
```diff
@@ -71,6 +71,11 @@ spec:
...
+            - name: NVIDIA_MIG_MONITOR_DEVICES
+              value: all
```
vprashar2929, line 78:
```suggestion
              value: "all"
```

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,386 @@
...
+			return nil
+		}
+		d.logger.Debug("dcgm-exporter endpoint not reachable", "endpoint", ep)
+	}
+
+	return fmt.Errorf("no reachable dcgm-exporter endpoint found")
+}
+
+// discoverLocalDCGMExporter finds the dcgm-exporter pod IP on the same node
+func (d *DCGMExporterBackend) discoverLocalDCGMExporter() string {
```
vprashar2929, line 201: Will this require rbac changes for Kepler?

internal/device/gpu/nvidia/nvml.go
```diff
@@ -368,32 +368,48 @@ func (d *nvmlDevice) GetMIGInstances() ([]MIGInstance, error) {
...
-	}
-
 	var instances []MIGInstance
-	for i := 0; i < maxCount; i++ {
-		migDevice, ret := d.handle.GetMigDeviceHandleByIndex(i)
+	seenIDs := make(map[uint]bool) // Deduplicate by GPU Instance ID
+
+	// Iterate through all possible profile types (0-14 covers standard profiles)
+	// Profile IDs: 0=1g.5gb, 1=2g.10gb, 2=3g.20gb, 3=4g.20gb, 4=7g.40gb, etc.
+	for profileID := 0; profileID < 15; profileID++ {
```
vprashar2929: From where did 15 come?

internal/device/gpu/nvidia/mig_types.go
```diff
@@ -0,0 +1,46 @@
...
+// the physical GPU is partitioned into isolated GPU Instances. We need to query
+// metrics at the GPU Instance level to attribute power to processes.
+//
+// The key metric is DCGM_FI_PROF_GR_ENGINE_ACTIVE (Field 1001) which measures
+// the graphics/compute engine activity ratio (0.0-1.0) for each MIG instance.
+//
+// Implementations:
+//   - DCGMExporterBackend: Queries dcgm-exporter's Prometheus endpoint (no library dependency)
+type DCGMBackend interface {
+	Init() error
```
vprashar2929: Suggestion: Adding context would be nice to have since Init() and GetMIGInstanceActivity() makes multiple HTTP requests as best practice

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,386 @@
...
+
+// IsInitialized returns whether the backend is ready
+func (d *DCGMExporterBackend) IsInitialized() bool {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	return d.initialized
+}
+
+// GetMIGInstanceActivity returns the GR_ENGINE_ACTIVE metric for a MIG instance.
+func (d *DCGMExporterBackend) GetMIGInstanceActivity(gpuIndex int, gpuInstanceID uint) (float64, error) {
```
vprashar2929: GetMIGInstanceActivity holds d.mu.RLock() but fetchMetrics() writes d.cachedMetrics on cache miss. This is a write under a read lock. Wouldn't this hit data race?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected status: %d", resp.StatusCode)
+	}
+
+	// Read a bit to verify it's valid
+	buf := make([]byte, 1024)
+	_, err = resp.Body.Read(buf)
```
vprashar2929: The body is only partially read (1024 bytes) before closing. This prevents HTTP keep-alive connection reuse. Minor since Init runs once, but the idiomatic fix is just io.Copy(io.Discard, resp.Body) instead of the manual Read. Thoughts?

internal/device/gpu/nvidia/collector.go
```diff
@@ -396,6 +434,219 @@ func (c *GPUPowerCollector) attributeTimeSlicing(deviceIndex int, result map[uin
...
+		}
+
+		dev, err := c.nvml.GetDevice(deviceIndex)
+		if err != nil {
+			c.logger.Warn("failed to get device for MIG enumeration",
+				"device", deviceIndex, "error", err)
+			continue
+		}
+
+		// Get totalGPUSlices once (same for all devices of same model)
```
vprashar2929: Are we sure about this?

internal/device/gpu/nvidia/collector.go
```diff
@@ -396,6 +434,219 @@ func (c *GPUPowerCollector) attributeTimeSlicing(deviceIndex int, result map[uin
...
+		if c.totalGPUSlices == 0 {
+			maxMigCount, err := dev.GetMaxMigDeviceCount()
+			if err != nil || maxMigCount == 0 {
+				c.logger.Warn("failed to get max MIG device count, using default 7", "error", err)
+				c.totalGPUSlices = 7
```
vprashar2929: How come default is 7? Is it true for all MIG supported GPU's?

internal/device/gpu/nvidia/collector.go
```diff
@@ -41,6 +41,12 @@ type GPUPowerCollector struct {
 	// When set (> 0), always used instead of observed idle power. 0 means auto-detect.
 	idlePower float64
 
+	// MIG support
+	dcgm                 DCGMBackend
+	dcgmEndpoint         string // pre-configured endpoint, set before Init()
+	totalGPUSlices       uint
```
vprashar2929: What is the use of `totalGPUSlices`? Are we using it in Power Attribution logic?

internal/device/gpu/nvidia/mig_types.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// GPUInstanceID is the unique identifier for this GPU Instance within the parent GPU
+	GPUInstanceID uint
+
+	// EntityID is the DCGM entity ID used for querying metrics
+	EntityID uint
+
+	// ProfileSlices is the number of GPU slices allocated to this instance
+	// (e.g., 3 for a 3g.20gb profile on A100)
+	ProfileSlices uint
```
vprashar2929: Are we using this anywhere?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,386 @@
...
+		d.initialized = true
+		d.logger.Info("DCGM exporter backend initialized", "endpoint", d.endpoint)
+		return nil
+	}
+
+	// Try to discover dcgm-exporter endpoint
+	// Priority:
+	// 1. Local dcgm-exporter pod on same node (discovered via K8s API)
+	// 2. localhost:9400 (if dcgm-exporter uses hostNetwork)
+	// 3. ClusterIP service (may route to wrong node - last resort)
```
vprashar2929: Should we even do this?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,386 @@
...
+	gpuIndex      int
+	gpuInstanceID uint
+}
+
+type migInstanceMetrics struct {
+	gpuIndex      int
+	gpuInstanceID uint
+	profile       string  // e.g., "1g.5gb"
+	activity      float64 // DCGM_FI_PROF_GR_ENGINE_ACTIVE
+	powerUsage    float64 // DCGM_FI_DEV_POWER_USAGE
```
vprashar2929: Are we using this any attribution?

manifests/helm/kepler/templates/daemonset.yaml
```diff
@@ -84,6 +84,11 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            # Required for NVML to see all MIG devices when running in a container
+            - name: NVIDIA_VISIBLE_DEVICES
```
vprashar2929, line 88: Shouldn't we document this requirement in our user docs?

vprashar2929:
@vimalk78  No test for a multi-GPU MIG scenario (two physical GPUs both in MIG mode)?

manifests/k8s/daemonset.yaml
vimalk78: no :( the precommit does not allow this

manifests/k8s/daemonset.yaml
vimalk78: same

internal/device/gpu/nvidia/mig_types.go
vimalk78: added. thx for pointing out

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: fixed.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: fixed

manifests/helm/kepler/templates/daemonset.yaml
vimalk78: added

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: removed

internal/device/gpu/nvidia/collector.go
```diff
@@ -41,6 +42,14 @@ type GPUPowerCollector struct {
 	// When set (> 0), always used instead of observed idle power. 0 means auto-detect.
 	idlePower float64
 
+	// MIG support
+	dcgm         DCGMBackend
+	dcgmEndpoint string // pre-configured endpoint, set before Init()
+	// totalGPUSlices is the max MIG slices on the GPU (e.g., 7 for A100).
+	// Used as denominator in power attribution: instancePower = gpuPower * (profileSlices / totalGPUSlices)
```
vprashar2929: Isn't this misleading?

internal/device/gpu/nvidia/mig_types.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// GPUInstanceID is the unique identifier for this GPU Instance within the parent GPU
+	GPUInstanceID uint
+
+	// ProfileSlices is the number of GPU slices allocated to this instance
+	// (e.g., 3 for a 3g.20gb profile on A100)
+	ProfileSlices uint
+
+	// TotalGPUSlices is the total number of slices on the parent GPU
+	// (e.g., 7 for A100, 8 for H100)
```
vprashar2929: Doesn't these contradict each other. [collector.go line 470](https://github.com/sustainable-computing-io/kepler/pull/2422/internal/device/gpu/nvidia/collector.go) says "A100=7, H100=7, A30=4"

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// SetEndpoint sets the dcgm-exporter endpoint URL
+func (d *DCGMExporterBackend) SetEndpoint(endpoint string) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	// Ensure endpoint has /metrics suffix
+	if !strings.HasSuffix(endpoint, "/metrics") {
+		endpoint = strings.TrimSuffix(endpoint, "/") + "/metrics"
```
vprashar2929: How does this hold up against Input "http://host:9400/metrics/" Wouldn't this produces "http://host:9400/metrics/metrics" ?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+	}
+
+	d.logger.Debug("no dcgm-exporter pod found on node", "node", nodeName)
+	return ""
+}
+
+// testEndpoint checks if an endpoint is reachable and returns valid metrics
+func (d *DCGMExporterBackend) testEndpoint(endpoint string) error {
+	resp, err := d.client.Get(endpoint)
```
vprashar2929: not using context here from init()?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,410 @@
...
+//   - dcgm-exporter is already deployed by NVIDIA GPU Operator as a DaemonSet
+//   - Exposes per-MIG-instance metrics with GPU_I_ID and GPU_I_PROFILE labels
+//   - No native library dependencies required
+//   - Uses standard Prometheus text format
+type DCGMExporterBackend struct {
+	logger      *slog.Logger
+	endpoint    string // e.g., "http://10.131.2.22:9400/metrics"
+	client      *http.Client
+	initialized bool
+	mu          sync.Mutex
```
vprashar2929, line 51: perhaps sync.RWMutex?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,410 @@
...
+		// Parse labels
+		labels := parseLabels(labelsStr)
+
+		// Only process metrics with GPU_I_ID (MIG instance metrics)
+		gpuIIDStr, hasMIG := labels["GPU_I_ID"]
+		if !hasMIG {
+			continue
+		}
+
+		gpuStr := labels["gpu"]
```
vprashar2929, line 436: Will there be scenario where we have non-numeric GPU labels ?

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: Discovery already searches all namespaces (`Pods("")` at line 203) and tries multiple label selectors covering both GPU Operator (`app=nvidia-dcgm-exporter`) and standalone Helm chart (`app.kubernetes.io/name=dcgm-exporter`). The labels are a struct field (`discoveryLabels`) for extensibility. Users can also set `DCGM_ENDPOINT` config explicitly to bypass discovery entirely. Happy to add label configurability in a follow-up if there is demand.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: Yes, `pods list` across namespaces is needed for K8s-based discovery. However, this is a **graceful degradation** path — if RBAC denies the call, `discoverLocalDCGMExporter` logs a debug message and returns empty, then Init falls back to static endpoints (localhost:9400, then ClusterIP service). So RBAC helps but is not strictly required. Will add the ClusterRole rule to manifests.

internal/device/gpu/nvidia/nvml.go
vimalk78: From NVML's `NVML_GPU_INSTANCE_PROFILE_*` constants defined in `nvml.h`. Profile IDs range from 0 to 14 (15 values total), corresponding to different MIG slice configurations across all GPU models (1g, 2g, 3g, 4g, 7g, and their memory variants). Updated the comment to reference `nvml.h` explicitly.

internal/device/gpu/nvidia/collector.go
vimalk78: Yes — this is consistent with how other methods handle device errors (e.g., `GetProcessInfo` at line 666). We skip the failed device and continue enumerating others. A warning is logged for debugging.

internal/device/gpu/nvidia/collector.go
vimalk78: 7 is the max MIG slices for A100/H100 (the most common MIG GPUs). A30 has 4. The default only triggers when `GetMaxMigDeviceCount()` fails, which is rare. Using 7 when the real value is smaller produces conservative (lower) power attribution ratios, which is the safe direction.

internal/device/gpu/nvidia/collector.go
vimalk78: Good catch — `totalGPUSlices` and `ProfileSlices` are stored as metadata from NVML enumeration but not used in the current activity-based attribution (which uses DCGM activity ratios exclusively). Updated the comment to clarify. They are available for a potential slice-ratio-based fallback but could be removed to keep things minimal.

internal/device/gpu/nvidia/mig_types.go
vimalk78: Same as `totalGPUSlices` — stored as metadata from NVML enumeration, not used in current activity-based attribution. Available for future slice-ratio-based fallback.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: Valid concern — ClusterIP might route to a dcgm-exporter on a different node, giving wrong per-MIG-instance metrics. However, this is the last-resort fallback (priority 3), only tried when both K8s pod discovery AND localhost fail. With DaemonSet deployment (the default), ClusterIP routes to the local pod. Added a comment noting this trade-off.

internal/device/gpu/nvidia/collector.go
vimalk78: Fixed — updated comment to clarify this is metadata from NVML MIG enumeration, not currently used as denominator in activity-based attribution. Available for a potential slice-ratio fallback.

internal/device/gpu/nvidia/mig_types.go
vimalk78: Good catch! H100 has 7 slices, not 8. Fixed the comment to `(e.g., 7 for A100/H100, 4 for A30)`.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: Good catch! This is a bug — input `http://host:9400/metrics/` would produce `http://host:9400/metrics/metrics`. Fixed by trimming trailing slash before the suffix check. Added test cases for trailing slash scenarios.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: Fixed — `testEndpoint` now accepts `context.Context` and uses `http.NewRequestWithContext()`. All callers in `Init()` pass the context through.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: `sync.Mutex` is fine here. `GetMIGInstanceActivity()` calls `fetchMetrics()` which may write `cachedMetrics`, so it needs a full lock. The only read-only method is `IsInitialized()` which is called infrequently. RWMutex would add complexity for negligible gain.

internal/device/gpu/nvidia/dcgm_exporter.go
vimalk78: The `gpu` label from dcgm-exporter is always the NVML device index (numeric). `strconv.Atoi` returns 0 on error, so it degrades gracefully (maps to GPU 0). A non-numeric `gpu` label would be a bug in dcgm-exporter itself.

internal/device/gpu/nvidia/collector.go
vprashar2929: Worth documenting it somewhere for future purposes?

internal/device/gpu/nvidia/dcgm_exporter.go
vprashar2929: In which situation will it write?

internal/device/gpu/nvidia/mig_types.go
vprashar2929: Worth mentioning so that it's less confusing?

vimalk78:
@vprashar2929  please resolve comments which do not require any rework. add comment about what rework is required on answered comments.

internal/device/gpu/nvidia/dcgm_exporter.go
vprashar2929: @vimalk78 This needs action

internal/device/gpu/nvidia/collector.go
vprashar2929: @vimalk78 This needs action

internal/device/gpu/nvidia/mig_types.go
vprashar2929: @vimalk78 🆙

internal/device/gpu/nvidia/dcgm_exporter.go
vprashar2929: @vimalk78 🆙

vprashar2929:
@vimalk78 Resolved the comments. Also tagged you in comments that needs action.

internal/device/gpu/nvidia/collector.go
vprashar2929: @vimalk78

internal/device/gpu/nvidia/dcgm_exporter.go
vprashar2929: @vimalk78 Can't see the ClusterRole changes to manifests

vimalk78: No ClusterRole changes needed — the existing rule already includes pods: [get, list, watch] at cluster scope, which covers dcgm-exporter pod discovery

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -0,0 +1,417 @@
...
+	// fails (e.g., RBAC issues, dcgm-exporter in a non-standard namespace)
+	// or dcgm-exporter uses hostNetwork (localhost:9400).
+	for _, ep := range d.fallbackEndpoints {
+		if err := d.testEndpoint(ctx, ep); err == nil {
+			d.endpoint = ep
+			d.initialized = true
+			d.logger.Info("DCGM exporter backend initialized", "endpoint", ep, "discovery", "fallback")
+			return nil
+		}
+		d.logger.Debug("dcgm-exporter endpoint not reachable", "endpoint", ep)
```
vprashar2929, line 186: Shouldn't this be warning/error?

internal/device/gpu/nvidia/dcgm_exporter.go
```diff
@@ -22,9 +22,18 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-// metricsCacheTTL is how long cached metrics are valid before refetching.
-// This prevents HTTP request storms when querying multiple MIG instances.
-const metricsCacheTTL = 2 * time.Second
+const (
+	// metricsCacheTTL is how long cached metrics are valid before refetching.
+	// This prevents HTTP request storms when querying multiple MIG instances.
+	metricsCacheTTL = 2 * time.Second
```
vprashar2929, line 28: Shouldn't we make this configurable like monitor interval?

vimalk78: next PR

vprashar2929 (approved):

PR #2427, opened 2026-02-24
chore(hwmon): add check for enabled sensor

KaiyiLiu1234:
It is possible in hwmon to have input files disabled or enabled by the existence of an *_enable file. It's best to use this as an additional check. If *_enable does not exist or is set to 1, then the enabled sensor check passes. Otherwise, it fails and the corresponding file is not read.

Note the *_enable file is not mandatory but is a part of the interface. I have confirmed the existence and active use of these files for disabling current/voltage readings in a nvidia jetson orion machine I provisioned.

### DIFF
--- a/.github/workflows/check-x-crypto-deps.yaml
+++ b/.github/workflows/check-x-crypto-deps.yaml
@@ -26,7 +26,7 @@ jobs:
             +${{ github.event.pull_request.base.ref }}:refs/heads/${{ github.event.pull_request.base.ref }}
 
       - name: setup go
-        uses: actions/setup-go@v6.2.0
+        uses: actions/setup-go@v6.3.0
         with:
           go-version-file: go.mod
           cache: false
--- a/.github/workflows/config-change.yaml
+++ b/.github/workflows/config-change.yaml
@@ -85,7 +85,7 @@ jobs:
       # NOTE: Uploading the message as an artifact so that PR Comment workflow can use it to
       # add comment on PR with the message.
       - name: Upload message
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: message
           path: /tmp/message-${{ github.event.pull_request.number }}.txt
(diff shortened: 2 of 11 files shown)


internal/device/hwmon_power_meter.go
```diff
@@ -715,6 +739,24 @@ type voltageSensorInfo struct {
 	label       string
 }
 
+// isSensorEnabled checks if a sensor channel is enabled.
+// Returns true if:
+// - The enable file doesn't exist (channel is always enabled)
```
vimalk78: can we assume this?

internal/device/hwmon_power_meter.go
```diff
@@ -715,6 +739,24 @@ type voltageSensorInfo struct {
...
+// - The enable file exists and contains "1"
+// Returns false if the enable file exists and contains "0" (disabled)
+func isSensorEnabled(hwmonPath, sensorType string, sensorNum int) bool {
+	enableFile := filepath.Join(hwmonPath, fmt.Sprintf("%s%d_enable", sensorType, sensorNum))
+	data, err := os.ReadFile(enableFile)
+	if err != nil {
```
vimalk78, line 752: what if file does not have proper permissions to be read, but is present?

internal/device/hwmon_power_meter.go
kaiyiliu-strala: @vimalk78 I double checked and actually this can be assumed. enable is required in the interface and only exists if the hardware is able to disable or enable a channel. However, assuming enable does not exist, kepler will try and check the channel. If the channel is not accessible (disabled), it will return an error that will be picked up by kepler and logged as a warning.

kaiyiliu-strala: To be safe, i have made a change in isSensorEnabled to explicitly write a warning log saying enable file not found to make it easier to debug.

internal/device/hwmon_power_meter.go
kaiyiliu-strala: Good question. I have updated now to actually do an open fail. We will log that the file was not readable due to permissions restriction (warning) but we will continue trying to read the channel file (we should at least try).

PR #2435, opened 2026-03-05
docs(proposal): Simplify Concurrency Model

vimalk78:
Document Kepler's concurrency design, the synchronization primitives
  it requires, related bug history, and propose a simplified alternative
  using a ticker loop with mutex-guarded freshness. Includes profiling
  data showing scrape-dominated collection pattern.


docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
```diff
@@ -0,0 +1,463 @@
...
+|------------|-------|----------------------------------------------------------------|------------------------------------------------------------------------------------------------|
+| `b43260a9` | —     | Removed goroutines that mutated maps concurrently              | Fan-out goroutines writing to shared maps without synchronization                              |
+| `99a271ec` | #2146 | Added context check after `select` in `scheduleNextCollection` | Recursive goroutine scheduling — `select` on timer + context can pick timer after cancellation |
+| `c5a2be5e` | —     | Replaced `RWMutex` with `atomic.Pointer` for snapshot access   | Lock contention between collection goroutine and scrape handler reading snapshots              |
+| `7c0ee313` | #2386 | Added synchronization to test cleanup                          | Recursive collection goroutine outlived test — no lifecycle tracking                           |
+| `a64511b4` | —     | Moved mock cleanup to safe point                               | Background collection goroutine read mock expectations during `t.Cleanup`                      |
```
vprashar2929, line 263: Does these exist?

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
```diff
@@ -0,0 +1,463 @@
...
+scrape handler (net/http goroutine)
+  → PowerCollector.Collect()
+    → monitor.Snapshot()
+      → ensureFreshData()
+        → synchronizedPowerRefresh()
+          → singleflight.Do("compute", ...)
+            → refreshSnapshot()          ← blocks here for full collection
+```
+
+The scrape handler blocks for the entire collection duration.
```
vprashar2929, line 160: Worth mentioning about prometheus scrape timeout?

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
```diff
@@ -0,0 +1,463 @@
...
+directly instead of spawning 2 goroutines per process.
+
+### Scrape-Time Freshness
+
+The scrape handler calls `collect()` directly when data is stale,
+guarded by a `sync.Mutex` to prevent overlap with the ticker:
+
+```go
+func (pm *PowerMonitor) Snapshot() (*Snapshot, error) {
+    if !pm.isFresh() {
```
vprashar2929, line 377: Doesn't this looks like a race condition?

docs/developer/proposal/EP-004-Simplify-Concurrency-Model.md
```diff
@@ -0,0 +1,463 @@
+<!-- SPDX-FileCopyrightText: 2025 The Kepler Authors -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# EP-004: Simplify Kepler's Concurrency Model
```
nikimanoledaki: Could we increment this proposal to `EP-005` please to separate it from the [pre-existing EP-004](https://github.com/sustainable-computing-io/kepler/pull/2304)? We would need to update all references including the PR title, images, etc. This would:
* preserve the history and unique reference to EPs
* help us create `EP-006` for the upcoming model training EP

nikimanoledaki: Following up with this - please ignore my previous message! Whichever one gets merged first will take `EP-004` and the next one will need to increment its version.

In the KEP process, unmerged EP PRs do not reserve a number so this can stay as is for now 👍

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
```diff
@@ -0,0 +1,463 @@
...
+Most steps are local I/O or computation. The one exception is pod
+resolution: when a container ID is not found in cache,
+`kubeletPodInformer.LookupByContainerID()` triggers an HTTP request
+to the kubelet API (`/pods`) within the collection path.
+
+### Design Intent: Scrape-Time Freshness
+
+The concurrency model serves a deliberate purpose: **data freshness at
+scrape time** (Architecture Principle #4). Prometheus scrapes at its own
+schedule (e.g., every 15-30s), which may not align with Kepler's
```
nikimanoledaki, line 48:
```suggestion
schedule (e.g. the [default](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) `scrape_interval` is `1m`), which may not align with Kepler's
```

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
nikimanoledaki: The default scrape timeout is 10s. Next paragraph mentions that:

> On a production node with thousands of processes,
collection can take hundreds of milliseconds

This is an order of magnitude margin, therefore timeout is not a risk

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
nikimanoledaki: +1 I could not find a64511b4  👀

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
```diff
@@ -0,0 +1,463 @@
...
+    for {
+        select {
+        case <-ctx.Done():
+            return nil
+        case <-ticker.C:
+            pm.collect()
+        }
+    }
+}
+```
```
nikimanoledaki, line 349: We use the ticker loop pattern in another project that I maintain, cloudcost-exporter. For example, take a look here: https://github.com/grafana/cloudcost-exporter/blob/main/pkg/google/vpc/vpc.go#L93

```
	go func() {
		ticker := time.NewTicker(PriceRefreshInterval)
		defer ticker.Stop()

		for {
			select {
			case <-ctx.Done():
				logger.Info("VPC pricing refresh cancelled")
				return
			case <-ticker.C:
				logger.Info("Refreshing VPC pricing data")
				if err := pricingMap.Refresh(ctx); err != nil {
					logger.Error("Failed to refresh VPC pricing data", "error", err)
				}
			}
		}
	}()
```

+1 for this approach!

docs/developer/proposal/draft-20260302-simplify-concurrency-model.md
nikimanoledaki: The outer `isFresh()` reads the snapshot timestamp via the atomic pointer, which should be safe without the mutex - @vimalk78 could you confirm this please?

nikimanoledaki (approved):
LGTM! Left some comments 👍

nikimanoledaki:
Hi @vimalk78, could you share an update on the status of this proposal and whether you will have the capacity to implement it, please? Thank you so much!

PR #2438, opened 2026-03-12
feat(hwmon): add IIO power meter fallback for Jetson Nano

noFlowWater:
## Summary

Add Linux IIO subsystem support as a fallback power data source when hwmon sensors are unavailable. This enables Kepler to monitor power on **NVIDIA Jetson Nano** (L4T R32) where the INA3221 chip uses the `ina3221x` IIO driver instead of the `ina3221` hwmon driver.

### Problem

Kepler's hwmon power meter scans `/sys/class/hwmon/` for power sensors. On Jetson Nano (L4T R32, kernel 4.9), the INA3221 power monitor is bound to NVIDIA's custom IIO driver (`ina3221x`) rather than the upstream hwmon driver (`ina3221`). This means Kepler finds no power zones and cannot attribute energy to containers.

| Jetson Device | Kernel | INA3221 Driver | Kepler Status |
|---|---|---|---|
| AGX Orin (L4T R35) | 5.10 | `ina3221` (hwmon) | ✅ Works |
| Xavier NX (L4T R35) | 5.10 | `ina3221` (hwmon) | ✅ Works |
| **Nano (L4T R32)** | **4.9** | **`ina3221x` (IIO)** | **❌ No zones found** |

### Solution

- **`sysfsIIOReader`**: Implements `hwmonReader` interface, scans `/sys/bus/iio/devices/` for `in_power{N}_input` sensors
- **`iioPowerZone`**: Implements `EnergyZone` interface with mW→µW conversion (IIO reports milliwatts, hwmon uses microwatts)
- **`compositeReader`**: Wraps both hwmon and IIO readers; tries hwmon first, falls back to IIO, combines zones from all available sources

No API changes — `NewHwmonPowerMeter()` signature and behavior are unchanged for existing hwmon-only systems.

### IIO Sensor Layout (Jetson Nano)

```
/sys/bus/iio/devices/iio:device0/
├── name              → "ina3221x"
├── rail_name_0       → "POM_5V_IN"    (total input)
├── rail_name_1       → "POM_5V_GPU"
├── rail_name_2       → "POM_5V_CPU"
├── in_power{0,1,2}_input   (mW, integer)
├── in_voltage{0,1,2}_input (mV)
└── in_current{0,1,2}_input (mA)
```

### Downstream Compatibility

IIO zones only support `Power()` (not `Energy()`). This works because `internal/monitor/node.go` already handles power-only zones:

1. Detects zone type via `MaxEnergy() > 0 || energyReading > 0`
2. For power-only zones: integrates `E = P × Δt` over collection intervals
3. Container attribution and Prometheus export use pre-computed snapshot data

No code path requires `Energy()` to succeed.

### Hardware Testing

Verified on 3 Jetson devices with cross-compiled arm64 binary:

| Device | Subsystem | Zones Discovered | Power Reading | Result |
|---|---|---|---|---|
| Jetson Nano (L4T R32) | **IIO fallback** | pom_5v_in, pom_5v_gpu, pom_5v_cpu | ~2.3W total (idle) | ✅ PASS |
| Jetson AGX Orin (L4T R35) | hwmon (unchanged) | vin_sys_5v0, vdd_gpu_soc, vdd_cpu_cv, vddq | ~2.6W total | ✅ PASS |
| Jetson Xavier NX (L4T R35) | hwmon (unchanged) | vdd_in, vdd_cpu_gpu_cv, vdd_soc | ~7.0W total | ✅ PASS |

### Changes

| File | Change |
|---|---|
| `internal/device/iio_power_meter.go` | New: `sysfsIIOReader` + `iioPowerZone` (~205 LOC) |
| `internal/device/iio_power_meter_test.go` | New: 13 tests covering discovery, conversion, fallback, errors (~243 LOC) |
| `internal/device/hwmon_power_meter.go` | Add `compositeReader`, update `NewHwmonPowerMeter` to use it (~80 LOC changed) |
| `internal/device/hwmon_power_meter_test.go` | Add `iioPowerZone` case to `TestHwmonPowerMeter_ZoneDetails` |
| `internal/device/testdata/sys/bus/iio/` | IIO testdata from real Jetson Nano values |

### Test Plan

- [x] All existing `internal/device/` tests pass with `-race`
- [x] 13 new IIO-specific tests pass
- [x] Full project `go build ./...` succeeds
- [x] Cross-compiled arm64 binary verified on real Jetson hardware (Nano, Orin, Xavier)
- [ ] CI pipeline (triggered by this PR)


kaiyiliu-strala:
@vimalk78 are you able to see if this works on your nvidia jetson machine? - specifically this: Jetson Nano (L4T R32)

vimalk78:
> @vimalk78 are you able to see if this works on your nvidia jetson machine? - specifically this: Jetson Nano (L4T R32)

I am not able to see

```
vimal@vimal-jetson:/sys/bus/iio/devices$ ll
total 0
drwxr-xr-x 2 root root 0 Mar 16 16:53 ./
drwxr-xr-x 4 root root 0 Mar 16 16:53 ../

vimal@vimal-jetson:~$ uname -a
Linux vimal-jetson 5.15.148-tegra #1 SMP PREEMPT Thu Sep 18 15:08:33 PDT 2025 aarch64 aarch64 aarch64 GNU/Linux

```

kaiyiliu-strala:
Can you also remove this from HWMON power meter and have it in its own device? This seems to be in an entirely different subsystem which would be great to leverage.

shellyco-code:
Hi @noFlowWater! Thanks for this great feature—fallback support for Jetson Nano is an awesome addition.

I was reviewing the code and noticed a cross-platform compatibility issue with the mock test data. The new test folders are named using the Linux convention with a colon (e.g., internal/device/testdata/.../iio:device0).

Since Windows file systems strictly forbid colons in file/folder names, merging this will cause git clone and git pull to crash with fatal errors for any contributors developing on Windows machines.

Could you rename the test folders to use an underscore or hyphen (e.g., iio_device0), and then update the test's mock path setup to map the colon to the underscore just for the unit tests? This will keep the repository clone-able for everyone!

iacker (commented):
Thanks for this @noFlowWater — I went through the IIO reader end to end. It's a genuinely useful fallback and the code is clean. A few things from a maintainer read, mostly to move the open questions along rather than block.

**Architecture works with the downstream monitor.** I traced how the merged zones are consumed in `internal/monitor/node.go`. Zone type is decided by `isEnergySensor := zone.MaxEnergy() > 0 || energyReading > 0`, so an `iioPowerZone` (which has `MaxEnergy() == 0` and returns an error from `Energy()`) correctly falls through to the power-sensor branch and gets integrated over time. So `Energy()` returning an error here is fine by design — no change needed on that front.

**One real concern: double counting in `compositeReader.Zones()`.** It appends zones from *every* reader that succeeds. On a normal host the IIO reader just returns "no iio power zones" and is skipped, so today it's fine. But on a board that exposes power through *both* hwmon and IIO for the same rails (some Jetson/L4T setups surface the INA sensors in both subsystems depending on kernel), you'd get each rail counted twice at the node level. Worth either deduplicating by rail/chip, or making the fallback strict (only use IIO when hwmon returned nothing), which also matches how "fallback" reads in the doc comment. The strict version is probably the smaller, safer change.

**The two open questions from earlier still need an answer:**
- @kaiyiliu-strala asked (Mar 18) to pull IIO out of the hwmon power meter into its own device since it's a separate subsystem. I'd lean the other way for now — keeping it behind the same `hwmonReader` interface via the composite is what makes the transparent fallback possible, and splitting it into a separate power meter would mean wiring a whole new device path for one board family. But that's a call worth confirming with the team rather than leaving the PR stuck on it.
- @shellyco-code flagged the test fixtures using `:` in folder names (the `iio:deviceN` convention), which breaks checkout on Windows/macOS-case setups. That one's worth fixing before merge — it'll trip CI for contributors on those platforms.

Not approving yet because of the double-count question and the unresolved fixtures, but this is close and the direction is right. Happy to re-review once the fallback semantics are pinned down.

PR #2452, opened 2026-03-31
fix(monitor): Fix stale resource utilization due to refresh ordering

HananAwwad:
## Summary
This PR fixes a bug where energy calculations used stale resource utilization
because resources.Refresh() was called after node power calculation.

## Fix
- Move resources.Refresh() before:
  - firstNodeRead()
  - calculateNodePower()

This ensures energy attribution uses current snapshot data.

## Tests
- Added integration test using a stateful SwitchingResourceInformer
- Verifies that:
  - Snapshot N uses utilization from snapshot N (not N-1)
  - No one-snapshot lag in energy attribution

## Validation
- Reproduced issue locally
- Confirmed correct behavior after fix
- All monitor tests pass

Fixes #2446

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
         entry: bash -c "commitlint --from $FROM --to $TO --verbose"
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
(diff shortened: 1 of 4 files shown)


vimalk78:
The fix in `calculatePower()` makes sense — `Refresh()` needs to happen before `calculateNodePower()` so that `pm.resources.Node().CPUUsageRatio` reflects the current cycle's utilization instead of the previous one.

However, I'm not sure the reorder is necessary. `firstReading()` is a seed/baseline — power is set to 0 for energy zones since there's no previous reading to compute ΔE/Δt. The Refresh() call there seeds the informer so the next calculatePower() cycle has resource data available. Moving it before firstNodeRead() changes the initial active/idle energy split, but that's just the baseline and not where the stale-data bug manifests.

Could the `firstReading()` change be reverted to keep the scope focused on the actual bug in `calculatePower()`?

Also, the `.pre-commit-config.yaml` version bumps should be a separate commit (or PR)

HananAwwad:
I have addressed all feedback: reverted the pre-commit version bumps, reverted the firstReading reorder, and added mock coverage

HananAwwad:
any updates about this pull request?

nikimanoledaki:
Thank you @HananAwwad, apologies for the delays.

@vimalk78 are you able to continue reviewing this PR? If not, please let us know. Thank you!

shellyco-code:
Hi @HananAwwad! Thanks for addressing the previous feedback.

I was taking a look to help get this moving, and I noticed the PR still has 21 commits. It looks like your branch accidentally pulled in a lot of unrelated commits from upstream main (for example: "chore: update maintainer spring 2026", "docs: add EP-003", etc.).

To get this merged cleanly, you'll need to clean up the Git history so this PR only contains your specific monitor fix. The easiest way to fix this is to run an interactive rebase (git rebase -i upstream/main) and drop all the commits that don't belong to your fix, then force push.

Once the commit history is clean, it should be super easy for the maintainers to approve and merge!

HananAwwad:
> Hi @HananAwwad! Thanks for addressing the previous feedback.
> 
> I was taking a look to help get this moving, and I noticed the PR still has 21 commits. It looks like your branch accidentally pulled in a lot of unrelated commits from upstream main (for example: "chore: update maintainer spring 2026", "docs: add EP-003", etc.).
> 
> To get this merged cleanly, you'll need to clean up the Git history so this PR only contains your specific monitor fix. The easiest way to fix this is to run an interactive rebase (git rebase -i upstream/main) and drop all the commits that don't belong to your fix, then force push.
> 
> Once the commit history is clean, it should be super easy for the maintainers to approve and merge!

Hi @shellyco-code, thanks for the guidance. I cleaned up the branch and force-pushed a focused three-commit history containing only the monitor refresh-ordering fix and its tests. The checks are running now; I’d appreciate a re-review when convenient.

HananAwwad:
any updates about this pull request?

vprashar2929 (approved):

PR #2467, opened 2026-05-03
docs(proposal): Improve extensibility of registering power meter backends

nikimanoledaki:
## Why

`cmd/kepler/main.go` constructs CPU and GPU power meters in two different ways:

* `createCPUMeter` is a strict RAPL-with-hwmon-fallback chain.
* `createGPUMeters` is a soft fan-out across registered vendors. 

Three problems follow: asymmetric patterns, hidden GPU init failures, and a strict CPU path that operators cannot extend easily.

## What

This proposal: 
* Promotes the unexported powerMeter interface in `internal/device/power_meter.go` to an exported PowerMeter (Name, Init). Shutdown is optional via service.Shutdowner.
* Adds cpu.meters config (default ["rapl", "hwmon"]) that operators can select.
* Replaces the hardcoded fallback chain with a switch dispatch: device.CreateCPUMeter walks the priority list, buildCPUMeter constructs each backend in a switch over backend names. 
* Splits gpu.Discover into three failure modes so real backend failures surface as errors rather than log lines (separable follow-up PR).
* Translates two legacy keys (experimental.hwmon.forceEnabled, dev.fake-cpu-meter.enabled) to cpu.meters at startup with a deprecation warning.

AI tooling disclosure: I used Claude Code to plan and generate some of these ideas. I prompted and heavily edited any AI-generated text and guarantee that I have read, reviewed, and edited every line.

### DIFF
--- a/docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
+++ b/docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
@@ -107,12 +107,12 @@ As part of this unification, `gpu.Discover` is refactored to split the three fai
 
 Result per scenario at startup:
 
-| Scenario            | Result                   |
-|---------------------|--------------------------|
-| Factory error       | Real failure. Aggregate. |
-| `Init()` error      | Real failure. Aggregate. |
-| Empty `Devices()`   | Soft skip. Not an error. |
-| Success             | Append to result.        |
+| Scenario          | Result                   |
+|-------------------|--------------------------|
+| Factory error     | Real failure. Aggregate. |
+| `Init()` error    | Real failure. Aggregate. |
+| Empty `Devices()` | Soft skip. Not an error. |
+| Success           | Append to result.        |
 
 Error aggregation uses `errors.Join`. `gpu.Discover` returns the joined error only when the GPU feature is explicitly enabled and every registered vendor returned a real failure. "No GPU on this node" stays `(nil, nil)`.
 
@@ -250,7 +250,7 @@ While this solves visibility for transient failures during runtime, it doesn't a
 
 Out of scope for this EP.
 
-### GPU failure-mode split
+### GPU failure-mode split (separate PR)
 
 Refactor `gpu.Discover` to split factory / `Init()` / empty-devices failure modes and return `(meters, error)`. Separate PR; independent of the CPU work.
 
--- a/docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
+++ b/docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
@@ -1,0 +1,263 @@
+# Draft: Extensible Power Meter Selection
+
+* **Status**: Draft
+* **Author**: Niki Manoledaki
+* **Created**: 2026-05-03
+
+## Problem
+
+For context, `cmd/kepler/main.go` constructs CPU and GPU power meters in two different shapes:
+
+* `createCPUMeter` hardcodes RAPL with hwmon as a fallback. It returns `(device.CPUPowerMeter, error)`. This is strict by design: Kepler must produce a CPU meter or fail startup.
+* `createGPUMeters` fans out across registered vendors via `gpu.DiscoverAll`. It returns `[]gpu.GPUPowerMeter` with no error, which is a soft failure mode, also by design, as it assumes that GPU monitoring is not required for Kepler.
+
+Once we agree on these assumptions, here are the problems that follow.
+
+1. **Asymmetry.** Two patterns for the same job (turn config into hardware abstractions) makes `cmd/kepler/main.go` harder to read and audit.
+2. **Hidden GPU failures.** `gpu.Discover` (`internal/device/gpu/registry.go:48`) collapses three failure modes into one `nil` return:
+   * factory error (driver missing or unsupported)
+   * `Init()` error (driver present but broken)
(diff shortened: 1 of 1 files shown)


docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
```diff
@@ -0,0 +1,263 @@
...
+```text
+INFO gpu meter discovery   ok=[nvidia] failed=[amd: factory: rocm-smi not found]
+```
+
+## Configuration
+
+```yaml
+cpu:
+  meters: ["rapl", "hwmon"]   # ordered priority
+```
```
vimalk78, line 176: the name `cpu.meters` doesnt convey that it is a ordered fallback chain. could we use `cpu.meterPriority` or `cpu.fallbackChain` ?

docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
```diff
@@ -0,0 +1,263 @@
...
+
+For context, `cmd/kepler/main.go` constructs CPU and GPU power meters in two different shapes:
+
+* `createCPUMeter` hardcodes RAPL with hwmon as a fallback. It returns `(device.CPUPowerMeter, error)`. This is strict by design: Kepler must produce a CPU meter or fail startup.
+* `createGPUMeters` fans out across registered vendors via `gpu.DiscoverAll`. It returns `[]gpu.GPUPowerMeter` with no error, which is a soft failure mode, also by design, as it assumes that GPU monitoring is not required for Kepler.
+
+Once we agree on these assumptions, here are the problems that follow.
+
+1. **Asymmetry.** Two patterns for the same job (turn config into hardware abstractions) makes `cmd/kepler/main.go` harder to read and audit.
+2. **Hidden GPU failures.** `gpu.Discover` (`internal/device/gpu/registry.go:48`) collapses three failure modes into one `nil` return:
```
vimalk78, line 17: there are some prerequisites for gpu initialization. 
- the node must have a gpu
- node feature discovery operator must discover gpu and add some labels to node
- nvidia (or AMD) gpu operator must be installed and some CRs created.
- wait for drivers to be installed and initialized etc. this ensures nvml is present
- now kepler's gpu discovery should succeed

if kepler inits gpu before all the steps are completed, gpu init will fail. 
if gpu is enabled, and gpu init fails, kepler cannot separate the two case
- gpu is not present on the node
- gpu is present but cannot initialize
and a cluster can be heterogeneous, i.e onlt few nodes may have gpu

so there was a deliberate decision to allow cpu power to continue to work if gpu side fails for any reason.
 @vprashar2929 do you remember same way or did i miss something?

this proposal will fail kepler startup if gpu is configured and gpu init fails, right?

may be there are ways to detect gpu's presence on the node, and if gpu is present, and if gpu init fails we treat it as kepler failure. but the method to detect gpu presence should work in VM nodes also.

docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
sunya-ch: I agree with fallbackChain field. 
And, in addition to log, Kepler resource of kepler-operator can have status to present the active (selected) meter.

docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
```diff
@@ -0,0 +1,263 @@
...
+Symmetric shape (`(meter(s), error)`), different failure semantics (CPU strict, GPU soft).
+
+## Detailed Design
+
+### Package layout
+
+```text
+internal/device/
+├── power_meter.go              # PowerMeter interface (promoted from private)
+├── cpu_power_meter.go          # CPUPowerMeter, CreateCPUMeter, buildCPUMeter
```
sunya-ch, line 140: We may have folder for cpu (same as what we have for gpu)

docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
vprashar2929: @vimalk78  Yes this is exact flow. 
>so there was a deliberate decision to allow cpu power to continue to work if gpu side fails for any reason.

IMO CPU is the base/primary power source for Kepler, and GPU is supplementary. Every node has a CPU, so Kepler should always be able to report CPU power. GPU power only makes sense once that whole prerequisite chain has actually converged on the node. In a heterogeneous cluster tying Kepler's startup to a successful GPU init would mean blocking CPU power monitoring on perfectly healthy nodes just because the GPU side hasn't caught up yet.

docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
nikimanoledaki: Yes! I agree. I plan to do this in a follow-up PR 👍 



I can update the proposal to reflect this

docs/developer/proposal/draft-20260503-extensible-power-meter-selection.md
nikimanoledaki: Thank you, I can rename this field 👍

nikimanoledaki: Please take a look: https://github.com/sustainable-computing-io/kepler/pull/2471

PR #2468, opened 2026-05-03
feat(device): add `cpu.meters` config and switch-based CPU meter selection

nikimanoledaki:
Part of #2467.

## Why

`createCPUMeter` in `cmd/kepler/main.go` hardcodes the RAPL → hwmon fallback chain. Operators on platforms where neither is the right source cannot reorder backend priority without forking. This PR makes the priority list config-driven without inventing new vocabulary (no registry, no factories, no init() side effects).

## What

Changes
- Promotes the unexported powerMeter interface to an exported PowerMeter (Name, Init). Shutdown is
optional via service.Shutdowner.
- Adds cfg.Cpu.Meters config (default ["rapl", "hwmon"]) — preserves prior fallback behaviour.
- device.CreateCPUMeter walks the list, builds each backend via buildCPUMeter, runs Init(), returns the
first that yields zones. Real failures aggregate via errors.Join; empty zones are a soft skip.
- buildCPUMeter is a switch over backend names — adding a backend means one new case + one constructor.
- cmd/kepler/main.go drops the old createCPUMeter / createHwmonMeter helpers (~70 lines).
- Two legacy keys translate at startup with a deprecation warning:
  - dev.fake-cpu-meter.enabled: true → cpu.meters: ["fake"]
  - experimental.hwmon.forceEnabled: true → cpu.meters: ["hwmon"]
- Sample configs (helm, k8s, hack, compose) and docs/user/configuration.md updated.

## Behaviour change

None for systems running the default config. Preserves the same RAPL to hwmon fallback. Operators using the legacy keys see a deprecation warning but continue to get the same selection.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -384,14 +384,12 @@ func DefaultConfig() *Config {
 // who set them today expect the legacy behavior. The legacy keys will stop
 // working in a future release.
 func (c *Config) ApplyCpuMeterDeprecations(logger *slog.Logger) {
-	if ptr.Deref(c.Dev.FakeCpuMeter.Enabled, false) {
-		logger.Warn("dev.fake-cpu-meter.enabled is deprecated; set cpu.meters: [\"fake\"] instead")
+	switch {
+	case ptr.Deref(c.Dev.FakeCpuMeter.Enabled, false):
+		logger.Warn(`dev.fake-cpu-meter.enabled is deprecated; set cpu.meters: ["fake"] instead`)
 		c.Cpu.Meters = []string{"fake"}
-		return
-	}
-
-	if c.Experimental != nil && ptr.Deref(c.Experimental.Hwmon.ForceEnabled, false) {
-		logger.Warn("experimental.hwmon.forceEnabled is deprecated; set cpu.meters: [\"hwmon\"] instead")
+	case c.Experimental != nil && ptr.Deref(c.Experimental.Hwmon.ForceEnabled, false):
+		logger.Warn(`experimental.hwmon.forceEnabled is deprecated; set cpu.meters: ["hwmon"] instead`)
 		c.Cpu.Meters = []string{"hwmon"}
 	}
 }
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ package config
 import (
 	"fmt"
 	"io"
+	"log/slog"
 	"net"
 	"net/url"
 	"os"
@@ -51,6 +52,13 @@ type (
 	Host struct {
 		SysFS  string `yaml:"sysfs"`
 		ProcFS string `yaml:"procfs"`
+	}
+
+	// Cpu configuration controls how CPU power meters are selected.
+	// Meters lists backends in priority order. The first backend that
+	// initializes successfully and reports zones is used.
+	Cpu struct {
+		Meters []string `yaml:"meters"`
 	}
 
 	// Rapl configuration
@@ -192,6 +200,7 @@ type (
 		Log      Log      `yaml:"log"`
 		Host     Host     `yaml:"host"`
 		Monitor  Monitor  `yaml:"monitor"`
+		Cpu      Cpu      `yaml:"cpu"`
 		Rapl     Rapl     `yaml:"rapl"`
 		Exporter Exporter `yaml:"exporter"`
 		Web      Web      `yaml:"web"`
@@ -264,6 +273,9 @@ const (
 	MonitorStaleness         = "monitor.staleness" // not a flag
 	MonitorMaxTerminatedFlag = "monitor.max-terminated"
 
+	// CPU
+	CpuMeters = "cpu.meters" // not a flag
+
 	// RAPL
 	RaplZones = "rapl.zones" // not a flag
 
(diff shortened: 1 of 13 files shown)


internal/device/cpu_power_meter.go
```diff
@@ -42,3 +51,92 @@ type CPUPowerMeter interface {
...
+		zones, _ := meter.Zones()
+		if len(zones) == 0 {
+			logger.Info("cpu meter reports no zones, trying next backend", "meter", name)
+			continue
+		}
+		logger.Info("using cpu power meter", "meter", name)
+		return meter, nil
+	}
+
+	if len(errs) == 0 {
```
leonorfmartins: errors gere don't include the case where the zone can be empty, which is something that can happen from this [line](https://github.com/sustainable-computing-io/kepler/pull/2468/changes#diff-96dfe5385007fa63c1238e4e614f8116c89f1ef590d2ead981528364d7116ad5R123). We only seem to log an info and continue. Should we include an error in this case or use a default zone in case it's empty?

internal/device/cpu_power_meter_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"log/slog"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sustainable-computing-io/kepler/config"
+)
+
+func discardLogger() *slog.Logger {
```
leonorfmartins, line 16: it seems that we can use `slog.DiscardHandler` which is a pattern already used in config_cpu_test.go tests. I think we could switch to use that format instead?

config/config.go
```diff
@@ -358,6 +373,27 @@ func DefaultConfig() *Config {
...
 
+// ApplyCpuMeterDeprecations translates legacy CPU-meter selectors into an
+// effective cpu.meters value and logs a deprecation warning per translation.
+//
+// Legacy selectors:
+//   - experimental.hwmon.forceEnabled=true → cpu.meters: ["hwmon"]
+//   - dev.fake-cpu-meter.enabled=true     → cpu.meters: ["fake"]
+//
+// Legacy selectors win over an explicit cpu.meters when set, since operators
+// who set them today expect the legacy behavior. The legacy keys will stop
```
leonorfmartins:
```suggestion
// who set them today expect the legacy behavior. When both legacy keys are set, fake takes precedence over hwmon. The legacy keys will stop
```

internal/device/cpu_power_meter.go
```diff
@@ -42,3 +51,92 @@ type CPUPowerMeter interface {
...
+// Returns the joined error only if no backend produced a usable meter.
+// CPU is mandatory; callers treat any returned error as fatal.
+func CreateCPUMeter(logger *slog.Logger, cfg *config.Config) (CPUPowerMeter, error) {
+	var errs []error
+	for _, name := range cfg.Cpu.Meters {
+		meter, err := buildCPUMeter(name, logger, cfg)
+		if err != nil {
+			logger.Warn("cpu meter not available, trying next backend", "meter", name, "error", err)
+			errs = append(errs, fmt.Errorf("cpu meter %q: %w", name, err))
+			continue
```
laurall974, line 74: Just a discussion: should unknown names hard-fail rather than logging a "warn"? As an example, in the case where the user puts ["rapll","hwmon"] (typo for RAPL), then rapll is unknown and hwmon would be use, and the user might not realize it.

internal/device/cpu_power_meter.go
nikimanoledaki: Good catch! Fixed. The skip appends to `errs` if there is no usable meter, handled with an early return at the top of the function

internal/device/cpu_power_meter_test.go
nikimanoledaki: Switched `discardLogger()` to `slog.New(slog.DiscardHandler)`

config/config.go
nikimanoledaki: Applied 👍

internal/device/cpu_power_meter.go
nikimanoledaki: Thank you - added some validation for this 👍

laurall974 (approved):

leonorfmartins (approved):

internal/device/cpu_power_meter.go
```diff
@@ -42,3 +51,93 @@ type CPUPowerMeter interface {
...
+		}
+		if err := meter.Init(); err != nil {
+			logger.Warn("cpu meter init failed, trying next backend", "meter", name, "error", err)
+			errs = append(errs, fmt.Errorf("cpu meter %q: init: %w", name, err))
+			continue
+		}
+		zones, _ := meter.Zones()
+		if len(zones) == 0 {
+			logger.Info("cpu meter reports no zones, trying next backend", "meter", name)
```
leonorfmartins, line 83: There might be an edge case here: if a meter is successfully initialised but has no zones, which seems to be possible, that meter will never be shutdown, because we end up just skipping it here. Maybe it's a good idea to call shutdown already here just in case?

vimalk78 (approved):
minor comment on the new config name already in the proposal PR, can be addressed in later PRs.

PR #2471, opened 2026-05-12
refactor(config): rename `cpu.meters` to `cpu.preferredMeters`

nikimanoledaki:
Follow up to #2468. Addresses some of the [feedback on #2467](https://github.com/sustainable-computing-io/kepler/pull/2467#discussion_r3186410605).

## Why

`cpu.meters: ["rapl", "hwmon"]` doesn't show clearly that it is an ordered preference with a priority & fallback mechanism. 

However this is not exactly a `fallbackChain` since `CreateCPUMeter` selects once at startup. There is no runtime failover.

Opting for `preferredMeters` since it matches the kube-apiserver `preferredAddressTypes` pattern of having an ordered preference list with the first viable option being selected.

## What

Rename config field `cpu.meters` to `cpu.preferredMeters` and update all references.

### DIFF
--- a/compose/default/kepler/etc/kepler/config.yaml
+++ b/compose/default/kepler/etc/kepler/config.yaml
@@ -32,6 +32,12 @@ host:
   sysfs: /host/sys # Path to sysfs filesystem (default: /sys)
   procfs: /host/proc # Path to procfs filesystem (default: /proc)
 
+cpu:
+  # Ordered preference list of CPU power-meter backends. The first backend
+  # that initializes successfully and reports zones is used.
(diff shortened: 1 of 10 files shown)


sunya-ch (dismissed):
lgtm

vprashar2929:
@nikimanoledaki  Can you also address this? https://github.com/sustainable-computing-io/kepler/pull/2471#issuecomment-4432709631 🙏

nikimanoledaki:
@vprashar2929 @sunya-ch could I have another review, please? thank you!

vprashar2929 (approved):

PR #2488, opened 2026-07-08
feat(exporter): expose container CPU time metric

Nachiket-Roy:
## Summary

Exposes the `kepler_container_cpu_seconds_total` Prometheus metric, reporting total container CPU time (user + system) in seconds.

### Changes

* **Exporter**: Declared, initialized, and registered the `containerCPUTimeDescriptor` on `PowerCollector`. Exposes `kepler_container_cpu_seconds_total` with labels (`container_id`, `container_name`, `runtime`, `state`, `pod_id`, `node_name`).
* **Tests**: Added unit tests in `power_collector_test.go` asserting metric values and labels for both running and terminated containers.
* **Docs**: Regenerated `docs/user/metrics.md` via `make gen-metrics-docs`.

### Verification

1. **Unit Tests**:
   ```bash
   go test ./internal/exporter/prometheus/collector/... -run TestPowerCollector -v
   go test ./internal/exporter/prometheus/collector/... -run TestTerminatedContainerExport -v
    ```
2. Quality Checks: gofmt -l . and go vet ./... passed.
3. Manual Smoke Test: Verified via local docker-compose run:
```bash
$ curl -s localhost:28283/metrics | grep kepler_container_cpu_seconds_total
kepler_container_cpu_seconds_total{container_id="...",container_name="...",node_name="kind-control-plane",pod_id="",runtime="docker",state="running"} 1.56
```

### DIFF
--- a/docs/user/metrics.md
+++ b/docs/user/metrics.md
@@ -185,6 +185,19 @@ These metrics provide energy and power information for containers.
 - **Constant Labels**:
   - `node_name`
 
+#### kepler_container_cpu_seconds_total
+
+- **Type**: COUNTER
+- **Description**: Total user and system time of cpu at container level in seconds
+- **Labels**:
+  - `container_id`
+  - `container_name`
+  - `runtime`
+  - `state`
+  - `pod_id`
+- **Constant Labels**:
+  - `node_name`
+
 #### kepler_container_cpu_watts
 
 - **Type**: GAUGE
--- a/internal/exporter/prometheus/collector/power_collector.go
+++ b/internal/exporter/prometheus/collector/power_collector.go
@@ -52,6 +52,7 @@ type PowerCollector struct {
 	// Container power metrics
 	containerCPUJoulesDescriptor *prometheus.Desc
 	containerCPUWattsDescriptor  *prometheus.Desc
+	containerCPUTimeDescriptor   *prometheus.Desc
 	containerGPUWattsDescriptor  *prometheus.Desc
 	containerGPUJoulesDescriptor *prometheus.Desc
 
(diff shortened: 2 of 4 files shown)


vprashar2929:
@Nachiket-Roy Couple of nits:
1. To keep things consistent for process we put CPU time next to the other CPU descriptors (before GPU). Here `containerCPUTimeDescriptor` lands after the GPU ones. Better to match the layout
2. `process seconds` intentionally don't include stale while with containers we do export stale metrics as well but worth highlighting this so in future we know it was deliberate (series will "move" from `state=running` → `state=terminated`)
3. Worth adding assertion for `kepler_container_cpu_seconds_total` in e2e tests

vprashar2929:
@Nachiket-Roy Will you also be working on adding the same for pod level as well?

Nachiket-Roy:
Hello @vprashar2929,
Thanks for the review i look into them and i'll also add the corresponding pod-level metric in a follow-up PR.

vprashar2929:
@Nachiket-Roy  Can you help resolve conflicts?

Nachiket-Roy:
@vprashar2929 I've resolved the conflict sorry for the delay.

iacker:
After the rebase, the metric itself is already on `main` via #2501 (July 15), same labels. What this diff still adds is the e2e coverage in `test/e2e-k8s`, which `main` lacks, plus unit assertions.

I would keep the PR at that scope and retitle it `test(e2e): cover kepler_container_cpu_seconds_total`. The descriptor comment can go, `docs/user/metrics.md` already covers it.

@Nachiket-Roy ok to trim? @vprashar2929 fine with you?

PR #2495, opened 2026-07-11
fix: refactor component initialization to return graceful errors instead of panicking

### LINKED ISSUE #2494: [Bug]: Component initialization functions panic on invalid or empty configurations
### Bug Description

In Go, it is a well-established architectural anti-pattern for library or component initialization functions to use panic(). A robust node-level daemon like Kepler should gracefully handle misconfigurations or dynamic hardware discovery failures by returning an error from constructors, allowing the caller to decide whether to fail the program, retry, or fallback.

Currently, there are two key components in Kepler that panic during initialization:

NewRedfishCollector (in platform_collector.go) panics if the RedfishDataProvider is nil.
NewAggregatedZone (in energy_zone.go) panics if the dynamically discovered zones list is empty.
If an unexpected hardware layout triggers these panics, the entire Kepler exporter crashes and enters a CrashLoopBackOff, dropping metrics for all other functional subsystems (e.g., RAPL, GPU) on the node.

### Steps to Reproduce

Attempt to initialize an AggregatedZone on a node where the sysfs discovery returns an empty list.
The Kepler daemon will panic with: panic: NewAggregatedZone: zones cannot be empty and crash.
Proposed Solution
I will submit a PR that:

### Expected Behavior

NewRedfishCollector should return (*PlatformCollector, error).
NewAggregatedZone should return (*AggregatedZone, error).
Callers (like prometheus.go and hwmon_power_meter.go) should handle these errors gracefully, log a warning, and continue collecting metrics from the available subsystems.
Actual Behavior
The daemon panics and crashes entirely.

### Environment

- OS: All OS / Linux
- Kubernetes Version: All supported versions
- Container Runtime: All runtimes
- Hardware: All hardware (Issue occurs when dynamic hardware discovery like sysfs or Redfish returns empty lists/nil)
- Deployment Method: Kubernetes DaemonSet

### Logs and Error Messages

```shell
If a user deploys Kepler on a node where sysfs energy zones are empty or the Redfish data provider fails to initialize, the daemon crashes entirely with the following panic signature, rather than logging the error and continuing with the remaining valid collectors:

panic: NewAggregatedZone: zones cannot be empty
goroutine 1 [running]:
github.com/sustainable-computing-io/kepler/internal/device.NewAggregatedZone(...)
        /workspace/internal/device/energy_zone.go:50

(or similarly for RedfishDataProvider cannot be nil in platform_collector.go)
```

### Additional Context

I found this architectural bug while reviewing the codebase. In Go, it is considered a severe anti-pattern for component initialization functions to use `panic()` on unexpected input. A robust node-level daemon like Kepler should gracefully handle misconfigurations or dynamic hardware discovery failures by returning an `error` from constructors, allowing the caller to decide whether to fail the program, retry, or fallback.

I am already working on a PR to fix this! The PR will convert the panics in `NewRedfishCollector` and `NewAggregatedZone` into graceful `error` returns, and update the collectors to log the error and safely continue.

shellyco-code:
Description
This PR addresses the architectural issue described in Issue #[INSERT_ISSUE_NUMBER].

In a robust daemon like Kepler, component initializations should not panic on bad input, as it can cause the entire node agent to crashloop if hardware discovery fails or configuration is missing.

This PR converts the panics in NewRedfishCollector and NewAggregatedZone into graceful error returns.

Changes Made
Platform Collector (platform_collector.go)
Refactored NewRedfishCollector to return (*PlatformCollector, error).
Handled the error in prometheus.go: if Redfish initialization fails, it now logs an error and gracefully skips the platform subsystem, allowing other collectors to continue functioning.
Updated gen-metric-docs/main.go to handle the error properly.
Energy Zone Aggregator (energy_zone.go)
Refactored NewAggregatedZone to return (*AggregatedZone, error).
Handled the error in both hwmon_power_meter.go and rapl_sysfs_power_meter.go. If an invalid/empty zone aggregation is attempted, it logs the error and gracefully continues iterating over the remaining valid power meters.
Tests
Updated all 50+ test instances in platform_collector_test.go and energy_zone_test.go to assert on the new error return signature (require.Error / require.NoError) rather than capturing panics with require.Panics.
Testing
make test passes successfully.
golangci-lint passes successfully.
Related Issues
Fixes #2494

### DIFF
--- a/hack/gen-metric-docs/main.go
+++ b/hack/gen-metric-docs/main.go
@@ -381,7 +381,7 @@ func main() {
 		bmcID:    "test-bmc",
 	}
 	fmt.Println("Creating platform collector...")
-	platformCollector := collector.NewRedfishCollector(mockRedfish, logger)
+	platformCollector, _ := collector.NewRedfishCollector(mockRedfish, logger)
 	fmt.Println("Created platform collector")
 
 	fmt.Println("Extracting metrics from platform collector...")
--- a/internal/device/energy_zone.go
+++ b/internal/device/energy_zone.go
@@ -4,6 +4,7 @@
 package device
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"sync"
@@ -43,11 +44,10 @@ type AggregatedZone struct {
 
 // NewAggregatedZone creates a new AggregatedZone for zones of the same type
 // The name is taken from the first zone
-// Panics if zones is empty or nil
-func NewAggregatedZone(zones []EnergyZone) *AggregatedZone {
-	// Panic on invalid inputs
+// Returns an error if zones is empty or nil
+func NewAggregatedZone(zones []EnergyZone) (*AggregatedZone, error) {
 	if len(zones) == 0 {
-		panic("NewAggregatedZone: zones cannot be empty")
+		return nil, errors.New("NewAggregatedZone: zones cannot be empty")
 	}
 
 	// Use the first zone's name as the aggregated zone name
@@ -73,7 +73,7 @@ func NewAggregatedZone(zones []EnergyZone) *AggregatedZone {
 		lastReadings:  make(map[zoneKey]Energy),
 		currentEnergy: 0,
 		maxEnergy:     totalMax, // Cache the combined MaxEnergy
-	}
+	}, nil
 }
 
 // Name returns the zone name
--- a/internal/exporter/prometheus/prometheus.go
+++ b/internal/exporter/prometheus/prometheus.go
@@ -166,8 +166,9 @@ func CreateCollectors(pm Monitor, applyOpts ...OptionFn) (map[string]prom.Collec
 	collectors["gpu_info"] = collector.NewGPUInfoCollector(pm, opts.nodeName)
 
 	// Add platform collector if platform data provider is available
-	if opts.platformDataProvider != nil {
-		collectors["platform"] = collector.NewRedfishCollector(opts.platformDataProvider, opts.logger)
+	platformCollector, err := collector.NewRedfishCollector(opts.platformDataProvider, opts.logger)
+	if err == nil {
+		collectors["platform"] = platformCollector
 	}
 
 	return collectors, nil
--- a/internal/device/energy_zone_test.go
+++ b/internal/device/energy_zone_test.go
@@ -69,7 +69,8 @@ func TestNewAggregatedZone(t *testing.T) {
 		&mockEnergyZone{name: "package", index: 1},
 	}
 
-	az := NewAggregatedZone(zones)
+	az, err := NewAggregatedZone(zones)
+	require.NoError(t, err)
 
 	assert.Equal(t, "package", az.Name())
 	assert.Equal(t, -1, az.Index())
@@ -77,10 +78,9 @@ func TestNewAggregatedZone(t *testing.T) {
 	assert.Len(t, az.zones, 2)
 	assert.NotNil(t, az.lastReadings)
 
-	// Test panic on empty zones
-	assert.Panics(t, func() {
-		NewAggregatedZone([]EnergyZone{})
-	})
+	// Test error on empty zones
+	_, err = NewAggregatedZone([]EnergyZone{})
+	require.Error(t, err)
 }
 
 // TestAggregatedZone_EnergyAggregation tests core energy aggregation functionality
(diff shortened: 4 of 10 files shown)


internal/device/energy_zone.go
```diff
@@ -43,11 +44,11 @@ type AggregatedZone struct {
 
 // NewAggregatedZone creates a new AggregatedZone for zones of the same type
 // The name is taken from the first zone
-// Panics if zones is empty or nil
-func NewAggregatedZone(zones []EnergyZone) *AggregatedZone {
+// Returns an error if zones is empty or nil
+func NewAggregatedZone(zones []EnergyZone) (*AggregatedZone, error) {
 	// Panic on invalid inputs
```
sunya-ch: This comment should be removed

sunya-ch (commented):
@shellyco-code  Thank you so much for the PR! Nice catch and very good start!

I agree that we shouldn't panic inside any *New* function. Most of the part in this PR looks good to me.

However, we should still fail the program because it doesn't work as expected. 
Zones shouldn't be empty at this stage. If I understand correctly, it should already return error at the service initialization if it is.

For example, RAPL initialization throws an error when detect no zone here https://github.com/sustainable-computing-io/kepler/blob/96430b41968b407bdb52d8d21218021e9d733ac8/internal/device/rapl_sysfs_power_meter.go#L82.

If the power monitor service is failed as no zone detected, it should be already exit.

https://github.com/sustainable-computing-io/kepler/blob/96430b41968b407bdb52d8d21218021e9d733ac8/internal/service/initializer.go#L15

If you have seen this panic, it mean the code has a bug that we need to fix.

I believe that it is panic by design by @sthaha .

Note that, we may also use log.Fatalf to print out reasonable error message instead of panic.

internal/device/energy_zone.go
shellyco-code: Good catch! I've removed the stale comment. I've also updated the callers so that the error propagates upwards to gracefully fail the service initialization, rather than swallowing it

hack/gen-metric-docs/main.go
```diff
@@ -381,7 +381,11 @@ func main() {
 		bmcID:    "test-bmc",
 	}
 	fmt.Println("Creating platform collector...")
-	platformCollector := collector.NewRedfishCollector(mockRedfish, logger)
+	platformCollector, err := collector.NewRedfishCollector(mockRedfish, logger)
+	if err != nil {
+		fmt.Printf("Failed to create platform collector: %v\n", err)
+		os.Exit(1)
```
sunya-ch: Actually we prefer log.Fatalf more than printf + exit as it introduces the timestamp prefix and thread safety.

```suggestion
        log.Fatalf("Failed to create platform collector: %v", err)
```

sthaha (commented):
Looks fine overall. 

I can't think of a reason to prefer `panic` instead of returning an error, albeit the behavior of the program itself will not change a lot; i.e. exits from main instead of `panic`. 

I do recommend that the PR itself have a good description of what it attempts to fix instead of AI written changelog.

internal/exporter/prometheus/prometheus.go
```diff
@@ -167,7 +167,11 @@ func CreateCollectors(pm Monitor, applyOpts ...OptionFn) (map[string]prom.Collec
 
 	// Add platform collector if platform data provider is available
 	if opts.platformDataProvider != nil {
-		collectors["platform"] = collector.NewRedfishCollector(opts.platformDataProvider, opts.logger)
+		platformCollector, err := collector.NewRedfishCollector(opts.platformDataProvider, opts.logger)
+		if err != nil {
+			return nil, err
+		}
+		collectors["platform"] = platformCollector
 	}
```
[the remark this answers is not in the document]
sunya-ch: This comment makes sense. @shellyco-code Please update the PR description.

sunya-ch (approved):
lgtm

Thank you for your contribution.

sunya-ch (changes requested):
Just notice that the test file is deleted. Please bring it back

sunya-ch (approved):
LGTM now. Thank you for the update.

sunya-ch:
The interval is stdout test is changed from 1s to 100ms noted by @shellyco-code that this is for mitigating the flaky test as context deadline is 1s. The interval should be set less than that value.

```diff
internal/exporter/stdout/stdout_test.go
-		exporter := NewExporter(mockMonitor, WithOutput(out), WithInterval(1*time.Second))
+		exporter := NewExporter(mockMonitor, WithOutput(out), WithInterval(100*time.Millisecond))
```

Issue #2491, opened 2026-07-09
CI fails when `go.mod` requires Go > 1.24


nikimanoledaki:
## Problem

Any PR that upgrades Go in `go.mod` above 1.24 fails CI. The Dockerfile and `golangci-lint` are both pinned to Go 1.24, so builds and lint fail.

Blocked dep upgrades:
* Dependabot PR #2481 upgrades `golang.org/x/net` to `v0.55.0`, which needs Go 1.25.
* https://github.com/sustainable-computing-io/kepler/pull/2480
* https://github.com/sustainable-computing-io/kepler/pull/2475

## Next Steps

Upgrade the `Dockerfile` base image and `golangci-lint` to a version built against the latest Go versions.

- [ ] Upgrade `Dockerfile`, `golangci-lint` in `.pre-commit-config.yaml` and `.github/workflows/pr-checks.yaml` to `golang:1.25`
- [ ] Upgrade `Dockerfile`, `golangci-lint` in `.pre-commit-config.yaml` and `.github/workflows/pr-checks.yaml` to `golang:1.26`
- [ ] Verify CI passes on 3 dep PRs linked above

shellyco-code:
Hi! I'd love to take this issue. If it's still available, I'll start working on it and open a PR.

PR #2492, opened 2026-07-09
fix: upgrade Dockerfile to golang:1.25 and golangci-lint to v2.9.0

molloyzak13:
kepler's CI build and lint jobs are pinned to Go 1.24. The Dockerfile uses `golang:1.24` and golangci-lint is on v2.1.6. When go.mod moves past 1.24 (dependabot bumped x/net to v0.55.0, which needs Go 1.25), both jobs break.

Bumped the Dockerfile base image to `golang:1.25`, and golangci-lint to v2.9.0 in `.pre-commit-config.yaml`. golangci-lint v2.9.0 builds against Go 1.25.

Covers the first checkbox from #2491. The Go 1.26 step needs goreleaser and other tooling to catch up, so splitting that out makes sense.

The CI workflow file (`.github/workflows/pr-checks.yaml`) also pins golangci-lint to v2.1.6 and should be bumped to v2.9.0 in the same pass. Same one-line change. Happy to push that as a follow-up commit or have a maintainer apply it directly.

PR #2496, opened 2026-07-11
chore: upgrade CI infrastructure to support Go 1.25

shellyco-code:
Description
This PR upgrades the Kepler Continuous Integration infrastructure to support Go 1.25 

Currently, several Dependabot and contributor PRs that bump go.mod dependencies requiring Go > 1.24 are blocked because the GitHub Actions pipeline uses a hardcoded Go 1.24 builder image, and an outdated golangci-lint version that lacks the AST parsers for newer Go language features.

This upgrade unblocks those dependency bumps and ensures future compatibility.

Changes Made
Dockerfile: Upgraded the builder base image from golang:1.24 to golang:1.25 so the binary compiles with the latest toolchain syntax.
GitHub Actions (.github/workflows/pr-checks.yaml): Bumped the golangci-lint version to v2.12.2 to support linting Go 1.25+ syntax.
Pre-commit Hooks (.pre-commit-config.yaml): Bumped the local golangci-lint hook to v2.12.2 to match the CI pipeline.
Testing
make test passes successfully.
golangci-lint passes successfully on the latest Go toolchain.
Related Issues
Part of #2491

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the binary
-FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
 
 # Build arguments for binary
 ARG VERSION
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -59,7 +59,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v9
         with:
-          version: v2.1.6
+          version: v2.12.2
           args: --timeout=3m --issues-exit-code=0 ./...
           only-new-issues: true
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
         exclude: go\.(sum|mod)$
 
   - repo: https://github.com/golangci/golangci-lint
-    rev: v2.1.6
+    rev: v2.12.2
     hooks:
       - id: golangci-lint
         args: [--timeout=5m]


vprashar2929:
@shellyco-code Isn't this duplicate to #2492 ?

shellyco-code:
> @shellyco-code Isn't this duplicate to #2492 ?

This is not a duplicate of #2492. PR #2492 by @molloyzak13 covers only the Go 1.25 step (Dockerfile + pre-commit), whereas this PR targets Go 1.26 and also updates .github/workflows/pr-checks.yaml, which #2492 does not touch. Issue #2491 was assigned to me by @sunya-ch, and both PRs address different checklist items from that issue.

vprashar2929:
@sunya-ch @shellyco-code I suggest we first decide which version we want to jump to. IMO we should switch to 1.25 and then proceed with 1.26 later once we cut a new release Thoughts?

shellyco-code:
> @sunya-ch @shellyco-code I suggest we first decide which version we want to jump to. IMO we should switch to 1.25 and then proceed with 1.26 later once we cut a new release Thoughts?

That makes sense! I've updated this PR to target Go 1.25 instead. The Dockerfile now uses golang:1.25, and the golangci-lint is at v2.12.2 which is compatible with Go 1.25. This PR also updates .github/workflows/pr-checks.yaml which #2492 does not cover. Happy to do a follow-up PR for Go 1.26 once the next release is cut.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the binary
-FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.26 AS builder
```
nikimanoledaki: Please upgrade iteratively starting with 1.25 first then following up in a separate PR to upgrade to 1.25 - thank you!

shellyco-code: Thanks for the feedback! Updated the Dockerfile to golang:1.25 — I'll open a separate PR for the 1.26 bump once this one lands 👍

bitflicker64:
Checked the three bumps against the CI logs on #2481 (the `x/net v0.55.0` bump, now closed, which set `go 1.25.0`). They fix real failures there:

- `lint` and `pre-commit`: `can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.0)`
- `build-images` / `build-and-deploy` / `compose-deploy`: `go: go.mod requires go >= 1.25.0 (running go 1.24.13; GOTOOLCHAIN=local)` inside the `golang:1.24` builder

v2.12.2's `go.mod` declares `go 1.25.0`, so it clears the first, and `golang:1.25` clears the second. Bumping `.pre-commit-config.yaml` and `pr-checks.yaml` to the same rev matters, since both hit the identical error. Item 1 on #2491 looks covered.

Three notes:

1. `docs/developer/multi-arch-builds.md:4` still says the base images are `golang:1.24` and `ubi9:latest`. Worth flipping to 1.25 here so the doc does not go stale with the Dockerfile change.
2. #2492 is the one to close, not to land alongside. It bumps `go.mod` to 1.25.0 but leaves `pr-checks.yaml` at v2.1.6, so its own `lint` job fails with that same config error. This PR has the piece #2492 is missing, and the two otherwise touch the same `Dockerfile:2` and `.pre-commit-config.yaml` rev line.
3. Heads up for item 2 on #2491: the dependabot PR open right now, #2534, sets `go 1.26.0`. v2.12.2 caps at a 1.25 target, so after this lands #2534's lint still fails, with `go1.25` where the log currently says `go1.24`. That step needs golangci-lint v2.13.x plus `golang:1.26`. Expected given the plan to do 1.25 first, just flagging that #2534 stays red.

One question: since this PR touches no Go files, the paths filter skips `lint`, `fmt`, `vet` and `test-and-codecov`, and the `pre-commit` hook runs with `--new-from-rev HEAD`, so nothing here actually linted the tree with v2.12.2. v2.1.6 to v2.12.2 is a long way in default linter behaviour. Did a plain `golangci-lint run ./...` come back clean locally? CI is shielded by `--issues-exit-code=0` and `only-new-issues: true`, but contributors running the hook are not.

PR #2497, opened 2026-07-12
nvidia: propagate error when no dcgm-exporter endpoint is reachable

### LINKED ISSUE #2430: Shouldn't this be warning/error?
Shouldn't this be warning/error?

_Originally posted by @vprashar2929 in https://github.com/sustainable-computing-io/kepler/pull/2422#discussion_r2853441206_

iacker:
The fallback endpoint loop in initLocked logged each failed endpoint at Debug level and discarded the underlying error from testEndpoint. When every fallback failed, the returned error was a generic message with no diagnostic detail, making it hard to tell why dcgm-exporter could not be reached.

Log each unreachable fallback endpoint at Warn level with its error (consistent with the sibling local-discovery path), and wrap the last error into the final failure so the root cause and the number of tried endpoints are surfaced.

Fixes #2430

### DIFF
--- a/internal/device/gpu/nvidia/dcgm_exporter.go
+++ b/internal/device/gpu/nvidia/dcgm_exporter.go
@@ -181,16 +181,22 @@ func (d *DCGMExporterBackend) initLocked(ctx context.Context) error {
 	// Fallback to static endpoints. This covers cases where K8s API discovery
 	// fails (e.g., RBAC issues, dcgm-exporter in a non-standard namespace)
 	// or dcgm-exporter uses hostNetwork (localhost:9400).
+	var lastErr error
 	for _, ep := range d.fallbackEndpoints {
-		if err := d.testEndpoint(ctx, ep); err == nil {
+		err := d.testEndpoint(ctx, ep)
+		if err == nil {
 			d.endpoint = ep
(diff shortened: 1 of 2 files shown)


iacker:
Thanks @nikimanoledaki! Tagging @vimalk78 for visibility since you're closest to the NVIDIA/GPU side — happy to iterate on any feedback.

sunya-ch (commented):
Thank you for the PR! 

I think, instead of last error, it might be more useful to include the appended list of error mapping to the endpoint. Sometimes, the last error is something we cannot fix such as hardware support but we may be able to fix the other error happening previously such as RBAC issue.

iacker:
Great point @sunya-ch — you're right that hiding the earlier errors loses the actionable ones (RBAC, etc.) behind a possibly-unfixable last error. Fixed in the latest commit: the fallback loop now collects every endpoint failure and joins them with `errors.Join`, prefixing each with its endpoint, so all causes are surfaced. Extended the existing test to assert both endpoints appear in the aggregated message. Validated locally (`go vet` clean, `go test ./internal/device/gpu/nvidia/` passing on linux/arm64).

nikimanoledaki (approved):

PR #2503, opened 2026-07-18
feat(gpu): configure DCGM metrics cache TTL

### LINKED ISSUE #2429: next PR
next PR

_Originally posted by @vimalk78 in https://github.com/sustainable-computing-io/kepler/pull/2422#discussion_r2857561667_

dipankardas011:
## Summary

- Add the `experimental.gpu.metrics-cache-ttl` flag and `experimental.gpu.metricsCacheTTL` YAML option for DCGM metrics used in MIG power attribution.
- Use an effective default of `2s` with precedence: CLI flag > YAML value > default. Set `0s` to disable caching; reject negative values.
- Apply TTL changes to an initialized DCGM exporter backend without reinitializing it, and document the option in deployment templates.

Fixes #2429

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -157,6 +157,15 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 		}
 	}
 
+	// Inject configured DCGM metrics cache TTL for MIG power attribution
+	if cfg.Experimental != nil {
+		for _, m := range gpuMeters {
+			if c, ok := m.(gpu.DCGMMetricsCacheTTLConfigurable); ok {
+				c.SetDCGMMetricsCacheTTL(*cfg.Experimental.GPU.MetricsCacheTTL)
+			}
+		}
+	}
+
 	var services []service.Service
 
 	var podInformer pod.Informer
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -158,7 +158,7 @@ func createServices(logger *slog.Logger, cfg *config.Config) ([]service.Service,
 	}
 
 	// Inject configured DCGM metrics cache TTL for MIG power attribution
-	if cfg.Experimental != nil {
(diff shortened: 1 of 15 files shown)


iacker:
Took a look at this one since it touches the same DCGM/MIG path I was working in for #2497 and #2498. Ran the config tests locally on macOS — `TestApplyGPUConfig` passes including both new cases (`yaml metrics cache TTL is preserved without CLI override`, `gpu enabled and metrics cache TTL flags`). The two failures I hit, `TestGPUConfigYAML` and `TestHwmonConfig_YAMLParsing`, reproduce identically on `main` without your changes — they want `/sys` and `/proc`, so that's a macOS artifact, not this PR.

The precedence chain reads correctly: flag > YAML > `defaultGPUMetricsCacheTTL`, with the `flagsSet` check making sure an unset flag doesn't clobber a YAML value, and negatives rejected in validation rather than silently clamped.

Two things I liked and want to call out, since they're easy to get wrong:

`SetMetricsCacheTTL` nils `cachedMetrics` when the TTL changes. Without that, lowering the TTL at runtime would still serve an entry cached under the old, longer window — you'd set `1s` and keep getting 2s-old data until it aged out.

The `d.metricsCacheTTL > 0 &&` guard in `fetchMetrics` is what actually makes `0s` mean "disabled" rather than "expire instantly". Worth having the test at `dcgm_exporter_test.go:263` pin that, because the naive version of this check looks equivalent and isn't.

One question rather than a request: `SetDCGMMetricsCacheTTL` on the collector falls back to `c.initDCGM()` when the backend isn't a `*DCGMExporterBackend`. Is there a second implementation that path is for, or is it defensive? If it's only ever the one type, the fallback re-init is dead weight — and re-initializing on a config setter is a heavier side effect than the method name suggests.

Nothing blocking. Helm side looks fine too — `configmap.yaml` renders `.Values.config` wholesale via `toYaml`, so the new `metricsCacheTTL` key flows through without needing a template change.

dipankardas011:
@iacker thanks for taking a look, will check on these, will let you know

PR #2508, opened 2026-07-22
fix: prevent informer panic on nil virtual machine process

### LINKED ISSUE #2507: Bug: Process Informer panics on nil virtual machine instead of gracefully skipping
### Kepler Version

0.10.0 or later (Current/Supported)

### Bug Description

In `internal/resource/informer.go`, when iterating and discovering processes, if a process is classified as a virtual machine but `proc.VM` is nil, the code explicitly triggers a `panic()`.

```go
panic(fmt.Sprintf("process %d of type %s (%s) has is nil virtual machine", proc.PID, proc.Type, proc.Comm))


### Steps to Reproduce

1. Run Kepler as a daemon on a node running containerized workloads alongside virtual machines (e.g. KubeVirt/QEMU).
2. Encounter a transient edge case during process reading where a discovered process is classified as `Type: VMProcess` but its `VirtualMachine` metadata fails to populate (`proc.VM == nil`).
3. See error: The entire `resourceInformer` panics with `process <pid> of type VMProcess (<comm>) has is nil virtual machine`, causing the daemon to crash.


### Expected Behavior

Instead of a hard panic, the informer should use its structured logger to record an error about the malformed process metadata and `continue` safely to the next process in the loop. A single malformed process state should not crash the entire metrics collection daemon.


### Environment

- Kepler version: `main` (0.10.0 post-rewrite)
- OS/Arch: Linux/amd64
- Kubernetes version: v1.30


### Logs and Error Messages

```shell
panic: process <pid> of type VMProcess (<comm>) has is nil virtual machine
```

### Additional Context

This happens because the `proc.VirtualMachine` struct can occasionally fail to populate depending on the environment or the timing of the procfs read, leaving it as `nil`. I have a local patch ready to fix this by introducing a nil check, logging the error, and using `continue` to safely skip the process instead of crashing. I'll open a PR momentarily.

shellyco-code:
Fixes #2507 

**Description**
This PR converts a hard `panic()` in `internal/resource/informer.go` into a structured log warning and a graceful skip.

Previously, if Kepler discovered a process of type Virtual Machine but the `proc.VM` struct was `nil`, it would crash the entire metrics informer during `refreshVMs()` and `updateVMCache()`. Given that transient or malformed process states can occur in production environments, crashing the daemon is dangerous. This change ensures Kepler simply logs the anomaly and moves on to the next process.

**Changes:**
- Replaced `panic(fmt.Sprintf(...))` with `ri.logger.Error(...)` in `updateVMCache` and returned `nil`.
- Added a `vm == nil` check with a `continue` in `refreshVMs` to safely skip the process.
- Updated the test `updateVMCache with nil VM` in `internal/resource/procfs_reader_test.go` to assert a `nil` return instead of a panic.

**Testing:**
- Existing unit tests updated and passing.
- Verified that the metrics informer loops correctly without panicking.

### DIFF
--- a/internal/resource/informer.go
+++ b/internal/resource/informer.go
@@ -254,6 +254,10 @@ func (ri *resourceInformer) refreshVMs(vmProcs []*Process) error {
 	// Build running VMs from pre-categorized VM processes
 	for _, proc := range vmProcs {
 		vm := proc.VirtualMachine
+		if vm == nil {
+			ri.logger.Error("process of type VM has nil virtual machine", "pid", proc.PID, "comm", proc.Comm)
+			continue
+		}
 		vmsRunning[vm.ID] = ri.updateVMCache(proc)
 	}
 
@@ -433,7 +437,8 @@ func (ri *resourceInformer) Pods() *Pods {
 func (ri *resourceInformer) updateVMCache(proc *Process) *VirtualMachine {
 	vm := proc.VirtualMachine
 	if vm == nil {
-		panic(fmt.Sprintf("process %d of type %s (%s)  has is nil virtual machine", proc.PID, proc.Type, proc.Comm))
+		ri.logger.Error("process of type VM has nil virtual machine", "pid", proc.PID, "comm", proc.Comm)
+		return nil
 	}
 
 	cached, exists := ri.vmCache[vm.ID]
--- a/internal/resource/procfs_reader_test.go
+++ b/internal/resource/procfs_reader_test.go
@@ -1101,7 +1101,7 @@ func TestResourceInformer_InitRefreshErr(t *testing.T) {
 		mockProcFS.AssertExpectations(t)
 	})
 
-	t.Run("updateVMCache with nil VM panic", func(t *testing.T) {
+	t.Run("updateVMCache with nil VM", func(t *testing.T) {
 		mockProcFS := &MockProcReader{}
 		informer, err := NewInformer(WithProcReader(mockProcFS))
 		require.NoError(t, err)
@@ -1109,12 +1109,11 @@ func TestResourceInformer_InitRefreshErr(t *testing.T) {
 		proc := &Process{
 			PID:            123,
 			Type:           VMProcess,
-			VirtualMachine: nil, // This should cause panic
+			VirtualMachine: nil,
 		}
 
-		assert.Panics(t, func() {
-			informer.updateVMCache(proc)
-		})
+		vm := informer.updateVMCache(proc)
+		assert.Nil(t, vm)
 	})
 }
 


sunya-ch (commented):
This is probably good for now but I believe that we should expose out the failed process through the metrics as well as some users may not be able to check the Kepler log.
Also this would need to clarify what should we do for idle power attribution. For example, should we count these nil processes or VMs.

I think this PR needs attention and discussion in Kepler community wide before approving.

cc: @nikimanoledaki

iacker:
@shellyco-code I duplicated this in #2532 before spotting yours, my fault, I closed mine. Yours came first and fixes the same two spots.

One thing you might want to fold in here. The test only calls `updateVMCache` directly, so it does not cover the path that actually crashed. `refreshVMs` reads `vm.ID` before calling it, so on main the daemon dies on a nil deref there and never prints the panic message. Without a test on that loop, someone can reintroduce the deref later and the suite stays green.

This one is red on main and green with your patch:

```go
t.Run("Refresh skips VM process with no virtual machine metadata", func(t *testing.T) {
	mockProcFS := &MockProcReader{}
	informer, err := NewInformer(WithProcReader(mockProcFS))
	require.NoError(t, err)

	assert.NotPanics(t, func() {
		require.NoError(t, informer.refreshVMs([]*Process{{
			PID:            4242,
			Comm:           "qemu-system-x86",
			Type:           VMProcess,
			VirtualMachine: nil,
		}}))
	})

	assert.Empty(t, informer.VirtualMachines().Running)
})
```

Take it or rework it, it is yours.

iacker:
@shellyco-code still up for this one? The branch is behind `main` now and the two points above are still open: the `refreshVMs` test, and sunya-ch's question on how a skipped VM process should show up (a metric, or at least a warning with the count).

If you are short on time, say so and I will rebase and add both on top of your commit, keeping you as author. I would rather not let a 10 line fix for a daemon crash sit another month.

internal/resource/informer.go
```diff
@@ -254,6 +254,10 @@ func (ri *resourceInformer) refreshVMs(vmProcs []*Process) error {
 	// Build running VMs from pre-categorized VM processes
 	for _, proc := range vmProcs {
 		vm := proc.VirtualMachine
```
sthaha, line 256: I am curious when this happens. I would argue that vmProcs must have non nil proc.VirtualMachine or must be handled at the time of creation.

meaning the contract the caller of refreshVMs must satisfy is that proc.VirtualMachine is not nil. If it is nil, then it shouldn't be even part of vmProcs.

internal/resource/informer.go
sthaha, line 203: I think there must be a bug somewhere here ..

internal/resource/procfs_reader_test.go
```diff
@@ -1101,20 +1101,19 @@ func TestResourceInformer_InitRefreshErr(t *testing.T) {
...
-	t.Run("updateVMCache with nil VM panic", func(t *testing.T) {
+	t.Run("updateVMCache with nil VM", func(t *testing.T) {
 		mockProcFS := &MockProcReader{}
 		informer, err := NewInformer(WithProcReader(mockProcFS))
 		require.NoError(t, err)
 
 		proc := &Process{
 			PID:            123,
 			Type:           VMProcess,
-			VirtualMachine: nil, // This should cause panic
```
sthaha, line 1112: IMHO, the original contract is correct and the right behavior. For an internal/private function, IMHO, it is totally fine to expect a contract to be satisfied by the caller.

PR #2511, opened 2026-07-29
Add liveness and readiness probe, with dedicated endpoint

### LINKED ISSUE #2282: API endpoints for health(z)

leonorfmartins:
Adds liveness and readiness probes with dedicated `probe` endpoint.

Seeing probes in logs like this:

```
msg="Initializing service" service=api-server
msg="Initializing service" service=monitor
msg="Initializing service" service=prometheus
msg="Initializing service" service=probe          <- registered last, sees all services
msg="Starting Kepler"
msg="Running service" service=api-server
msg="Running service" service=probe               <- readyz flips to 200 here
msg="Running service" service=monitor
msg="Listening on" address=[::]:28282
```

Fixes https://github.com/sustainable-computing-io/kepler/issues/2282

### DIFF
--- a/manifests/helm/kepler/values.yaml
+++ b/manifests/helm/kepler/values.yaml
@@ -54,16 +54,23 @@ daemonset:
     #   cpu: 100m
     #   memory: 128Mi
 
+  # Probes use the dedicated /probe/* endpoints, which are cheap to serve.
+  # Avoid pointing them at /metrics: every scrape triggers a power computation.
   livenessProbe:
     httpGet:
-      path: /metrics
+      path: /probe/livez
       port: http
     initialDelaySeconds: 10
     periodSeconds: 60
 
   startupProbe: {}
 
-  readinessProbe: {}
+  readinessProbe:
+    httpGet:
+      path: /probe/readyz
+      port: http
+    initialDelaySeconds: 5
+    periodSeconds: 10
 
 config:
   log:
--- a/README.md
+++ b/README.md
@@ -102,10 +102,20 @@ make build && sudo ./bin/kepler
 cd compose/dev && docker compose up -d
 
 # 🐳 Kubernetes with Kustomize
+git clone https://github.com/sustainable-computing-io/kepler.git
+cd kepler && git checkout v0.11.4 # pin to a release tag, see note below
 kubectl kustomize manifests/k8s | \
-  sed -e "s|<KEPLER_IMAGE>|quay.io/sustainable_computing_io/kepler:latest|g" | \
+  sed -e "s|<KEPLER_IMAGE>|quay.io/sustainable_computing_io/kepler:v0.11.4|g" | \
   kubectl apply --server-side --force-conflicts -f -
 ```
+
+> **⚠️ Important:** Always check out a matching release tag (rather than `main`) before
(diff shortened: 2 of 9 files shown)


manifests/helm/kepler/values.yaml
```diff
@@ -54,16 +54,23 @@ daemonset:
     #   cpu: 100m
     #   memory: 128Mi
 
+  # Probes use the dedicated /probe/* endpoints, which are cheap to serve.
+  # Avoid pointing them at /metrics: every scrape triggers a power computation.
   livenessProbe:
     httpGet:
-      path: /metrics
+      path: /probe/livez
```
JulianLegler, line 61: Is there any regression we might introduce by changing this? I mean, it's good that we change it to the new path, but is there a case where someone might pull the new Kubernetes YAML files but still use an old Kepler image version? Because that would then probably lead to Kepler not starting at all. 
If this is not the case or not a real concern, then I am fine.

JulianLegler (commented):
Implementation is very clean and in line with the initial issue. Documentation also appears to be fine. 
Only some questions on a possible breaking manifest change from my side.

iacker:
Went through this from the docs side and checked the prose against `probe.go` — it lines up: `alive` / `ok` / `unavailable`, the `failures` map, GET+HEAD with 405 otherwise, and both endpoint paths all match the implementation. Nice to see the docs land in the same PR rather than after the fact.

Two things I noticed while reading:

**1. The version gap that Julian's Helm question also touches.** The docs say the endpoints are always exposed, which is true for this build, but someone pulling the updated manifests against an older Kepler image gets probes hitting paths that don't exist yet — and a failing liveness probe restart-loops the pod, which is a lot louder than a 404. Worth a one-liner in `configuration.md` saying which release first ships `/probe/*`, so people upgrading manifests and image separately know the ordering matters.

**2. The TLS note could go one step further.** `> **Note**: When TLS is enabled via web.configFile, probes must use scheme: HTTPS` is correct, but the shipped `values.yaml` snippet right above it has no `scheme`, so anyone enabling TLS has to work out that they need to edit the probe block too. Showing the `scheme: HTTPS` line in context would save that round trip.

Neither is blocking. The `/metrics` vs probe rationale is a genuinely useful thing to have written down — I'd been wondering why a scrape was doing real work.

JulianLegler:
We discussed this PR in the last community meeting. It would probably be required to have the README changed to mention using a specific release tag for cloning the repo if they plan to deploy it via the Kubernetes YAML as a normal user. As this would introduce the regression that when using the main branch with the changed daemonset.yaml and the latest release image would lead to Kepler not being able to get a positive liveness response. 

So this README change blocks this PR currently. Do you (@leonorfmartins) want to add it to this PR, or should I open a new PR for this?

leonorfmartins:
@JulianLegler apologies as I was away for summer holidays, but I can absolutely update the README, as this can be considered a breaking change!

JulianLegler (approved):
Thanks a lot! 

Should be fine to merge now.

laurall974 (dismissed):
lgtm!  could you rebase?

laurall974 (dismissed):

laurall974 (approved):

PR #2515, opened 2026-08-02
feat(resource): detect VirtualBox, VMware and Xen VM processes

### LINKED ISSUE #2514: Detect VirtualBox, VMware and Xen VM processes
### Target Version

### Feature Description

Extend VM process detection in `internal/resource` to recognize VirtualBox, VMware and Xen guests, in addition to the QEMU/KVM guests that are detected today.

The `Hypervisor` type in `internal/resource/types.go` already declares `VirtualBoxHypervisor`, `VMwareHypervisor` and `XenHypervisor` constants, and `internal/resource/vm.go` carries a TODO for this:

```go
// TODO: add patterns for virtual box,  VMware, Xen
```

### Problem Statement

On hosts running VirtualBox, VMware Workstation/Player or Xen, the per-VM processes are currently classified as regular processes instead of VM workloads. As a result their energy is not attributed at the VM level and the `kepler_vm_*` metrics are missing for those guests, even though the process-level building blocks are already in place.

### Proposed Solution

Add detection patterns and metadata extraction for the three hypervisors:

- VirtualBox: match the per-VM `VBoxHeadless`, `VBoxSDL` and `VirtualBoxVM` processes; take the VM ID from the `--startvm` argument and the VM name from `--comment`
- VMware: match the per-VM `vmware-vmx` process; take the VM name from the `.vmx` config file path passed as an argument
- Xen: the device model runs through QEMU, so distinguish it from a plain QEMU/KVM guest via the `-xen-domid` flag; use the domain ID as VM ID and the QEMU style `-name` argument as VM name

While adding this, two robustness fixes fall out naturally:

- `vmProcessPatterns` becomes an ordered slice instead of a map so classification stays deterministic when a command line matches more than one pattern
- the generated fallback VM name no longer panics when the VM ID is shorter than 8 characters (a Xen domain ID like `5` would have hit `vmID[:8]`)

I have a patch ready with unit tests covering the new patterns and will open a PR shortly.

### Alternatives Considered

Detecting guests via hypervisor management APIs (VBoxManage, vmrun, xl) was considered but rejected: command line based detection matches the existing QEMU/KVM approach, needs no extra dependencies and works read-only from procfs.

### Additional Context

Xen domain IDs change across guest reboots, so the VM ID label for Xen guests is stable only for the lifetime of a domain. This matches the behavior of the existing hash based fallback IDs.

younsl:
## Summary

Closes #2514

Extends VM process detection in `internal/resource` beyond QEMU/KVM to the remaining hypervisors already declared in the `Hypervisor` type, resolving the TODO in `vm.go`:

```go
// TODO: add patterns for virtual box,  VMware, Xen
```

## Changes

- **VirtualBox**: matches the per-VM `VBoxHeadless`, `VBoxSDL` and `VirtualBoxVM` processes. VM ID comes from the `--startvm` argument (single or double dash), VM name from `--comment`.
- **VMware**: matches the per-VM `vmware-vmx` process. VM name comes from the `.vmx` config file path passed as an argument.
- **Xen**: the Xen device model runs through QEMU, so the `-xen-domid` flag distinguishes it from a plain QEMU/KVM guest. The domain ID becomes the VM ID and the QEMU style `-name` argument the VM name.

Robustness fixes in the shared detection path:

- `vmProcessPatterns` is now an ordered slice instead of a map, so classification stays deterministic when a command line matches more than one pattern.
- The generated fallback VM name no longer panics when the VM ID is shorter than 8 characters. A Xen domain ID like `5` would have hit the previous `vmID[:8]` slice.
- Patterns require a path prefix (`virtualbox/`, `bin/`) to avoid classifying unrelated processes that merely mention a hypervisor binary in their arguments.

## Testing

- Added table-driven unit tests for all new patterns and extraction helpers: VirtualBox headless/GUI/SDL variants, single and double dash arguments, VMware `.vmx` name extraction, Xen domain ID extraction and the QEMU name fallback.
- `go test -race ./internal/resource/` passes on Linux.
- `go vet` and `gofmt` clean.

## Notes

- Xen domain IDs change across guest reboots, so the VM ID label for Xen guests is stable only for the lifetime of a domain. This matches the behavior of the existing hash based fallback IDs.
- Behavior change for Xen hosts only: the Xen device model path (`lib/xen/bin/qemu-system-*`) already matched the QEMU pattern, so those guests were previously reported with `hypervisor="kvm"` and a name based VM ID. They now report `hypervisor="xen"` with the domain ID, which changes those label values. QEMU/KVM detection on non-Xen hosts is unchanged, covered by the pre-existing tests.
- VMware VM IDs derive from the `.vmx` file basename, so two powered-on VMs with identical config file names would share an ID. This mirrors the existing name based fallback for QEMU guests without a UUID.
- Xen PV guests that run without a QEMU device model have no host process and remain undetectable by design.

### DIFF
--- a/internal/resource/vm.go
+++ b/internal/resource/vm.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 )
 
@@ -14,11 +15,21 @@ var (
 	// QEMU/KVM patterns - matches both qemu-system-* and qemu-kvm variants
 	qemuPattern = regexp.MustCompile(`(bin/qemu-system-\w+|libexec/qemu-kvm|bin/kvm)`)
 
-	// TODO: add patterns for virtual box,  VMware, Xen
+	// VirtualBox patterns - matches the per-VM headless, SDL and GUI processes
+	virtualBoxPattern = regexp.MustCompile(`virtualbox/(VBoxHeadless|VBoxSDL|VirtualBoxVM)`)
 
-	// VM process name patterns
-	vmProcessPatterns = map[*regexp.Regexp]Hypervisor{
-		qemuPattern: KVMHypervisor,
+	// VMware patterns - vmware-vmx runs as one process per powered-on VM
+	vmwarePattern = regexp.MustCompile(`bin/vmware-vmx`)
+
+	// VM process name patterns, checked in order so that classification stays
+	// deterministic even if a command line matches more than one pattern
+	vmProcessPatterns = []struct {
+		pattern    *regexp.Regexp
+		hypervisor Hypervisor
+	}{
+		{qemuPattern, KVMHypervisor},
+		{virtualBoxPattern, VirtualBoxHypervisor},
+		{vmwarePattern, VMwareHypervisor},
 	}
 )
 
@@ -48,7 +59,11 @@ func vmInfoFromProc(proc procInfo) (*VirtualMachine, error) {
 	vm.Name = vmNameFromCmdLine(cmdline, hypervisor)
 
 	if vm.Name == "" {
-		vm.Name = fmt.Sprintf("%s-%s", hypervisor, vmID[:8])
+		shortID := vmID
+		if len(shortID) > 8 {
+			shortID = shortID[:8]
+		}
+		vm.Name = fmt.Sprintf("%s-%s", hypervisor, shortID)
 	}
 
 	return vm, nil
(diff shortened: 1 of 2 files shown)


younsl:
Addressed the Codecov patch coverage gap in 67f6c73.

The 4 uncovered lines were the VirtualBox and VMware arms of the vmNameFromCmdLine switch. The helper functions behind them were unit tested, but no test reached the arms through the dispatching switch itself.

What the new commit adds:

- Direct cases in TestVMNameFromCmdLine for the VirtualBox, VMware and Xen arms
- End-to-end cases in TestVMInfoFromProc for all three new hypervisors, including a Xen guest without a name argument whose short domain ID exercises the generated fallback name path

Verified on linux/amd64 with race detection enabled: all resource package tests pass, vm.go has no uncovered lines left and package coverage rises to 95.4 percent. The Codecov report should update once the workflows are approved to run.

JulianLegler (approved):
This PR seems to heavily follow the existing detection structure implemented for qemu/kvm and extend it with VirtualBox, VMware, and Xen. 

The impacted code areas are limited to one file, and all new calls are well guarded, limiting the introduction of possible regressions with this PR. 

Overall, I think this is a good looking PR. I didnt run the tests, but if they are all green, then I am all fine.

younsl:
Rebased onto `main` (was 8 commits behind, no conflicts, diff unchanged).

Workflows on the new head `88921a9` are at `action_required`. @sunya-ch @vprashar2929 — could one of you approve the run and review when you get a chance?

sunya-ch (approved):
LGTM. Thank you!

PR #2519, opened 2026-08-10
fix(resource): skip processes with unreadable /proc entries

### LINKED ISSUE #2518: One unreadable /proc/PID/exe blocks all power metrics from being exported
I ran Kepler on a cloud GPU host to test the experimental GPU power monitoring flag. NVML detected the GPU fine. But no metrics ever appeared on /metrics, not even basic CPU ones.

I traced it to internal/resource/informer.go. populateProcessFields returns a hard error when readlink on /proc/PID/exe fails with permission denied. That happens for a process running under a different uid, in my case a system nginx with pid 80. refreshProcesses joins that single error and returns it. monitor.go then calls resources.Refresh in calculatePower and firstReading, and both bail out on any non nil error before computing power for anything. Node power, process power, container power, GPU power, none of it runs.

So one process the exporter cannot introspect silently kills the entire metrics pipeline every cycle, forever, with no retry or skip.

Steps to reproduce. Run kepler with experimental gpu enabled and dev fake cpu meter enabled on a host where at least one running process has an exe symlink the kepler process cannot read, which is common on shared or sandboxed hosts. Query /metrics. Only go runtime metrics and kepler_build_info show up.

Log line for reference.
level=ERROR source=internal/monitor/monitor.go:387 msg="snapshot rebuild failed to refresh resources" service=monitor error="failed to get process executable: readlink /proc/80/exe: permission denied"

I would expect a process the informer cannot introspect to be skipped and logged, not to abort the whole refresh. Happy to send a PR if that matches the intended behaviour, roughly treating permission denied the same way os.IsNotExist is already handled a few lines above in refreshProcesses, just skip and continue instead of joining into refreshErrs.

iacker:
Fixes #2518

## What

One process with an unreadable /proc/PID/exe (permission denied) made refreshProcesses return an error. The monitor then aborted every power computation cycle at resources.Refresh, so no power metrics were ever exported. Not node, not process, not GPU.

I hit this while testing the experimental GPU power monitoring on a cloud A100 host. NVML detected the GPU fine, but /metrics only ever showed build info and cpu info. A sandboxed nginx process on the host had an exe symlink kepler could not read.

## Change

Treat permission errors in the refresh loop the way the not-exist case is already handled a few lines above. Skip the process instead of joining the error into refreshErrs.

The skip is not silent. refreshProcesses counts the skipped processes and warns with the count and the total. An operator can then see that part of the node power is being redistributed. It only logs when the count changes, since Refresh runs every 5s and an unreadable entry is usually a steady state.

## Testing

go test -race ./internal/resource/ passes, package at 95.7%. Tests cover the skip, the repeated refresh staying silent, and the count going back to zero. Also verified end to end on the A100 host where the bug was found, details in #2518.

### DIFF
--- a/internal/resource/informer.go
+++ b/internal/resource/informer.go
@@ -189,6 +189,15 @@ func (ri *resourceInformer) refreshProcesses() ([]*Process, []*Process, error) {
 				continue
 			}
 
+			if errors.Is(err, os.ErrPermission) {
+				// Processes owned by other users (e.g. in sandboxed or
+				// shared environments) may have unreadable /proc entries.
+				// Skip them instead of failing the whole refresh, which
+				// would block all power metrics from being exported.
+				ri.logger.Debug("Skipping process with unreadable /proc entry", "pid", pid, "error", err)
+				continue
+			}
+
 			ri.logger.Debug("Failed to get process info", "pid", pid, "error", err)
 			refreshErrs = errors.Join(refreshErrs, err)
 			continue
--- a/internal/resource/informer.go
+++ b/internal/resource/informer.go
@@ -179,6 +179,7 @@ func (ri *resourceInformer) refreshProcesses() ([]*Process, []*Process, error) {
 
 	// Refresh process cache and update running processes
 	var refreshErrs error
+	unreadable := 0
 	for _, p := range procs {
 		pid := p.PID()
 		// start by updating the process
@@ -194,6 +195,7 @@ func (ri *resourceInformer) refreshProcesses() ([]*Process, []*Process, error) {
 				// shared environments) may have unreadable /proc entries.
 				// Skip them instead of failing the whole refresh, which
 				// would block all power metrics from being exported.
+				unreadable++
 				ri.logger.Debug("Skipping process with unreadable /proc entry", "pid", pid, "error", err)
 				continue
 			}
@@ -211,6 +213,14 @@ func (ri *resourceInformer) refreshProcesses() ([]*Process, []*Process, error) {
 		case VMProcess:
 			vmProcs = append(vmProcs, proc)
 		}
+	}
+
+	// Surface permission-skipped processes: they are excluded from
+	// attribution, so an operator needs more than a debug line to notice
+	// that a subset of the node's power is being redistributed.
+	if unreadable > 0 {
+		ri.logger.Warn("Skipped processes with unreadable /proc entries; their power will be attributed to the remaining processes",
+			"count", unreadable, "total", len(procs))
 	}
 
 	// Find terminated processes
--- a/internal/resource/informer.go
+++ b/internal/resource/informer.go
@@ -91,6 +91,10 @@ type resourceInformer struct {
 	pods        *Pods
 
 	lastScanTime time.Time // Time of the last full scan
+
+	// Number of processes skipped for unreadable /proc entries in the previous
+	// refresh, used to warn on transitions only instead of on every refresh.
+	lastUnreadableCount int
 }
 
 var _ Informer = (*resourceInformer)(nil)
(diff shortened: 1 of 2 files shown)


JulianLegler (commented):
Thanks for the PR. It appears to be reasonable to catch the error that is responsible for making the whole /metrics scrape call unresponsive or let it fail just because of maybe one unreadable process.

The trade-off of this approach, from how I understand it (please correct me if I am wrong or this is not an issue):
This enables the silent exclusion (only visible via a debug log message) of maybe energy-heavy processes that then lead to wrongful energy attribution as the energy is then only distributed to the working/accessible subset of processes.  

So we probably have to decide how valuable a wrong energy attribution is compared to a non-energy attribution in this situation. Or how likely this will happen or if this is even a problem at all.

iacker:
That trade-off is real. `/proc` entries disappearing during enumeration is normal process-exit churn. Failing the whole refresh makes metrics unavailable, while skipping keeps node totals available but can undercount process and container attribution for that interval. I think the exclusion should be observable rather than debug-only, either with a warning or a skipped-process counter.

iacker:
@JulianLegler pushed the observability part in 29dc1381.

The skip is no longer debug-only. `refreshProcesses` counts the permission-denied processes and warns with the count and the total, so the message says what the exclusion costs you.

On the trade-off, I think the current behavior is the worse one. One unreadable /proc entry makes `refreshProcesses` return an error. That fails the whole scrape. No node total, no container attribution, nothing. Skipping keeps those correct. You only lose the share of the skipped PIDs, which gets spread over the rest.

I can turn the counter into a metric like `kepler_process_skipped_total` if you want it alertable instead of log-only. Say the word.

iacker:
Correction to my own comment above, I had the log frequency wrong.

`Refresh` runs every 5s, and an unreadable `/proc` entry owned by another user is a steady state. Warning on every refresh would have been around 17k lines a day. That is the kind of log people mute.

Fixed in ce41ef78, it only logs when the count changes:

```
0 -> n   warn once, with the count
n -> n   silent
n -> 0   one info line, /proc readable again
```

Tests cover both directions, ff26cf76 adds the reset case. Package is at 95.7%.

vprashar2929:
@iacker Might be worth looking into but we can also have a self metric for skipped/unreadable processes. Something that can give us indication apart from logs

vprashar2929 (dismissed):

iacker:
Agreed, a counter for skipped processes would be more useful than log lines. I would rather keep this PR to the crash fix. The metric naming and the collector it belongs in need their own discussion.

iacker:
Yes, that is the `kepler_process_skipped_total` metric I offered above. I will open a follow-up PR for it rather than grow this one, since it is already approved.

ghost (commented):
From memory, I think this breaks the per process power computation which is the core for aggregating the power usage.

Could you please validate how the number of processes and per process stat is used to allocate the power use before merging this PR? 

cc: @vprashar2929

sthaha:
From memory, I think this breaks the per process power computation which is the core for aggregating the power usage.

Could you please validate how the number of processes and per process stat is used to allocate the power use before merging this PR? 

cc: @vprashar2929

iacker:
I checked the attribution path, and your concern is valid. Skipped CPU time is removed from the denominator, so its power gets redistributed to visible workloads. I can limit this PR to `os.ErrNotExist` and handle permission errors separately.

iacker:
Limited to `os.ErrNotExist` only as discussed. Permission errors now return to the error path.

JulianLegler:
Can you explain in which situation the ErrNotExist error appears? Is this solving your initial issue intended with this PR?

JulianLegler:
In the community meeting we discussed that we might close this PR, as the intended behavior should be to fail fast instead of maybe risking dealing with wrong attribution data because of secretly dropping processes. 
If there are other opinions we can discuss here again. Just want to keep you updated about the discussions in the todays community meeting.

iacker:
Correction: my previous comment misstated the current patch. The PR does skip permission-denied `/proc` entries and reports the count in a warning. Julian's concern is valid: this can redistribute node power across readable processes and affect attribution.

The original failure is the `/proc` race where a process exits between PID enumeration and `/proc/<pid>/exe` access. I will not claim the attribution trade-off is resolved.

iacker:
Closing this, following the community meeting decision to fail fast rather than skip entries and risk skewed attribution. The /proc race from #2518 still exists, so I will leave the issue open for a fail fast approach that does not silently drop processes.

PR #2528, opened 2026-08-13
docs: fix broken star history chart

Dessalines39394:
The star history chart shown in the README is currently broken due to the previous provider hitting GitHub stargazer API restrictions. This switches the chart to star-history.dera.page, which uses a different data source that needs no API token, so the chart renders correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Our project adheres to the Linux Foundation's Generative AI Policy, which can be
 
 ## ⭐ Star History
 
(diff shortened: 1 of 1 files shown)


vprashar2929:
@Dessalines39394  Can you address the DCO and CI failure?

nikimanoledaki:
It looks like potentially a bot account (apologies if this is wrong). I researched this to confirm that the website itself is safe and that the fix solves the issue. Given the lack of response, I'll take on the PR to get this merged 👍

nikimanoledaki (approved):

sunya-ch:
FYI, reference for this change https://jlai.lu/post/41713315

Issue #2523, opened 2026-08-11
chore(helm): Archive Kepler Helm repo


nikimanoledaki:
We decided to archive the Kepler Helm repo since the Helm Chart is part of the Kepler repo. 

## Next Steps
- [x] Add a notice of deprecation in the Helm repo's README and redirect folks to up-to-date documentation about the new Helm Chart 
- [x] Archive the repo

laurall974:
- [x] Investigate https://github.com/sustainable-computing-io/kepler-metal-ci. @iacker pointed out  it also installs from the deprecated kepler-helm-chart

laurall974:
```
 helm install kepler kepler/kepler --namespace kepler --create-namespace --dry-run --devel
WARNING: This chart is deprecated
```

laurall974:
@sunya-ch could you archive it https://github.com/sustainable-computing-io/kepler-helm-chart ? I dont have the access to do it

sunya-ch:
Thanks to 
- https://github.com/sustainable-computing-io/kepler-helm-chart/pull/108 by @laurall974 

I have archived the repo and closed stale and resolved issues.

Here are transferred issues which might need revisit:

- https://github.com/sustainable-computing-io/kepler/issues/2539
- https://github.com/sustainable-computing-io/kepler/issues/2540
- https://github.com/sustainable-computing-io/kepler/issues/2538

PR #2531, opened 2026-08-14
chore(helm): archive Kepler Helm repo

laurall974:
Related to https://github.com/sustainable-computing-io/kepler/issues/2523

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -107,9 +107,9 @@ significant contributions including:
 
 - Core Kepler development and maintenance
 - Test suite maintenance
-- Deployment and integration (including
-  the [Kepler Operator](https://github.com/sustainable-computing-io/kepler-operator),
-  [Helm Chart](https://github.com/sustainable-computing-io/kepler-helm-chart))
+- Deployment and integration including
+  - the [Kepler Operator](https://github.com/sustainable-computing-io/kepler-operator),
+  - the [Helm Chart](https://github.com/sustainable-computing-io/kepler/tree/main/manifests/helm)
 - [Model server](https://github.com/sustainable-computing-io/kepler-model-server)
   and model development
 - [Continuous integration](https://github.com/sustainable-computing-io/kepler-action)


iacker:
Thanks for picking this up. I checked the new links, manifests/helm, docs/user/installation.md and the quay OCI chart all resolve fine.

One render nit on GOVERNANCE.md. The "(the standalone ...)" line is indented 2 spaces, so it attaches to the parent bullet instead of the Helm Chart item. 4 spaces should fix it.

The repo is also still open, so maybe "is being archived" until the archive actually lands. Your call.

vprashar2929:
@laurall974  Can you address the DCO failure? Otherwise LGTM

nikimanoledaki (dismissed):

GOVERNANCE.md
```diff
@@ -107,9 +107,10 @@ significant contributions including:
 
 - Core Kepler development and maintenance
 - Test suite maintenance
-- Deployment and integration (including
-  the [Kepler Operator](https://github.com/sustainable-computing-io/kepler-operator),
-  [Helm Chart](https://github.com/sustainable-computing-io/kepler-helm-chart))
+- Deployment and integration including
+  - the [Kepler Operator](https://github.com/sustainable-computing-io/kepler-operator),
+  - the [Helm Chart](https://github.com/sustainable-computing-io/kepler/tree/main/manifests/helm)
+  (the standalone `kepler-helm-chart` repo has been archived)
```
sunya-ch: may not need this line

sunya-ch (commented):
lgtm mostly.
Just thinking that we may not need to mention about deprecated repo here.

No strong objection to merge as is.

GOVERNANCE.md
laurall974: I am happy to change whatever feels the best option. Removing it now.

vprashar2929 (approved):

sunya-ch (approved):

PR #2532, opened 2026-08-15
fix(resource): skip VM processes with nil metadata instead of panicking

### LINKED ISSUE #2507: Bug: Process Informer panics on nil virtual machine instead of gracefully skipping

iacker:
Closes #2507.

A process classified as a VM with no VM metadata crashed the whole informer, and with it the daemon and all metrics collection.

Worth noting the panic in `updateVMCache` was never actually reached. `refreshVMs` reads `vm.ID` one line before calling it, so the process died on a nil pointer dereference first, without the message the panic was there to give. Both are fixed:

- `refreshVMs` skips the process and logs an error with the pid, comm and type
- `updateVMCache` returns nil instead of panicking

This matches what `updateContainerCache` already does for a nil container, so VMs and containers now behave the same way.

Tests: the old test asserted the panic, so it is now an assertion that it does not panic, plus a new case covering the `refreshVMs` path. Both are red on main and green here. `go test -race ./internal/resource/...` passes.

### DIFF
--- a/internal/resource/informer.go
+++ b/internal/resource/informer.go
@@ -254,6 +254,14 @@ func (ri *resourceInformer) refreshVMs(vmProcs []*Process) error {
 	// Build running VMs from pre-categorized VM processes
 	for _, proc := range vmProcs {
 		vm := proc.VirtualMachine
+		if vm == nil {
+			// A malformed process should not bring the whole informer down.
+			// Containers are already skipped this way in updateContainerCache.
+			ri.logger.Error("Skipping VM process with no virtual machine metadata",
+				"pid", proc.PID, "comm", proc.Comm, "type", proc.Type)
+			continue
+		}
+
 		vmsRunning[vm.ID] = ri.updateVMCache(proc)
 	}
 
@@ -433,7 +441,7 @@ func (ri *resourceInformer) Pods() *Pods {
 func (ri *resourceInformer) updateVMCache(proc *Process) *VirtualMachine {
 	vm := proc.VirtualMachine
 	if vm == nil {
-		panic(fmt.Sprintf("process %d of type %s (%s)  has is nil virtual machine", proc.PID, proc.Type, proc.Comm))
+		return nil
 	}
 
 	cached, exists := ri.vmCache[vm.ID]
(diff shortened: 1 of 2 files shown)


iacker:
Closing this, #2508 got there first on 22 July and covers the same two spots. I missed it before opening, sorry for the noise.

For anyone landing here from #2507, one detail the issue gets wrong: the daemon never actually reached that `panic`. `refreshVMs` reads `vm.ID` one line before calling `updateVMCache`, so it died on a nil pointer dereference first and the panic message was never printed. #2508 fixes both spots, so the outcome is the same either way.

The only thing here that #2508 does not have is a test for the `refreshVMs` path, since its test only covers `updateVMCache` directly. I left it as a suggestion over there rather than keeping a competing PR open.

Issue #2524, opened 2026-08-11
dcgmEndpoint is ignored and autoDetect is always active


cloudymax:
### Kepler Version

0.10.0 or later (Current/Supported)

### Bug Description

I am unable to specify a `dcgmEndpoint`

when added to the config eg.
```yaml
            gpu:
              enabled: true
              idlePower: 0
              dcgmEndpoint: "http://192.168.2.70:9400/metrics"
```

The container ignores this value and auto-detect is still attempted:

```log
2026-08-11 21:24:28.718Configuration
2026-08-11 21:24:28.718━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
2026-08-11 21:24:28.718log:
2026-08-11 21:24:28.718debug    level: debug
2026-08-11 21:24:28.718    format: text
2026-08-11 21:24:28.718host:
2026-08-11 21:24:28.718    sysfs: /sys
2026-08-11 21:24:28.718    procfs: /proc
2026-08-11 21:24:28.718monitor:
2026-08-11 21:24:28.718    interval: 5s
2026-08-11 21:24:28.718    staleness: 500ms
2026-08-11 21:24:28.718    maxTerminated: 100
2026-08-11 21:24:28.718    minTerminatedEnergyThreshold: 10
2026-08-11 21:24:28.718rapl:
2026-08-11 21:24:28.718    zones: []
2026-08-11 21:24:28.718exporter:
2026-08-11 21:24:28.718    stdout:
2026-08-11 21:24:28.718        enabled: false
2026-08-11 21:24:28.718    prometheus:
2026-08-11 21:24:28.718        enabled: true
2026-08-11 21:24:28.718        debugCollectors:
2026-08-11 21:24:28.718            - go
2026-08-11 21:24:28.718        metricsLevel:
2026-08-11 21:24:28.718            - node
2026-08-11 21:24:28.718            - process
2026-08-11 21:24:28.718            - container
2026-08-11 21:24:28.718            - vm
2026-08-11 21:24:28.718            - pod
2026-08-11 21:24:28.718web:
2026-08-11 21:24:28.718    configFile: ""
2026-08-11 21:24:28.718    listenAddresses:
2026-08-11 21:24:28.718        - :28282
2026-08-11 21:24:28.718debugdebug:
2026-08-11 21:24:28.718    pprof:
[...]

PR #2536, opened 2026-08-18
fix(config): warn about unknown config fields instead of dropping them silently

iacker:
Config loading uses `yaml.Unmarshal`, which ignores keys that match no struct field. A typo, or an option that only exists on main, is dropped with no feedback and never appears in the startup config dump. That is what happened to `dcgmEndpoint` on 0.11.4 in #2524.

This adds a second decode with `KnownFields(true)` to collect the ignored keys and log them as a warning at startup. cloudymax asked for a warning rather than a hard error, so unknown keys stay non fatal and existing configs keep working.

Example output:

```
level=WARN msg="ignoring unknown config fields; check for typos or options from another release" fields="[dcgmEndpoint]"
```

Refs #2524

Tests: new `config/unknown_test.go` covers a clean config, an unknown nested key, an unknown top level key, and several at once. `go test -race ./config/...` passes in a Linux container. Two failures in `TestInvalidConfigurationValues` also fail on unpatched main, they are unrelated.

AI assistance was used, mostly because my English is poor. I reviewed the change and ran the tests above.

### DIFF
--- a/cmd/kepler/main.go
+++ b/cmd/kepler/main.go
@@ -46,6 +46,7 @@ func main() {
 	printConfigInfo(logger, cfg)
 
 	cfg.ApplyCpuMeterDeprecations(logger)
+	cfg.LogUnknownFields(logger)
 
 	services, err := createServices(logger, cfg)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


sthaha (approved):
This is a is a great catch @iacker ! Thank you for the PR.

Issue #2363, opened 2025-12-01
DRAM metrics not supported in 0.10.0 post rewrite? kepler_container_joules_total missing


MA3CIN:
### Target Version

Current version (0.10.0+) - New Architecture

### Feature Description

Comparing documentations of versions 0.9.0 and 0.10.0, as well as their deployments on my clusters, the archive version has more metrics per container and process than the newest one.

I'm specifically talking about the "kepler_container_joules_total " metrics, which included "CPU, dram, gpus, and other host components" ( 0.9.0 docs - https://sustainable-computing.io/archive/design/metrics/#kepler-metrics-for-container-energy-consumption).
The newest version only supports cpu related metrics for containers - kepler_container_cpu_joules_total  (0.10.0 - https://sustainable-computing.io/kepler/design/metrics/)

I've read the Slack announcement, including the  new "limitations" section, which talks about only supporting the RAPL/powercap framework. I've also read the Slack comments regarding the "past models being way off on power consumption"

<img width="350" height="192" alt="Image" src="https://github.com/user-attachments/assets/c97a9955-3c6c-4516-9e1e-0e4acc76dc82" />

Will the DRAM related metrics be back eventually in newer Kepler versions? 

### Problem Statement

Less metrics available than before in 0.10.0 (specifically dropping DRAM support)

### Proposed Solution

A comment regarding DRAM/GPU power metrics's place in the Kepler Roadmap

vprashar2929:
@MA3CIN Thanks for raising this. Quick rundown of what changed and why:

In the archived version (e.g., v0.9), Kepler exposed separate, static metrics for different components (like `package`, `core`, `dram`), sometimes even when the underlying hardware support was missing. This could lead to misleading data, as you noted, where metrics were produced but were either empty or based on estimations that might not reflect reality.

In the current version (v0.10.0+), we have shifted to a more dynamic approach that strictly reflects the available hardware sensors (RAPL). Instead of separate metric names for each component, we now use a single metric `kepler_container_cpu_joules_total` with a `zone` label.

This `zone` label is populated dynamically based on the RAPL domains available on your specific machine:
- If your machine supports **Package** and **Core** power domains, you will see `zone="package"` and `zone="core"`.
- If your machine also supports **DRAM** power domain, you will additionally see `zone="dram"`.

[...]

PR #2537, opened 2026-08-18
docs(metrics): add more information about zones and how to work with them in the kepler rewrite

JulianLegler:
The goal of this PR is to make the role of `zones` in metrics more clear in the Kepler remake (>= 0.10.0) by shortly addressing them right in front of the metrics documentation and giving some advice and warnings. 

This hopefully resolves the existing discussion about the lacking documentation as mentioned in #2363 and thereby also closes this issue. 

Do I understand correctly that we also would have to create a PR in kepler-docs to make this go live?

### DIFF
--- a/docs/user/metrics.md
+++ b/docs/user/metrics.md
@@ -11,6 +11,20 @@ Kepler exports metrics in Prometheus format that can be scraped by Prometheus or
 - **COUNTER**: A cumulative metric that only increases over time
 - **GAUGE**: A metric that can increase and decrease
 
+### Energy Zones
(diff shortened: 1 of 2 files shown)


sthaha (dismissed):
LGTM! Thanks for the patch @JulianLegler

JulianLegler:
pre-commit failed because of a trailing white space. Fixed it hopefully now in the last commit. But now I think I need a new approval now @sthaha

sthaha (approved):
LGTM!

PR #2551, opened 2026-09-08
chore: upgrade CI infrastructure to support Go 1.25

leonorfmartins:
## Description

Upgrades Kepler's CI infrastructure to support Go 1.25, based on #2496 (stalled, reopened here).

- `Dockerfile`: builder base image `golang:1.24` → `golang:1.25`
- `.github/workflows/pr-checks.yaml`: golangci-lint action `v2.1.6` → `v2.12.2`
- `.pre-commit-config.yaml`: golangci-lint hook `v2.1.6` → `v2.12.2`

Ran in Linux containers matching CI (`ubuntu-latest`). `make image` command + tests pass locally.

Part of https://github.com/sustainable-computing-io/kepler/issues/2491


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the binary
-FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
```
bitflicker64: Confirmed safe by CI here: `build-images`, `build-and-deploy` and `compose-deploy` all pass on this head, and they are the jobs that consume this stage.

One follow-on in the same tree: `docs/developer/multi-arch-builds.md:4` still reads "images (`golang:1.24` and `ubi9:latest`) are already multi-arch", so it goes stale the moment this merges. Cheaper to flip here than in a separate docs PR.

`CONTRIBUTING.md:78` and `docs/user/installation.md:7` also say 1.24, but those track the `go` directive in go.mod, which this PR leaves at `1.24.0`, so they should stay.

.pre-commit-config.yaml
```diff
@@ -54,7 +54,7 @@ repos:
         exclude: go\.(sum|mod)$
 
   - repo: https://github.com/golangci/golangci-lint
-    rev: v2.1.6
+    rev: v2.12.2
```
bitflicker64: Worth knowing that this hook and the workflow pin resolve golangci-lint by different means: `language: golang` makes pre-commit build it from source, so the binary's Go version follows golangci-lint's own go.mod rather than the release build. v2.12.2 declares `go 1.25.0`, so this rev produces a go1.25 binary and still fails a 1.26 target. That is what the `pre-commit` job hits on #2548 today, at go1.24. v2.13.2 declares `go 1.26.0` and clears it.

Separately, the green `pre-commit` check here does not show the tree is clean under v2.12.2. This hook id is `golangci-lint run --new-from-rev HEAD --fix` with `pass_filenames: false` (upstream `.pre-commit-hooks.yaml`, same entry at both revs), so under `--all-files` only issues introduced on top of the checked-out commit can surface. The `lint` job did not cover it either: no Go file changed here, so the paths filter skipped `lint`, `fmt`, `vet` and `test-and-codecov`. v2.1.6 is 2025-05-04 and v2.12.2 is 2026-05-06. Did a plain `golangci-lint run ./...` at v2.12.2 come back clean locally?

.github/workflows/pr-checks.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v9
         with:
-          version: v2.1.6
+          version: v2.12.2
```
bitflicker64: Correcting myself from #2496: I read the version gate off golangci-lint's go.mod, which is the wrong source. `pkg/goutil/version.go` compares `runtime.Version()` of the binary against the target `go` directive, and the action installs the release binary. Measured with `go version -m`:

- v2.1.6 -> go1.24.2, matching the `go1.24` in #2548's failure text
- v2.12.2 -> go1.26.2
- v2.13.2 -> go1.27.0

So on the action side v2.12.2 already clears a 1.26 target and nothing more is needed here. The pre-commit hook is the half that does not, since it builds from source against golangci-lint's `go 1.25.0`.

Suggestion: take v2.13.2 on both lines instead. Same effect for 1.25, and it closes the golangci-lint half of item 2 on #2491 outright, leaving that step as the one Dockerfile line. Not proposing you move the Dockerfile to 1.26 here, the 1.25-first sequencing on #2496 was deliberate.

For reference, #2548 stays red either way: its Docker builds want `golang:1.26` (`go.mod requires go >= 1.26.0 (running go 1.24.13; GOTOOLCHAIN=local)` in its `profiling` job), and its `vet` failures are unrelated mock drift from the bumped deps.

Dockerfile
bitflicker64, line 2: Confirmed safe by CI here: `build-images`, `build-and-deploy` and `compose-deploy` all pass on this head, and they are the jobs that consume this stage.

One follow-on in the same tree: `docs/developer/multi-arch-builds.md:4` still reads "images (`golang:1.24` and `ubi9:latest`) are already multi-arch", so it goes stale the moment this merges. Cheaper to flip here than in a separate docs PR.

`CONTRIBUTING.md:78` and `docs/user/installation.md:7` also say 1.24, but those track the `go` directive in go.mod, which this PR leaves at `1.24.0`, so they should stay.

.pre-commit-config.yaml
bitflicker64, line 57: Worth knowing that this hook and the workflow pin resolve golangci-lint by different means: `language: golang` makes pre-commit build it from source, so the binary's Go version follows golangci-lint's own go.mod rather than the release build. v2.12.2 declares `go 1.25.0`, so this rev produces a go1.25 binary and still fails a 1.26 target. That is what the `pre-commit` job hits on #2548 today, at go1.24. v2.13.2 declares `go 1.26.0` and clears it.

Separately, the green `pre-commit` check here does not show the tree is clean under v2.12.2. This hook id is `golangci-lint run --new-from-rev HEAD --fix` with `pass_filenames: false` (upstream `.pre-commit-hooks.yaml`, same entry at both revs), so under `--all-files` only issues introduced on top of the checked-out commit can surface. The `lint` job did not cover it either: no Go file changed here, so the paths filter skipped `lint`, `fmt`, `vet` and `test-and-codecov`. v2.1.6 is 2025-05-04 and v2.12.2 is 2026-05-06. Did a plain `golangci-lint run ./...` at v2.12.2 come back clean locally?

.github/workflows/pr-checks.yaml
bitflicker64, line 62: Correcting myself from #2496: I read the version gate off golangci-lint's go.mod, which is the wrong source. `pkg/goutil/version.go` compares `runtime.Version()` of the binary against the target `go` directive, and the action installs the release binary. Measured with `go version -m`:

- v2.1.6 -> go1.24.2, matching the `go1.24` in #2548's failure text
- v2.12.2 -> go1.26.2
- v2.13.2 -> go1.27.0

So on the action side v2.12.2 already clears a 1.26 target and nothing more is needed here. The pre-commit hook is the half that does not, since it builds from source against golangci-lint's `go 1.25.0`.

Suggestion: take v2.13.2 on both lines instead. Same effect for 1.25, and it closes the golangci-lint half of item 2 on #2491 outright, leaving that step as the one Dockerfile line. Not proposing you move the Dockerfile to 1.26 here, the 1.25-first sequencing on #2496 was deliberate.

For reference, #2548 stays red either way: its Docker builds want `golang:1.26` (`go.mod requires go >= 1.26.0 (running go 1.24.13; GOTOOLCHAIN=local)` in its `profiling` job), and its `vet` failures are unrelated mock drift from the bumped deps.

bitflicker64 (commented):
Change is correct and I would take it. The image jobs (`build-images`, `build-and-deploy`, `compose-deploy`) all pass here, which is the direct evidence that the `golang:1.25` bump is safe.

One suggestion runs through the two golangci-lint lines: v2.13.2 rather than v2.12.2 would also clear the Go 1.26 step, for both the action and the pre-commit hook, and those two need it for different reasons. Details on each line.

Thanks for crediting #2496 up front in the description. The hunk text is identical across all three files; `Dockerfile` and `.pre-commit-config.yaml` even have the same pre- and post-image blob hashes in both PRs, and `pr-checks.yaml` differs only because main moved underneath it. Whichever of the two lands, #2492 and #2496 want closing in the same pass so the next dependabot bump does not arrive to three competing fixes.

.github/workflows/pr-checks.yaml
leonorfmartins: I think the goal is to move to go1.26.2 in the end, this PR is just a measured step in that direction, so that will make `golangci-lint` version get updated anyways. I didn't understand from your comment if you're suggesting another approach here?

bitflicker64: Yeah, I buried the ask, sorry. It's just: `v2.13.2` on both lines instead of `v2.12.2`. Dockerfile stays `golang:1.25`.

The action side is fine either way, it pulls the release binary. The pre-commit hook builds from source, so it follows golangci-lint's own go.mod: v2.12.2 is `go 1.25.0` and needs bumping again at 1.26, v2.13.2 is `go 1.26.0` and doesn't.

bitflicker64:
checked the open question from my review myself. ran a full `golangci-lint run ./...` on main (`a6eacd80`) in a `golang:1.25` linux container, with both versions built from source the way the pre-commit hook builds them. `v2.1.6` and `v2.12.2` report the same 3 `errcheck` warnings in `internal/server/probe_test.go` (422, 430, 462), all pre-existing, and nothing new. so the bump is clean.

the `v2.13.2` suggestion and the `multi-arch-builds.md` line are both optional. fine to merge without them, and my 3 threads can be resolved whenever it goes in. per vibhu in slack, this should land after the next release.

